PR #3, opened 2017-05-17
Created SSH Service

cathoang:


erinix:
Overall this looks good, however one request, could you make it where this page goes back to the services page after an update.  Entity-config already has a method in it to do this, see its onSubmit() function.

PR #18, opened 2017-10-12
Card table titles

dszidi:
**CSS Conflict Resolution:**
I added some style rules to egret-overrides.css and a class selector to any page that uses entity-table selector and needs the tab header fixed to top of page.

**FAB Buttons (WIP):**
For now it is stuck to top right hand corner of parent element. For cards and tables, I added a title/toolbar area to the element. This branch gives FABs a home on cards, but FABS will still need more work in the following areas ...

- Not every control needs to be a FAB button. 
- FAB buttons with only a single action shouldn't require a hover or click to be presented to the user
- As per the material design spec, FAB buttons should also have different animations when sliding between tabs. Meaning the FAB should not slide with the tab.

### DIFF
--- a/src/app/pages/account/groups/group-list/group-list.component.ts
+++ b/src/app/pages/account/groups/group-list/group-list.component.ts
@@ -5,10 +5,11 @@ import {RestService} from '../../../../services/rest.service';
 
 @Component({
   selector : 'app-group-list',
-  template : `<entity-table [conf]="this"></entity-table>`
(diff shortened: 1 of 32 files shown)


lola27:
Hi @dszidi, it looks nice, the fab button is on the top right of the page. but when the page scroll up/down, the fab button will scrolls on that page too. can you make it fixed at the same position? Thanks.

lola27:
Jenkins: retest

kmoore134:
Jenkins: retest this please

PR #28, opened 2017-10-18
Fix 26082

vaibhav-rbs:


kmoore134:
Looks like there's a merge conflict here with master, can you fix and update the PR?

PR #43, opened 2017-10-24
Revert "Ticket 25825"

lola27:
You have circular dependency error. 
```WARNING in Circular dependency detected:
src/app/services/app-loader/app-loader.component.ts -> src/app/services/app-loader/app-loader.service.ts -> src/app/services/app-loader/app-loader.component.ts```

Please check out the app-loader.component and app-loader.service. There is circular dependency over there.


lola27:
@Nesma1988 please check this out.

PR #49, opened 2017-10-26
Feature jails

lola27:
Added new jails UI with new iocage WebSocket call


kmoore134:
Looks good, thanks Lola!

PR #50, opened 2017-10-27
Ticket: 26372

Nesma1988:
Bug Description: New/Beta UI - login screen does not behave properly when entering the wrong password
Fixes: https://bugs.freenas.org/issues/26372


kmoore134:
Tested here, looks good!

PR #53, opened 2017-10-28
Ticket: 25452

Nesma1988:
Bug Desc: Add red warning when stopping a service
Ticket: https://bugs.freenas.org/issues/25452


kmoore134:
Nesma, I didn't see exactly what this change did. The previous warning made the background (not the dialog box) all RED to let the user know something dangerous was happening.

Nesma1988:
@kmoore134 
I added a warning alert when the admin tries to stop any of the service. Previously there were no warning alert or messages when admin tries to stop the service, it stopped it anyway.
Let me know your thoughts about this.

kmoore134:
Understood, in that case I think that will work in this instance. Thanks!

PR #57, opened 2017-10-30
Chentao fix 25954

Allwithyou999:
I completed https://bugs.freenas.org/issues/25954


kmoore134:
Jenkins: add to whitelist

PR #63, opened 2017-11-01
Subject: Link to the old UI in login page

mactanxin:
Done: Link added
Ticket: #26435


kmoore134:
Is this on the login dialog or the about dialog? The subject says login, but the commit is to the about dialog class?

PR #65, opened 2017-11-02
Subject: "Reboot" and "Shutdown" buttons do not work in new Gui

mactanxin:
Done: "Reboot" and "Shutdown" fixed
Ticket: #26478


lola27:
LGTM

PR #71, opened 2017-11-05
Entity table fixes

psouthwell:
Fixed column issue saw in demo where a string like this: "lkjdlkafjlkajlkefjlkaewjrlkawjlkfjlkasjfllakdjflkadlfljkaslfjklaskjdfkladfj"   was not cutting off appropriately.  The overflow was wrong.  This works now.

REPLICATION Added drop downs for periodic snapshot file paths for which to replicate.. Rather then just an edit box.  The appropriate ones are pre-populated now.  This is for both new/edit forms.

Was able to do this.. Without a more complicated special case plugin.. Found out a simple way to do dynamic form building... without having to tweak entity-form to make it happen.

### DIFF
--- a/src/app/pages/common/entity/entity-table/entity-table.component.css
+++ b/src/app/pages/common/entity/entity-table/entity-table.component.css
@@ -1,7 +1,0 @@
-.padding_all {
-    padding: 20px;
-}
-
-.padding_top {
-    padding-top: 30px;
-}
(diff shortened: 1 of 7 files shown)


psouthwell:
can we try this again?

vaibhav-rbs:
build autostarted after I merged webui:master to this branch,

PR #73, opened 2017-11-06
Entity table fixes

psouthwell:
Fixed protected field that needed to be public.


psouthwell:
Ready for re-jenkins.

PR #87, opened 2017-11-09
Entity table fixes

psouthwell:
Ready.   

Had to snap actions in place. Cleaned up the CSS a bit.  

Looks pretty darn close to mdTable.. But with all the abilities and ACTUAL flexibility of ngxDataTable.

Actions
Sort
Pagination
Filtering
Overflow Fix
Sizeable Columns

AND... The feature code for snapshots.

    Drop down Volumes List for add.  (edit not supported)
    Utilized the existing list/clone/delete

Should be parity complete with Original UI.  Found out the check box was irrelavant.. Code was setting vmshare based on it's occurrence in vm snapshot count.  The code checks by default.  And I see no reason not to vmsharesync when that's the case.  Almost seems like "cause a problem == true/false"  Just set it to (no problem = true) all the time.


psouthwell:
Per review from vaibhav-rbs  

1) Bring back the [5, 10, 25, 100]    (Totally agreed)

2) Edit actions don't work in firefox... Just in chrome.  firefox will be fixed...!!!

THEN.... Ill resubmit this.

psouthwell:
Ill reopen this one in a bit.

PR #91, opened 2017-11-10
Improved SHELL UI to be flexible according to browser size. Ticket: #26542

Allwithyou999:
This ticket was completed. https://bugs.freenas.org/issues/26542

When SHELL page is loaded, automatically, terminal will be resized according to Browser size flexibly.


lola27:
Jenkins: retest this please

PR #98, opened 2017-11-14
Entity pagination demo

psouthwell:
Got the pagination control working in unison with external pagination.

Gives you.. Pagination at top... (and item range selector)    <<  >>

At bottom... Second pagination option by page.

Still will fix the  artifact behind actions pop on chrome.. But this change is separate from that.


psouthwell:
This branch also fixes the missing resources that weren't being inlcuded with entity-table.. Mainly the icons for sort,  <<   >> etc...

psouthwell:
pagination enhancements

PR #109, opened 2017-11-17
UI issue and re-asking of credential on High Sierra Safari(Chentao fix #26574)

Allwithyou999:
https://redmine.ixsystems.com/issues/26574

- On Safari of High Sierra MAC, some CSS styles are supported. For example, font-smoothing: antialiased, transform: translate3D, etc. 
I coded so that they could be hidden on safari regarding the custom CSS, but didn't for node libraries. I think that Safari of itself is a matter.

- Regarding re-credential, when I clicked "remember me", I could never see the alert of credential.
As well, I think that Safari of itself is a matter.

I think the current modification needs to be merged in master branch.


vaibhav-rbs:
LGTM

PR #113, opened 2017-11-20
Ticket:26538

Nesma1988:
Fixes: https://redmine.ixsystems.com/issues/26538
Desc: Add zooming to graphs 

### Guidelines:
Zooming is achieved through two ways..
1. In the mini-map under the actual graph you can select the portion (which was present in the system)
2. Mouse scrolling in the graphs would enable the same functionality; upto the extent the graph can be zoomable (added in this PR).


vaibhav-rbs:
LGTM

PR #115, opened 2017-11-20
Entity action fixes 2

psouthwell:
Fixed the artifacdt that you saw in chrome behind the menu..... AND kept it fixed in the firefox case.


vaibhav-rbs:
LGTM

PR #128, opened 2017-11-28
Fix 26226

lola27:
Finished Plugin UIs - list all available plugins, enable to install plugins, enable to stop/delete installed plugins, and enable to go plugin management UI.

https://redmine.ixsystems.com/issues/26226
https://redmine.ixsystems.com/issues/24544


erinix:
LGTM

PR #129, opened 2017-11-28
FIX-26868

lola27:
Only show freenas/truenas logo once ui get response from middleware

Ticket: #26868
https://redmine.ixsystems.com/issues/26868


erinix:
LGTM

PR #132, opened 2017-11-29
Themservice default fix

psouthwell:
The way to see if the bug is fixed.. Take a fresh system that has never run the theme changer. (So the variable is never set)... It should pick a default of ix-systemblue.. With the Icon working :)

Also.. Theme changeing should continue to work.. after saved.


lola27:
LGTM

PR #135, opened 2017-12-01
Fix 24419

lola27:
- enable entity-table to do multi-select
- enable to start, stop, update, and delete multiple jails at once
- enable to start, stop, and delete multiple installed plugins at once

https://redmine.ixsystems.com/issues/24419


erinix:
LGTM

PR #141, opened 2017-12-05
Entity group table

psouthwell:
Enhanced the filter ability for volumes page

You can now filter the table using a hierarchy browsing tree control.

When you select a node.. It filters the table to show only that dataset.. and it's children.


psouthwell:
Gonna add some a splitter window.. Then Ill re-open anohter pull request

PR #142, opened 2017-12-06
Entity group table

psouthwell:
Test.  Just seeing if this builds after upgrading to angular-cli@latest

Please don't actually review this.  Im just seeing if it passes.


psouthwell:
Ok.. This can be merged.

erinix:
I noticed you commented out the tour, could you fix that and then I will merge it?

psouthwell:
cool Erin.   Its been done :)

psouthwell:
wait.. one more inny tinny binny thing in title

psouthwell:
ok Tis Ready Erin.

erinix:
Upon further inspection I notice that I don't see the pool themselves in the in the normal table (perhaps because it isn't a dataset/zvol?), because of this there are no action buttons that would let me delete it or anything.  We need those action buttons on the tree so that the pool can be managed (and maybe just make the tree the default) before this can be pulled in as it

erinix:
But then I refreshed the page and they appeared, not sure why.

psouthwell:
Ok Erin Id like to push this now.. Think the look and feel is more inline.. Wen't a different route to solve it.. but I think the end result is nice.. and close enough to what your saying that you'll also like.. Ive run it in chrome and firefox.. Looks pretty good.. .To me anyways...   Think it's ready for master.

psouthwell:
Lets push this when your comfortable erin :)

erinix:
LGTM - Angular 5!!

PR #149, opened 2017-12-08
asking users to check Notifications if boot-pool condition is DEGRADE…

vaibhav-rbs:
…D, changed variable from res to wres for cleaner code.


erinix:
LGTM

PR #156, opened 2017-12-13
FIX-27105

erinix:
Remove unused tooltip service from network global configuration  to fix the page

Ticket: #27105


lola27:
LGTM

PR #170, opened 2017-12-18
Entity group table2

psouthwell:
Ive got the control much closer.. BUT I had to make compromises due to the amount of effort vs feature reward.

Do have a ZFS_POOL based grouping columnn.
But...... Had to KEEP including it as a row.. Or I had no way in the UI of presenting the clickable actions that pertains only to the zfs_pool type.. And not the same root node named dataset.

Sorting broke.. (Doesn't funciton during group column being present.. It's either or).  Sort.. Or group by... Pick one... (Is how the control thinks).

CanNot.. Hide the columnb in the rows... That are already in the group colun.. So... zfs_pool_parent appears as a columnb.. And a group panel.  Sort of redundant.

I had to disable pagination.. As it worked poorly with the group control..   You can have 2 groups and 40 items... (20 each 5 per page)... I have to scroll to 3rd page to see the second group.. even when the group panels are collapased.  So. I took paging out entirely.. Hoping the grouping does the noise reduction of so pagination.. Which.. I can see some cases.. Would be a very very lot of items.. If your looking at massive dataset count in a gievn zfs_pool. 

So..... Id like some feed back about how much further to take this feature beyond this check in... :)

Apologies.. Wish I had a better ansswer... But the VERY simpler VERY reasonable request.> Became a rat whole in the ngx-datatable and angular framework.


erinix:
Jenkins: retest this please

PR #175, opened 2017-12-20
Alternative group table

psouthwell:
This is the new volumes page.

Don't show ZFS as a data row with type.. It became instead an expansion panel header.

Page lists collapsed expansion panels by zfspool.... Expand each panel.. Antd there's an entity table showiong just that pools dataset... Theres a global menu above table and under expansion column for the zfspool actions (delete,status).    Believe this should be ready to merge use etc.

Defaulted pagination to be 20.. And have a new.. Probably will see most anything 1000 possible.. More then that.. Forcing pagination.


erinix:
I have one request, if there is only one pool could the single pool be by default expanded?

PR #188, opened 2017-12-26
Rename volumes to zfs pools

erinix:
Ticket: #26111


erinix:
This will have to be redone, too much changed in this section

PR #191, opened 2017-12-27
Subject: Network summary dashboard widget

mactanxin:
Done: new widget is ready
Ticket: #26955


erinix:
It does not appear to show my nameservers even though they do show in the old ui's network summary tab does (perhaps because I am using DHCP?)

PR #193, opened 2017-12-28
New GUI log out operation shouldn't have confirmation checkbox in it(Chentao fix 27446)

Allwithyou999:
https://redmine.ixsystems.com/issues/27446

Done


erinix:
the button for logging out appears to still be greyed out even though the checkbox is hidden, you should add some logic to the [disabled] field for the button to not be greyed out when hideCheckBox = true;

Allwithyou999:
@erinix right, That is not the latest version and a commit was missed.Sorry

Allwithyou999:
Resolved.

erinix:
LGTM

PR #195, opened 2017-12-29
Fixes for update/shutdown/reboot/login pages

erinix:
Make reboot page show up on reboot events, add shutdown page for
shutting down, hide login page while websockets are not available and
add loader to update pages

Ticket: #27357
        #27453
        #27068
        #27444


lola27:
LGTM

PR #196, opened 2017-12-29
Tooltip adjustment: AWS alert service

Mrt134:
Added an explanation and example of an ARN.
Tested locally, no issues/errors.


lola27:
LGTM

PR #202, opened 2017-12-29
Diskpage visual improvement

psouthwell:
Quick improvement for bug I saw in tabs.. Table would randomly degrade the rows.. Under extreme tab swithcing... Had same issue once upon a time in the Reportdashboard... Works nice now.  :) "Well behaved"


psouthwell:
small bug fix for t

PR #203, opened 2017-12-30
System Advanced MOTD field should use fixed-size font(Chentao fix 27465)

Allwithyou999:
https://redmine.ixsystems.com/issues/27465

Done.


lola27:
LGTM

PR #218, opened 2018-01-03
Tooltips: add 3 / 3 messages to the Storage -> Snapshots -> Add page

Mrt134:
There is one additional to add in the 'clone' file.
Build test: no issues locally.


erinix:
LGTM

PR #221, opened 2018-01-03
Change requirement for login button so it doesn't get arbitrarily disabled by tin foil hat browsers.

erinix:
Ticket: #27444


lola27:
LGTM

PR #222, opened 2018-01-03
Save Debug in System-Advanced Page

mactanxin:
ticket: #27466


william-gr:
@mactanxin redirecting to "/legacy/system/debug/download/" can't work.

Specially if the user is not logged in to the old UI.

You have to download the file using the same mechanism used in "Save Config", calling `file.download`

erinix:
LGTM

PR #235, opened 2018-01-08
Fixed out of order names.

a-stjohn:
Fixed a typo where "Guest account" was named as the checkbox when it should have been the select type.


erinix:
LGTM

PR #250, opened 2018-01-09
ADDED: set font on page

mactanxin:


erinix:
A couple recommendations, first it should probably have a default, 14 px looks pretty reasonable for the smallest on my screen, second it shouldn't  allow negative px, perhaps it would be better as a dropdown with some values between 10 px and 32 px or so?

erinix:
also I notice with larger fonts the text tends to go off the screen, is there a way to refresh the terminal when you readjust the font?

erinix:
LGTM

PR #256, opened 2018-01-10
Fix the field for vlan parent interface in edit mode and add priority code point

erinix:
Ticket: #27508
        #27554


lola27:
LGTM

PR #265, opened 2018-01-11
Added more information about how to find the service key and the…

sambonham:
the client name from the PagerDuty web interface.


lola27:
LGTM

PR #266, opened 2018-01-12
Fix 27468

lola27:
Fixed bind interfaces list is empty for SSH service

https://redmine.ixsystems.com/issues/27468


erinix:
LGTM

PR #269, opened 2018-01-12
Add Time Machine Quota

erinix:
Ticket: #27597


lola27:
should be ticket#27662

lola27:
LGTM

PR #279, opened 2018-01-15
FIX-27642

lola27:
Added form-array css file, enable to flex shink fieldset

https://redmine.ixsystems.com/issues/27642


erinix:
LGTM

PR #282, opened 2018-01-16
Found an edge case.. Where I need to check if the avail is instead a …

psouthwell:
…string instead of a number.. Like when it's locked


psouthwell:
edge-fix-for-when-volume-locked

PR #284, opened 2018-01-16
FIX save debug confirm dialog doesn't work

mactanxin:
UPDATED: confirm dialog works as expected, need user to confirm to start download
ticket: #27808


erinix:
LGTM

PR #300, opened 2018-01-19
fixed the browser (not website) login that pops up when you logout.. …

psouthwell:
…basically just force a simple quick page reload after calling the logout code via.... window.location.reload();

One line of code in ws.service.ts     .....  method    logout()


psouthwell:
Please don't merge this.. May have found a bug when updating.

psouthwell:
this may not be a bug.. Still verifying

PR #305, opened 2018-01-20
Fix 27059

psouthwell:


psouthwell:
Retrying this change worked fine...  IIssue was unrelated.

PR #306, opened 2018-01-20
Use SVGs for UI graphics. Implements #27828.

davidjb:
This changes FreeNAS/TrueNAS/iX logos and icons, as well as the Beastie
Shark graphic into their SVG equivalents for scalability and greatly
improving their appearance on zoomed interfaces, high-DPI screens,
mobile devices etc.  This implements https://redmine.ixsystems.com/issues/27828.

Several image tags within the HTML have also had their alt text added or improved for better accessibility at the same time.

`cancel.png` from Hopscotch has been left alone as the graphic scales well enough and it's an issue for that upstream project to resolve.  An issue was raised over there with https://github.com/linkedin/hopscotch/issues/357.

Beastie Shark's artwork was re-obtained from https://forums.freenas.org/index.php?threads/freenas-artwork.40541/ (which is where the original PNG was presumably sourced).

Note that images have **not** been optimised.  Though they could easily be run through `svgo` to save around 50% of their size, this limits their ease of / ability to be edited later.  Ideally, `angular-cli` would do the optimisation automatically as part of the build but this isn't currently a feature they have implemented (https://github.com/angular/angular-cli/issues/2935 & https://github.com/angular/angular-cli/issues/5356).  Webpack has loaders that will help optimise images (eg https://github.com/tcoopman/image-webpack-loader) but using this would involve ejecting from `angular-cli`, so I'll leave this decision up to someone else.


erinix:
LGTM

PR #316, opened 2018-01-23
fixed service page to show service/<servicename> instead of service/s…

vaibhav-rbs:
…ervice.


erinix:
cifs breadcrumb (and probably the route too) should say smb intsead of cifs (cifs is so 90s)

vaibhav-rbs:
erin reviewed the changes over google hangouts, so merging the changes,

PR #319, opened 2018-01-24
Fix 27754

vaibhav-rbs:


vaibhav-rbs:
Jenkins: retest this please

erinix:
email address appears to be a required field and it should probably validate the input to make sure it is given a proper email address before saving, could you add some validation for that field?

erinix:
LGTM

PR #329, opened 2018-01-26
Use a 1-minute ttl token with keepalive instead of storing username/password for authentication

erinix:
Ticket: #27817


lola27:
LGTM

PR #332, opened 2018-01-26
Attempt to update most of the node_modules

erinix:
(supposedly we can even use typescript 2.6 now)


erinix:
Don't actually pull this yet

erinix:
now aot devmode should work again so please test and push if it is good

erinix:
devmode has a refreshing problem still, trying to see if I can find a better combination of deps that fixes this

PR #339, opened 2018-01-29
Temporarily reverting so rest calls can work again, will revert the r…

erinix:
…evert once they are fixed

Revert "Use a 1-minute ttl token with keepalive instead of storing username/password for authentication"

This reverts commit 2b72852d0c8c8c21635da828dce565ef4d3a7913.


lola27:
LGTM

PR #350, opened 2018-02-01
added rekey for encrypted volumes menu option... Also found and fixed…

psouthwell:
… a global problem with filesize method.. where importing it no longer works.. So.. I added it to .angular-cli.json under scripts and access it via (<any>window).filesize.  This works.  Removed the import verified the many pages worked


psouthwell:
Fixed obvious mistake

PR #360, opened 2018-02-05
Subject: System process not working in prod build

mactanxin:
Done: Fixed with removed dependencies from component to angular-cli
Ticket: #28191


erinix:
LGTM

PR #377, opened 2018-02-08
Fix 27929

psouthwell:
Special request erin.

Ive removed dialogService from the WebsocketService.

Reason being that it comes up at a time in my case. .I didn't want to display it...

It can be invoked again.. Using hte subscription error method.. For example..

websocketservice.call("someapi", {}).subscrit((response)=>{
                   // Normal code we all use now... Handles the resposne
}, (error)=> {
           // A method that I dont' see many of us using.. but is here and where I think 
           // the dialog service invoke should be.. So.. The UI comnponents.. Can customize
           // whats done with the error.
            this.dialogService.error(....);
});

as opposed to having the error being internal.. and un overrideable.. in the bowels of websocektservice.onmessage(...).... If that makes sense... 

So... Lets see if you agree...  

the original bug

https://redmine.ixsystems.com/issues/27929

Is not reporoducable.. I thought it might be on a totally fresh... But I took nightly from today.. Made sure it's updated by trying an update (It said it was already updated)...

And the UI works fine.. DOn't get the missing aggreation-cpu-sum path error.

However.. I did put this in.. To protect against it.. And in the chart cases... Log the error instead of showing a dialog.. Whhen a dataset isn't found ;)

### DIFF
--- a/src/app/components/common/lineChart/lineChart.service.ts
+++ b/src/app/components/common/lineChart/lineChart.service.ts
@@ -96,6 +96,9 @@ export class LineChartService {
       });
 
       dataHandlerInterface.handleDataFunc(linechartData);
+    }, (error)=>{
+      console.error("Application got error in LineChart.service.ts calling data for list observce objects in log ->", dataList, error);
+      return false;
     });
   }
 
--- a/src/app/services/ws.service.ts
+++ b/src/app/services/ws.service.ts
@@ -16,15 +16,15 @@ export class WebSocketService {
   pendingSub: any;
   pendingMessages: any[] = [];
   socket: WebSocket;
-  connected: boolean = false;
-  loggedIn: boolean = false;
+  connected = false;
+  loggedIn = false;
   @LocalStorage() token;
-  redirectUrl: string = '';
+  redirectUrl = '';
   shuttingdown = false;
 
   public subscriptions: Map<string, Array<any>> = new Map<string, Array<any>>();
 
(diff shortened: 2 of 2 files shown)


psouthwell:
If needed Erin.. What I can do.. Is go to the places ws.call is being called...

and snap in

, (error) => {

this.dialogService.error(.....);

});

Then there'd be no worries...

erinix:
Not a bad idea but one issue, it makes error messages for pages other than the dashboard show up.  For instance try creating a user with a duplicate name, you won't see on the page why the task failed.

psouthwell:
I nkow Erin. But putting a UI in a Service is the wrong IDea... You can't turn it off....    Services should never have UI... Else.. They'd be components.

psouthwell:
It's like.. Rather then using good error handling pracitices.. Were going around it..   By using a deeply entrenched Artifact in the Service... Be like.. If a REST service used webpages to express internal server errors.. Just not done.

psouthwell:
Im wrong erin.. thought for a moment.. I was arguing the validity of taking the dialogService out of WebSocketService.   But I realize.. You'd of course understand it..

Your point is... I didn't write code to support the artifact that the other portions of code were expecdting.. So.. This PR would basically.. Break a bunch of stuff....

Gotcha..  Ill leave this open.. And push that code into there. .I agree.. The artifact even if not uber coding style is expected.. It's essentially a feature how it's working.. and needs to be replaced not removed.

PR #386, opened 2018-02-11
Fix 28429

lola27:
https://redmine.ixsystems.com/issues/28429


erinix:
I have one suggestion, perhaps have the box get auto-checked when dhcp becomes checked

lola27:
Yes.  The vnet box will auto-checked when DHCP becomes checked, but user can uncheck it. So if user uncheck vnet box by themselves, a notification dialog will be pop up.

PR #391, opened 2018-02-13
Petes weekly 2 7 2018

psouthwell:
added
1.7.0-beta.3

For the angular-cli version.. This should put us back where we were before going to RC0 and some of that versions assocaited CSS stranges.


psouthwell:
Fixed AOT issues.

psouthwell:
finalized.
Only show lock unlock.. If  volume has a passphrase
Show unlock if locked... 
Show lock if unlocked
Show status for all actions (sense many of the encrypt actions take some time)

PR #411, opened 2018-02-22
Add translations to reboot and shutdown pages

erinix:
Ticket: #23139


mactanxin:
found an error on rebooting
ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'role: undefined'. Current value: 'role: dialog'. It seems like the view has been created after its parent and its children have been dirty checked. Has it been created in a change detection hook ?

PR #424, opened 2018-02-27
Change action labels to be capital letter followed by lowercase

erinix:
(as it was before they were turned into keys, this should make it consistent again and actually easier to translate)

Ticket: #28842


mactanxin:
LGTM
LGTM

PR #435, opened 2018-03-01
activate pool page for Jails

mactanxin:
this is the for the init install with no jails pool activated, should prompt a config setup
ticket: #28864


erinix:
LGTM

PR #438, opened 2018-03-01
Warning when switch Update train

mactanxin:
ticket: #28714


erinix:
This looks good so far but one request, the old UI outright prevents you from switching from a nightly to a stable, we should probably mimic this behavior too

erinix:
Looking better, the error message should be different for when the  user goes from the current stable branch to an older stable though since you are not switching from "nightlies"

erinix:
LGTM

PR #456, opened 2018-03-06
Use ngx-translate instead of ng2-translate

erinix:
Ticket: #23139


mactanxin:
I got a  `ERROR in src/app/core/directives/animation.directive.ts(37,21): error TS2554: Expected 2 arguments, but got 1.`

PR #460, opened 2018-03-07
Fix 24107

lola27:
Enable to access jail console via UI

https://redmine.ixsystems.com/issues/24107
https://redmine.ixsystems.com/issues/29129


erinix:
LGTM

PR #467, opened 2018-03-09
Add tooltips to Jail Add, Edit, and Configure pages.

Mrt134:
Also, RCTL property tooltips are added, although that feature has not been implemented yet - do we need to comment that section out for now?


erinix:
rebuild

lola27:
LGTM

PR #484, opened 2018-03-13
this is a commit brings in changes for VM pages

vaibhav-rbs:
this PR is was a result of my commit to the master directly and then reverting the said commit to master. this would have been avoided I has created FIX-29817 after I reverted my commit to the master or if I would have created FIX-29817 right on `13eeb99cc8fefcc88fea6ee65cf7d5a54cc436ef`. but anyhow things are in order and please review this change,


erinix:
LGTM

PR #493, opened 2018-03-15
Jails forms reworks:

Mrt134:
Adjusted the tooltips and option organisation for the Jails/Add and Jails/Edit forms to be more informative and better guide the user through the jail creation process.

Merge in translation elements and fix a few merge conflicts with my in-progress work.

Reworked elements of the Jail Wizard to better avoid user confusion.

Added a "Cancel" button to the basic wizard form that will allow the user to easily leave a Wizard form.

Local build testing with npm install and npm start: no issues or breakage detected.


erinix:
The cancel button is a good idea but it should probably only show up on the first page (i.e. in place of the back button)

Mrt134:
Thanks, I reworked it a bit so the "Cancel" button should now only appear on that first section of the wizard.

Mrt134:
Closed to consolidate with the other pull request.

PR #495, opened 2018-03-15
Remove redundant text that always displays, linkify logos, Jail forms reworks, beginning error review, add "Cancel" button to wizards

Mrt134:
Per Warren's request, I removed the text that is always displaying in the upper right area of the screen and in its place added relevant links to the company/project logos that are always displayed.


erinix:
The cancel button is a good idea but it should probably only show up on the first page (i.e. in place of the  back button)

erinix:
This pull request appears to have commits from your other PR in it, could you perhaps consider consolidating these two?

Mrt134:
Yep, I'll reconfig this to the single pull request.

erinix:
LGTM

PR #500, opened 2018-03-16
Revert "Remove redundant text that always displays, linkify logos, Ja…

erinix:
…il forms reworks, beginning error review, add "Cancel" button to wizards"

On second thought removing this legal text is not a good idea, reverting until I get clarification


lola27:
LGTM

PR #501, opened 2018-03-16
Entity table's spinner for all the things

dszidi:
Reverted template on entity-table-add-actions.component to previous state. Spinner is back for everything


erinix:
LGTM

PR #505, opened 2018-03-17
Fix 29018

vaibhav-rbs:
adding Marcelo as he requested to be part of this review.


erinix:
LGTM

src/app/pages/vm/dockervm-wizard/dockervm-wizard.component.ts
```diff
@@ -0,0 +1,238 @@
...
+        ( < FormGroup > entityWizard.formArray.get([2])).controls['NIC_mac'].setValue(mac_res);
+      });
+  
+  }
+  getRndInteger(min, max) {
+    return Math.floor(Math.random() * (max - min + 1) ) + min;
+}
+
+async customSubmit(value) {
+  const path = value.raw_file_directory+ '/' + value.raw_filename+ '_'+ value.name;
```
araujobsd: I think here would be safer to do something like:
const path = value.raw_file_directory+ '/' + value.id + '_' + value.name + '_' +value.raw_filename;

Or otherwise users will face out an issue if they try to create two VMs with the same name, vm.name are not unique, but vm.id must be unique.

What you think?

src/app/pages/vm/vm-cards/vm-cards.component.ts
```diff
@@ -413,60 +399,111 @@ export class VmCardsComponent implements OnInit {
...
+          if (res[0].vm_type === "Container Provider"){
+            this.dialogRef = this.matdialog.open(EntityJobComponent, { data: {title: 'Fetching RancherOS'}, disableClose: false});
+            this.dialogRef.componentInstance.progressNumberType = "nopercent";
+            this.dialogRef.componentInstance.setCall('vm.fetch_image', ['RancherOS']);
+            this.dialogRef.componentInstance.submit();
+            this.dialogRef.componentInstance.success.subscribe((sucess_res) => {
+              this.loader.open();
+              this.ws.call('vm.image_path', ['RancherOS']).subscribe((img_path)=>{
+                if(!img_path){
+                  this.dialog.Info('CHECKSUM MISMATCH', 'system checks failed to verify ISO, please try rebooting Virtual Machine');
```
araujobsd: If for some reason the download has a checksum problem the VM won't start, but if you try to start it again, it will remove the image problem and download it again.

Maybe a better error message would be something like:
this.dialog.Info('CHECKSUM MISMATCH', 'system checks failed to verify ISO, please try to start again');

PR #510, opened 2018-03-19
FIX 29629

lola27:
Ticket: #29629
Add translation makers to the Accounts section


erinix:
The "save" and "clear" button text should be "Save" and "Clear" so it will use the same translations for those words as other buttons do

lola27:
Thanks Erin! Changed it :)

erinix:
LGTM

PR #512, opened 2018-03-20
Volume translation

psouthwell:
added translation to STORAGE UIs (Snapshots, vmsnapshots, disks and volumes)

some fixes along the way.  Included a volume fix where undefined columns that could pop up were causing an error so improved check verified it worked.

Added dialogService.info to replace an alert()  being called ;)

fixed an error case handlign in add new snap shot.


erinix:
The placeholders labeled "passphrase" in volumeunlock-form.component.ts should probably be "Passphrase", i.e. capitalized for consistency.

erinix:
rebuild

PR #513, opened 2018-03-20
Translation markers for Service components

mactanxin:
ticket: #29685


mactanxin:
should be fixed by import module

erinix:
LGTM

PR #514, opened 2018-03-20
Shell page translation

mactanxin:
ticket: #30348

### DIFF
--- a/src/app/pages/shell/shell.component.html
+++ b/src/app/pages/shell/shell.component.html
@@ -8,7 +8,7 @@
 	<mat-card-content>
 		<mat-divider class="mb-1"></mat-divider>
 		{{"Set font size" | translate}}: <mat-slider min="10" max="20" step="1" value="14"  [(ngModel)]="font_size"></mat-slider>
-		<button mat-raised-button (click)="resetDefault()" matTooltip={{"Reset font size to default value" | translate }}>{{"Restore default" | translate}}</button>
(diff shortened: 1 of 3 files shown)


erinix:
I am seeing this when I try to build it with aot

```ERROR in : The pipe 'translate' could not be found ("
  <mat-card-title class="mat-bg-primary m-0">
  	 <div class="card-title-text">
  	 	{{[ERROR ->]"Shell" | translate}}
  	 </div>
  </mat-card-title>
")
: The pipe 'translate' could not be found ("
	<mat-card-content>
		<mat-divider class="mb-1"></mat-divider>
		{{[ERROR ->]"Set font size" | translate}}: <mat-slider min="10" max="20" step="1" value="14"  [(ngModel)]="font_s")
: The pipe 'translate' could not be found ("	<button mat-raised-button (click)="resetDefault()" matTooltip="Reset font size to default value">{{[ERROR ->]"Restore default" | translate}}</button>
		<tooltip class="shellTooltip" [message]=shell_tooltip></to")
```

You need to import the translate module into the shell module

mactanxin:
fixed

erinix:
one more thing, the matTooltips should also be piped to translate

erinix:
also this one ```matTooltip="Reset font size to default value"```, you should be able to do something like ```[maTooltip]={{"Reset font size to default value" | translate}}```

mactanxin:
Ah ok, it's a little tricky to make it work with quotes within quotes instead of using variables

erinix:
LGTM

PR #516, opened 2018-03-20
FIX 29429

lola27:
Comment out 'RCTL Limits' section of the jail add and jail edit since it hasn't been added yet

Ticket: #29429
Ticket: #29353


erinix:
LGTM

PR #520, opened 2018-03-21
Removed random un-needed logs... Gave a confirmation boxy to upgradin…

psouthwell:
…g a volume that warns doing so could make some features of the pool incompatible with older versions of freenas

Hi Erin.. Sorry for submitting the last  PR prematurely here are the extra bits you wanted in addition.


erinix:
The confirmation box works and the upgrade option shows up (although you should probably have some kind of visual representation that shows the volume is out of date, maybe an icon that shows up) but the upgrade rest call itself seems to fail

erinix:
LGTM

PR #521, opened 2018-03-21
ftp service check if no certificate installed

mactanxin:
ticket: #30402


erinix:
I am still getting the traceback data.forEach is not a function, consider rewriting that forEach into a standard for loop.

erinix:
LGTM

PR #525, opened 2018-03-22
reimplement VM clone button using custActions

vaibhav-rbs:


erinix:
you should probably import the loader to the vm-card-edit directly rather than call it from the conf, everything else looks good

PR #528, opened 2018-03-22
FIX 30369

lola27:
Ticket: #30369


erinix:
Could you also put markers on the plugins-available and plugins-installed components?

PR #541, opened 2018-03-27
Fixed ticket #30321

dszidi:
Added conditionals in widgetsysinfo html and ts files to detect ixsystems as manufacturer. Should now show serial for any hardware with ixsystems in the manufacturer attribute. Also fixed a few flex-layout issues in the dashboard and removed debug messages from widget


dszidi:
As far as I can tell, I don't see why this wouldn't work properly. I have is a mini and it works. However, if somebody with access to more iX Systems hardware could test this, that would be great.

erinix:
LGTM

PR #548, opened 2018-03-28
Tkt 29477

dszidi:
Fixes pool widget layout on dashboard. It should fall into the regular flow of the cards. Also added some adjustments to flexlayout. Also adjusted widgetpool's css file to use css variables from theme . If any weird readability issues come of these color adjustments, the fix will have to be via adjusting the color palette in the theme itself.


dszidi:
I'd like somebody with access to multiple pools to please verify this works as intended. I don't have multiple pools available here.

erinix:
LGTM

PR #553, opened 2018-03-29
FIX 30888

lola27:
Ticket: #30888


lola27:
rebuild

erinix:
could these maybe just be regular buttons?  tooltips help but they are still kind of inconsistent with the rest of the UI

erinix:
LGTM

PR #555, opened 2018-03-29
Figured out.. The API was returning M (for megabytes) although the in…

psouthwell:
…put for the api only accepts K... So.... If it's M.. I reconvert it back to K.. Which then selects an actual choice for the drop down.. Of which.. All are K.

Simple test case... Edit a Dataset.  recordsize  should be working now.  (Sticks to the Select)


erinix:
rebuild

erinix:
rebuild

erinix:
LGTM

PR #556, opened 2018-03-29
Added a Read nonly name field indicating the name of the dataset bein…

psouthwell:
…g delete.. Actually the path

Should be simple to test.  Create a dataset that can be deleted.

Then..  Navigate to it in the tree... And "Delete Dataset"

The name field should now indicate to the user.. the item being deleted.


erinix:
LGTM

PR #557, opened 2018-03-29
Removed successful box in datasetform-save.. not needed.  Also fixed …

psouthwell:
…the disk to show WIPE for only disks being used in pools

So... only disks being used by a pool should show the wipe option.  Other wise.. Wipe is not there.

This is on the Storage/disks page.


erinix:
It still presents wipe for my bootpool disk, don't wanna do that lol

psouthwell:
Ok Erin..

So now.. Root Pool Disks.. Shouldn't have the Wipe Menu..


And there's a new tab called "Root Pool"  which should show only the disks involved in your root pool.

erinix:
LGTM

PR #558, opened 2018-03-29
Fix 30225

vaibhav-rbs:


erinix:
Seems to be broken for jail shell, getting this.xterm is undefined

vaibhav-rbs:
LGTM

PR #579, opened 2018-04-04
Fix 31176 31152

vaibhav-rbs:


erinix:
you need both required: true and validators.required for all the required fields otherwise it traces back or doesn't show the * for that field

erinix:
LGTM

PR #581, opened 2018-04-05
fix vm title on device list page. fixed a traceback as well.

vaibhav-rbs:
https://stackoverflow.com/questions/39787038/how-to-manage-angular2-expression-has-changed-after-it-was-checked-exception-w


erinix:
LGTM

PR #584, opened 2018-04-05
Feature 28712 api finished

psouthwell:
There's geting a bunch of stuff in here..  Want to push

delete's all replaced with confirms.  ONly detach is left.  only volume-delete is left.. And that's becaue it's really a Detach/Or Delete and it's   template : `<entity-form [conf]="this"></entity-form>`  Not entity-delete.

So testing this.. Is going through the storage places.. making sure entity-delete does not appear.


psouthwell:
Had to fix two issues.

1) I caused  BUG and wipe wasn't showing.. Fixed that.

2) Found out Not my bug.. but an older bug was in the Wipe form.. So fixed that aswell.

erinix:
LGTM

PR #590, opened 2018-04-06
Change text-align and padding for tooltips

dkmullen:
Very small change to CSS to make tooltips left-align and to even out the padding.


erinix:
I don't think you merged master properly, it shouldn't show all those commits like that

PR #592, opened 2018-04-06
Using inputType='password' in storage form configruaitons when ever t…

psouthwell:
Used the inputType password on all passphrase input within the storage section.

To see them...  Create an encrypted volume.

Then try stuff like.. Create Passphrase
                                Lock/Unlock

etc..  Under volumes section.  Password INPUTS should look like  "****"  instead of seeing the characters.


erinix:
LGTM

PR #597, opened 2018-04-07
FIX 32193

lola27:
Ticket: #31293


erinix:
Since Jail Edit has the same fields it should probably have this logic too

erinix:
LGTM

PR #605, opened 2018-04-09
Animation service

dszidi:
Created animation service to eventually replace animation directive. Implemented stats service that acts as a proxy for components that require stats. This way the stats api calls won't be made unless the stats service confirms that stat source is available. This service also fetches stats at regular intervals so dashboard widgets get updated in a staggered fashion which looks nice and also helps the middleware not get overloaded with calls. Implemented stat service as a global singleton and dashboard widgets currently make use of it.


vaibhav-rbs:
rebuild

erinix:
The first build had this error in the log
```
ERROR in src/app/core/services/animation.service.ts(45,27): error TS2554: Expected 0-1 arguments, but got 2.
```

dszidi:
Oh is that fixed? I have been putting in that bogus second parameter because a while ago you said it wouldn't build otherwise

PR #610, opened 2018-04-09
Faeture 28712 sunday

psouthwell:
Got the API volume/id/detach    body:  JSON.stringify( {  label:  disk.disk_name } )

This works.  the api retuns and says "Disk successfully detached"

But.. I see no way to then detect this.. The getting disks by pool.. The disk I detached still comes up?

I notice in disks.query There's no status in da1 vs da2 (I detached da2 in my pool).  Same is true of v1.0/storage/disks

So.. How do you detech the status that this api toggled?

OH.. I removed the uneeded Succesfully Un locked volume blah blah.. because you can see the difference in the parent list when itupdates (Not locked)


psouthwell:
Ok.. Now show offline/online (Volume per disk status) on disks page.
OFLINE by label.. But  I feel there's a middle wier bug.. It reports the disk was offlined.. But the volume status api does not reflect that.. Still reporting "ONLINE" but I don't think that'a a UI BUg.

erinix:
LGTM

PR #611, opened 2018-04-09
Optimize the loading of dataset data so volumes-list won't take as long

erinix:


psouthwell:
LGTM... Nice one Erin.

PR #613, opened 2018-04-10
Tour does not start by default

mactanxin:
the automatic tour to only appear when the user requests it.


mactanxin:
ticket: #31620

PR #616, opened 2018-04-11
Fixed wording problem casuing confusion... 'Create Key'  Was supposed…

psouthwell:
… to say 'Create Passphrase'  because that's what it did... Corrected the label so it says what people wold expect to see... 'Create Passphrase' for a newly constructed pool/volume


erinix:
LGTM

PR #618, opened 2018-04-11
Change textarea on info and error dialogs to readonly

dkmullen:
Ticket: #31704
Info and error dialog boxes were editable.


erinix:
LGTM

PR #621, opened 2018-04-12
Fix 31680

vaibhav-rbs:


lola27:
Hi, could you make a placeholder to the mat-select or set a default value?

PR #627, opened 2018-04-13
Feature 31158

psouthwell:
Finished adding

,
      validation: [Validators.required],
      required: true

To all the forms in storage.  

Also to Detach Volume added a confirmation.  In addition to the "Delete All Contents" thing..  Now you have to confirm before button allows pushng.  Did that for all things.  Don't let form poppy if there's no default parameter(s) values required.

Easy enough to do this.. but testing requires trucking around the Storage Section and trying all forms.  Ive done this.. But should be verified by another.


erinix:
LGTM

PR #646, opened 2018-04-18
Make dynamic instructions for console dialog

dkmullen:
Ticket: #31380
Instructions for console are on the tooltip where it is launched, but since the console is active throughout the app, I added dynamic instructions to the console.


erinix:
LGTM

PR #648, opened 2018-04-18
Tkt #27457: 

dszidi:
This removes delete button and replaces it with an options icon/popup menu with more start options and a delete option. For quick toggling of basic power states the old power button is left where it was.


erinix:
LGTM

PR #650, opened 2018-04-18
catching up with ws API changes.

vaibhav-rbs:


dszidi:
I got an error but I wasn't running the latest nightly. Updating and then will try again

dszidi:
Checked it with the latest nightly and it works for me :)

PR #651, opened 2018-04-18
Update README.md

wblock:
Show installing npm only, which will pull in the correct version of node as a dependency. (Verified.)

Add note that as long as the webserver is kept running, changes to files will be incrementally compiled.  (Not verified.)


erinix:
LGTM

PR #663, opened 2018-04-20
Update Jail mount point "Destination" tooltip.

Mrt134:
- Rewrite message to reflect that a directory named in "Destination" that does
  not exist is automatically created.
- Local build testing: no issues.


erinix:
LGTM

PR #667, opened 2018-04-21
Fix WebGUI HTTP -> HTTPS Redirect

erinix:
Fix 30183


william-gr:
@erinix this also has to happen when port changes, redirecting user to right port (it could not be the default)

vaibhav-rbs:
@erinix how do you want me to test this?

PR #680, opened 2018-04-28
enable to active/create pool on plugins ui

lola27:


erinix:
Looks good although you should add pipe the mat-select title and the no pools exist messages to translate

PR #681, opened 2018-04-30
Fix/ftp config 32400

dkmullen:
Ticket: #32400
Add emphasis to the selected input on Services/ftp/configure so that the item selected isn't overlooked.


erinix:
The big red text isn't really necessary (also only doing it on ftp doesn't help the usr with this on any other place where the file explorer is used), I think what would be more helpful would be an indication that the directory was selected, i.e. highlight the node on the file explorer.

PR #693, opened 2018-05-02
Raise tooltips that drop off bottom of screen

dkmullen:
Ticket: #689 
Tooltips now revert to original style when screen scrolls up. :)


erinix:
LGTM

PR #703, opened 2018-05-04
Fix/pool manager 32481

dkmullen:
Ticket# is actually 32508 for this one, not the one in the branch name


dkmullen:
Also in this PR I restored the check-all box for tables on the right side of pool manager. Maybe there is a good reason to leave them commented out???

erinix:
LGTM

PR #705, opened 2018-05-04
Increase token timeout from 1 minute to 5 minutes

erinix:
(this helps with the issue of getting logged out prematurely)

Ticket: #31425

### DIFF
--- a/src/app/views/sessions/signin/signin.component.ts
+++ b/src/app/views/sessions/signin/signin.component.ts
@@ -66,7 +66,7 @@ export class SigninComponent implements OnInit {
   }
 
   successLogin() {
-    this.ws.call('auth.generate_token', [60]).subscribe((result) => {
+    this.ws.call('auth.generate_token', [300]).subscribe((result) => {
       if (result) {
         this.ws.token = result;
 


dark-swordsman:
Firstly, I want to apologize in advance if I'm missing something, but based on what I'm reading, I don't think I am.

Secondly, sorry since this is definitely not the right place, but given that [issues with little traction get closed](https://jira.ixsystems.com/browse/NAS-100275) regardless of what they are, I kinda feel like I have no other option.

To this day, it's the same amount of time and uses the same call. 5 minutes seems arbitrary and is trying to impose pseudo security, especially when it's an API endpoint. It causes a major workflow issue for me and there's no reason I shouldn't be able to define this in my settings.

So right out of the gate, the feature that I would request: Allow the user to change this token expiration in the UI. You can clearly modify it here very easily, so I don't see why there can't just be a UI section to do this.

Now, forgive me if there's some ignorance here, but it seems that there's no actual authorization happening when a token is requested. It's just [checking for a boolean](https://github.com/truenas/webui/blob/a98de752dbd27c2fbffeeeb6bbbb929c4a1aedd5/src/app/views/sessions/signin/signin.component.ts#L351) and then [requesting a session token](https://github.com/truenas/webui/blob/a98de752dbd27c2fbffeeeb6bbbb929c4a1aedd5/src/app/views/sessions/signin/signin.component.ts#L377). What's stopping me from seeing exactly how this works (like I just did) and abusing it? A 5 minute session login would do nothing to stop me, when I can just request a token?

Usually, you send the username and password to a backend server, and then it gives you a token. It doesn't send login success locally, and then request a token from the client. Unless I'm missing something, this is actually a core issue with your security, and a 5 minute token expiration will do nothing from someone genuinely trying to abuse your software.

undsoft:
Hi @dark-swordsman,
This is definitely not the best place for this :) 
I agree that how we handle timeouts can be improved. I have created a ticket that you can track https://jira.ixsystems.com/browse/NAS-113146

You are incorrect in saying that no actual authorization happens when token is requested. The `loginCallback` method you were looking at is being called after sending authentication token to API.

What I can offer you now is to manually adjust timeout value on your machine.
You can:
1. Go to shell on a TrueNAS machine.
2. Navigate into webui directory: Core: `cd /usr/local/www/webui`, Scale `cd /usr/share/truenas/webui/`.
3. Find JS file with timeout value: `grep -rl . -e 'auth.generate_token",\[300'`
4. Update the value to something else in your text editor.
5. Relogin to webui.

You will of course need to do this every time after an update.

henvic:
I've been using TrueNAS Core for hosting my photography library and am bothered by its low timeout, especially with 2FA enabled.

Some options I see to improve this:

1. Add option to remain signed for 24h / browser isn't closed.
2. Make it user-configurable and increase the default value to 30min.
3. If you improve support to 2FA beyond OTP, ask for the second factor alone.

jhavron:
I found the code in the files quickly using [Midnight Commander](https://github.com/adamyg/mcwin32/releases) to copy the whole directory to my PC and [grepWin](https://github.com/stefankueng/grepWin) to search the folder.

26-es5.ac2edb595650fe92bfa8.js
and
26-es2016.ac2edb595650fe92bfa8.js
image


TrueNAS-13.0-U2

PR #717, opened 2018-05-10
Reword "system is not yet available" to "Attempting to connect"

wblock:
Make this message more clear about what is going on and what might cause connection problems.


erinix:
This message will eventually show up on TrueNAS too so perhaps try wording it without "FreeNAS" in the name

wblock:
I considered that, but being too generic won't help the user much.  Investigating whether there is some kind of NAS type variable that can be used here to avoid hardcoded assumptions.

erinix:
There is but if the websocket service isn't available there is no way the UI will know which it is

dlavigne:
Now that there is an associated ticket, please also remove the other 2 lines from the message as they are irrevelant.

erinix:
please merge master to bring in the build fix

erinix:
LGTM

PR #719, opened 2018-05-10
Fix mat-checkbox alignment on Mozilla

dkmullen:
Ticket: #33366


erinix:
Using &nbsp might work but it is kind of janky (does it still work ok in all other browsers?).  A css approach would be preferable

erinix:
LGTM

erinix:
rebuild

PR #731, opened 2018-05-14
Reviwed Sharing -> NFS tooltips:

sambonham:
-Replaced "volume" references with "pool."
-Replaced "checked"  references with "set"
-Cleaned up the formatting so all the tooltip lines are aligned with the first line. 
- Tested the build.


erinix:
LGTM

PR #734, opened 2018-05-15
Bug: 33567 Add regex to not allow invalid characters in Jail Names

miker:
Only Aa-Zz, 0-9, -, and _ are allowed in Jail names
The Jail Wizard allowed all characters
Added the regex /^[a-zA-Z0-9-_]+$/ to only allow valid chars

This closes bug 33567


miker:
I triple checked this regex works as expected.
Hope this PR is ok.

miker:
This bug was resolved in https://github.com/freenas/webui/pull/748

PR #735, opened 2018-05-15
Add confirm dialogs to options to switch to legacy UI

dkmullen:
Ticket: #33426


erinix:
This doesn't work on mine, I tried building it and hotpatching my box but the sending to /legacy/ doesn't work.  I think the reason is that you are trying to use the router which would try to send the user to /ui/legacy rather than /legacy since the router only works within the app itself.  Try using window.location.href = /legacy/ instead.

erinix:
rebuild

erinix:
LGTM

PR #737, opened 2018-05-15
Fix 33042

dszidi:
Gets rid of dashboard console message generated by system_manufacturer value being null in widgetsysinfo. Also removed unwanted console messages and added debug options to message bus so basic event tracing can be done.


erinix:
LGTM

PR #746, opened 2018-05-16
Fix/sort issues pool manager 33558

dkmullen:
- Replaced lodash string sorting with 'natural' sorting
- Removed lodash from the two files since I'm 99% sure my stuff was the only thing using it
- Unable to duplicate any problem with Suggest Disk function


erinix:
You should put the sorter into a service, perhaps the storage service so that you don't have reused code

erinix:
LGTM

PR #747, opened 2018-05-16
reflect parent dataset values

vaibhav-rbs:
please test


erinix:
Looks good, you shouldn't need any of those things where you delete the INHERIT now (or the remaining commented out INHERITs anymore though

erinix:
LGTM

PR #748, opened 2018-05-16
Add jail jail name validation

miker:
Also added the import for the regexValidator


miker:
I have no idea how the jail-wizard PR got added to this PR...

PR #750, opened 2018-05-16
fix for VM wizard hard-disk creation step.

vaibhav-rbs:


erinix:
you shouldn't remove the loader.close from the vm error, otherwise the loader will never stop spinning

PR #762, opened 2018-05-18
Clarify wording and default choices in pool wizard

wblock:


erinix:
I see what you are doing there and it looks good, one thing though you should set isNew = true in order for the "Create Pool" button to work properly, it still defaults to false so the button still defaults to next instead of create pool

erinix:
LGTM

PR #767, opened 2018-05-18
Rename Volumes to Pools for table views on three Tasks pages:

Mrt134:
Also clean up some whitespace.
Local testing with NPM: no issues.


erinix:
LGTM

PR #785, opened 2018-05-23
Fix tour show logic in Account/Users, remove "you just closed User To…

wblock:
…ur" annoybot


erinix:
please merge master to bring in the build fix

erinix:
rebuild

erinix:
the showTour !== "false" seems to make the tour show up every time I click into that page.

wblock:
> the showTour !== "false" seems to make the tour show up every time I click into that page.

That is what it did for me with "true" and new user profiles (my Selenium setup starts with a new profile each time).  This might be an interaction with the preferences thing.  Which should be a button anyway.  People aren't going to watch the tour every time they log in.  At most, they will want to see it once when it is new to them.  I'll try to figure out what this test really means.

erinix:
rebuild

PR #791, opened 2018-05-24
Fix/pool screen 33345

dkmullen:
This table now duplicates (mostly) the legacy UI as requested in the ticket.
- Added Comments, Compression Ratio, renamed two others
- Inherited properties (compression, readonly, etc) show as such except...
- Inherited comments do not repeat (comments can be added to any pool)
- All data fields are filled in from the db (which wasn't the case before0 EXCEPT...
- Compression ratio is NOT being filled in, and I can't find it (or the means to calc it) on the db object!!!
- There's a question in the ticket I can't answer:

"Path" isn't in the legacy UI, so what is it supposed to represent? Shouldn't it start with "/mnt"?


erinix:
This looks good although I almost think we should remove path since it is not in the old UI and it it is redundant information and it makes this table too wide after you've expanded a few datasets and forces you to scroll horizontally on some displays

PR #794, opened 2018-05-24
Fix/pools detach dialog 33717

dkmullen:
This branch is one possible answer to tickets 33717 AND 33372. I need some guidance on which way to go.

Another one that I have been working on is dialog only - a good idea, but requires changes to the standard dialog box that I'm still trying to figure out: fix/dialog-only-pools-detach-33717 (not complete, but working up to a point)

The one I started with is no dialog at all: fix/detach-pool-warning-33372


erinix:
Lola recently introduced an entity-dialog which I think would be a good way to make this page into a dialog as well.  See her new pool status page to see how that is done.

PR #799, opened 2018-05-25
Fix/defaults show before data 33858

dkmullen:
Current behavior for tables and forms is to show defaults (such as "No data" or empty fields) until data loads and is filled in.This can be misleading. So this branch...
- Hides defaults until data requests complete - for entity tables, forms, tasks and also Pools list page
- Includes a brief timeout for the entity docs after which the defaults load even if the data isn't back. This timeout is the max. time the display will wait. If data comes back quicker, the page loads quicker. 
- Adds a loader (spinner) on the pools page, since it often takes noticeable time to load.

### DIFF
--- a/src/app/pages/common/entity/entity-form/entity-form.component.html
+++ b/src/app/pages/common/entity/entity-form/entity-form.component.html
@@ -4,7 +4,7 @@
 </p>
 <mat-card class="form-card">
     <div class="mat-content">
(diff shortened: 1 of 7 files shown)


erinix:
Looks good but you should also make sure you do a loader.close on the error conditions as well as success otherwise the loader will never stop spinning.

PR #802, opened 2018-05-25
Fix/sort pool list 33801

dkmullen:
Used the natural sorting function in storage service to sort pool list. Also, noticed that pools in the manager weren't sorted on page- load and fixed this. Tested this before upload but unable to do final test because of trouble running the current build.


erinix:
LGTM

PR #805, opened 2018-05-26
Add various validation improvements to the pool manager

erinix:
Ticket: #33336
            #28139
            #33333
            #32511


erinix:
Please test the heck out of this PR, these improvements will hopefully prevent lots of weird pool configurations and make pool setup easier but it is a big change so it might mess it up too.

PR #809, opened 2018-05-29
Update unclear field name and tooltip:

Mrt134:
- Update the "Home Directory" name and related tooltip to better reflect the field's functionality/role in the task.
- NPM testing: no issues


lola27:
LGTM

PR #823, opened 2018-05-30
Improve prompts and tooltips

wblock:


erinix:
rebuild

src/app/pages/vm/vm-wizard/vm-wizard.component.ts
```diff
@@ -245,9 +245,9 @@ export class VMWizardComponent {
           type: 'upload',
           name: 'upload_iso',
           placeholder : 'ISO upload location',
-          tooltip: 'Browse to the .iso file and click <b>Upload</b>.',
+          tooltip: 'Browse to the installer image file and click <b>Upload</b>.',
           isHidden: true,
-          acceptedFiles: ',.iso',
+          acceptedFiles: '.img,.iso',
```
vaibhav-rbs: @wblock did you test this change ?

wblock: Yes, tested with and without the leading comma, worked without, listing both .iso and .img files in the file dialog.

PR #826, opened 2018-05-31
Add compressratio data to pool list

dkmullen:
Ticket: #34342


erinix:
rebuild

erinix:
I am getting this error '''dataset_data2[k].compressratio is undefined''', you probably should check to see if that field is there before trying to do something with it

PR #828, opened 2018-05-31
Fix/dialog only pools detach 33717

dkmullen:
- Everything seems to work without breaking other entity dialog boxes
- I'm thinking I should make snackbars into dialogs and have dynamic info maybe?
- I know someone will look carefully to see that detach detaches and destroys and not vice  versa!


erinix:
Looks good, "DownloadKey" should probably be "Download Key" also is the wording for "if there is no passphrase" correct?  I think it should say something like "if the passphrase is lost" or something, the current wording makes it sound like if you never gave it a passphrase it won't be importable which is not true.

erinix:
rebuild

erinix:
rebuild

erinix:
rebuild

erinix:
rebuild

erinix:
rebuild

erinix:
It looks like there is actually a build failure in here on my side ```ERROR in ./node_modules/raw-loader!./node_modules/postcss-loader/lib?{"ident":"embedded","sourceMap":true}!./node_modules/sass-loader/lib/loader.js?{"sourceMap":true,"precision":8,"includePaths":[]}!./src/assets/styles/themes/ix-blue.scss
Module build failed: 
}
^
      Invalid CSS after "}": expected "}", was ""
      in /Users/erin/devel/webui/src/assets/styles/themes/ix-blue.scss (line 573, column 2)```

src/app/pages/storage/volumes/volumes-list/volumes-list.component.ts
```diff
@@ -230,13 +228,94 @@ export class VolumesListTableConfig implements InputTableConf {
...
+              name: 'pool_detach_warning',
+              paraText: T("WARNING: You are about to detach '" + row1.name + "'. \
+                Detaching a pool makes the data unavailable. \
+                Be sure that you understand the risks.\
+                In addition to detaching the pool, you may also choose to destroy its data."),
+              isHidden: false
+            }, {
+              type: 'paragraph',
+              name: 'pool_detach_warning',
+              paraText: T("'" + row1.name + "' is encrypted!. If the passphrase for \
```
erinix: The sentence should just end with a !, not !.

PR #831, opened 2018-05-31
Fixed breakage with device-edit form

dszidi:
The device-edit component was using only the template from entity-form and nothing else. When fieldsets were added to entity-form, this caused some breakage. Wasn't noticed since technically this component is not using entity-form, only the template. Copied scss and html files from entity-form so device-edit could safely have it's own modifications without causing trouble. Wrapped fieldConfig inside a fieldset so look and feel and features can be consistent with entity-form. At some point this and all other forms should be converted to entity-form.


erinix:
rebuild

erinix:
LGTM

PR #840, opened 2018-06-04
style buttons and selector on update page

dkmullen:
Ticket: #34348


erinix:
Your fix changes the Download update button back into a Check Now button.  We had changed the Check Now button into a Download update button because we decided to have the check happen when you enter the page and when you change train thus removing the need for a "Check" button.  (There's now no way to actually download updates)

PR #848, opened 2018-06-06
Fix/pool unlock dialog 31707

dkmullen:
Ticket: 31707
Replaces Pool Unlock form with a dialog to match Pool Lock dialog, duplicates function of the old form.


dkmullen:
rebuild

erinix:
If the old unlock is no longer used it should be removed

erinix:
LGTM

PR #854, opened 2018-06-06
Comment out idle detection and auto logout for now

erinix:
Ticket: #34243


vaibhav-rbs:
@rishabh27892 is testing this change with nightlies hot patched machine running with code compiled from this branch.

PR #856, opened 2018-06-07
Fix/vert align table headers 34585

dkmullen:
- aligns text vertically in entity table headers
- aligns pager in entity table footers
- sets all headers to same height as rows and footers for consistent appearance
- makes footer flexible in height to allow for pager expansion


erinix:
This looks good to me, what do you think @dszidi ?

erinix:
rebuild

dszidi:
Don't have a way to check it out on a working FreeNAS system ATM, but I don't see anything that would indicate any problems. LGTM

dkmullen:
rebuild

PR #858, opened 2018-06-07
fixed summary and validation; add T marker

lola27:


erinix:
We have a common IPv6 validator in the network service

erinix:
rebuild

PR #865, opened 2018-06-08
Fix 27133

lola27:


erinix:
Could you add a snackbar or something that tells me if it was successful?

erinix:
rebuild

PR #875, opened 2018-06-12
Fix issues in TFTP with non-usable exec permissions and no users in dropdown

erinix:
Ticket: #34783
            #34786


erinix:
rebuild

PR #879, opened 2018-06-13
Fix/entity cols 34077 34083

dkmullen:
Ticket: #34083
This branch simply changes true/false to yes/no in entity tables, which does help a little bit on crowded tables, in English anyway. I experimented with a horizontal scroll bar and headers that allow word-wrap but the results vary widely across all the entity tables.


erinix:
rebuild

erinix:
rebuild

src/app/pages/common/entity/entity-table/entity-table.component.ts
```diff
@@ -241,7 +241,13 @@ export class EntityTableComponent implements OnInit, AfterViewInit {
     for (let i = 0; i < rows.length; i++) {
       for (const attr in rows[i]) {
         if (rows[i].hasOwnProperty(attr)) {
-          rows[i][attr] = this.rowValue(rows[i], attr);
+          if (rows[i][attr] === true) {
+            rows[i][attr] = 'yes';
```
wblock: If we use "yes" and "no", I think a T() call is needed.  But checkboxes would be (or should be) consistent with the edit forms and don't need translation at all.

erinix: Use translate.get('yes') to accomplish the translated effect, if you need help finding that let me know.

PR #882, opened 2018-06-13
Attempt to fix the build with a older npm version

miwi-fbsd:


kmoore134:
Does that freenas/npm port exist now in ports tree?

miwi-fbsd:
@kmoore134 i took a downgraded port and moved it to freenas/npm and linked the webui to freenas/npm, as for now it seems to fix the build in jenkins

PR #887, opened 2018-06-14
Feature/modify entity table footer 33579

dkmullen:
Ticket: 33579
Modify the entity table footer with custom info along with info that shows/hides depending on whether rows are selected. Also fixed an alignment issue.


dkmullen:
rebuild

dkmullen:
rebuild

PR #897, opened 2018-06-15
Feat/selectable columns 34077

dkmullen:
Ticket: 34077
The ticket simply says that the User page is unreadable, suggests scrollbars. Instead, this PR offers a simplified version of the check box project seen earlier - check boxes to select columns on the three most crowded tables (User, Disks, Cron Jobs), some presets but no ability for the user to save new ones. I look at this as a possible short-term solution to at least make these pages usable.


dkmullen:
rebuild

PR #900, opened 2018-06-15
Clarify "select pool or dataset" field.

vaibhav-rbs:


vaibhav-rbs:
@lola27 your concern for ```No value accessor for form control with name: 'pool_detach_warning'``` was fixed as a part of #906, which is merged to master earlier in the today,

PR #904, opened 2018-06-18
Fix 34789

lola27:


erinix:
jails created by the wizard still have that box checked

PR #907, opened 2018-06-19
Add Periodic Snapshot Task setting to SMB Add

erinix:
Ticket: #34957


erinix:
This is awaiting some smb fixes on the middleware side to make creating shares work properly again

PR #913, opened 2018-06-20
Make filter function case insensitive

dkmullen:
Ticket: #33954
I think the filter works great - a few keystrokes usually gets you down to just a result or two. One improvement is that it should be case insensitive. We could easily make the filter ignore booleans, but I think this might cause more headaches than it solves.


dkmullen:
rebuild

dkmullen:
rebuild

PR #915, opened 2018-06-22
Fix 34735

lola27:


themylogin:
Please remove *Auxiliary arguments* from everywhere, it's been decided that users should not see this until https://redmine.ixsystems.com/issues/36633 is implemented

PR #918, opened 2018-06-22
Fix/remodel form explorer 35269

dkmullen:
Ticket: #35269
On form explorer, puts button next to the input field, deals with multiple layout issues when resizing the page, groups all the elements (including file tree) in a theme-specific color box on small screens when the input has to wrap below the button.


erinix:
rebuild

dkmullen:
Had to change this b/c the original solution failed on the New User form

erinix:
This one looks good but I am confused by the moving the browse button back under the file explorer, I thought we wanted it to be on top

dkmullen:
The ticket said that the arrangement we had didn't make it clear that things were grouped. I originally had the button beside the input form and the tree below, and it looked great on the wizards but not on forms where it is kinda crowded into one column. I moved it below, but I think it makes sense above too, and would be glad to put it there.

erinix:
rebuild

PR #930, opened 2018-06-26
Fix/click visual indicator 34624

dkmullen:
Ticket: 34624
After experimenting with different options, this fix adds only a simple visual effect on clicking a nav item. Similar to the Gmail sidnav, which doesn't do anything except show the click if you click on the current link.


dkmullen:
rebuild

PR #936, opened 2018-06-27
FIX 35797

lola27:


william-gr:
Can we hold on to merging this please?

I am thinking thatt option should not be available in UI, i will add more to the ticket.

PR #937, opened 2018-06-27
Fix 34618

lola27:
ticket: #34618 and ticket: #34822


erinix:
rebuild

erinix:
The breadcrumb for Jail / Add does not go anywhere now, Jail / Add / Wizard and Jail / Add / Advanced work but if I click on the intermediate Add it goes to a blank page

PR #950, opened 2018-07-02
Feature/color icon by theme

dkmullen:
Ticket: #36408
Colors the ix logo with the theme's primary color. This branch does pull in a new node module that pulls ix.svg inline in the DOM so it can be styled with CSS.


package.json
```diff
@@ -58,6 +58,7 @@
     "angular-calendar": "0.23.2",
     "angular-draggable-droppable": "^2.0.0",
     "angular-resizable-element": "^2.0.0",
+    "angular-svg-icon": "^5.1.1",
```
erinix: Remove the ^ from this line, looking at angular-svg-icon it seems newer versions are meant for angular 6, the ^ could introduce a non-working version of this dependency.

erinix: Remove the ^ from this line, looking at angular-svg-icon it seems newer versions are meant for angular 6, the ^ could introduce a non-working version of this dependency.

erinix:
rebuild

dkmullen:
rebuild

dkmullen:
rebuild

PR #957, opened 2018-07-03
do not allow allow users to shrink zvol with GUI.

vaibhav-rbs:


vaibhav-rbs:
I have included fix for https://redmine.ixsystems.com/issues/35221 in this ticket as well as its a simple text change.

PR #977, opened 2018-07-11
Fix 37478

erinix:


erinix:
rebuild

src/app/pages/task-calendar/cron/cron-form/cron-form.component.ts
```diff
@@ -61,38 +61,40 @@ export class CronFormComponent implements AfterViewInit {
...
          {
            type: 'scheduler',
            name: 'cron_picker',
            placeholder: T('Schedule a Cron Job'),
            tooltip: T('Choose one of the convenient presets\
              or choose <b>Custom</b> to trigger the advanced scheduler UI'),
-             required: true
+           validation: [ Validators.required ],
+           required: true,
+           value: "0 0 * * *"
```
vaibhav-rbs: if you change `value` to from `0 0 * * *` to `59 23 * * *` , I see that I am not getting `Custom Preset` which it should. I think this should be fixed

erinix:
rebuild

PR #986, opened 2018-07-12
Remove ellipsis on overflow to allow data to show

dkmullen:
Ticket: #37666 (also addresses #37453). 

Removes ellipsis on text overflow for data tables.

With horizontal scroll bars on data tables, there are few reasons to keep the ellipsis on text overflow, and the ellipsis is causing a new problem - making data invisible until user scrolls over - on TrueOS, at least). 

Ticket 37453 reports shifting check boxes, which I haven't been able to duplicate but which the ticketer says is solved by removing the ellipsis (which was a known issue with the ellipsis in the past).

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "angular-calendar": "0.23.2",
     "angular-draggable-droppable": "^2.0.0",
     "angular-resizable-element": "^2.0.0",
-    "angular-svg-icon": "^5.1.1",
+    "angular-svg-icon": "5.1.1",
     "angular-tree-component": "7.1.0",
     "angular2-markdown": "2.2.2",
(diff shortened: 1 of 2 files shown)


erinix:
rebuild

package.json
```diff
@@ -58,7 +58,7 @@
     "angular-calendar": "0.23.2",
     "angular-draggable-droppable": "^2.0.0",
     "angular-resizable-element": "^2.0.0",
-    "angular-svg-icon": "5.1.1",
+    "angular-svg-icon": "^5.1.1",
```
erinix: remove dat ^

erinix:
rebuild

erinix:
rebuild

PR #998, opened 2018-07-17
Add option to unlock pool using recovery key and hide actions that cannot be accessed when pool is locked

erinix:
Ticket: #35793
        #35789


vaibhav-rbs:
rebuild

PR #1017, opened 2018-07-19
Fixed widgetnetinfo html template

dszidi:
Fixed widgetnetinfo html template so it only shows information for the Primary NIC (the interface the UI is connected to.)


erinix:
Remove those console.logs and it is good

PR #1020, opened 2018-07-19
Remove screen saver choice from sys-advanced

dkmullen:
Ticket: #37423
Removes screen saver option from System > Advanced


erinix:
rebuild

PR #1023, opened 2018-07-19
Put a label over the swap size field on system > advanced

dkmullen:
Ticket: #37846
Adds a label of sorts over the swap size input box on System > Advanced


erinix:
rebuild

erinix:
rebuild

erinix:
rebuild

PR #1035, opened 2018-07-20
Hide collapsed toggle when sidenav is offscreen to keep it from getting out of sync

dkmullen:
Ticket: #39095
Fix for glitch in sidenav when it goes from collapsed state to full - the bug is that it covers up the main toggle button. This fixes that and also hides the little arrow button when it isn't useful, to keep it from getting the menu out of sync.


erinix:
rebuild

erinix:
rebuild

PR #1054, opened 2018-07-25
Fix 39428

vaibhav-rbs:


vaibhav-rbs:
rebuild

erinix:
This should just work now after my other fix and no longer need to look at "yes" and "no" (which never would have worked anyway because once you change language they will be whatever that language's version of "yes" and "no" are)

PR #1066, opened 2018-07-27
Feature/change ok buttons 32758

dkmullen:
Ticket: #32758
Changes 'Ok' to an action word, changes styling of buttons


dkmullen:
rebuild

erinix:
sorry man, that ix-blue just gets you every time :(

PR #1076, opened 2018-07-31
Add a rule to get checkboxes on SMART tests to pick up theme color

dkmullen:
Ticket: #40206
Adds a CSS rule to color 'pseudo-checkboxes' which appear without the right theme color in SMART Tests


erinix:
rebuild

erinix:
rebuild

PR #1084, opened 2018-08-02
add a short delay to scroll bar initializing

dkmullen:
Ticket: #37188
Fix an issue with the scroll bar not showing up at first by slightly delaying the initialization


erinix:
rebuild

PR #1085, opened 2018-08-02
Update System:Advanced tooltips and fields

miker:
Fields with updated tooltips:
Show Text Console without Password Prompt
Swap Size in GiB
Enable autotune
Enable Debug Kernel
Show advanced fields by default
MOTD Banner
Periodic Notification User
Use FQDN for logging
Report CPU usage in percentage
ATA Security User
SED Password

Added text "SED (Self-Encrypting-Drive) Options" above the SED options
This is to help clarify what a SED is.

webui built with no errors
Redmine: https://redmine.ixsystems.com/issues/37606


erinix:
rebuild

PR #1087, opened 2018-08-03
Correct System System Dataset and Tunables breadcrumbs

miker:
Ticket for change:
https://redmine.ixsystems.com/issues/40656


miker:
webui built without errors

PR #1093, opened 2018-08-06
System -> Tunable tooltips

miker:
Tooltips updated:
Variable
Value
Type drop-down
Comment field
Enabled checkbox

webui built with no errors

https://redmine.ixsystems.com/issues/37606

### DIFF
--- a/src/app/pages/system/tunable/tunable-form/tunable-form.component.ts
+++ b/src/app/pages/system/tunable/tunable-form/tunable-form.component.ts
@@ -36,7 +36,13 @@ export class TunableFormComponent {
       type: 'input',
       name: 'tun_var',
       placeholder: T('Variable'),
-      tooltip: T('The name of the sysctl or driver to load.'),
+      tooltip: T('The name of the loader, sysctl, or rc.conf variable\
+                  to configure.<br>\
+                  <b>loader</b>: Tunables that can be set <b>only</b> at\
+                  boot and not later.<br>\
+                  <b>rc.conf</b>: Enable or disable system services and\
+                  daemons.<br>\
+                  <b>sysctl</b>: Tunables that can be set anytime.'),
       required: true,
       validation : [ Validators.required ]
     },
@@ -44,11 +50,13 @@ export class TunableFormComponent {
       type: 'textarea',
       name: 'tun_value',
       placeholder: T('Value'),
-      tooltip: T('Set a value for the variable. Refer to the man page\
-                  for the specific driver or the <a\
-                  href="https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/"\
-                  target="_blank">FreeBSD Handbook</a> for suggested\
-                  values.'),
+      tooltip: T('Set a value to use for the <a\
+                  href="https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/boot-introduction.html#boot-loader-commands"\
+                  target="_blank">loader</a>, <a\
+                  href="https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/configtuning-sysctl.html"\
+                  target="_blank">sysctl</a>, or <a\
+                  href="https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/config-tuning.html"\
+                  target="_blank">rc.conf</a> variable.'),
       required: true,
       validation : [ Validators.required ]
     },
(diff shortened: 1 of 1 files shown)


src/app/pages/system/tunable/tunable-form/tunable-form.component.ts
```diff
@@ -36,50 +36,61 @@ export class TunableFormComponent {
...
     {
       type: 'input',
       name: 'tun_comment',
       placeholder: T('Comment'),
-      tooltip: T('Optional. Enter a substantive description or\
-                  explanation of this tunable.'),
-    },
-    {
+      tooltip: T('Enter a description of the tunable. What the tunable\
+                  does and why it\'s needed is helpful for future\
```
Mrt134: > it\'s

Recommend replacing or fully writing out the contraction.

miker: fixed

src/app/pages/system/tunable/tunable-form/tunable-form.component.ts
```diff
@@ -36,50 +36,60 @@ export class TunableFormComponent {
       type: 'input',
       name: 'tun_var',
       placeholder: T('Variable'),
-      tooltip: T('The name of the sysctl or driver to load.'),
+      tooltip: T('Enter the name of the loader, sysctl, or rc.conf\
+                  variable to configure.<br>\
+                  <b>loader</b>: Tunables that can be set only at boot\
```
wblock: It's not when they are set but when they take effect.  Also, this switches back and forth to telling when a setting takes effect to what it does.  It should really do both.  We should avoid hard-coded line breaks in tooltips, and also bold.  Is it loader, or loader.conf?
"<i>loader</i> tunables are for... and only take effect after a restart. <i>rc.conf</i> tunables are for enabling system services and daemons, and only take effect after a reboot. <i>sysctl</i> tunables are for... and generally take effect immediately."

"any time" is two words.

src/app/pages/system/tunable/tunable-form/tunable-form.component.ts
```diff
@@ -36,50 +36,60 @@ export class TunableFormComponent {
...
       type: 'select',
       name: 'tun_type',
       placeholder: T('Type'),
-      tooltip: T('Creating or editing a <i>Sysctl</i> immediately\
-                  updates the <b>Variable</b>. A reboot is required when\
-                  a <i>Loader</i> or <i>rc.conf</i> value changes.\
-                  A tunable remains at boot and across upgrades unless\
-                  it is deleted or <b>Enabled</b> is unset.'),
+      tooltip: T('Creating or editing a <i>sysctl</i> immediately\
+                  updates the <b>Variable</b> to the configured\
```
wblock: Remove bold.

src/app/pages/system/tunable/tunable-form/tunable-form.component.ts
```diff
@@ -36,50 +36,60 @@ export class TunableFormComponent {
...
+                  <b>Value</b>. When using a <i>loader</i> or\
```
wblock: Simplify and avoid pauses: "A restart is required to apply <i>loader</i> or <i>rc.conf</i> tunables. Configured tunables remain in effect until deleted or <i>Enabled</i> is unset."

src/app/pages/system/tunable/tunable-form/tunable-form.component.ts
miker: The intended use of set was when they take effect. Rereading the tooltip I can see the potential confusion especially with how we use set/unset for checkboxes. I'll reword this using your suggestions.

The bold was added before we discussed to eradicating the bolditis. I'll remove the bold. 

loader and sysctl are the variables, sysctl.conf and loader.conf are where the variables are set. I don't think it makes sense to talk about sysctl.conf or loader.conf here as the user isn't adding a new sysctl.conf or loader.conf, they are adding a new sysctl or loader variable and value. Also they are adding the variable using the webui, not to the respective conf. 
rc.conf is a tad different as one isn't adding an "rc", they enable/disable services and options to services. Changing rc.conf to "system service or system service options" would not only be confusing it's a tad wordy. I was going to say link to the user guide but we don't talk about rc.conf specifically there so I'd say link to the FreeBSD man page here. rc.conf is also well known and established. I'd argue if a user doesn't know what rc.conf is they maybe shouldn't be editing this based on the tooltip description. If enabling/disabling a service, or service options, calling this rc.conf would make the section easier to find and also how the options to set here would be explained if one was searching the internet for assistance.

sysctl, loader, and rc.conf are the options that were already in the tooltip and used in the user guide as well.

I know you don't like the use of hard-coded line breaks. I used them purposely here to visually separate the info. Without the line breaks the info ran together and was hard to quickly see there are multiple terms being explained. I'm not using the linebreak to force wrapping of a sentence but to visually separate the distinct items.

miker: when the variable is applied is specified/described in the Type tooltip.

src/app/pages/system/tunable/tunable-form/tunable-form.component.ts
miker: done

src/app/pages/system/tunable/tunable-form/tunable-form.component.ts
miker: done.
Removed the bold and the italics from Enabled. Filed names are identified with capitalization, no italics needed.

src/app/pages/system/tunable/tunable-form/tunable-form.component.ts
miker: Updated this section to be more paragraph like rather than item:description.

PR #1099, opened 2018-08-07
display correct VM summary when existing HDD is used for vm creation.

vaibhav-rbs:


src/app/pages/vm/vm-wizard/vm-wizard.component.ts
```diff
@@ -312,9 +312,29 @@ export class VMWizardComponent {
       ( < FormGroup > entityWizard.formArray.get([2])).get('memory').valueChanges.subscribe((memory) => {
         this.summary[T('Memory')] = memory + ' Mib';
       });
+
       ( < FormGroup > entityWizard.formArray.get([3])).get('volsize').valueChanges.subscribe((volsize) => {
         this.summary[T('Hard Disk Size')] = volsize + ' GiB';
       });
+
+      ( < FormGroup > entityWizard.formArray.get([3])).get('disk_radio').valueChanges.subscribe((disk_radio)=>{
+        if(this.summary[T('Hard Disk')] || this.summary[T('Hard Disk Size')]){
```
erinix: Using a translatable field as a key looks dangerous, do you know for a fact that it will still work if "Hard Disk" is translated?

vaibhav-rbs: I can take out T marker for harddisk if you want

PR #1123, opened 2018-08-09
Fix spelling error

Mrt134:


src/app/pages/plugins/plugin-add/plugin-add.component.ts
```diff
@@ -339,7 +339,7 @@ export class PluginAddComponent implements OnInit {
             }
           );
         } else {
-          this.snackBar.open(T("Plugin sucessfully installed"), T("Close"), { duration: 5000 });
+          this.snackBar.open(T("Plugin successfully installed"), T("Close"), { duration: 5000 });
```
wblock: Do we need "successfully" at all?  I mean, if it installed, it was successful.

Mrt134: fixed

PR #1139, opened 2018-08-14
reword legacy ui dialog

dkmullen:
Ticket: #39383


dkmullen:
The commented out version of onGoToLegacy() is more like the original, but I wonder if the active version is too simple?

PR #1149, opened 2018-08-15
edit more buttons into material compliance

dkmullen:
Ticket: #37258
Failed testing - this commit picks up more buttons


dkmullen:
Commit 2d2d977 'flattens' all raised buttons except the example buttons on the Custom theme page. Also, it renders all buttons in Caps. A few pages (VM, Services) have a three-button 'toggle group' at the top, which kind of looks like the raised buttons, but it is a separate kind of component and I'm not sure if it should be rendered like regular buttons.

PR #1158, opened 2018-08-16
remove `/mnt/` when new disk is created.

vaibhav-rbs:
please create a new vm requesting to create a disk as this case was missed in previous fix,


erinix:
rebuild

erinix:
rebuild

PR #1181, opened 2018-08-21
Fix 39732

dkmullen:
Ticket: #39732
Fixes for cron custom setup screen. Actually, the requested fix was already done, probably by the CSS changes yesterday.


erinix:
rebuild

PR #1205, opened 2018-08-28
Update VM Disk Device help text:

Mrt134:
Also add basic help text for the new Device Order field for each VM Device.
NPM testing: no issues.


src/app/pages/vm/devices/device-edit/device-edit.component.ts
```diff
@@ -254,14 +258,17 @@ export class DeviceEditComponent implements OnInit {
         {
           name : 'DISK_sectorsize',
           placeholder : 'Disk sector size',
-          tooltip : 'Enter the sector size in bytes. The default <i>0</i>\
-                     leaves the sector size unset.',
+          tooltip : 'A fixed amount of data that is the minimum storage\
+                     unit of a hard drive. Enter the sector size in\
```
wblock: Reverse the two sentences about the default and entering a value, or it sounds like the user is commanded to enter a value.  "the default is the best to use. if a custom size is needed, enter the value in bytes..."

PR #1216, opened 2018-08-31
do not call split on non string items, if it's empty send an empty ar…

vaibhav-rbs:
…ray.


vaibhav-rbs:
rebuild

erinix:
rebuild

erinix:
rebuild

erinix:
rebuild

erinix:
rebuild

PR #1221, opened 2018-09-04
display Alerta when power| turn  off VM's.

vaibhav-rbs:


src/app/pages/vm/vm-cards/vm-cards.component.ts
```diff
@@ -522,12 +523,22 @@ export class VmCardsComponent implements OnInit,OnDestroy {
     }
      else {
       if(poweroff){
-        eventName = "VmPowerOff";
-      } else {
-        eventName = "VmStop";
+         this.off_text =  `Do you wish to Power Off ${vm.name} ?`;
```
wblock: Please do not use "you", and no reason to capitalize "power off" specially.  Also, set off the name of the VM so the user can see it is a name:
`` this.off_text =  `Turn off power to virtual machine "${vm.name}"?`; ``

src/app/pages/vm/vm-cards/vm-cards.component.ts
```diff
@@ -522,12 +523,22 @@ export class VmCardsComponent implements OnInit,OnDestroy {
...
+      }
+      else {
+        this.off_text =  `Do you wish to Turn Off ${vm.name} ?`;
```
wblock: As above, although I don't see what the difference is between "power off" and "turn off".  I suspect this second one should just be "stop".
`` this.off_text = Stop virtual machine "${vm.name}"?;` ``

src/app/pages/vm/vm-cards/vm-cards.component.ts
```diff
@@ -523,12 +524,14 @@ export class VmCardsComponent implements OnInit,OnDestroy {
     }
      else {
       if(poweroff){
-         this.off_text =  `Do you wish to Power Off ${vm.name} ?`;
+         this.off_text =  `Power off to ${vm.name} ?`;
```
wblock: Please see my earlier comment on this, which included the suggested code that addresses the problems this still has, like the extraneous "to", the lack of delimiters around the vm name, and the extraneous space before the question mark.

src/app/pages/vm/vm-cards/vm-cards.component.ts
```diff
@@ -523,12 +524,14 @@ export class VmCardsComponent implements OnInit,OnDestroy {
...
+         this.off_button_text = 'Power off';
       }
       else {
-        this.off_text =  `Do you wish to Turn Off ${vm.name} ?`;
+        this.off_text =  `Stop virtual machine ${vm.name} ?`;
```
wblock: As above, I pasted the actual code to be copied.

src/app/pages/vm/vm-cards/vm-cards.component.ts
vaibhav-rbs: what `delimiters` do you want ?

src/app/pages/vm/vm-cards/vm-cards.component.ts
vaibhav-rbs: what delimiters do you want ?

PR #1236, opened 2018-09-14
45849: Update help text for Sharing/NFS/Add/_Path_ field:

Mrt134:
NPM build test: no issues.


src/app/pages/sharing/nfs/nfs-form/nfs-form.component.ts
```diff
@@ -29,8 +29,9 @@ export class NFSFormComponent {
       formarray: [{
         name: 'path',
         placeholder: T('Path'),
-        tooltip: T('Browse to the pool or dataset to be shared. Click\
-                    <b>Add extra path</b> to select multiple paths.'),
+        tooltip: T('The required full path to the pool or dataset to\
```
wblock: "The required full path" is a weird construction.  "The" doesn't add anything.  "required" can be either at the beginning or the end, and can be "mandatory" or punctuated to be less of an aside:
"Required: full path to the pool or dataset to share."

PR #1240, opened 2018-09-18
Fix/console 46711

dkmullen:
Ticket: 46711
Make pop-up console much wider and taller by default, and responsive to window size


lola27:
@dkmullen This PR works well except for one thing, the dialog expands once the console message is ready. Is that possible to make the width of console dialog fixed when it pops up?

PR #1249, opened 2018-09-20
explicitly set cursor focus in shell to true

dkmullen:
Ticket: #43778


vaibhav-rbs:
I think cursor is always in focus, like solid, its not blinking is this behaviour expected ?

dkmullen:
Yes, cursor blink is set to false. I prefer blinking. Also, focus is currently the default behavior anyway, but it produces a warning if you don't set it to true that default will change in the next version.

PR #1255, opened 2018-09-24
Added VM menu item to restart VM

dszidi:


lola27:
@dszidi it looks good to me except the console.logs. could you remove those?

PR #1257, opened 2018-09-24
Fix/ca identifier 46194

dkmullen:
Ticket: #46194
Provides a live error check and immediate error message as user types into the Identifier field. The error message is simply the allowed characters, same as before.

### DIFF
--- a/src/app/pages/system/ca/ca-add/ca-add.component.ts
+++ b/src/app/pages/system/ca/ca-add/ca-add.component.ts
@@ -320,12 +320,9 @@ export class CertificateAuthorityAddComponent {
     })
 
     entity.formGroup.controls['name'].statusChanges.subscribe((res) => {
-      if (this.identifier === '' && res === 'INVALID') {
-        _.find(this.fieldConfig).hasErrors = false;
-      } else if (this.identifier && res === 'INVALID') {
+      if (this.identifier && res === 'INVALID') {
         _.find(this.fieldConfig).hasErrors = true;
-      }
-      else {
+      } else {
         _.find(this.fieldConfig).hasErrors = false;
       }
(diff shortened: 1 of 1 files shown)


src/app/pages/system/ca/ca-add/ca-add.component.ts
```diff
@@ -27,7 +27,9 @@ export class CertificateAuthorityAddComponent {
       placeholder : T('Identifier'),
       tooltip: T('Enter a description of the CA.'),
       required: true,
-      validation : [ Validators.required ]
+      validation : [ Validators.required, Validators.pattern('[A-Za-z0-9_-]+$') ],
+      hasErrors: false,
+      errors: 'Use alphanumeric characters, "_" and "-".',
```
wblock: The quotes can imply that quote marks can be used in the field.  Suggested rewording:
` errors: 'Allowed characters: letters, numbers, underscore (_), and dash (-).', `

vaibhav-rbs:
rebuild

src/app/pages/system/ca/ca-add/ca-add.component.ts
```diff
@@ -312,6 +314,12 @@ export class CertificateAuthorityAddComponent {
         }
       }
     })
+
+    // Live error check on Identifier (name) field
+    entity.formGroup.controls['name'].statusChanges.subscribe((res) => {
+      res === 'INVALID' ? _.find(this.fieldConfig).hasErrors = true : _.find(this.fieldConfig).hasErrors = false;
```
lola27: an empty value is also invalid for this field, but the error message is not correct.

lola27:
rebuild

PR #1262, opened 2018-09-25
Fix/vm card adjust 48063

dkmullen:
Ticket: #48063
Move content up on slim cards (VM) and change positioning of three-dot icon to fix Safari display issue


vaibhav-rbs:
rebuild

PR #1264, opened 2018-09-25
Fix Vm Cards Edit and Save

dszidi:
In the VmCardsComponent some extra properties were added to the VmProfile type that were meant for UI use only. These were being sent to middleware and causing errors. There is now a method that strips these properties before sending new VmProfile data to middleware.


vaibhav-rbs:
rebuild

vaibhav-rbs:
rebuild

PR #1265, opened 2018-09-25
if vmclone fails display error message.

vaibhav-rbs:


dszidi:
LGTM

dlavigne:
The error message is not grammatically correct, it also does not give an indication of why it failed. Is it possible to give the user a reason why? If not, the message should read "VM clone failed."

PR #1271, opened 2018-09-27
Fix/update page 39977

dkmullen:
Ticket: #39977
Rearranges the update page
Gives a custom message depending on whether the train is stable or not
Only shows the update buttons if there is an update available


lola27:
rebuild

dkmullen:
Nightly build: Also would include a changelog card if there is a changelog.
screenshot from 2018-09-29 19-03-06

No updates available
screenshot from 2018-09-29 19-25-30

The message on a stable update.
screenshot from 2018-09-29 19-15-36

PR #1275, opened 2018-09-28
constrain description size and width of display

dkmullen:
Ticket: #48523
Limit size of description on VM cards, and limit display size too.


dkmullen:
New design, puts button at top and Description at the bottom. Damian should probably sign off on this.

screenshot from 2018-09-29 16-45-05

dkmullen:
The problem with the current layout:
screenshot from 2018-10-01 11-08-25

PR #1294, opened 2018-10-03
Fix sort by date 49084

dkmullen:
Ticket: #49084
Adds sort by date function that works with dates on BootEnv

### DIFF
--- a/src/app/services/storage.service.ts
+++ b/src/app/services/storage.service.ts
@@ -93,6 +93,7 @@ export class StorageService {
     
     sorter = bytes.concat(kbytes, mbytes, gbytes, tbytes)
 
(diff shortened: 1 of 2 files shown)


src/app/services/storage.service.ts
```diff
@@ -123,14 +126,28 @@ export class StorageService {
...
+         sorter.push(moment(elem).format('l LT'));
+        }
+      }
     else {
       sorter = tempArr.sort(myCollator.compare);
     }
       // Rejoins the sorted keys with the rest of the row data
       let v;
       // ascending or decending
+      console.log(sorter)
```
lola27: Could please remove the console.log?

PR #1296, opened 2018-10-03
Fix/iocage in treelist 48790

dkmullen:
Ticket: #48790
For the Pool list, this gives iocage a label to make sure it takes the right place in the tree


william-gr:
Wait, you guys are using filesystem listing to show the pools UI listing? why?

PR #1309, opened 2018-10-05
Fixed labels on virtual memory chart in vm-summary in table view.

dszidi:
Also removed cpu uage and pool usage charts since they were not appropriate and there wasn't enough room to accommodate them anyway. Changed legend location in viewchartdonut. You can now set legendPosition property to place legend to top or to the right.


dszidi:
This PR was created before the multiple PR requirement. Please also merge in PR 1353
for the branch  `FIX-40844-stable`

PR #1310, opened 2018-10-05
Fix/certs 47246

dkmullen:
Ticket: 47246
Renames passphrase fields to match API, thus fixing inability to submit CAs and Certs with passphrases

### DIFF
--- a/src/app/pages/system/ca/ca-add/ca-add.component.ts
+++ b/src/app/pages/system/ca/ca-add/ca-add.component.ts
@@ -194,17 +194,17 @@ export class CertificateAuthorityAddComponent {
     },
     {
       type : 'input',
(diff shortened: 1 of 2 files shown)


src/app/pages/system/ca/ca-add/ca-add.component.ts
```diff
@@ -341,6 +341,13 @@ export class CertificateAuthorityAddComponent {
     } else {
       data.san = _.split(data.san, ' ');
     }
-  }
 
+    // Addresses non-pristine field being mistaken for a passphrase of ''
+    if (data.Passphrase == '') {
```
lola27: data.Passphrase should be data.passphrase

PR #1311, opened 2018-10-05
Fix/certs 47246

dkmullen:
Ticket: 47246
Renames passphrase fields to match API, thus fixing inability to submit CAs and Certs with passphrases


lola27:
reopened this pr https://github.com/freenas/webui/pull/1388

PR #1315, opened 2018-10-05
FIX 45233

lola27:


lola27:
Thanks Dennis. pushed a fix commit.

PR #1324, opened 2018-10-09
Adjust method for adding up capacity of striped array

dkmullen:
Ticket: #48966
Adjusts method for calculating capacity of stripe only, assuming that the stripe option means using the real capacity of all disks combined, unlike mirror or raidz.


dkmullen:
https://github.com/freenas/webui/pull/1377
is same PR, for 11.2 stable

PR #1330, opened 2018-10-09
Fix/50431

dkmullen:


dkmullen:
https://github.com/freenas/webui/pull/1350
is matching pr for stable

PR #1338, opened 2018-10-10
Fixed text issue on ssh config screen (11.2-stable).

a-stjohn:


dlavigne:
Root should definitely not be uppercase as it is a lowercase username. Why the uppercase in Password?

PR #1340, opened 2018-10-11
Fix 45222 stable

lola27:


vaibhav-rbs:
rebuild

vaibhav-rbs:
rebuild

lola27:
rebuild

PR #1353, opened 2018-10-15
Duplicate of PR 1309

dszidi:
Fixed labels on virtual memory chart in vm-summary in table view. Also removed cpu uage and pool usage charts since they were not appropriate and there wasn't enough room to accommodate them anyway. Changed legend location in viewchartdonut. You can now set legendPosition property to place legend to top or to the right.


dszidi:
I'm going to delete this one and create a new branch based off of freenas/11.2 like we discussed

PR #1356, opened 2018-10-16
Removes arrows from number fields for moz and webkit

dkmullen:
Ticket: #35272
Uses CSS to remove up/down arrows on number input fields


dkmullen:
https://github.com/freenas/webui/pull/1398
is the matching PR for 11.2 stable

PR #1377, opened 2018-10-19
vdev adding for stable branch

dkmullen:
Ticket: #48966


dkmullen:
https://github.com/freenas/webui/pull/1324
is the same PR, for master

PR #1389, opened 2018-10-23
Update page further enhancements

dkmullen:
Ticket: #40940

- In this branch, the update page train selector displays only trains that are declared 'allowed' by update.component.ts. 
- If only the current branch is allowed, its name/description displays, no selector. 
- Also, descriptions from the response object are now displayed in the selector;
- The wording of the descriptions will be changed in trains.txt
- Plus, some style changes.


dkmullen:
https://github.com/freenas/webui/pull/1404
is matching pr for 11.2

PR #1391, opened 2018-10-23
Add warning and link to Legacy UI to jails and installed plugins

dkmullen:
Ticket: #47598


dkmullen:
https://github.com/freenas/webui/pull/1402
is the same PR, for 11.2 stable

PR #1392, opened 2018-10-24
Added a tooltip to warn users 

sambonham:
to avoid using uppercase usernames when creating an SMB home share. 
This is in response to bug #45387.
Passed build test with no errors.


src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -44,7 +44,13 @@ export class SMBFormComponent implements OnDestroy {
       placeholder: T('Use as home share'),
       tooltip: T('Set to allow this share to hold user home\
                   directories. Only one share can be\
-                  the <i>homes</i> share.')
+                  the <i>homes</i> share.\
```
wblock: s/homes/home/

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -44,7 +44,13 @@ export class SMBFormComponent implements OnDestroy {
...
+                  <b>Note:</b> Samba converts %u to the lowercase\
```
wblock: This is confusing to the end user.  Where does %u come from?  It has not been mentioned elsewhere.

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -44,7 +44,13 @@ export class SMBFormComponent implements OnDestroy {
...
+                  username, so a directory name with an uppercase\
+                  letter will not match. If an uppercase username\
```
wblock: Will not match what?

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -44,7 +44,13 @@ export class SMBFormComponent implements OnDestroy {
...
+                  was used, create a new user with a lowercase\
+                  <b>Username</b> in <b>Accounts --> Users</b>,\
```
wblock: I thought we were not using bold in tooltips any more.  We should not be, anyway.

Creating a new user with the same name but lowercase will cause confusion.

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -44,7 +44,13 @@ export class SMBFormComponent implements OnDestroy {
...
+                  and try creating the home share again.'),
```
wblock: Avoid using "try" when telling users what to do, it suggests that this might not work even if they do it correctly.

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -44,13 +44,18 @@ export class SMBFormComponent implements OnDestroy {
...
-                  was used, create a new user with a lowercase\
-                  <b>Username</b> in <b>Accounts --> Users</b>,\
-                  and try creating the home share again.'),
+                  the home share.\
+                  Note: Lower case names for user home\
+                  directories are strongly recommended, as Samba\
+                  maps usernames to all lower case. For example, the\
+                  username John will be mapped to a home directory\
+                  named john. If the <i>Path</i> to the home share\
+                  includes an upper case username, delete the existing user,\
```
wblock: No comma after "user".

PR #1393, opened 2018-10-24
adjust safeUrl settings

dkmullen:
Ticket: #44641
Removes the specific reference to 'http' to perhaps make our url more flexible; This will not, however, let an unencrypted link through a security filter.


dkmullen:
https://github.com/freenas/webui/pull/1405
is matching PR for stable

PR #1408, opened 2018-10-24
ctl-tpc report label changed to match old UI. SCSI Target Port (tpc)

dszidi:
(cherry picked from commit 4dc172c7b6fe076c7ff0137a677fe4107e7f65da)

### DIFF
--- a/src/app/components/common/lineChart/lineChart.service.ts
+++ b/src/app/components/common/lineChart/lineChart.service.ts
@@ -347,7 +347,7 @@ export class LineChartService {
         });
 
         let divideBy: number;
-        let title: string = prop;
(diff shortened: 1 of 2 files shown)


src/app/components/common/lineChart/lineChart.service.ts
```diff
@@ -347,7 +347,7 @@ export class LineChartService {
         });
 
         let divideBy: number;
-        let title: string = prop;
+        let title: string = prop == "ctl-tpc" ? "SCSI Target Port (tpc)" : prop; // Put in ugly override. Wasn't really a better place for this one change.
```
amotin: I am not sure how other target ports named there in the new UI, but if there is some flexibility, it would be nice to decode the "tpc" here as Third-Party Copy, i.e. traffic caused by XCOPY operations.

PR #1429, opened 2018-10-26
use new middleware call jail.restart to restart jail/plugin

lola27:


lola27:
Please don't merge this one till we start to work on 11.3 tickets.

PR #1430, opened 2018-10-26
use new middleware call jail.restart to restart jail/plugin

lola27:


lola27:
close this pr since middleware change only for 11.3.

PR #1434, opened 2018-10-28
Fix 28870

dszidi:
Fixes overloaded disks tab by providing the user with controls. The user can now specify which disks and which metrics they'd like shown via a new report options UI.


dszidi:
I have one more commit coming. I just need to the the form fields to required in the fieldConfig.

PR #1451, opened 2018-10-29
do not call filesystem.statfs when gui value undefined or empty

vaibhav-rbs:
(cherry picked from commit d7f57b14e79d3fb43000a6b0631f9ae49f596eef)


erinix:
rebuild

PR #1461, opened 2018-10-30
Added an example for the authorized hosts and ip addresses field.

sambonham:
Passed build test with no issues.
Related to: https://redmine.ixsystems.com/issues/23923#change-362826.
Added a correct example of an IP address/hostname to the Authorized Hosts and IP addresses field in response to a customer adding an incorrect address  (192.168.2.1/24).


dlavigne:
We should use a better example as that one is ambigous (it could be a host address, it could be a network address). Also, does this field require the CIDR mask? (You will probably have to sync with VB to get him to explain what validation his regex enforces) Finally, rather than making up an IP, we should use an RFC1918 address, with Class C (192.168.1.x) being the least ambiguous.

PR #1463, opened 2018-10-31
Fix 54420

dszidi:
Fixes issues with general preferences form. Form now loads the current preference values for field values. Saved preferences now persist after logging out and back in.


erinix:
rebuild

PR #1473, opened 2018-11-01
fixed firefox UI websocket disconnet.

vaibhav-rbs:
On initial first time load, If preferences does exist return
after saving Preferences so that UI can retry.

remove redundant Download Sucessful.


erinix:
rebuild

erinix:
rebuild

vaibhav-rbs:
rebuild

PR #1477, opened 2018-11-02
QA-Test tkt-54945 fix tests/runtest.py grid option

ericbsd:


tests/runtest.py
```diff
@@ -110,19 +107,22 @@
...
     from driverG import webDriver
     print("Running Selenium Grid")
     runDriver = webDriver(grid_server_ip)
-
 else:
-    if (driver_v == "U"):
+    if driver_v == "F":
         from driverU import webDriver
-        print ("Running Ubuntu driver")
+        print ("Running webDriver driver")
```
rishabh27892: driverU is for Ubuntu driver

ericbsd: Actually, I meant Firefox I forgot to change it.

ericbsd: Actually, driverU should support all system. I did some change.

PR #1482, opened 2018-11-04
Remove tooltip stating password must be <= 9 chars

remyjette:
The validator forcing the max password length to <= 9 characters was
removed in 94e555c96f17bcb10f466948f5f1203ebea619ff, but the tooltip
was not. This change adjusts the tooltip accordingly.


remyjette:
Opened #1483 against `freenas/11.2-stable`.

PR #1495, opened 2018-11-06
Clarified the docs by removing the fullstop at the end of the 

sambonham:
characters list and the A before the $ symbol.
Sent these text changes to Jeff E and he approved them. 
Passed build test.


erinix:
rebuild

PR #1504, opened 2018-11-08
Remove redundant "SAVE CONFIG" button in System/Update/Manual Update

vaibhav-rbs:
show Save Config dailoge when user clicks  INSTALL MANUAL UDPATE FILE


vaibhav-rbs:
please do npm run build:prod:aot build

PR #1581, opened 2018-11-28
Dszidi redo reports

dszidi:
Reports now use the new entity-toolbar in place of entity-form and tabbed navigation. Charts also have min/max/avg etc in addition to the old legend


vaibhav-rbs:
LGTM

PR #1598, opened 2018-12-03
Set the default theme to ix-dark in preferences.service. This should …

dszidi:
…ensure that ix-dark is used on a clean install


erinix:
rebuild

erinix:
rebuild

PR #1631, opened 2018-12-12
Sign-in card background image improvements; fixes #63630

davidjb:
This puts the background image actually behind the logo (it was in front
before, obscuring the logo) as well as a significant reduction in file
size.  The stars bg image was originally 6000x4000px, taking up 4.8MB of
download when the displayed size is set at 400px.  This adds optional
sized images for different display resolutions, saving up to 99% (eg was
4.8MB, now 22KB).

Lastly, the accessibility of the background image is improved as well,
now having more detailed alt text.

See issue: https://redmine.ixsystems.com/issues/63630 for further details.  

Before (4.8MB):
before-4 8mb

After (22KB, with z-index fixed):
after-22kb

PS This PR was opened against what I think is the right branch, but I'm guessing.  `master` seems to be behind `11.2-stable` and I don't know the relationship between this and other `ux-*` named branches.  Perhaps contributing guidelines (https://redmine.ixsystems.com/issues/27829) could be added so contributing is easier and more obvious?


erinix:
rebuild

PR #1635, opened 2018-12-13
Fix issue with saved value not being populated in cron permissions component, fix tracebacks and remove console.logs

erinix:
Ticket: #60057


erinix:
rebuild

erinix:
rebuild

erinix:
rebuild

PR #1637, opened 2018-12-13
FIX 63483 STABLE

lola27:


src/app/pages/system/CloudCredentials/cloudcredentials-form/cloudcredentials-form.component.ts
```diff
@@ -149,6 +149,38 @@ export class CloudCredentialsFormComponent {
         }
       ]
     },
+    {
+      type: 'checkbox',
+      name: 'skip_region-S3',
+      placeholder: T('Endpoint does not support regions'),
```
wblock: The negative logic makes this confusing.  Can we reverse the logic and meaning of the variable?
`` placeholder:  T('Endpoint Supports Regions'), ``

src/app/pages/system/CloudCredentials/cloudcredentials-form/cloudcredentials-form.component.ts
```diff
@@ -149,6 +149,38 @@ export class CloudCredentialsFormComponent {
...
+            name: 'provider',
+            value: 'S3',
+           }]
+        }
+      ]
+    },
+    {
+      type: 'checkbox',
+      name: 'signatures_v2-S3',
+      placeholder: T('Use v2 signatures'),
```
wblock: `Use V2 Signatures`

src/app/pages/system/CloudCredentials/cloudcredentials-form/cloudcredentials-form.component.ts
erinix: The logic is coming from the middleware so it would not be something we can/should invert on our side.  After investigating what this checkbox does I propose it say something like "Skip region autodetect", how does that sound to you?

wblock: Works for me. :)

PR #1638, opened 2018-12-13
Fix 63234

lola27:


src/app/pages/plugins/plugin-advanced-add/plugin-advanced-add.component.ts
```diff
@@ -0,0 +1,1232 @@
...
+      const config = this.formFileds[i];
+      if (config.relation.length > 0) {
+        this.setRelation(config);
+      }
+    }
+
+    this.formGroup.controls['dhcp'].valueChanges.subscribe((res) => {
+			console.log(res);
+			
+      if (res) {
```
wblock: Seems like this type of construct would be easier to read with inline ifs and maybe just overriding defaults.  For example, this replaces that whole if/else block:
```
if (res) {
  this.formGroup.controls['vnet'].setValue(true);
  this.formGroup.controls['bpf'].setValue(true);
}
_.find(this.basicfieldConfig, { 'name': 'vnet' }).required = (res ? true : false);
_.find(this.basicfieldConfig, { 'name': 'bpf' }).required = (res ? true: false);
```
Could be there is an easier/shorter way in TS/JS to convert res to a boolean, which would be even better.  Some quick searching suggests `!!res`, but that's not very readable.

src/app/pages/plugins/plugin-advanced-add/plugin-advanced-add.component.ts
```diff
@@ -0,0 +1,1232 @@
...
+		value['props'] = property;
+		value['name'] = this.plugin_name;
+		console.log(value);
+
+		// only for plugin bru-server
+    if (this.plugin_name == 'bru-server') {
+      value['accept'] = true;
+    }
+		
+		this.dialogRef = this.dialog.open(EntityJobComponent, { data: { "title": T("Install") }, disableClose: true });
```
wblock: Indentation?

src/app/pages/plugins/plugin-advanced-add/plugin-advanced-add.component.ts
```diff
@@ -0,0 +1,1232 @@
...
+    this.dialogRef.componentInstance.setDescription(T("Installing plugin..."));
+    this.dialogRef.componentInstance.setCall(this.addCall, [value]);
+    this.dialogRef.componentInstance.submit();
+    this.dialogRef.componentInstance.success.subscribe((res) => {
+      this.dialogRef.componentInstance.setTitle(T("Plugin installed successfully"));
```
wblock: Use title case for dialog titles: Plugin Installed Successfully

PR #1639, opened 2018-12-13
Fix 63234 stable

lola27:


wblock:
Please see my review of #1638

PR #1642, opened 2018-12-14
Update "Save Config" dialog and add it to the button in System/General

Mrt134:
Cherry picked the commit from PR1570 to port to 11.2-stable branch.
Could not get local NPM or Yarn builds to function, there is some kind of error with the 'node-sass' module that won't allow either to build. This is not related to my changes.


erinix:
rebuild

erinix:
rebuild

PR #1669, opened 2018-12-21
30495 tooltips for 11.3

dkmullen:
Ticket: #30495
Allows user to temporarily pin tooltips (and unpin them), for reading longer ones without having to hold the mouse in place. Tooltips reset to normal on refresh.


erinix:
rebuild

erinix:
I really like this piece of functionality, having it able to keep the tooltip open while you type in a field is a very nice touch.  One thing though, it looks like the button is getting cut off, at least on entity-form based things like account/users
tooltip

dkmullen:
Wasn't able to duplicate the issue, but I think it will solve it to move right-aligned tips over a bit to keep them out of the padding
screenshot from 2019-01-03 15-10-23

erinix:
rebuild

erinix:
rebuild

PR #1670, opened 2018-12-21
Changed the 'Wait to boot' tool tip to be clearer.

a-stjohn:


src/app/helptext/vm/devices/device-add-edit.ts
```diff
@@ -107,8 +107,8 @@ vnc_port_tooltip : T('Can be set to <i>0</i>, left empty for FreeNAS to\
  fixed, preferred port number.'),
 
 wait_placeholder : T('Wait to boot'),
-wait_tooltip : T('Set for the VNC client to wait until the VM has\
- booted before attempting the connection.'),
+wait_tooltip : T('Enable this setting to restrict VNC connections until\
```
wblock: "Set" is better terminology.  But also, this setting does not restrict connections.  We should improve it by changing the placeholder also.  What it seems to do is really: "Wait to connect with VNC until the VM is done booting."  The placeholder should say something similar but shorter.

PR #1684, opened 2019-01-02
63636 update page for 11.2 stable

dkmullen:
Sets the train of current OS as default, to fix issue where config file from older version causes wrong train options to load

Adds ‘MINOR_UPGRADE’ and ‘MAJOR_UPGRADE’ to the legit upgrade list, which seems consistent with 11.1

Fixes it so if user switches to another train, but then switches back before actually updating, the display resets and then reloads properly

Moves the first check for updates to later in ngOnInit, AFTER the value of this.trains is set; otherwise, the update check runs on a default value which can be wrong, at least when you have loaded a config file from an older version.

### DIFF
--- a/ports/freenas/freenas-webui/Makefile
+++ b/ports/freenas/freenas-webui/Makefile
@@ -14,7 +14,8 @@ EXTRACT_ONLY=
 
 FETCH_DEPENDS=	npm:freenas/npm5 \
 		git>0:devel/git \
-		python2>0:lang/python2
+		python2>0:lang/python2 \
+		gmake:devel/gmake
 BUILD_DEPENDS=	rsync>0:net/rsync \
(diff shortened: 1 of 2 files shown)


ports/freenas/freenas-webui/Makefile
```diff
@@ -14,7 +14,8 @@ EXTRACT_ONLY=
 
 FETCH_DEPENDS=	npm:freenas/npm5 \
 		git>0:devel/git \
-		python2>0:lang/python2
+		python2>0:lang/python2 \
+		gmake:devel/gmake
```
erinix: looks like this snuck into your PR

dkmullen: I'll replace this with:
https://github.com/freenas/webui/pull/1697

PR #1699, opened 2019-01-04
position and size tooltips to stay within margins

dkmullen:
Ticket: #62550 and duplicates
Positions tooltips and dynamically resizes some of them to keep all of them within the margins of the card where they originate; to fix a bug where tooltips sometimes get cut off at the margin or padding.
screenshot from 2019-01-04 11-53-53


erinix:
rebuild

PR #1707, opened 2019-01-08
Fix 67420 stable

vaibhav-rbs:


src/app/core/services/api.service.ts
```diff
@@ -283,8 +283,19 @@ export class ApiService {
         namespace:"vm.delete",
         args:[],
         responseEvent:"VmDeleted"
-      }
+      },
+      async preProcessor(def:ApiCall, self) {
+        return await self.dialog.confirm("Delete", `Delete ${def.args[1].name} ?`).toPromise().then((res)=>{
```
wblock: It would be nice to help the user verify what they are deleting (Delete VM {name}), but I don't know if this is a VM-specific routine or just generic.

vaibhav-rbs: screen shot 2019-01-08 at 10 49 19 am
this is the current behaviour, in which user are told in a dialog box that they are deleting a specific Virtual Machine.(FreeNASRC2 in this case)

vaibhav-rbs: @wblock let me know if this is desired behaviour

wblock:
```suggestion
        return await self.dialog.confirm("Delete VM", `Delete VM ${def.args[1].name} ?`).toPromise().then((res)=>{
```

PR #1709, opened 2019-01-08
edit style for more visible error state

dkmullen:
Ticket #59490
Makes invalid fields more visible (we were accidentally overwriting the red color on the label)
The material spec for helping the visually impaired is to have the red label and underline, and maybe a warning message ('email is required', etc.) We could consider adding messages, as much for catching the eye as for their content.

screenshot from 2019-01-08 14-11-24

screenshot from 2019-01-08 09-04-17


erinix:
I like the current changes, I think we should investigate adding messages below the field for future versions too

PR #1718, opened 2019-01-09
Use websocket call instead of rest to get > 20 snapshot task entries and perform minor cleanup to replication form

erinix:
Ticket: #67879


erinix:
rebuild

PR #1734, opened 2019-01-10
Fix 59904 stable alternative solution

dszidi:
Since this is tricky to test I've left some debug messages in the signin component. Once the reviewer has verified this to be working I will remove the unwanted console messages for the final merging.


erinix:
rebuild

PR #1753, opened 2019-01-11
Fix 57780 stable

erinix:


src/app/helptext/storage/volumes/datasets/dataset-permissions.ts
```diff
@@ -33,4 +33,8 @@ dataset_permissions_mp_recursive_placeholder: T('Apply permissions recursively')
 dataset_permissions_mp_recursive_tooltip: T('Apply permissions recursively to all directories\
  and files within the current dataset'),
 
+dataset_permissions_dialog_warning: T('Warning'),
+dataset_permissions_dialog_warning_message: T('Warning: Changing a pool or dataset with Windows\
```
wblock: For the message (with Andrew's help):
`` Warning: Changing dataset permission mode can severely affect existing permissions, particularly when going from Windows to Unix permissions. ``

PR #1764, opened 2019-01-14
Basic IE support (master)

dszidi:
This PR does the following...

Loads polyfills only for IE. This fixes problem where legacy IE11 browsers can at least log in and navigate around. Browsers with native support for ES6 (everything that's not IE) don't load unnecessary polyfills and get a significant UI performance improvement (everything loads faster and is generally more responsive)

Fixes css variables fallback rules. Browsers that don't support CSS custom properties still won't get theme switching support, but will at least get the default theme colors.

SPECIAL NOTE: Although the Angular UI loads, there are significant problems with layout and css. Compensating for so many language features with polyfills results in a big performance hit in IE11. I would strongly recommend IE users fall back to legacy UI.


dszidi:
image

PR #1782, opened 2019-01-15
FIX 69760 stable

lola27:


src/app/pages/system/CloudCredentials/cloudcredentials-form/cloudcredentials-form.component.ts
```diff
@@ -889,6 +903,25 @@ export class CloudCredentialsFormComponent {
...
+        if (res === '') {
+          this.setFieldRequired('key_file-SFTP', true, entityForm);
+        } else {
+          this.setFieldRequired('key_file-SFTP', false, entityForm);
+        }
+      }
+    });
+    entityForm.formGroup.controls['key_file-SFTP'].valueChanges.subscribe((res) => {
+      if (res !== undefined) {
+        if (res === '' ) {
```
wblock: I think this can all be replaced with:
```
if (res !== undefined) {
    this.setFieldRequired('key_file-SFTP', (res === ''), entityForm);
}
```

PR #1790, opened 2019-01-16
Fixed trailing zeros charts on front page dashboard (stable)

dszidi:
Changed how stats service formats the start and end times for API call. Using epoch time previously led to null values and some sort of clock mismatch. Switching to 'now-xxx' syntax and limiting the end param to 20 seconds from present fixes things.

NOTE: requires latest middleware.


dszidi:
Sorry, just noticed some commits from another branch wound up in here somehow

PR #1793, opened 2019-01-17
Pool manager dialog button should say "extend pool" when extending and "create pool" when creating.

erinix:
Ticket: #69663


dkmullen:
rebuild

PR #1795, opened 2019-01-17
Fix-48295-datatables-adjustments

dkmullen:
Ticket: #48295
Adds min and max sizes (set by developer) to entity table
Sets a few col widths for Jails, Storage Snapshots, NTP Servers, User List, Avail PlugIns
Makes resize handles always visible, which I think makes resizing a much better experience

This branch doesn't address all the issues in the ticket, but some...
screenshot from 2019-01-17 11-05-46


erinix:
rebuild

erinix:
rebuild

dkmullen:
Replaced by PR 1991

PR #1812, opened 2019-01-21
Add some traditional Chinese words

jasoncheng7115:


src/assets/i18n/zh-hant.po
```diff
@@ -418,13 +418,13 @@ msgid "Account Name"
 msgstr "帳戶名稱"
 
 msgid ""
-"Account to be used for guest access. Default is nobody. Account is required "
+"Account to be used for guest access. ；Default is nobody. Account is required "
```
erinix: You appear to have changed one of the string keys, could you change this one back so it does not break this key?

erinix:
Looks good, if you want these changes to go into 11.2 you will also need to create a PR against the freenas/11.2-stable branch

jasoncheng7115:
> Looks good, if you want these changes to go into 11.2 you will also need to create a PR against the freenas/11.2-stable branch

Thanks for your guidance. I have submitted PR to the 11.2-stable.

PR #1822, opened 2019-01-23
Fix 71022

erinix:


erinix:
I created a separate ticket to do that in, going to merge this as is for the sake of keeping master in sync with stable for now https://redmine.ixsystems.com/issues/72089

PR #1829, opened 2019-01-24
FIX 71038

lola27:


src/app/pages/jails/jail-list/jail-list.component.ts
```diff
@@ -146,16 +146,17 @@ export class JailListComponent implements OnInit {
     },
   ];
 
+  public tooltipMsg: any = T("Choose a pool where the iocage jail manager \
```
erinix: this tooltip message should probably go in a helper file rather than being in the component itself

PR #1834, opened 2019-01-28
63483: Add help text to new Cloud Credentials/S3 fields:

Mrt134:
- Rename fields to Disable Enpoint Region and Use Signature Version 2.
- Add simple help text describing function and usage of these fields.
- Local yarn build test: no issues.


src/app/pages/system/CloudCredentials/cloudcredentials-form/cloudcredentials-form.component.ts
```diff
@@ -152,8 +152,10 @@ export class CloudCredentialsFormComponent {
     {
       type: 'checkbox',
       name: 'skip_region-S3',
-      placeholder: T('Skip Region Autodetect'),
-      // tooltip: T(''),
+      placeholder: T('Disable Endpoint Region'),
+      tooltip: T('Skips automatic detection of the Endpoint URL\
```
wblock:
```suggestion
      tooltip: T('Skip automatic detection of the Endpoint URL\
```

PR #1851, opened 2019-01-30
Fix 70824

erinix:


src/app/pages/plugins/plugins-installed/plugins-installed.component.ts
```diff
@@ -154,6 +155,7 @@ export class PluginsInstalledListComponent {
   public selectedPool;
   public activatedPool: any;
   public availablePools: any;
+  protected publicIp = '';
```
skarekrow: This isn't necessarily a public IP, it should probably just read as privateIP

erinix: but it is a public IP, it grabs it using a tool that queries whatismyip.com

skarekrow: Oops nevermind me, crossed this up with jail IP and asigra

PR #1860, opened 2019-01-31
Bump up copyright year to 2019

erinix:
Ticket: #72777


erinix:
> This really needs to be a constant so it only needs to be changed once and to avoid the possibility of missing some instances.

I agree, I'll make a separate ticket for that (will need testing to make sure it always does the right thing in scenarios like rebooting, etc)

PR #1865, opened 2019-02-01
Update fr.po

MDVAurelien:
Progress of the French translation


erinix:
If you want your changes to show up in an 11.2 release, you will need to bring your changes over to the freenas/11.2-stable branch as well FYI :)

PR #1869, opened 2019-02-01
Update fr.po

MDVAurelien:
Progression of the French translation


MDVAurelien:
Please do not accept this commit because there are errors on the transmitted file

PR #1882, opened 2019-02-07
add . as valid char for username

vaibhav-rbs:
#71281


mibux:
The dot char needs to be escaped for regex

[a-z_A-Z_][a-zA-Z0-9_\.-]*[$]?

would be usefull if this could be added to the group name as well.

mibux:
Sorry, this should be right. Had to escape the backslash as well.

[a-z_A-Z_][a-zA-Z0-9_\\.-]*[$]?

vaibhav-rbs:
@erinix I have applied changes please take a  look and approve.

PR #1883, opened 2019-02-07
Fix 59850

lola27:
please merge #1846 first.


lola27:
@wblock would you please review and provide the jailtype_tooltip? thanks

src/app/pages/jails/jail-add/jail-add.component.ts
```diff
@@ -53,6 +54,23 @@ export class JailAddComponent implements OnInit {
...
       parent: this
     },
+    {
+      type: 'select',
+      name: 'jailtype',
+      placeholder: helptext.jailtype_placeholder,
+      tooltip: helptext.jailtype_tooltip,
+      options: [
+        {
+          label: 'Default (clone jail)',
```
wblock:
```suggestion
          label: 'Default (Clone Jail)',
```

src/app/helptext/jails/jails-add.ts
```diff
@@ -6,6 +6,9 @@ uuid_placeholder: T('Name'),
 uuid_tooltip: T('Required. Can only contain alphanumeric characters \
 (Aa-Zz 0-9), dashes (-), or underscores (\_).'),
 
+jailtype_placeholder: T('Jail Type'),
+jailtype_tooltip: T(''),
```
wblock:
```suggestion
jailtype_tooltip: T('Select <i>Default</i> or <i>Basejail</i>. The default is \
for normal, cloned jails. A basejail is a read-only filesystem of a \
release shared by one or more normal jails.'),
```

skarekrow: We should drop the verbage normal. As the jails that share the base are not actually clone jails.

wblock: Okay, please expand.  What is a "default" type jail, and what do we call it other than "normal".  "Typical"?

skarekrow: Default are clone jails like the message says. Which is why saying normal sounds redundant. As for what else do we call it, I'm suggesting just dropping it. Saying it's shared by one or more jails is correct. As they could be a template, or a basejail.

wblock: We need to help the reader choose a type.  (For that matter, I'm not sure why we are adding this field to the UI.  I suspect other things need to be accessible to create and use a basejail.)

Anyway, we give the user a choice of two types of jails to create here.  We need to differentiate and explain what each type does.  We say that one is a basejail, but we also need to say that the other is the one they will use 99.9% of the time, and "standard" or "normal" or "typical" jail.  This avoids defining things negatively: "Basejail is probably not the one you want!"

What should this help text say?  At present, all I can come up with is "Default (always pick this) or Basejail (explanation of what basejails are goes here, which will only help to confuse the user)."

skarekrow: Right but the word normal doesn't help that. We are already telling the user that it is a clone jail. That is the default type. Normal doesn't apply to the basejail portion, and yeah that is all we need to create basejails, it's already present in the MW.

wblock: I'm still not able to write something clear describing this field.  There are two options, we should describe them both, because "Type" doesn't tell anything about the choice they are making here.

It sounds like my guessed description of basejail is okay.  How do we describe the Default choice, so the user knows that is what they should almost always choose?  Please try to write a sentence that could go in here, give me something to work from.

skarekrow:
```
Default (Clone Jail): These are clones of the specified RELEASE. They will forever be linked to this RELEASE, even if they are upgraded.

Basejails: Basejails mount the specified RELEASE directories as nullfs mounts over the jail's directories, these are not linked to the original RELEASE when upgraded.
```

Something more similar to that. The notion of 'normal' doesn't really apply to these, as it's odd for default jails and incorrect for basejails.

wblock: That helps!  I've rewritten the original help text.  Does it look reasonably accurate?

skarekrow: Yep!

wblock: Thank you!

erinix:
This PR has a build error, please fix it so I can merge

PR #1887, opened 2019-02-08
Remove redundant add option from vm list (wizard should let user pick what type they want)

erinix:
Ticket: #74011
(cherry picked from commit def4471e06c3b4262216dfea560eef2cbaf55b57)


wblock:
Forgot to add that I tested this.

PR #1894, opened 2019-02-08
FIX 72529

lola27:


src/app/pages/storage/snapshots/snapshot-list/snapshot-list.component.ts
```diff
@@ -151,7 +151,9 @@ export class SnapshotListComponent implements OnInit {
 
   doRollback(item) {
     const warningMsg = T("<b>WARNING:</b> Rolling back to this snapshot will permanently delete later snapshots of this dataset. Do not roll back until all desired snapshots have been backed up!");
-    this.entityList.dialogService.confirm(T("Warning"), warningMsg, false, T('Rollback')).subscribe(res => {
+    const msg = T("<br><br>You're attempting to roll back to an early snapshot, would you like to rollback the snapshot <b>") + item.snapshot_name + '</b>( ' + item.creation + ' )?';
```
Mrt134: As an alternative, place the affected item earlier in the message:
`Revert (item) back to (snapshot name)?`

wblock: From #1893 
```suggestion
    const msg = T("<br><br>Roll back to snapshot <i>") + item.snapshot_name + '</i> from ' + item.creation + '?';
```

PR #1896, opened 2019-02-08
FIX 73650 

lola27:


src/app/helptext/sharing/smb/smb.ts
```diff
@@ -70,6 +70,9 @@ export const helptext_sharing_smb = {
     tooltip_guestonly: T('Requires <b>Allow guest access</b> to also be set.\
  Forces guest access for all connections.'),
 
+    placeholder_abe: T('Access Based Share Enumeration'),
+    tooltip_abe: T(''),
```
Mrt134:
```suggestion
    tooltip_abe: T('Restricts share visibility to users with read or write access to the share. See <a href="https://www.freebsd.org/cgi/man.cgi?query=smb.conf" target=_blank>smb.conf(5)</a>.'),
```
Formatting might need cleanup.

wblock: Yes, but s/Restricts/Restrict/

PR #1915, opened 2019-02-14
 A 20% Simplified Chinese Translation

dingyifei:
for the 11.2-stable


dingyifei:
hmmmm I guess I need to fix this

PR #1937, opened 2019-02-20
FIX-75631-stable

dkmullen:
Ticket 75631
Display update percentage as integer not float. We apparently don't label file size or download rate on the front end, so this PR doesn't relabel MB to MiB, etc. as suggested in the ticket.


dkmullen:
We need at least two decimal places. Sometimes it takes a minute to get up to 1%, and at least you can see that something is happening. The fix returns at least one digit before the point, and a min/max of two after.

PR #1978, opened 2019-02-25
Fix 67359 stable

dkmullen:
Ticket: #67359
Adds a message to make it clear that password (and matching PW) are always needed to submit the form (if SMTP auth is selected), even when you are just editing.

### DIFF
--- a/src/app/pages/system/email/email.component.ts
+++ b/src/app/pages/system/email/email.component.ts
@@ -140,7 +140,7 @@ export class EmailComponent implements OnDestroy {
     },
     {
       type: 'paragraph',
-      name: 'email_form_password_message',
+      name: 'em_pwmessage',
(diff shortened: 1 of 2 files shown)


src/app/pages/system/email/email.component.ts
```diff
@@ -138,6 +138,11 @@ export class EmailComponent implements OnDestroy {
       required: true,
       validation : [ Validators.required ]
     },
+    {
+      type: 'paragraph',
+      name: 'em_pwmessage',
+      paraText: T('Matching passwords are REQUIRED to submit the form.'),
```
wblock:
```suggestion
      paraText: T('Matching passwords must be entered to submit the form, even when editing settings.'),
```

dkmullen:
Replaced by PR #1988

PR #1979, opened 2019-02-25
Fix 67359 master

dkmullen:
Ticket: #67359
Adds a message to make it clear that password (and matching PW) are always needed to submit the form (if SMTP auth is selected), even when you are just editing.


dkmullen:
replaced by pr 1989

PR #1983, opened 2019-02-25
FIX 75432

lola27:


src/app/helptext/task-calendar/cloudsync/cloudsync-form.ts
```diff
@@ -31,6 +31,11 @@ folder_tooltip: T('Enter the name of the destination folder.'),
 encryption_placeholder: T('Server Side Encryption'),
 encryption_tooltip: T('Choose <i>AES-256</i> or <i>None</i>.'),
 
+b2_chunk_size_placeholder:  T('Upload Chunk Size (in megabytes)'),
+b2_chunk_size_tooltip: T('Upload chunk size. Must fit in memory. Note that these chunks are buffered in \
+ memory and there might be a maximum of «--transfers» chunks in progress at once. Also, your largest file\
+ must be split in no more than 10 000 chunks.'),
```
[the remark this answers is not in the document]
Mrt134: I tested the escaped comma locally, and it does appear to work properly.

src/app/helptext/task-calendar/cloudsync/cloudsync-form.ts
```diff
@@ -31,6 +31,11 @@ folder_tooltip: T('Enter the name of the destination folder.'),
 encryption_placeholder: T('Server Side Encryption'),
 encryption_tooltip: T('Choose <i>AES-256</i> or <i>None</i>.'),
 
+b2_chunk_size_placeholder:  T('Upload Chunk Size (in megabytes)'),
```
wblock:
```suggestion
b2_chunk_size_placeholder:  T('Upload Chunk Size (MiB)'),
```

src/app/helptext/task-calendar/cloudsync/cloudsync-form.ts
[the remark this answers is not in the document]
wblock: Escaping the comma should not be necessary, it's inside quotes.

I don't understand all the details that this is trying to say.  "must fit in memory" seems unnecessary, unless it means the entire file.  Let's try to simplify the whole thing:
"Files are split into chunks of this size before upload. Up to «--transfers» chunks can be in progress at one time. The single largest file being transferred must fit into no more than 10,000 chunks."

src/app/helptext/task-calendar/cloudsync/cloudsync-form.ts
```diff
@@ -31,6 +31,11 @@ folder_tooltip: T('Enter the name of the destination folder.'),
 encryption_placeholder: T('Server Side Encryption'),
 encryption_tooltip: T('Choose <i>AES-256</i> or <i>None</i>.'),
 
+b2_chunk_size_placeholder:  T('Upload Chunk Size (MiB)'),
+b2_chunk_size_tooltip: T('Files are split into chunks of this size before upload.\
+ Up to «--transfers» chunks can be in progress at one time. The single largest file\
```
wblock: This «--transfers» string seems to be a literal.  Is this value set in another field?  How does the user find out what the value is?

themylogin: @wblock yes, this value is set in another field, in legacy UI it's here: https://github.com/freenas/freenas/blob/master/gui/tasks/forms.py#L45

lola27: @wblock The transfers field is already in new UI.

PR #2001, opened 2019-02-27
69625 (11.2-stable): Update Init/Shutdown Scripts > Add help text:

Mrt134:
(cherry-pick from PR 2000)
Sync help text for "Command", "Script", and "When" fields with updated docs.
PEP8 cleanup.

### DIFF
--- a/src/app/helptext/task-calendar/initshutdown/initshutdown.ts
+++ b/src/app/helptext/task-calendar/initshutdown/initshutdown.ts
@@ -4,25 +4,27 @@ import {Validators} from '@angular/forms';
 export default {
     ini_type_placeholder: T('Type'),
     ini_type_tooltip: T('Select <i>Command</i> for an executable or\
-                <i>Script</i> for an executable script.'),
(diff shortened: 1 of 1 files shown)


src/app/helptext/task-calendar/initshutdown/initshutdown.ts
```diff
@@ -4,25 +4,27 @@ import {Validators} from '@angular/forms';
 export default {
     ini_type_placeholder: T('Type'),
     ini_type_tooltip: T('Select <i>Command</i> for an executable or\
-                <i>Script</i> for an executable script.'),
+                         <i>Script</i> for an executable script.'),
```
erinix: the text to these should be as far to the left as they possibly can with only a single space indention (otherwise the string extractor will add the extra spaces)

Mrt134: Ah, ok. I didn't realize there was a change from the previous help text standard.

PR #2013, opened 2019-02-28
Fix 59904 master alt

dszidi:
Clears the browser cache when it detects an update. (Compares localstorage value against new PlatformInfo constant)


erinix:
rebuild

PR #2047, opened 2019-03-05
Update and merge zh-hans.po

dingyifei:
update from 11.2-stable branch's *yarn run extract* and merge translation from the master branch

Is the code freeze means a new release is on its way? 11.2 U3 or something?


erinix:
There is a possibility that the strings from master branch and stable branch are a 1 to 1 matching, if you want to do the extract I would recommend you do that from stable itself rather than master (you could just run the extract on this file on freenas/11.2-stable and it will merge/prune the strings).  Otherwise this looks good.

dingyifei:
I ran the extract from 11.2-stable and then merge the translation from on master branch I did. The msgid changes are from 11.2 stable. 
The 11.2-Stable branch’s .po files are all out of date xD.

PR #2059, opened 2019-03-07
Fix 77917

erinix:


src/app/pages/storage/volumes/datasets/dataset-form/dataset-form.component.ts
```diff
@@ -672,41 +672,65 @@ export class DatasetFormComponent implements Formconfiguration{
...
+            let sync_value = pk_dataset[0].sync.value;
+            if (pk_dataset[0].sync.source === 'DEFAULT') {
+              sync_value = 'INHERIT';
             }
-            else{
-              entityForm.formGroup.controls['compression'].setValue(pk_dataset[0].compression.value);
+            entityForm.formGroup.controls['sync'].setValue(sync_value);
 
+            let compression_value = pk_dataset[0].compression.value;
+            if (pk_dataset[0].compression.source === 'INHERITED') {
```
william-gr: I think it should be true if INHERITED or DEFAULT, same for all the others below.

PR #2061, opened 2019-03-07
Update endpoint URL to be less confusing.

a-stjohn:


src/app/helptext/system/cloudcredentials.ts
```diff
@@ -53,12 +53,11 @@ export const helptext_system_cloudcredentials = {
...
-      '<a href="https://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteEndpoints.html"\
- target="_blank">Endpoint URL</a> for a bucket\
- configured for website hosting. Leave blank when using \
- AWS. Enter endpoint URL if using custom S3 API. URL \
- general format: \
- <i>bucket-name.s3-website.region.amazonaws.com</i>.\
+      'Leave blank when using AWS. The available buckets are fetched \
+ dynamically. Enter a \
+ <a href="https://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteEndpoints.html"\
+ target="_blank">Endpoint URL</a> if using <i>custom</i> S3 API. URL \
```
wblock: s/if/when/

Not clear what "custom" means here.  Is it the contents of a field set elsewhere (might need to be capitalized to match, if so), or does it mean using your own S3 implementation rather than AWS?

PR #2083, opened 2019-03-12
Feat 59085 master

dkmullen:
Tickets 59085, 72347, and 73783
For CA Add and Certificate Add:
- Adds Organizational Unit field 
- Adds option to choose between RSA and EC key types
- Makes the form disable/enable fields based on key type selection
- Hides fields not needed for chosen cert or CA type

### DIFF
--- a/src/app/helptext/system/ca.ts
+++ b/src/app/helptext/system/ca.ts
@@ -27,7 +27,8 @@ export const helptext_system_ca = {
     key_type: {
       placeholder: T("Key Type"),
       tooltip: T(
-        'See <a href="https://crypto.stackexchange.com/questions/1190/why-is-elliptic-curve-cryptography-not-widely-used-compared-to-rsa"></a>\
+        'See <a href="https://crypto.stackexchange.com/questions/1190/why-is-elliptic-curve-cryptography-not-widely-used-compared-to-rsa" target="blank">\
+ Why is elliptic curve cryptography not widely used, compared to RSA?</a>\
  for more information about key types.'
       ),
       validation: [Validators.required]
@@ -36,8 +37,9 @@ export const helptext_system_ca = {
     ec_curve: {
       placeholder: T("EC Curve"),
       tooltip: T(
-        'Brainpool* curves can be more secure, while secp* curves can be faster. See\
- <a href="https://tls.mbed.org/kb/cryptography/elliptic-curve-performance-nist-vs-brainpool">\
+        'Brainpool curves can be more secure, while secp curves can be faster. See\
+ <a href="https://tls.mbed.org/kb/cryptography/elliptic-curve-performance-nist-vs-brainpool" target="blank">\
+ Elliptic Curve performance: NIST vs Brainpool\
  </a> for more information.')
     },   
 
--- a/src/app/helptext/system/ca.ts
+++ b/src/app/helptext/system/ca.ts
@@ -77,6 +77,11 @@ export const helptext_system_ca = {
       validation: [Validators.required]
     },
 
+    organizational_unit: {
+      placeholder: T("Organizational Unit"),
+      tooltip: T("Enter the name of the organizational unit."),
+    },
+
     email: {
       placeholder: T("Email"),
       tooltip: T(
--- a/src/app/helptext/system/ca.ts
+++ b/src/app/helptext/system/ca.ts
@@ -24,6 +24,23 @@ export const helptext_system_ca = {
       validation: [Validators.required]
     },
 
+    key_type: {
+      placeholder: T("Key Type"),
+      tooltip: T(
+        'See <a href="https://crypto.stackexchange.com/questions/1190/why-is-elliptic-curve-cryptography-not-widely-used-compared-to-rsa"></a>\
+ for more information about key types.'
+      ),
+      validation: [Validators.required]
+    },
+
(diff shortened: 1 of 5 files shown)


src/app/helptext/system/ca.ts
```diff
@@ -24,6 +24,23 @@ export const helptext_system_ca = {
...
+        'See <a href="https://crypto.stackexchange.com/questions/1190/why-is-elliptic-curve-cryptography-not-widely-used-compared-to-rsa"></a>\
+ for more information about key types.'
+      ),
+      validation: [Validators.required]
+    },
+
+    ec_curve: {
+      placeholder: T("EC Curve"),
+      tooltip: T(
+        'Brainpool* curves can be more secure, while secp* curves can be faster. See\
```
wblock: Remove the stars, this is likely to only confuse the reader rather than tell them to pattern match.
```suggestion
        'Brainpool curves can be more secure, while secp curves can be faster. See\
```

src/app/helptext/system/ca.ts
```diff
@@ -24,6 +24,23 @@ export const helptext_system_ca = {
       validation: [Validators.required]
     },
 
+    key_type: {
+      placeholder: T("Key Type"),
+      tooltip: T(
+        'See <a href="https://crypto.stackexchange.com/questions/1190/why-is-elliptic-curve-cryptography-not-widely-used-compared-to-rsa"></a>\
```
wblock: These links should have link text:
```suggestion
        'See <a href="https://crypto.stackexchange.com/questions/1190/why-is-elliptic-curve-cryptography-not-widely-used-compared-to-rsa">Why is elliptic curve cryptography not widely used, compared to RSA?</a>\
```

src/app/helptext/system/ca.ts
```diff
@@ -24,6 +24,23 @@ export const helptext_system_ca = {
...
+ for more information about key types.'
+      ),
+      validation: [Validators.required]
+    },
+
+    ec_curve: {
+      placeholder: T("EC Curve"),
+      tooltip: T(
+        'Brainpool* curves can be more secure, while secp* curves can be faster. See\
+ <a href="https://tls.mbed.org/kb/cryptography/elliptic-curve-performance-nist-vs-brainpool">\
```
wblock:
```suggestion
 <a href="https://tls.mbed.org/kb/cryptography/elliptic-curve-performance-nist-vs-brainpool">Elliptic Curve performance: NIST vs Brainpool\
```

src/app/helptext/system/certificates.ts
```diff
@@ -24,6 +24,23 @@ export const helptext_system_certificates = {
       validation: [Validators.required]
     },
 
+    key_type: {
+      placeholder: T("Key Type"),
+      tooltip: T(
+        'See <a href="https://crypto.stackexchange.com/questions/1190/why-is-elliptic-curve-cryptography-not-widely-used-compared-to-rsa"></a>\
```
wblock: Can't we just reuse the help string variable from above, rather than duplicating the contents?

src/app/helptext/system/certificates.ts
```diff
@@ -24,6 +24,23 @@ export const helptext_system_certificates = {
...
+ for more information about key types.'
+      ),
+      validation: [Validators.required]
+    },
+
+    ec_curve: {
+      placeholder: T("EC Curve"),
+      tooltip: T(
+        'Brainpool* curves can be more secure, while secp* curves can be faster. See\
+ <a href="https://tls.mbed.org/kb/cryptography/elliptic-curve-performance-nist-vs-brainpool">\
```
wblock: As above, let's try to avoid duplication and just reuse variables or fields.

src/app/pages/common/entity/entity-form/entity-form.component.ts
```diff
@@ -557,12 +557,15 @@ export class EntityFormComponent implements OnInit, OnDestroy, OnChanges, AfterV
     return this.fb.control({disabled, value}, validation);
   }
 
-  setDisabled(name: string, disable: boolean, hide: boolean = false, status?:string) {
+  setDisabled(name: string, disable: boolean, hide?: boolean, status?:string) {
     // if field will be hide, disabled it too
```
wblock:
```suggestion
    // if field is hidden, disable it too
```

src/app/pages/common/entity/entity-form/entity-form.component.ts
```diff
@@ -557,12 +557,15 @@ export class EntityFormComponent implements OnInit, OnDestroy, OnChanges, AfterV
...
     if (hide) {
       disable = hide;
+    } else {
+      hide = false;
```
wblock: Sorry, don't understand the reason for this.  hide is already false or the if would not get here.  Unless this is to set it to a particular known value (false rather than zero, say).

dkmullen: 'hide' was being set to 'false' as a parameter to setDisabled() which meant that the condition 'if(hide)' would never evaluate to true. So I made it an optional parameter, which means that if it isn't set in the component, then it is 'undefined' when it gets here, so I explicitly set it to 'false'.

We could eliminate all of these lines if we made sure every place calling setDisabled, passed in all the arguments.

PR #2088, opened 2019-03-13
FIX-80508-stable

dkmullen:
Ticket: #80508
Hides fields not in use on CA and Cert add pages

### DIFF
--- a/src/app/pages/common/entity/entity-form/entity-form.component.ts
+++ b/src/app/pages/common/entity/entity-form/entity-form.component.ts
@@ -557,10 +557,12 @@ export class EntityFormComponent implements OnInit, OnDestroy, OnChanges, AfterV
     return this.fb.control({disabled, value}, validation);
   }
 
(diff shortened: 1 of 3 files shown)


src/app/pages/common/entity/entity-form/entity-form.component.ts
```diff
@@ -557,10 +557,12 @@ export class EntityFormComponent implements OnInit, OnDestroy, OnChanges, AfterV
     return this.fb.control({disabled, value}, validation);
   }
 
-  setDisabled(name: string, disable: boolean, hide: boolean = false, status?:string) {
+  setDisabled(name: string, disable: boolean, hide?: boolean, status?:string) {
```
lola27: I think you don't need to change the setDisabled function.

erinix: Apparently this is how master already is, so merging it will make stable the same as master.

PR #2092, opened 2019-03-14
FIX-72387-stable

dkmullen:
Ticket: #72387
Remove DOS Charset as an option in SMB form


dkmullen:
This change isn't going into 11.2

PR #2095, opened 2019-03-14
FIX-59112-master

dkmullen:
Ticket: #59112
Makes a custom validator to prevent 'global' from being used as a SMB share name


src/app/helptext/sharing/smb/smb.ts
```diff
@@ -8,6 +8,7 @@ export const helptext_sharing_smb = {
     placeholder_path: T('Path'),
     tooltip_path: T('Select pool, dataset, or directory to share.'),
     validators_path: [ Validators.required ],
+    errormsg_name: T("'global' is a reserved section name. Please select another one"),
```
wblock:
```suggestion
    errormsg_name: T("<i>global</i> is a reserved name that cannot be used as a share name. Please enter a different share name."),
```

dkmullen:
Screenshot from 2019-03-21 11-43-46

PR #2103, opened 2019-03-15
FIX 81210

lola27:


src/app/helptext/sharing/iscsi/iscsi.ts
```diff
@@ -129,6 +129,9 @@ export const helptext_sharing_iscsi = {
  information.'
   ),
 
+  globalconf_placeholder_alua: T('Enable iSCSI ALUA'),
+  globalconf_tooltip_alua: T(''),
```
Mrt134: Draft suggestion:
Requires ALUA-capable, Highly Available (HA) hardware. Set for the initiator to discover paths to both storage controllers on the target and increase storage traffic efficiency.

wblock: Always tell what it is first, then tell requirements and other things like how to disable.  Also, our terminology is "High Availability". So:
"Set for the initiator to discover paths to both storage controllers on the target and increase storage traffic efficiency. Requires ALUA-capable, High Availability (HA) hardware."

PR #2115, opened 2019-03-20
Fix 82263 stable

dkmullen:
Ticket: #82263
Brings all TrueNAS logos up to date, formats the ones that always display. (May not need an 11.2 PR for this, but I made one since 11.2 has references to the old TN logos).


erinix:
rebuild

PR #2118, opened 2019-03-20
Fix 80112

lola27:


Mrt134:
Working with UI team to create initial help text.

src/app/helptext/system/ssh-connections.ts
```diff
@@ -0,0 +1,38 @@
...
+    host_tooltip: T(''),
+
+    port_placeholder: T('Port'),
+    port_tooltip: T(''),
+
+    username_placeholder: T('Username'),
+    username_tooltip: T(''),
+
+    private_key_placeholder: T('Private Key'),
+    private_key_tooltip: T(''),
```
Mrt134: Connection key used to authenticate with the remote system. Choose a private key generated from a local Certificate Authority.

src/app/helptext/system/ssh-connections.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    remote_host_key_placeholder: T('Remote Host Key'),
+    remote_host_key_tooltip: T(''),
```
Mrt134: Connection key generated from the remote system used to authenticate with this system. Copy the contents of the remote system private key and paste the contents in this field.

src/app/helptext/system/ssh-connections.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    cipher_placeholder: T('Cipher'),
+    cipher_tooltip: T(''),
```
Mrt134: From Replication Task Options table:

Encryption algorithm to secure the connection. <i>Standard</i> provides the best security. <i>Fast</i> is less secure, but has better transfer rates for devices with limited cryptographic speed. <i>Disabled</i> is for networks where the entire path between sources and destinations is trusted.

src/app/helptext/system/ssh-connections.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    connect_timeout_placeholder: T('Connect Timeout'),
+    connect_timeout_tooltip: T(''),
```
Mrt134: from ssh_config(5):

Time to continuously attempt connecting to the remote system. Enter a number of seconds.

src/app/helptext/system/ssh-connections.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    url_placeholder: T('FreeNAS URL'),
+    url_tooltip: T(''),
```
Mrt134: From Guide Example: FreeNAS to FreeNAS Semi-Automatic Setup section

IP address or hostname of the remote FreeNAS system.

src/app/helptext/system/ssh-connections.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    token_placeholder: T('Auth Token'),
+    token_tooltip: T(''),
```
Mrt134: From Guide Example: FreeNAS to FreeNAS Semi-Automatic Setup section

An replication authorization string that originates on the remote system. Copy a temporary replication token from the remote system and enter or paste the string in this field.

wblock:
@william-gr Can we please have some help with this?  We can make educated guesses for some of these fields, but it's slow and we need to get these PRs moving. 2119 is the same thing for another branch.

src/app/helptext/system/ssh-connections.ts
```diff
@@ -0,0 +1,38 @@
+import { T } from '../../translate-marker';
+
+export default {
+    name_placeholder: T('Name'),
+    name_tooltip: T(''),
```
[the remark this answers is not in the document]
william-gr: I dont think there is a hard requirement for it to be unique, the allowed chars you guessed are right.

src/app/helptext/system/ssh-connections.ts
william-gr: Private key is local, the keys are generated locally and the public key is what goes to the remote system.

src/app/helptext/system/ssh-connections.ts
william-gr: No, thats the host key, its the key of the remote ssh server. There should be a button to scan the host key of the remote node.

src/app/helptext/system/ssh-connections.ts
themylogin: This can mislead people. Nothing in our system "continuously attempts connecting", it is one `connect` call and that's it. I don't think we can and should explain what is a "connect timeout" in this tooltip, instead, just say "in seconds" or, better, have a widget with placeholder and remove tooltip completely.

src/app/helptext/system/ssh-connections.ts
themylogin: It's not just IP/hostname anymore, it's URL with `http://` or `https://` at the beginning

PR #2119, opened 2019-03-20
Fix 80121

lola27:


src/app/helptext/system/ssh-keypairs.ts
```diff
@@ -0,0 +1,14 @@
+import { T } from '../../translate-marker';
+
+export default {
+    name_placeholder: T('Name'),
+    name_tooltip: T(''),
+
+    private_key_placeholder: T('Private Key'),
+    private_key_tooltip: T(''),
```
wblock: Will the private key be visible after generation?
```suggestion
    private_key_tooltip: T('See <i>Public key authentication</i> in <a href="https://www.freebsd.org/cgi/man.cgi?query=ssh" target="_blank">SSH/Authentication</a>.'),
```

src/app/helptext/system/ssh-keypairs.ts
```diff
@@ -0,0 +1,14 @@
...
+
+    public_key_placeholder: T('Public Key'),
+    public_key_tooltip: T(''),
```
wblock:
```suggestion
    public_key_tooltip: T('See <i>Public key authentication</i> in <a href="https://www.freebsd.org/cgi/man.cgi?query=ssh" target="_blank">SSH/Authentication</a>.'),
```

src/app/pages/system/ssh-keypairs/ssh-keypairs-form/ssh-keypairs-form.component.ts
```diff
@@ -0,0 +1,107 @@
+import { Component } from '@angular/core';
```
wblock: Why are these "keypairs" going in System--aren't they for replications?  For that matter, why are we calling them "keypairs" and not "Known Hosts" or some other clearer term?

src/app/helptext/system/ssh-keypairs.ts
lola27: @wblock yes. the private key is visible.

src/app/pages/system/ssh-keypairs/ssh-keypairs-form/ssh-keypairs-form.component.ts
william-gr: Because Key pairs is not the same as Known Hosts, completely different things.

They are for replications but they could eventually be used by some other subsystem, so they are shared instead of repeating the same thing for all. (I would be OK if UI want to restrict it to replication screen for now and abstract it away).

wblock: The problem is that this is being seen as SSH only.  There are other authentication methods, and we should plan for those being used eventually even if it is not possible now.  SSH is just a special case.  The "keypair" term does not suggest that this is really a way of defining a known remote host and the authentication for it.

william-gr: Known Remote Host is not how SSH authentication works, you should read on that a bit.

You seem to have terms confused.

wblock: I am not using the term Known Hosts as it is used by SSH.  The mistake is in thinking that this should be SSH-specific.  Corral used the term Peers, which is also terrible, but perhaps that will clarify.  The idea is that there is a list of predefined systems with pre-entered authentication information.  It might be SSH, it could be something else, and we should not box ourselves into a corner by making it SSH-specific.

william-gr: This one specifically is exclusive for SSH authentication.

PR #2120, opened 2019-03-21
Fix 79995 master

dkmullen:
Ticket: #79995
Fixes color issues on date picker that were especially bad on the hi-contrast theme. Also improves the calendar's display on Firefox.


erinix:
It looks better but I notice the month April still looks a little funky
april

PR #2127, opened 2019-03-27
Fix 74902 - Network Interfaces Rewrite with TrueNAS features

erinix:


src/app/helptext/network/interfaces/interfaces-form.ts
```diff
@@ -60,4 +56,65 @@ alias_netmaskbit6_tooltip : T('Select the prefix length used on the network.'),
 delete_placeholder6: T('Delete'),
 delete_tooltip6: T('Set to delete this alias.'),
 
+bridge_members_placeholder: T('Bridge Members'),
+bridge_members_tooltip: T(''),
```
Mrt134: From FreeBSD documentation:

"Network interfaces to include in the bridge."

src/app/helptext/network/interfaces/interfaces-form.ts
```diff
@@ -60,4 +56,65 @@ alias_netmaskbit6_tooltip : T('Select the prefix length used on the network.'),
...
+
+failover_critical_placeholder: T('Failover Critical'),
+failover_critical_tooltip: T(''),
```
Mrt134: From TrueNAS Guide "Interface Configuration Settings" table:

"Set this interface as critical. This allows logging in to the web interface available at the Virtual IP address after a failover. Warning: At least one interface must have this option set or the web interface will become unavailable. This can also be set when configuring network interfaces in the Console Setup Menu."

src/app/helptext/network/interfaces/interfaces-form.ts
```diff
@@ -60,4 +56,65 @@ alias_netmaskbit6_tooltip : T('Select the prefix length used on the network.'),
...
+
+failover_group_placeholder: T('Failover Group'),
+failover_group_tooltip: T(''),
```
Mrt134: From TrueNAS Guide "Failover" section:

"Combine multiple critical interfaces into a single Failover group. All interfaces in the group must fail before triggering a failover."

src/app/helptext/network/interfaces/interfaces-form.ts
```diff
@@ -60,4 +56,65 @@ alias_netmaskbit6_tooltip : T('Select the prefix length used on the network.'),
...
+
+failover_vhid_placeholder: T('Failover VHID'),
+failover_vhid_tooltip: T(''),
```
Mrt134: From TrueNAS Guide "Failover" section:

"Virtual Host ID for the failover interface". Must be unique on the broadcast segment of the network. Can be any unused number between <i>1</i> and <i>255</i>."

src/app/helptext/network/interfaces/interfaces-form.ts
```diff
@@ -60,4 +56,65 @@ alias_netmaskbit6_tooltip : T('Select the prefix length used on the network.'),
...
+
+failover_alias_address_placeholder: T('Failover Alias'),
+failover_alias_address_tooltip: T(''),
```
Mrt134: From TrueNAS Guide "Failover" section:

"IPv4 address to use for the failover interface. Set a manual address here if DHCP will not be used."

src/app/helptext/network/interfaces/interfaces-form.ts
```diff
@@ -60,4 +56,65 @@ alias_netmaskbit6_tooltip : T('Select the prefix length used on the network.'),
...
+
+failover_virtual_alias_address_placeholder: T('Failover Virtual Alias'),
+failover_virtual_alias_address_tooltip: T(''),
```
Mrt134: From TrueNAS Guide "Failover" section:

"IP Address used for administrative access to this array."

william-gr: We should probably call IPs in the UI.

This is the same as Virtual IP in legacy UI, its the CARP address (shared between nodes)

src/app/helptext/network/interfaces/interfaces-form.ts
william-gr: Should probably call IPs.

These are the IPs of the controller 2/B.

erinix: These fields are able to take either IPv4 or IPv6 with the new page

william-gr: The backend does not really support IPv6 for failover yet ;) It may accept but does not work

src/app/helptext/network/interfaces/interfaces-form.ts
```diff
@@ -1,7 +1,16 @@
 import { T } from '../../../translate-marker';
 import { Validators } from '@angular/forms';
+import { rangeValidator } from '../../../pages/common/entity/entity-form/validators/range-validation';
 
 export default {
+int_type_placeholder: T('Type'),
+int_type_tooltip: T(''),
```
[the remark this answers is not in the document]
erinix: "Often used with Jails or Virtual Machines.", I believe this wording only applies to bridges iirc.

src/app/helptext/network/interfaces/interfaces-form.ts
erinix: Ah :)

src/app/helptext/network/interfaces/interfaces-form.ts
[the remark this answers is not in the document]
william-gr: Yes, VLAN are used to isolate networks, nut necessarily related to jails or vms.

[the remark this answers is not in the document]
william-gr: Actually very rarely used in them.

william-gr:
I am not seeing how you actually manage the interfaces changes here.

Calling interface.create does not actually create the interface, only queues it up.

You have to call interface.commit to apply then and after that interface.checkin to confirm the changes were applied correctly otherwise they will be reverted.

src/app/components/common/topbar/topbar.component.ts
```diff
@@ -181,6 +183,23 @@ export class TopbarComponent implements OnInit, OnDestroy {
...
+  checkNetworkChangesPending() {
+    this.ws.call('interface.has_pending_changes').subscribe(res => {
+      this.pendingNetworkChanges = res;
+    });
+  }
+
+  showNetworkChangesPending() {
+    this.dialogService.confirm(
+      T("Pending Network Changes"),
+      T("You currently have uncommited network interface changes pending, go to the network interfaces section and review now?"),
```
wblock: Avoid using "you":
```suggestion
      T("There are uncommited network interface changes pending. Review them now?"),
```

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -0,0 +1,16 @@
+import { T } from '../../../translate-marker';
+import { Validators } from '@angular/forms';
+import { rangeValidator } from '../../../pages/common/entity/entity-form/validators/range-validation';
+
+export default {
+pending_changes_text : T('You currently have pending network changes, do you wish to commit now?\
```
wblock: As above, avoid "you":
'There are uncommitted network changes. Commit them now?\\'

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -0,0 +1,16 @@
...
+ Any changes that are not committed will be automatically rolled back.'),
```
wblock: 'Unsaved changes will be lost.'

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -0,0 +1,16 @@
...
+commit_changes_title: T("Commit Network Changes"),
+commit_changes_warning: T("Do you wish commit your network changes?  Network connectivity will be interrupted."),
```
wblock: 'Commit network changes? Network connectivity will be interrupted.'

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -0,0 +1,16 @@
...
+changes_saved_successfully: T("Network changes saved successfully"),
```
wblock: If this not a title, but is in a dialog, it should end in a period.

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -0,0 +1,16 @@
...
+commit_button: T("Commit"),
+rollback_button: T("Rollback"),
```
wblock: I think these should be treated like other verbs: Roll Back

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -0,0 +1,16 @@
...
+rollback_changes_title: T("Rollback Network Changes"),
```
wblock: 'Roll Back Network Changes'

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -0,0 +1,16 @@
...
+rollback_changes_warning: T("Do you wish rollback your network changes?"),
```
wblock: 'Roll back network changes?'

Should we use the term "roll back", or just "Discard unsaved network changes?"

src/app/pages/network/interfaces/interfaces-list/interfaces-list.component.ts
```diff
@@ -22,6 +25,11 @@ export class InterfacesListComponent {
     message: T("Network connectivity will be interrupted. "),
   }
   protected hasDetails = true;
+  protected entityList: any;
+  protected checkChangesSubscription: any;
+  public hasPendingChanges = false;
+  pending_changes_text = T('You currently have pending network changes, do you wish to commit now?\
```
wblock: As above.  Isn't this a repeated string?

src/app/pages/network/interfaces/interfaces-list/interfaces-list.component.ts
```diff
@@ -1,55 +1,146 @@
...
+
+  commitPendingChanges() {
+    this.entityList.dialogService.confirm(
+      helptext.commit_changes_title,
+      helptext.commit_changes_warning,
+      false, helptext.commit_button).subscribe(confirm => {
+        if (confirm) {
+          this.entityList.loader.open();
+          this.entityList.loaderOpen = true;
+          this.ws.call('interface.commit').subscribe(res => {
```
william-gr: That method has a `checkin_timeout` option, it defaults to 60 seconds. It means if the user doesnt come back confirming the changes within that period it will get automatically rolled back.

We should mention that, or make it configurable.

erinix: I will add that :)

src/app/helptext/network/interfaces/interfaces-list.ts
erinix: I kind of like discard better

src/app/pages/network/interfaces/interfaces-list/interfaces-list.component.ts
erinix: That should be removed now and only in the helptext file

src/app/helptext/network/interfaces/interfaces-list.ts
erinix: I changed to Discard Network Changes as per your next comment

src/app/components/common/topbar/topbar.component.ts
```diff
@@ -188,16 +192,51 @@ export class TopbarComponent implements OnInit, OnDestroy {
...
+    );
+  }
+
+  showNetworkChangesPending() {
+    if (this.waitingNetworkCheckin) {
+      this.showNetworkCheckinWaiting();
+    } else {
+      this.dialogService.confirm(
+        T("Pending Network Changes"),
+        T("There are uncommited network interface changes pending.  Review them now?"),
```
wblock: 'There are unsaved network interface settings. Review them now?'

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -13,4 +13,10 @@ rollback_button: T("Discard"),
 rollback_changes_title: T("Discard Network Changes"),
 rollback_changes_warning: T("Discard unsaved network changes?"),
 changes_rolled_back: T("Network changes discarded."),
+
+checkin_title: T("Network Changes Check-in"),
```
wblock: 'Save Network Settings'

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -13,4 +13,10 @@ rollback_button: T("Discard"),
...
+checkin_message: T("Network changes have been applied, check in now?  Settings will be automatically discarded if not checked in."),
```
wblock: 'Network settings have been changed. These changes will be lost if not saved. Save them now?'

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -13,4 +13,10 @@ rollback_button: T("Discard"),
...
+checkin_complete_title: T("Network Check-in Complete"),
```
wblock: 'Network Settings Saved'

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -13,4 +13,10 @@ rollback_button: T("Discard"),
...
+checkin_complete_message: T("Network changes checked in successfully."),
```
wblock: 'Network setting changes have been saved.'

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -13,4 +13,10 @@ rollback_button: T("Discard"),
...
+checkin_button: T('Check-in'),
```
wblock: 'Save Settings'

src/app/pages/network/interfaces/interfaces-list/interfaces-list.component.ts
```diff
@@ -29,7 +29,13 @@ export class InterfacesListComponent implements OnDestroy {
   protected entityList: any;
   protected checkChangesSubscription: any;
   public hasPendingChanges = false;
-  pending_changes_text: string; 
+  public checkinWaiting = false;
+  pending_changes_text: string;
+  pending_checkin_text: string;
+  checkin_text: string = T("Settings will revert ");
+  checkin_text_2: string = T(" seconds after commit if you do not check back in.");
```
wblock: How does this even work?  If someone saves changes ("commit"), that means keep them, do not revert.

erinix: Basically they are applied but not permanent until the check-in, kind of like graphics settings in some OSes, that way you don't risk locking yourself out of your box

src/app/helptext/network/interfaces/interfaces-form.ts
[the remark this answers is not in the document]
erinix: @wblock @william-gr @Mrt134  does "Choose the kind of interface. Bridge adds this interface as part of an existing network configuration. Link Aggregation combines multiple network connections into a single interface. A virtual LAN (VLAN) partitions and isolates a segment of the connection. "  work?  (Basically leave off the jails/virtual machines part?

src/app/helptext/network/interfaces/interfaces-form.ts
erinix: @wblock @william-gr this seems like a resonable tooltip label to me, what do you say?

src/app/helptext/network/interfaces/interfaces-form.ts
erinix: @wblock @william-gr this seems like a reasonable tooltip label to me, what do you say?

src/app/helptext/network/interfaces/interfaces-form.ts
erinix: @wblock @william-gr this seems like a reasonable tooltip label to me, what do you say?

src/app/helptext/network/interfaces/interfaces-form.ts
erinix: @wblock @william-gr this seems like a reasonable tooltip label to me, what do you say?

src/app/helptext/network/interfaces/interfaces-form.ts
erinix: how about we call it Failover Virtual IP Address @william-gr @wblock  (and perhaps just change the word Alias to IP Address on all fields that say Alias?)

erinix: The tooltip sounds fine to me though, what do you say @wblock

src/app/helptext/network/interfaces/interfaces-form.ts
erinix: Same for this one as below, "IP Address" instead of "Alias"? @wblock  @william-gr

src/app/helptext/network/interfaces/interfaces-form.ts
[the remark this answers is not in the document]
william-gr: "Bridge adds this interface as part of an existing network configuration." reads very strangely.

I would go with "Bridge creates a logical link between two or more networks."

src/app/helptext/network/interfaces/interfaces-form.ts
william-gr: yes

src/app/helptext/network/interfaces/interfaces-form.ts
william-gr: That option more importantly means that if the interface is preempted via VRRP protocol it will cause a failover.

src/app/helptext/network/interfaces/interfaces-form.ts
william-gr: yes

src/app/helptext/network/interfaces/interfaces-form.ts
william-gr: yes

src/app/helptext/network/interfaces/interfaces-form.ts
william-gr: yes

src/app/helptext/network/interfaces/interfaces-form.ts
william-gr: yes

william-gr: The intent if this Ip is not just to administrative access it, its the IP shared between the controllers.

src/app/helptext/network/interfaces/interfaces-form.ts
```diff
@@ -60,4 +56,65 @@ alias_netmaskbit6_tooltip : T('Select the prefix length used on the network.'),
...
+],
+
+lagg_interfaces_placeholder : T('Lagg Interfaces'),
+lagg_interfaces_tooltip : T('Select the interfaces to use in the aggregation.<br>\
+ Warning: Lagg creation fails if any of the selected\
+ interfaces have been manually configured.'),
+lagg_interfaces_validation : [ Validators.required ],
+
+mtu_placeholder: T('MTU'),
+mtu_tooltip: T(''),
```
wblock:
```suggestion
mtu_tooltip: T('Maximum Transmission Unit, the largest protocol data unit that can be communicated. The largest workable MTU size varies with network interfaces and equipment. *1500* and *9000* are standard Ethernet MTU sizes.'),
```

src/app/helptext/network/interfaces/interfaces-form.ts
william-gr: "If this interface is considered critical for failover it will cause a failover event in case it is preempted by the other controller via VRRP/CARP protocol and there arent other interfaces in that same failover group."

wblock:
@erinix @william-gr The MTU range limit is currently 68-9000.  However, some interfaces do not support 9000-byte MTUs, and some support more (em(4)). Fibre Channel is probably different, too.  I don't know if we can query the maximum MTU from the driver. The suggestion from Performance was to allow any value but warn the user if the value entered was not 1500 or 9000.

william-gr:
> @erinix @william-gr The MTU range limit is currently 68-9000. However, some interfaces do not support 9000-byte MTUs, and some support more (em(4)). Fibre Channel is probably different, too. I don't know if we can query the maximum MTU from the driver. The suggestion from Performance was to allow any value but warn the user if the value entered was not 1500 or 9000.

Warning sounds reasonable enough.

src/app/helptext/network/interfaces/interfaces-form.ts
```diff
@@ -60,4 +56,65 @@ alias_netmaskbit6_tooltip : T('Select the prefix length used on the network.'),
 delete_placeholder6: T('Delete'),
 delete_tooltip6: T('Set to delete this alias.'),
```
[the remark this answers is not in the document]
erinix: Lola has been working on a solution like that for jails, once she has that implemented I'll see about making that work here too (probably in a separate PR)

PR #2151, opened 2019-04-03
Added info in the helptext about creating a friendly name in the from email field.

a-stjohn:


src/app/helptext/system/email.ts
```diff
@@ -6,9 +6,10 @@ export const helptext_system_email = {
   em_fromemail: {
     placeholder: T("From E-mail"),
     tooltip: T(
-      "The envelope From address shown in the email.\
- This is set to assist with filtering mail on the\
- receiving system."
+      'The envelope <i>From</i> address shown in the email.\
+ This can be set to make filtering mail on the\
```
Mrt134: "This can be set to make filtering mail on the ..." Is this missing a word?

PR #2158, opened 2019-04-03
Fix up dyndns - remove password requirement and use middleware call for providers

erinix:
(should only have to change password if a new one needs adding)

Ticket: #83363


erinix:
MW code is ready and in there but it was merged for 11.2u4 so doesn't exist on an 11.2u3 system (I tested this on a box that has an internal)

PR #2175, opened 2019-04-09
Another Documentation URL Fix

aidik:
This time it is `account.html` vs. `accounts.html`

I also took the time and check the logs and it seems that this is the last of the erroneous URLs, _unless the URL is really well hidden and no one clicks on it._


wblock:
Erin: this needs to be in 11.3 also.

PR #2185, opened 2019-04-11
Update de.po

FaMux:
Some german translations


erinix:
Please also translate for the branch 'freenas/11.2-stable' if you would like to see your translations in an 11.2 release as well, thank you :)

PR #2212, opened 2019-04-17
Update helptext to mention the script is run using sh(master).

a-stjohn:


src/app/helptext/task-calendar/initshutdown/initshutdown.ts
```diff
@@ -12,7 +12,8 @@ export default {
 
     ini_script_placeholder: T('Script'),
     ini_script_validation : [ Validators.required ],
-    ini_script_tooltip: T('Select the script from an existing pool.'),
+    ini_script_tooltip: T('Select the script from an existing pool.\
+     The script is run using <b>sh</b>'),
```
wblock: This does not tell the user which version of sh.  Link to the man page, and avoid bold.

PR #2228, opened 2019-04-19
Update es.po

FelixSabando:
I don't translate some key words like pool because it don't have a clear translation and it can cause confusion (if you search "pool" you will find a lot of information, for the translation I don't think so) and if I change 1 instance I should change all to prevent confusions.
In some other words (like snapshot) I put the word in English between parentheses to be more clear, because it is commonly used.


erinix:
yea not sure how you should translate "pool" since it is a zfs term, thanks for your work :)

PR #2243, opened 2019-04-22
NAS 101139

lola27:


src/app/pages/system/CloudCredentials/cloudcredentials-form/cloudcredentials-form.component.ts
```diff
@@ -793,14 +797,59 @@ export class CloudCredentialsFormComponent {
...
   protected entityForm: any;
+  protected oauthClentIdField: any;
+  protected oauthClentSecretField: any;
 
   public custActions: Array<any> = [
+    {
+      id: 'authenticate',
+      name: T('Authenticate'),
+      function: () => {
+        window.open(this.oauthURL+ "?origin=" + encodeURIComponent(window.location.toString()), "_blank", "width=640,height=480");
```
wblock: Why a fixed 640x480 window?  We really should scale dialogs and new browser windows to like 80% of the main browser window. It might be much larger (big monitor) or much smaller (phone) than a fixed number.

lola27: Hi @wblock , this one is a new scalable window. it will show as a new tab on the phone.

src/app/helptext/system/cloudcredentials.ts
```diff
@@ -15,6 +15,16 @@ export const helptext_system_cloudcredentials = {
     validation: [Validators.required]
   },
 
+  client_id: {
+    placeholder: T('OAuth Client ID'),
+    tooltip: T(''),
```
wblock: @william-gr we need some starting point on the tooltips for these two fields.

william-gr: https://developers.google.com/identity/protocols/OAuth2

PR #2246, opened 2019-04-23
NAS-101524 make sure topmost dialog is selected for resizing

dkmullen:
NAS-101524
Another fix for the resizable error dialog, this time dealing with multiple dialogs being open at once by using ids where possible, and selecting the last one in an array which is hopefully our error dialog.


dkmullen:
One more change: When the err message at the top is several lines long, it can push the backtrace msg down too far, so we calc and adjust the size of both messages.

erinix:
rebuild

PR #2248, opened 2019-04-23
NAS-100807 Match err msg from MW on pool unlock to err dialog format

dkmullen:
NAS-100807
Re-works the err message on error dialog in Unlock Pool action to make use of MW error message (currently shows Object Object)


erinix:
rebuild

PR #2254, opened 2019-04-25
NAS-100496 101549 Updated helptext because the removal "Protocol" field.

a-stjohn:


src/app/helptext/system/general.ts
```diff
@@ -64,8 +63,7 @@ export const helptext_system_general = {
     placeholder: T("WebGUI HTTP -> HTTPS Redirect"),
     tooltip: T(
       "Check this to redirect <i>HTTP</i> connections to\
-        <i>HTTPS</i>. <i>HTTPS</i> must be selected in\
-        <b>Protocol</b>."
+ <i>HTTPS</i>. A <b>GUI SSL Certificate</b> must be selected."
```
wblock:
```suggestion
 <i>HTTPS</i>. A <i>GUI SSL Certificate</i> must be selected."
```

PR #2255, opened 2019-04-25
NAS-100496 Updated helptext because the removal "Protocol" field.

a-stjohn:
(cherry picked from commit 7c36d1c070b307da93b05325ca19246a2e2be812)


a-stjohn:
The associated ticket is only for 11.3. An 11.2 PR isn't needed.

PR #2265, opened 2019-04-28
Update de.po

FaMux:
Some more german translations


erinix:
If you would like to see your changes in an 11.2 release please also create a PR for the branch 'freenas/11.2-stable'

PR #2273, opened 2019-05-01
QE tkt-86715 Add wait_on_element to create pool and fix xpath FreeNAS 11.3 UI testing

ericbsd:


tests/selenium/test_006_store_pool.py
```diff
@@ -102,16 +100,15 @@ def test_03_create_newpool2(wb_driver):
...
+    # Click OK Button
     wb_driver.find_element_by_xpath(xpaths['createpoolButton']).click()
-    time.sleep(1)
+    wait_on_element(wb_driver, xpaths['addAction'], script_name, test_name)
     # taking screenshot
-    test_name = sys._getframe().f_code.co_name
     take_screenshot(wb_driver, script_name, test_name)
     error_check(wb_driver)
-    time.sleep(2)
+    time.sleep(1)
```
wblock: Watch out for these nonspecific waits. If anything changes, like moving the testing system to new hardware, they can all break because the timing changes.  If possible, wait for specific conditions.  See https://selenium-python.readthedocs.io/waits.html.

ericbsd: @wblock thanks for this info I am still learning selenium.

PR #2280, opened 2019-05-02
Nas 101541 stable checkbox align

dkmullen:
NAS-101541
Aligns checkboxes with the center of their labels rather than the baseline


erinix:
rebuild

PR #2300, opened 2019-05-14
NAS 101316 replication wizard

lola27:


src/app/helptext/task-calendar/replication/replication-wizard.ts
```diff
@@ -0,0 +1,13 @@
+import { T } from '../../../translate-marker';
+
+export default {
+    step1_label: T('SSH Keypair/Connection'),
+
+    name_placeholder: T('Name'),
+    name_tooltip: T(''),
```
a-stjohn:
```suggestion
    name_tooltip: T('Replication Task Name'),
```

src/app/helptext/task-calendar/replication/replication-wizard.ts
```diff
@@ -0,0 +1,13 @@
...
+
+    cipher_placeholder: T('Replication Performance'),
+    cipher_tooltip: T(''),
```
a-stjohn: I don't know what "Replication Performance" is/does/means. Could you put some initial text here describing what it is?

lola27: @a-stjohn This is the cipher field. We use `Replication Performance` instead of `Cipher` here.

src/app/helptext/task-calendar/replication/replication-wizard.ts
lola27: @a-stjohn It is not only for Replication task. The name also use for new created ssh keypair/ ssh connection/period snapshot task.

erinix:
Docs review will continue after merging

PR #2303, opened 2019-05-15
Nas 100746 support page

dkmullen:
NAS-100746
Support page for TrueNAS - also a rework of the layout for FN, since they are built on the same model
IMPORTANT: A small feature on this page, to update the license, is blocked by NAS-101741
https://jira.ixsystems.com/browse/NAS-101724


erinix:
rebuild

PR #2306, opened 2019-05-15
Feat 55986 master

dkmullen:
NAS 100502 Adds support for ACME certificates


src/app/pages/system/acmedns/acmedns-list/acmedns-list.component.ts
```diff
@@ -0,0 +1,63 @@
...
+  public columns: Array < any > = [
+    { name: T('Authenticator'), prop: 'authenticator' },
+    { name: T('Name'), prop: 'name' }
+  ];
+
+  public config: any = {
+    paging: true,
+    sorting: { columns: this.columns },
+    deleteMsg: {
+      title: 'Name',
```
lola27: Currently, the delete message is like 'Delete Name XXX'. You may want to change the title to be like 'ACME DNS Authenticator', then the delete message will be 'Delete ACME DNS Authenticator XXX'.

dkmullen: Good catch!

PR #2314, opened 2019-05-20
nest ws calls to make sure all data is in

dkmullen:
NAS-101841
Fixes a condition where available updates (sometimes) don't load. Nests the second ws call inside the first to make sure the first is done before check() is called.


erinix:
rebuild

PR #2318, opened 2019-05-21
Properly fetch remote host key when managing replication task

themylogin:
@lola27 please check if it works properly, I don't have an ability to build it.


themylogin:
@lola27 also please change "Remote HostKey" input to textarea because it is multiline value in 11.2

lola27:
Thanks, @themylogin, your PR works properly!  I added the change request :)

PR #2331, opened 2019-05-28
NAS-100844 adjust temp size esp for ffx

dkmullen:
NAS-100844 - adjust temp size on Pools widget


dkmullen:
Should be fixed by PR 2333

PR #2332, opened 2019-05-28
NAS-101117 adjust height of input elements

dkmullen:
NAS-101117
Makes inputs 15px high (previously 14.5px on Chrome, 13.5 on FFx, which hid underscores)


dkmullen:
Should be fixed by NAS-101876, installing the Roboto font

PR #2338, opened 2019-05-30
Fix hold_pending_snapshots_tooltip

themylogin:


src/app/helptext/task-calendar/replication/replication.ts
```diff
@@ -105,7 +105,7 @@ snapshots will be stored. Example: Poolname/Datasetname, not the mountpoint or f
  destination system snapshots and do a full copy of all configured source system snapshots.'),
 
     hold_pending_snapshots_placeholder: T('Hold Pending Snapshots'),
-    hold_pending_snapshots_tooltip: T('Prevent copied source snapshots from being deleted on the destination\
+    hold_pending_snapshots_tooltip: T('Prevent snapshots from being cleaned up by retention on the source\
```
Mrt134: The wording is still a little confusing to me, so I'm going to make a suggestion based on these assumptions about what will happen when this setting is enabled:
- The system identifies source snapshots that should be, but have not been replicated to the destination.
- A ZFS hold is placed on these snapshots to prevent the source system snapshot retention policy from automatically removing these unreplicated snapshots.

```suggestion
    hold_pending_snapshots_tooltip: T('Prevent source system snapshots that have failed replication from being automatically removed by the snapshot retention policy.'),
```

src/app/helptext/task-calendar/replication/replication.ts
```diff
@@ -105,7 +105,7 @@ snapshots will be stored. Example: Poolname/Datasetname, not the mountpoint or f
...
  system if the replication task fails.'),
```
Mrt134:
```suggestion

```

PR #2355, opened 2019-06-04
NAS-101924 allow minutes to be updated on periodic snapshot tasks

dkmullen:
NAS-101924
Allow minutes to be edited on periodic snapshot tasks, fits the schedule list to the mat-card; also, I took the liberty of removing seconds from the time display since it is always 00 and never adjustable(?)


erinix:
rebuild

PR #2358, opened 2019-06-05
NAS-101996

lola27:


lola27:
@Mrt134 the auto-created ssh keypair will be named as NAME_keypair (NAME is the name field value). You may want to update the tooltip of that field.

Mrt134:
Ok, I added an extra explanation to the Name field help text. The rest of the PR looks ok to me.

PR #2374, opened 2019-06-07
NAS-100084 -- fetching/displaying snapshot name only

aervin:


aervin:
Closing. Doesn't address the ticket.

PR #2375, opened 2019-06-07
Nas 100084 -- snapshot details in summary row

aervin:


aervin:
I think this one's ready for another review @erinix

PR #2403, opened 2019-06-14
NAS-101989 make adjustment for a long tooltip

dkmullen:
NAS-101989
Raises really long tool tips up a few lines to avoid being cut off at bottom


erinix:
The wizard still has some cut off
Screen Shot 2019-06-14 at 2 30 24 PM

erinix:
looks good, just need to fix your merge conflict :)

PR #2413, opened 2019-06-17
NAS-100948 -- auto expand pool panels in Storage / Pools

aervin:


erinix:
looking good so far but we also want to have the first node of the dataset tree expanded too

PR #2418, opened 2019-06-18
NAS-100348 -- moves smb default perm field to Basic mode

aervin:


anodos325:
> There's a little more to this one than just unhiding it, the alert still doesn't show up (because it is already checked upon form load). We might want to wait until Warren is back next week to discuss if we want to use that same warning message or another one for a new share.

This parameter should remain an advanced option, but should be changed to default to False.

aervin:
Adding to discussion ticket list

wblock:
"Setting default permissions will change the permissions of every file in this share directory and all subdirectories. Finer control of permissions is available through the <i>Edit ACL</i> option for each dataset."

PR #2422, opened 2019-06-18
Nas 100203 -- right-click message about copy/paste shortcuts

aervin:
Leaving the context menu component even though it's not being used. Could come in handy later.


package.json
```diff
@@ -122,6 +122,7 @@
     "@types/jasmine": "2.8.11",
     "@types/lodash": "4.14.134",
     "@types/node": "10.12.7",
+    "@types/xterm": "^3.0.0",
```
erinix: is this dependency being used somewhere?

src/app/pages/shell/shell.component.ts
```diff
@@ -97,6 +85,20 @@ export class ShellComponent implements OnInit, OnChanges, OnDestroy {
     }
   }
 
+  onRightClick(): false {
+    this._snackbar.open(
+      T(
```
erinix: I like how this works but I think we may want to get docs opinion on the message, also if we cannot display OS specific cut and paste (since it is different for Mac OS vs Windows and Unix) we should maybe tell the user something like "If you are on Mac OS use Control C and Control V, otherwise use Ctrl Insert and Shift Insert"

a-stjohn: Checkout #2434

aervin:
`yarn remove <dep>` seems to sort dependencies automatically?

PR #2432, opened 2019-06-20
NAS-101596

lola27:


lola27:
@freenas/docs-team Jail name able to include dots (.), could you please provide an update the tooltip?

PR #2443, opened 2019-06-21
NAS-101955 -- adds cache meta tags to index.html

aervin:


aervin:
Closing until I've resolved an issue I've found in FF

PR #2469, opened 2019-06-28
NAS-102306 -- use jails approach for multipath mgmt

aervin:


erinix:
Also I believe the code in clean() can be removed as that is no longer necessary

PR #2482, opened 2019-07-02
NAS-102218 -- email page updates

aervin:
* using ws calls
* password field no longer required
* this does not strip the password field on load. users are able to configure other fields and save as usual without stripping the password field. let me know if that's not desirable behavior.


erinix:
This is failing the build test because of a mismatched )

src/app/pages/common/entity/entity-form/entity-form.component.ts
```diff
@@ -492,6 +492,7 @@ export class EntityFormComponent implements OnInit, OnDestroy, OnChanges, AfterV
       this.busy = this.submitFunction(value)
                     .subscribe(
                         (res) => {
+                          console.log({ res });
```
erinix: remove your console.log

PR #2488, opened 2019-07-05
Nas 102136

aervin:
https://jira.ixsystems.com/browse/NAS-102136


aervin:
Closing this as obsolete now that this can be achieved with simple column definitions in disk list.

PR #2495, opened 2019-07-09
Nas 102362

dkmullen:
Provides warnings when changing keys/passphrases on pools, and also forces the download-key dialog to open so the user has to download a copy of the new key (requested by docs)

### DIFF
--- a/src/app/pages/storage/volumes/volumerekey-form/volumerekey-form.component.ts
+++ b/src/app/pages/storage/volumes/volumerekey-form/volumerekey-form.component.ts
@@ -93,7 +93,7 @@ export class VolumeRekeyFormComponent  implements Formconfiguration {
     return this.rest.post(this.resource_name + "/" + this.pk + "/rekey/", 
       { body: JSON.stringify({passphrase: value.passphrase}) }).subscribe((restPostResp) => {
         this.loader.close();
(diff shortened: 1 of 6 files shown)


src/app/pages/storage/volumes/volumerekey-form/volumerekey-form.component.ts
```diff
@@ -82,12 +90,18 @@ export class VolumeRekeyFormComponent  implements Formconfiguration {
...
-    return this.rest.post(this.resource_name + "/" + this.pk + "/rekey/", { body: JSON.stringify({passphrase: value.passphrase}) }).subscribe((restPostResp) => {
-      this.loader.close();
-      this.dialogService.Info(T("Re-keyed Pool"), T("Successfully re-keyed pool ") + value.name);
-
-      this.router.navigate(new Array('/').concat(
-        this.route_success));
+    return this.rest.post(this.resource_name + "/" + this.pk + "/rekey/", 
+      { body: JSON.stringify({passphrase: value.passphrase}) }).subscribe((restPostResp) => {
+        this.loader.close();
+        this.snackBar.open(T('Successfully re-keyed pool ' + value.name), T("Close"), {
```
erinix: Moving the ) of the translation marker actually breaks the translation grabbing for this string

PR #2497, opened 2019-07-09
Update zh-hant.po

jasoncheng7115:


erinix:
I notice this PR brings some things from master that are not necessarily in stable.  I would suggest mainly focusing your efforts on master at this point because we are trying to avoid churn on the 11.2 branch.

jasoncheng7115:
Hoo, I see. I'm going to focus on master. Thank you!

PR #2501, opened 2019-07-09
Nas 102098

aervin:
https://jira.ixsystems.com/browse/NAS-102098


src/app/pages/task-calendar/replication/replication-list/replication-list.component.ts
```diff
@@ -19,22 +18,32 @@ export class ReplicationListComponent {
...
     protected route_edit: string[] = ['tasks', 'replication', 'edit'];
     protected route_success: string[] = ['tasks', 'replication'];
-    protected entityList: any;
+    public entityList: any;
+    protected hasDetails = true;
     protected asyncView = true;
 
     public columns: Array<any> = [
         { name: 'Name', prop: 'name' },
-        { name: 'Direction', prop: 'direction', hidden: true},
```
erinix: These columns should remain as long as they are hidden (it allows for filtering and the user can still check them as not hidden if they want)

PR #2511, opened 2019-07-11
NAS-102298 hide checkbox unless needed

dkmullen:
Hides the rrd destroy checkbox except when needed, enforces unchecked state on load, adds validation (numbers only) to two fields

### DIFF
--- a/src/app/pages/system/reporting/reporting.component.ts
+++ b/src/app/pages/system/reporting/reporting.component.ts
@@ -19,9 +19,22 @@ export class ReportingComponent {
   protected queryCall = 'reporting.config';
   public entityForm: any;
   private settings_saved = T("Settings saved.");
-  public graphAge = [];
-  public graphPoints = [];
   public rrd_checkbox: any;
+  custActions: any[] = [
+    {
+      id:'reset',
+      name:'Reset',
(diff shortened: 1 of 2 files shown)


src/app/pages/system/reporting/reporting.component.ts
```diff
@@ -65,22 +65,40 @@ export class ReportingComponent {
     public snackBar: MatSnackBar,
   ) {}
 
-
   afterInit(entityEdit: any) {
     this.entityForm = entityEdit;
+    this.rrd_checkbox = _.find(this.fieldConfig, {'name' : 'confirm_rrd_destroy'});
+    entityEdit.formGroup.controls['graph_age'].valueChanges.subscribe((res) => {
```
lola27: 1. need to check both `graph_age` and `graph_points` no matter which one triggered valueChanges event. 
2. one more suggestion, you are able to access the query result (original config) by `entityEdit.wsResponse`, so just need to compare the changed value with it :)

dkmullen: Oh wow, good catch! Also, I'm glad to know about wsResponse, and I added a reset button using it

PR #2520, opened 2019-07-15
Nas 101943 master

dkmullen:
NAS-101943
Uses new MW calls to provide user, when Exporting/Disconnecting a pool, with info on what services are using it; also gives feedback if processes need to be restarted or can't be stopped by the system


lola27:
@dkmullen could you please resolve the conflicts? thanks

PR #2522, opened 2019-07-16
NAS-102196

dkmullen:
Adds dynamic labels to identify controllers in Interfaces, IPMI components


src/app/pages/network/ipmi/ipmi.component.ts
```diff
@@ -16,9 +16,16 @@ import { T } from '../../../translate-marker';
...
+    diameter='25'
+    class="form-select-spinner"
+    id="ipmi_controller-spinner"
+    *ngIf="!currentControllerLabel">
+  </mat-spinner>
   <mat-select *ngIf="is_ha" #storageController name="controller" placeholder="Controller" (selectionChange)="loadData()" [(ngModel)]="remoteController">
-    <mat-option [value]="false">Current controller</mat-option>
-    <mat-option [value]="true">Failover controller</mat-option>
+
+    <mat-option [value]="false">Current Controller: TrueNAS Controller {{currentControllerLabel}}</mat-option>
```
wblock: This is pretty redundant.  "Current" and "Failover" are kind of ambiguous in meaning.  Let's just implement the correct terminology here:
```suggestion
    <mat-option [value]="false">Active: TrueNAS Controller {{currentControllerLabel}}</mat-option>
```

src/app/pages/network/ipmi/ipmi.component.ts
```diff
@@ -16,9 +16,16 @@ import { T } from '../../../translate-marker';
...
+    <mat-option [value]="true">Failover Controller: TrueNAS Controller {{failoverControllerLabel}}</mat-option>
```
wblock:
```suggestion
    <mat-option [value]="true">Standby: TrueNAS Controller {{failoverControllerLabel}}</mat-option>
```

src/app/pages/network/ipmi/ipmi.component.ts
```diff
@@ -132,6 +141,17 @@ export class IPMIComponent {
     if (window.localStorage.getItem('is_freenas') === 'false') {
       this.ws.call('failover.licensed').subscribe((is_ha) => {
         this.is_ha = is_ha;
+        if (this.is_ha) {
+          this.ws.call('failover.node').subscribe((node) => {
+            if (node === 'A') {
```
wblock: This whole block can be replaced with the ternary operator.  That assumes we will only ever have two controllers, but the if/else if block also does that.
```
this.currentControllerLabel = (node === 'A') ? '1' : '2';
this.failoverControllerLabel = (node === 'A') ? '2' : '1';
```

lola27:
@dkmullen could you please resolve the conflicts? thanks

PR #2525, opened 2019-07-16
Nas 102087

aervin:


src/app/helptext/sharing/afp/afp.ts
```diff
@@ -138,6 +138,12 @@ export const helptext_sharing_afp = {
   dialog_message: T("Enable this service?"),
   dialog_button: T("Enable Service"),
 
+  dialog_delete_title: T("Delete"),
+  dialog_delete_message: T(
+    "Stop sharing AFP (Apple File Protocol)? The sharing configuration\
+ will be deleted (but the underlying dataset will not)."
```
wblock: I thought we came up with a wording for this already for the SMB service, but maybe not.
```suggestion
 will be deleted. Data in the former share dataset will not be affected."
```

src/app/helptext/sharing/nfs/nfs.ts
```diff
@@ -90,7 +90,11 @@ export const helptext_sharing_nfs = {
 
   dialog_enable_service_title: T("Enable service"),
   dialog_enable_service_message: T("Enable this service?"),
-  dialog_enable_service_button: T("Enable Service"),
+  dialog_enable_service_button: T("Enable Service"),    
+  
+  dialog_delete_title: T("Delete"),
+  dialog_delete_message: T("Stop sharing NFS? The sharing configuration will be deleted\
+(but the underlying dataset will not)."),
```
wblock: As above.  Also, can't we factor out this dialog and pass in the message as a parameter?  It seems like a lot of work to maintain a different one for each sharing system.

src/app/helptext/sharing/smb/smb.ts
```diff
@@ -113,6 +113,10 @@ export const helptext_sharing_smb = {
     dialog_edit_acl_title: T("Configure ACL"),
     dialog_edit_acl_message: T("Configure permissions for this share's dataset now?"),
     dialog_edit_acl_button: T("Configure now"),
+
+    dialog_delete_title: T("Delete"),
+    dialog_delete_message: T("Stop sharing Windows (SMB)? The sharing configuration will be deleted\
+ (but the underlying dataset will not)."),
```
wblock: As above.

src/app/pages/sharing/smb/smb-list/smb-list.component.ts
```diff
@@ -64,9 +65,19 @@ export class SMBListComponent {
...
+          this.dialogService
+            .confirm(
+              helptext_sharing_smb.dialog_delete_title,
+              helptext_sharing_smb.dialog_delete_message.replace('?', ` ${row.cifs_name}?`)
+            )
+            .pipe(filter(ok => !!ok))
+            .subscribe(
+              () => this.entityList.delete(row.id),
+              error => new EntityUtils().handleWSError(this, error, this.dialogService)
+            )
```
wblock: These routines also look identical.  Can we factor to reduce code duplication?

aervin:
There's some code in entity-table that gets us there if we're okay with a more generic message.

fn-share-msg-01

wblock:
The last sentence kind of overrides the reassurance of the first two sentences.  Can that be moved?  Otherwise, the only thing that would be nice is having the dialog title tell what will be deleted: "Delete AFP Share d0".

Actually... why not call this what it is, "Unshare" rather than delete?

aervin:
I like "unshare," but the user _is_ deleting something (the configuration). The final sentence is tacked on after the fact... maybe I can add a flag to remove it conditionally. "Delete" is also hardcoded in some places.

wblock:
It's the term "delete" and what it implies to the user that is the problem. I think we can phrase this to make it clear.
```
Unshare d0

Stop sharing d0. The sharing configuration will be removed.
Data in the share dataset will not be affected.

X Confirm                                   CANCEL   UNSHARE
```

aervin:
I've not included the share identifier in the dialog title to avoid even more conditional logic in entity-table. Otherwise, the dialog is as suggested in the last comment.

src/app/helptext/sharing/index.ts
```diff
@@ -1,5 +1,11 @@
+import { T } from "app/translate-marker";
+
 export * from "./afp/afp";
 export * from "./iscsi/iscsi";
 export * from "./nfs/nfs";
 export * from "./smb/smb";
 export * from "./webdav/webdav";
+
+export const delete_share_message = (shareName: string) =>
+  T(`Stop sharing. The sharing configuration will be removed.\
```
wblock: Can this be simplified to
```suggestion
  T(`Stop sharing ${shareName}. The sharing configuration will be removed.\
```

src/app/helptext/sharing/index.ts
```diff
@@ -1,5 +1,11 @@
...
+ Data in the share dataset will not be affected.`).replace('sharing.', `sharing ${shareName}.`);
```
wblock:
```suggestion
 Data in the share dataset will not be affected.`);
```

src/app/helptext/sharing/index.ts
aervin: How does this interact with the app's translation scheme? I think negatively?

aervin: The `replace` won't work correctly either for a translated language, so the attempt was futile anyway...

wblock: Good point.  I don't know whether that will be a problem.  But then, the replacement only works if 'sharing.' is translated the same in both places, and some languages do not necessarily do that (Czech has seven different ways of using nouns, for instance).  Maybe make it more explicit?
```suggestion
  T(`Stop sharing --sharename--. The sharing configuration will be removed.\
```

src/app/helptext/sharing/index.ts
aervin: Haha alright how about Unshare `<share_name>` for the dialog title. Then a generic dialog message: `The sharing configuration will be removed. Data in the share dataset will not be affected.`?

aervin:
Haha alright how about Unshare <share_name> for the dialog title. Then a generic dialog message: The sharing configuration will be removed. Data in the share dataset will not be affected.?

aervin:
fn-share-dialog-02

PR #2526, opened 2019-07-16
Nas 100172

dkmullen:
Replaces 'storage controllers' with TrueNAS controllers (controllers is capitalized in labels, not capitalized in sentences or otherwise. Thought about making a global var, but not sure if that breaks translation.

### DIFF
--- a/src/app/helptext/system/failover.ts
+++ b/src/app/helptext/system/failover.ts
@@ -4,11 +4,11 @@ import { T } from "app/translate-marker";
 export const helptext_system_failover = {
   dialog_initiate_failover_title: T("Initiate Failover"),
   dialog_initiate_failover_message: T("Initiating failover will temporarily disable services while failing over.  Failover now?"),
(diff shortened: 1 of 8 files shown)


src/app/helptext/system/failover.ts
```diff
@@ -1,14 +1,15 @@
 import { Validators } from "@angular/forms";
 import { T } from "app/translate-marker";
+import globalHelptext from '../../helptext/global-helptext';
 
 export const helptext_system_failover = {
   dialog_initiate_failover_title: T("Initiate Failover"),
   dialog_initiate_failover_message: T("Initiating failover will temporarily disable services while failing over.  Failover now?"),
-  dialog_initiate_failover_checkbox: T("Reboot current storage controller?"),
+  dialog_initiate_failover_checkbox: T(`Reboot current ${globalHelptext.ctrlr}?`),
```
wblock:
```suggestion
  dialog_initiate_failover_checkbox: T(`Reboot active ${globalHelptext.ctrlr}?`),
```

PR #2533, opened 2019-07-18
NAS 102551

lola27:


src/app/pages/jails/jail-add/jail-add.component.ts
```diff
@@ -1027,70 +1026,37 @@ export class JailAddComponent implements OnInit, AfterViewInit {
...
       (res) => {
-        if (res.result) {
-          for (const i in res.result) {
-            this.template_list.push(res.result[i][1]);
-            this.releaseField.options.push({ label: res.result[i][1] + '(template)', value: res.result[i][1] });
-          }
-        }
-        if (res.error) {
-          this.dialogService.errorReport(T('Error: Displaying templates failed.'), res.error, res.exception);
+        for (let i = 0; i < res.length; i++) {
```
aervin: `for (const template of templates) {...}` is easier to read imo. not a dealbreaker though :wink:

lola27: Good suggestion! I will change it :) thanks

PR #2542, opened 2019-07-22
NAS-102395 Add function to read fuzzy data measurements, use it in re…

dkmullen:
Thanks to Warren Block for writing the functions to convert human readable-input (45M, 20GiB, etc) to bytes. These functions (convertHumanStringToNum() and its helper functions) are in the storage service and can be used elsewhere in the app. 

This PR applies those functions to replication form speed limit and proposes some new labeling and tooltip text to reflect the goals stated in the ticket


src/app/services/storage.service.ts
```diff
@@ -240,4 +240,26 @@ export class StorageService {
...
     return path.indexOf('/') < 0;
   }
+
+  // Takes a string with fuzzy units like 20Mib, 20k, 20Gb, returns number in bytes
+  convertDataMeasurementStrings(str: string) {
+    if(!str) { return 0}; // if empty, return the default 0, which is unlimited
+    const values = {b: 1, k:1024, m:Math.pow(1024, 2), g:Math.pow(1024, 3), t: Math.pow(1024, 4)};
+    const letters = str.replace( /[^a-zA-Z]/g, '');
+    const number = parseInt(str.replace(/\D/g,''));
+    let tempArr = [];
```
wblock: Easier, just use the unit letter as the array index:
```suggestion
        return value * powersOf1024[unit];
```

src/app/services/storage.service.ts
```diff
@@ -240,4 +240,26 @@ export class StorageService {
 
     return path.indexOf('/') < 0;
   }
+
+  // Takes a string with fuzzy units like 20Mib, 20k, 20Gb, returns number in bytes
+  convertDataMeasurementStrings(str: string) {
+    if(!str) { return 0}; // if empty, return the default 0, which is unlimited
+    const values = {b: 1, k:1024, m:Math.pow(1024, 2), g:Math.pow(1024, 3), t: Math.pow(1024, 4)};
```
wblock:
```suggestion
    const powersOf1024 = {b: 1, k:1024, m:Math.pow(1024, 2), g:Math.pow(1024, 3), t: Math.pow(1024, 4)};
```

src/app/services/storage.service.ts
```diff
@@ -240,4 +240,26 @@ export class StorageService {
...
+    const letters = str.replace( /[^a-zA-Z]/g, '');
```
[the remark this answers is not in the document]
dkmullen: Some thing that should work don't in Angular, such as the i for case-insensitive (hence the 'toLowerCase()s). It still ignores anything after the first letter...

src/app/pages/task-calendar/replication/replication-form/replication-form.component.ts
```diff
@@ -784,6 +786,18 @@ export class ReplicationFormComponent {
                 }
             }
         )
+
+        entityForm.formGroup.controls['speed_limit'].valueChanges.subscribe((value) => {
+            const speedLimitField = _.find(this.fieldConfig, {name: "speed_limit"});
+            const filteredValue = this.storageService.convertDataMeasurementStrings(value);
+                    if (isNaN(filteredValue)) {
+                        speedLimitField['hasErrors'] = true;
+                        speedLimitField['errors'] = 'Invalid value'
```
wblock: If this error string is shown to the user, it should show an example of correct usage.  If formatting can be included in that message, the values should be in <i> tags or a monospace font or both.
```suggestion
                        speedLimitField['errors'] = 'Invalid value. Valid values are numbers followed by unit letters, like 256k or 1G.'
```

src/app/pages/task-calendar/replication/replication-form/replication-form.component.ts
```diff
@@ -784,6 +786,18 @@ export class ReplicationFormComponent {
                 }
             }
         )
+
+        entityForm.formGroup.controls['speed_limit'].valueChanges.subscribe((value) => {
+            const speedLimitField = _.find(this.fieldConfig, {name: "speed_limit"});
+            const filteredValue = this.storageService.convertDataMeasurementStrings(value);
+                    if (isNaN(filteredValue)) {
```
wblock: Might be clearer to do it this way:
```
const speedLimitField = _.find(this.fieldConfig, {name: "speed_limit"});
const filteredValue = this.storageService.convertDataMeasurementStrings(value);
speedLimitField['hasErrors'] = false;
speedLimitField['errors'] = '';
if (isNaN(filteredValue)) {
    speedLimitField['hasErrors'] = true;
    speedLimitField['errors'] = 'Invalid value'
}

File: src/app/pages/task-calendar/replication/replication-form/replication-form.component.ts [hunk 6]
```diff
@@ -784,6 +786,18 @@ export class ReplicationFormComponent {
...
+                        speedLimitField['hasErrors'] = true;
```
wblock: These two fields could be combined, using the presence of a string as the boolean value. In other words:
```
if (speedLimitField['errors']) { print('there are errors') }
```

src/app/services/storage.service.ts
```diff
@@ -240,4 +240,16 @@ export class StorageService {
...
   }
+
+  // Takes a string with fuzzy units like 20Mib, 20k, 20Gb, returns number in bytes
+  convertDataMeasurementStrings(str: string) {
+    if(!str) { return 0 }; // if empty, return the default 0, which is unlimited
+    if(str[0] && !str[0].match(/[0-9]/)) { return NaN };
+    const letters = str.replace( /[^a-zA-Z]/g, '');
+    if (letters[0] && !letters[0].toLowerCase().match(/b|k|m|g|t/i)) { return NaN };
+    const powersOf1024 = {b: 1, k: 1024, m: 1024**2, g: 1024**3, t: 1024**4};
+    const unit = str.toLowerCase().match(/[b|k|m|g|t]/i) || 'b';
```
wblock: Neat!  That's a nice way to set a default.  Might as well remove the "i" flag, since it's not being used.
```suggestion
    const unit = str.toLowerCase().match(/[b|k|m|g|t]/) || 'b';
```

src/app/services/storage.service.ts
```diff
@@ -240,4 +240,16 @@ export class StorageService {
...
+    const value = str.match(/^\D*(\d+)/) ? str.match(/^\D*(\d+)/)[1] : '0';
```
wblock: Hm.  Can't really skip over non-digit characters (\D), because those could be unit numbers that should not be allowed before a number, like "g 1".  Better to only match what is allowed.  Any number of whitespace characters followed by one or more digits is what I'd use.This works (although only trivially tested):
```
const value = (num = str.match(/^\s*(\d+)/)) ? num[1] : '0';
```

dkmullen: I cannot get this to accept the leading spaces.

src/app/pages/task-calendar/replication/replication-form/replication-form.component.ts
dkmullen: If I understand, 'errors' and 'hasErrors' are to separate properties that have to be set on our form field.

src/app/services/storage.service.ts
dkmullen: As written, this validates anything that doesn't match by replacing it with zero. If you change '0' at the end to NaN, it throws the error if you enter a space, but will clear the err message and accept it if you eventually follow with a number.

src/app/helptext/task-calendar/replication/replication.ts
```diff
@@ -123,8 +123,10 @@ snapshots will be stored. Example: Poolname/Datasetname, not the mountpoint or f
     compression_placeholder: T('Stream Compression'),
     compression_tooltip: T('Select a compression algorithm to reduce the size of the data being replicated.'),
 
-    speed_limit_placeholder: T('Limit (KiB/s)'),
-    speed_limit_tooltip: T('Limit replication speed to the specified value in kilobytes/second. The default 0 is unlimited.'),
+    speed_limit_placeholder: T('Limit (Ex. 500 KiB/s, 500M, 2 TB)'),
+    speed_limit_tooltip: T('Limit replication speed to this number of bytes per second.'),
+    speed_limit_errors: T('Invalid value. Valid values are numbers followed by unit letters, \
```
wblock:
```suggestion
    speed_limit_errors: T('Invalid value. Valid values are numbers followed by optional unit letters, \
```

src/app/services/storage.service.ts
```diff
@@ -240,4 +240,16 @@ export class StorageService {
 
     return path.indexOf('/') < 0;
   }
+
+  // Takes a string with fuzzy units like 20Mib, 20k, 20Gb, returns number in bytes
+  convertHumanStringtoNum(str: string) {
+    if(!str) { return 0 }; // if empty, return the default 0, which is unlimited
+    const letters = str.replace( /[^a-zA-Z]/g, '');
```
wblock: The match should be that a letter is only allowed after numbers.  A letter alone is not valid ("G"), and a letter before numbers is also not valid ("g10").

src/app/services/storage.service.ts
```diff
@@ -240,4 +240,16 @@ export class StorageService {
...
+    if (letters[0] && !letters[0].toLowerCase().match(/b|k|m|g|t/)) { return NaN };
+    const powersOf1024 = {b: 1, k: 1024, m: 1024**2, g: 1024**3, t: 1024**4};
+    const unit = str.toLowerCase().match(/[b|k|m|g|t]/) || 'b';
```
wblock: The square brackets make that into a character class, so the vertical bar "or"s are not needed.

src/app/services/storage.service.ts
```diff
@@ -240,4 +240,16 @@ export class StorageService {
 
     return path.indexOf('/') < 0;
   }
+
+  // Takes a string with fuzzy units like 20Mib, 20k, 20Gb, returns number in bytes
+  convertHumanStringtoNum(str: string) {
+    if(!str) { return 0 }; // if empty, return the default 0, which is unlimited
+    const letters = str.replace( /[^a-zA-Z]/g, '');
+    if (letters[0] && !letters[0].toLowerCase().match(/b|k|m|g|t/)) { return NaN };
```
wblock: This can be a character class like was used above.  The error handling probably needs to be more involved, though.

src/app/services/storage.service.ts
```diff
@@ -240,4 +240,16 @@ export class StorageService {
...
+    const powersOf1024 = {b: 1, k: 1024, m: 1024**2, g: 1024**3, t: 1024**4};
+    const unit = str.toLowerCase().match(/[b|k|m|g|t]/) || 'b';
+    let num;
+    const value = (num = str.match(/^\s*(\d+)/)) ? num[1] : NaN;
+    return parseInt(value) * powersOf1024[unit.toString()];
```
wblock: Not sure why the toString call is here.  That list should return a number.

dkmullen: This is a Typescript thing. TS says that unit is of type RegExpMatchArray. I can't use parseInt() to make it an integer either, but I can change it to a string, and JS knows what to do with a string here.

image

src/app/services/storage.service.ts
dkmullen: I'll do that...

src/app/services/storage.service.ts
dkmullen: ...and that.

src/app/services/storage.service.ts
dkmullen: This doesn't allow a letter first, it just says that the first letter that appears has to be one of our set, and if it isn't, we return NaN and stop there. The reason for adding it is that the const unit = ... regex below seems to turn any letter not in our list to a 'b'.

PR #2545, opened 2019-07-23
Nas 102290 -- jail delete "force" checkbox

aervin:


aervin:
BRB determined this feature does not need to be exposed in the UI.

PR #2561, opened 2019-07-25
Nas 102627

dkmullen:
Hides Ui pointers to legacy UI in Freenas, and shows them in TrueNAS only if user selects this option in System Advanced. This PR contains some dialogs and a tooltip and needs docs input.

### DIFF
--- a/src/app/components/common/topbar/topbar.component.ts
+++ b/src/app/components/common/topbar/topbar.component.ts
@@ -323,8 +323,8 @@ export class TopbarComponent implements OnInit, OnDestroy {
 
   onGoToLegacy() {
     this.dialogService.confirm(T("Warning:"), 
-      T("The Legacy User Interface is deprecated. The system should be managed from the new UI."), 
-      true, T("I understand")).subscribe((res) => {
+      helptext.legacyUIWarning, 
+      true, T("Continue to Legacy UI")).subscribe((res) => {
       if (res) {
         window.location.href = '/legacy/';
       }
--- a/src/app/components/common/topbar/topbar.component.ts
+++ b/src/app/components/common/topbar/topbar.component.ts
@@ -322,7 +322,9 @@ export class TopbarComponent implements OnInit, OnDestroy {
   }
 
   onGoToLegacy() {
-    this.dialogService.confirm(T("Switch to Legacy User Interface?"), T(""), true, T("Continue")).subscribe((res) => {
+    this.dialogService.confirm(T("Warning:"), 
+      T("The Legacy User Interface is deprecated. The system should be managed from the new UI."), 
+      true, T("I understand")).subscribe((res) => {
       if (res) {
         window.location.href = '/legacy/';
       }
--- a/src/app/components/common/topbar/topbar.component.ts
+++ b/src/app/components/common/topbar/topbar.component.ts
@@ -1,5 +1,5 @@
 
-import {interval as observableInterval,  Observable ,  Subscription } from 'rxjs';
+import {interval as observableInterval,  Subscription } from 'rxjs';
 import { Component, EventEmitter, Input, OnDestroy, OnInit, Output } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
 import * as domHelper from '../../../helpers/dom.helper';
@@ -13,7 +13,6 @@ import { TaskManagerComponent } from '../dialog/task-manager/task-manager.compon
 import { DirectoryServicesMonitorComponent } from '../dialog/directory-services-monitor/directory-services-monitor.component';
 import { NotificationAlert, NotificationsService } from '../../../services/notifications.service';
 import { MatSnackBar, MatDialog, MatDialogRef } from '@angular/material';
-import * as hopscotch from 'hopscotch';
 import { RestService } from "../../../services/rest.service";
 import { LanguageService } from "../../../services/language.service"
 import { TranslateService } from '@ngx-translate/core';
@@ -89,6 +88,7 @@ export class TopbarComponent implements OnInit, OnDestroy {
         this.getHAStatus();
       });
       this.sysName = 'TrueNAS';
+      this.checkLegacyUISetting();
     }
     let theme = this.themeService.currentTheme();
     this.currentTheme = theme.name;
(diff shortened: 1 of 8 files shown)


src/app/components/common/topbar/topbar.component.ts
```diff
@@ -313,7 +322,9 @@ export class TopbarComponent implements OnInit, OnDestroy {
   }
 
   onGoToLegacy() {
-    this.dialogService.confirm(T("Switch to Legacy User Interface?"), T(""), true, T("Continue")).subscribe((res) => {
+    this.dialogService.confirm(T("Warning:"), 
+      T("The Legacy User Interface is deprecated. The system should be managed from the new UI."),
```
wblock: Slight rewording.  Can we make the "new user interface" a link to the new user interface?
```suggestion
      T("The legacy user interface is deprecated. Please manage the system with the new user interface."), 
```

src/app/components/common/topbar/topbar.component.ts
```diff
@@ -313,7 +322,9 @@ export class TopbarComponent implements OnInit, OnDestroy {
   }
 
   onGoToLegacy() {
-    this.dialogService.confirm(T("Switch to Legacy User Interface?"), T(""), true, T("Continue")).subscribe((res) => {
+    this.dialogService.confirm(T("Warning:"), 
+      T("The Legacy User Interface is deprecated. The system should be managed from the new UI."), 
+      true, T("I understand")).subscribe((res) => {
```
wblock: Not clear what the "I understand" does--enable the legacy UI, or just make the warning go away?

src/app/helptext/system/advanced.ts
```diff
@@ -36,6 +36,10 @@ export const helptext_system_advanced = {
  DISCOURAGED</b>'),
   swapondrive_validation: [ Validators.required, Validators.min(0), Validators.max(99) ],
 
+  enable_legacy_placeholder: T('Enable Legacy UI'),
+  enable_legacy_tooltip: T('WARNING: The Legacy UI is deprecated and may lack functionality. Please perform \
```
wblock:
```suggestion
  enable_legacy_tooltip: T('WARNING: The legacy user interface is deprecated and may lack functionality. Please perform \
```

src/app/helptext/system/advanced.ts
```diff
@@ -36,6 +36,10 @@ export const helptext_system_advanced = {
...
+ all management through the new UI.'),
```
wblock: As above, can we make "new user interface" a link to the new UI?

```suggestion
 all management through the new user interface.'),
```

src/app/pages/system/advanced/advanced.component.ts
```diff
@@ -246,42 +252,61 @@ export class AdvancedComponent implements OnDestroy {
...
+        if (parseInt(value) === 0) {
+          this.swapondrive.warnings = helptext_system_advanced.swapondrive_warning;
+        } else {
+          this.swapondrive.warnings = null;
+        }
+      });
+      setTimeout(() => {
+        entityEdit.formGroup.controls['legacy_ui'].valueChanges.subscribe((value) => {
+          if (value) {
+            this.dialog.confirm('Warning', 'The Legacy UI is deprecated and may \
```
wblock: This should use the helptext string rather than repeat a literal string. And the button is now "I accept the risks", different from the one above.

src/app/views/sessions/signin/signin.component.ts
```diff
@@ -260,7 +263,9 @@ export class SigninComponent implements OnInit {
   }
 
   onGoToLegacy() {
-    this.dialogService.confirm(T("Log in to Legacy User Interface?"), "", true, T('Continue')).subscribe((res) => {
+    this.dialogService.confirm(T("Warning:"), 
+      "The Legacy User Interface is deprecated. The system should be managed from the new UI.",
```
wblock: As above, use the global helptext to avoid repeating a literal string.

src/app/views/sessions/signin/signin.component.ts
```diff
@@ -260,7 +263,9 @@ export class SigninComponent implements OnInit {
...
+       true, T('I understand')).subscribe((res) => {
```
wblock: And here.

wblock:
It doesn't make sense to put this in System/Advanced.  The option is only available from the new UI, so this should be a setting under the user preferences menu (Gear in upper right, Preferences).

src/app/pages/system/advanced/advanced.component.ts
dkmullen: Regarding the button message, I'm changing the two dialogs that lead to the legacy UI to 'Continue to the..." but this one is different because it is a checkbox chainging a setting, not actually leading to the UI

src/app/components/common/topbar/topbar.component.ts
dkmullen: Changed to 'Continue to Legacy UI' on the button (hope UI is ok here, since the message above calls it the user interface.

src/app/components/common/topbar/topbar.component.ts
dkmullen: Remember these items are all in the new UI. I think a link has been added to the warning in the old UI

src/app/views/sessions/signin/signin.component.ts
```diff
@@ -260,7 +263,9 @@ export class SigninComponent implements OnInit {
   }
 
   onGoToLegacy() {
-    this.dialogService.confirm(T("Log in to Legacy User Interface?"), "", true, T('Continue')).subscribe((res) => {
+    this.dialogService.confirm(T("Warning:"),
```
Mrt134:
```suggestion
    this.dialogService.confirm(T("Warning"),
```
The `:` appears to be unnecessary for this dialog.

william-gr:
> @dkmullen please make sure we offer user to restart web server after changing this option (as we do when setting/unsetting HTTP->HTTPs redirect in General Settings).

Restart webserver is not necessary when changing legacy UI option.

wblock:
The wording of the message was meant to match the message shown in the legacy UI.  The new wording changes the meaning from "do not use the legacy UI any more" to suggest that the old UI is fine and the new UI just adds new features.

PR #2569, opened 2019-07-26
NAS-102652

lola27:


lola27:
@dkmullen the tooltip is fine since this form is only used to update config

PR #2570, opened 2019-07-26
NAS-102657

lola27:


lola27:
@freenas/docs-team could you please provide a tooltip for the added field?

PR #2581, opened 2019-07-30
NAS-102709

lola27:


lola27:
@dkmullen that call is not a job anymore, and they do return a job id (which is for debug generation) and a download link. but the generation job is always failed with a `Broken pipe` error. I will add it if that is fixed.

PR #2582, opened 2019-07-30
NAS-102155 

aervin:
* "No snapshots sent yet" if rep task does not have a previous run
* Unfortunately, the column widgets cannot support show/hide on a per-row basis, so future runs cannot be supported at this time. Perhaps we could target for 12?
* Moves task parsing from `dataHandler` to `resourceTransformIncomingRestData` so that the column data updates dynamically as the task runs

QUICK UPDATES:
* Fix for tooltip formatting bug
* Adds support for mdi icons in table widgets
* Make sure that table detail actions are visible at the right times

### DIFF
--- a/src/app/helptext/task-calendar/replication/replication.ts
+++ b/src/app/helptext/task-calendar/replication/replication.ts
@@ -46,7 +46,7 @@ export default {
 
     target_dataset_placeholder: T('Target Dataset'),
     target_dataset_tooltip: T('Enter the dataset on the remote or destination system where\
-snapshots will be stored. Example: Poolname/Datasetname, not the mountpoint or filesystem path'),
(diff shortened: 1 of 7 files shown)


src/app/helptext/task-calendar/replication/replication.ts
```diff
@@ -46,7 +46,7 @@ export default {
 
     target_dataset_placeholder: T('Target Dataset'),
     target_dataset_tooltip: T('Enter the dataset on the remote or destination system where\
-snapshots will be stored. Example: Poolname/Datasetname, not the mountpoint or filesystem path'),
+ snapshots will be stored. Example: Poolname/Datasetname, not the mountpoint or filesystem path'),
```
wblock:
```suggestion
 snapshots will be stored. Example: Poolname/Datasetname, not the mountpoint or filesystem path.'),
```

PR #2596, opened 2019-08-02
NAS-102736

aervin:
* Adds ID and name props to all action buttons in volume-list.component
* Quick update to changelog parsing so that newlines work and the response is formatted in a readable way.

### DIFF
--- a/src/app/pages/storage/volumes/volumes-list/volumes-list.component.ts
+++ b/src/app/pages/storage/volumes/volumes-list/volumes-list.component.ts
@@ -340,6 +340,8 @@ export class VolumesListTableConfig implements InputTableConf {
     if (rowData.type === 'zpool') {
 
       actions.push({
+        id: rowData.name,
+        name: 'export_disconnect',
         label: T("Export/Disconnect"),
         onClick: (row1) => {
           this.loader.open();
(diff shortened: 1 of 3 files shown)


src/app/pages/storage/volumes/volumes-list/volumes-list.component.ts
```diff
@@ -340,6 +340,8 @@ export class VolumesListTableConfig implements InputTableConf {
     if (rowData.type === 'zpool') {
 
       actions.push({
+        id: rowData.name,
+        name: 'export_disconnect',
```
wblock: This should match the label precisely, so maybe:
```suggestion
        name: 'Export/Disconnect',
        label: T(rowData.name),
```
For that matter, menu options ought to all be in the global helptext files, but maybe we haven't done that yet.

aervin: Agree the actual button text ought to be in the helptext files, but those will be translated. I'm guessing we'd like the Ids to be static for automation purposes?

lola27:
rebuild

src/app/pages/storage/volumes/volumes-list/volumes-list.component.ts
```diff
@@ -340,6 +340,9 @@ export class VolumesListTableConfig implements InputTableConf {
     if (rowData.type === 'zpool') {
 
       actions.push({
+        id: rowData.name,
+        name: 'Export/Disconnect',
+        label: T(rowData.name),
```
lola27: should remove the duplicate label (line 345)

PR #2631, opened 2019-08-14
NAS-102777

lola27:


src/app/pages/storage/volumes/volume-status/volume-status.component.ts
```diff
@@ -175,6 +178,8 @@ export class VolumeStatusComponent implements OnInit {
     }, {
       label: T("Offline"),
       onClick: (row) => {
+        const encryptPoolWarning = T('<br><b>Warnning: online disk is not supported for encrypted pool!</b></br>');
```
wblock: Typo, and not clear that "online" is a verb.  Not sure whether the hard breaks or bold are needed or should be done elsewhere.
```suggestion
        const encryptPoolWarning = T('<br><b>Warning: Disks cannot be onlined in encrypted pools.</b></br>');
```

PR #2634, opened 2019-08-14
NAS 102863

lola27:


src/app/pages/plugins/available-plugins/available-plugins.component.html
```diff
@@ -25,7 +25,8 @@
         <div fxFlex="25" fxFlexOffset="3" fxFlex.xs="100">
             <div class="plugin-vinfo">
                 <p *ngIf="isSelectedOffical">Official Plugin <mat-icon class="official-plugin">verified_user</mat-icon></p>
-                <p>Version: {{selectedPlugin.revision}}</p>
+                <p>Version: {{selectedPlugin.version}}</p>
+                <p>Revision: {{selectedPlugin.revision}}</p>
```
wblock: This might be clearer at the end of the previous line.  So when people are asked which version they have, the revision will be right there implying that it goes along with the version.  Might also need to only display revision if it has a value to avoid confusion.
```suggestion
                <p>Version: {{selectedPlugin.version}} Revision {{selectedPlugin.revision}}</p>
```

PR #2646, opened 2019-08-15
NAS 102745

lola27:


src/app/helptext/storage/volumes/datasets/dataset-permissions.ts
```diff
@@ -1,38 +1,41 @@
...
 
-dataset_permissions_mp_acl_placeholder: T('ACL Type'),
-dataset_permissions_mp_acl_tooltip: T('Select the type that matches the type of client\
+dataset_permissions_acl_placeholder: T('ACL Type'),
+dataset_permissions_acl_tooltip: T('Select the type that matches the type of client\
  accessing the dataset.'),
 
-dataset_permissions_mp_user_en_placeholder: T('Apply User'),
-dataset_permissions_mp_user_en_tooltip: T('Apply changes to the user.'),
+dataset_permissions_user_en_placeholder: T('Apply User'),
```
wblock: Not clear what "_en_" means here and in some later fields.  I'd suspect it means English, but that's not usually how we do things.  They are not used in or are removed from the other files in this diff.  Are they duplicates that can be removed?

lola27: thanks @wblock. They should be removed.

PR #2667, opened 2019-08-22
Nas 102879

dkmullen:
Checks (from topbar) for a running update process (freenas or truenas), sends a message through SystemGeneral service to System Widget and Update page to warn the user that an update has been started, possibly by TrueCommand (the most likely use for this), and also displays a dialog (once).


dkmullen:
The method for checking if an update is running doesn't return 'true' if the user is only downloading the update, so the dialog only appears if the update process begins - so, the message says that the system WILL restart, and doesn't need to be conditional.

PR #2675, opened 2019-08-23
NAS-100942 Add warning when saving WebDAV shares

dkmullen:


src/app/helptext/sharing/webdav/webdav.ts
```diff
@@ -24,5 +24,11 @@ export const helptext_sharing_webdav = {
     column_comment: T('Comment'),
     column_path: T('Path'),
     column_ro: T('Read Only'),
-    column_perm: T('Change User & Group Ownership')
+    column_perm: T('Change User & Group Ownership'),
```
wblock: Only use "&" in the name of a law firm.
```suggestion
    column_perm: T('Change User and Group Ownership'),
```

PR #2699, opened 2019-08-30
Removed option to hide help text in User Preferences.

dszidi:
Removed option to hide help text icons


src/app/pages/preferences/page/forms/general-preferences-form.component.ts
```diff
@@ -65,14 +65,14 @@ export class GeneralPreferencesFormComponent implements OnInit, OnChanges, OnDes
             tooltip: 'Preserve screen space with icons and tooltips instead of text labels.',
             class:'inline'
           },
-          {
+          /*{
```
wblock: Why comment rather than just remove?

dszidi: Oops. I usually comment things out while testing and then remove. I seemed to have missed this one.

PR #2717, opened 2019-09-05
New support

dkmullen:
NAS-100746
Rebuild of support page; Breaks out Prod. Status, Proactive Support, FN and TN support into seperate Entity Forms; Other stuff is divided among components to make layout more flexible.


src/app/helptext/system/support.ts
```diff
@@ -107,6 +135,38 @@ export const helptext_system_support = {
...
+  is_production_error_dialog: {
+    title: T('Error Updating Production Status'),
+  },
+
+  is_production_snackbar: {
+    message: T('Production status successfully updated'),
+    action: T('Close')
+  },
+
+  FN_instructions: 'Search the <a href="https://jira.ixsystems.com/projects/NAS/issues/" \
```
erinix: Forgot a T()

PR #2758, opened 2019-09-13
Nas 103111

dkmullen:
Refreshes page via a dialog when a TN license is accepted; also removes the old License Update button!


erinix:
rebuild

PR #2766, opened 2019-09-16
NAS-103089: Add help text to new "Region" field

Mrt134:
- Yarn build test: no issues.

### DIFF
--- a/src/app/helptext/system/cloudcredentials.ts
+++ b/src/app/helptext/system/cloudcredentials.ts
@@ -62,9 +62,11 @@ export const helptext_system_cloudcredentials = {
     placeholder: T("Region"),
     tooltip: T('<a href="https://docs.aws.amazon.com/general/latest/gr/rande-manage.html" \
  target="_blank">AWS resources in a geographic area</a>. Leave empty to \
- automatically detect the correct public region for the bucket. Enter a \
- private region name to allow the credential to find Amazon buckets \
- created in that region.'),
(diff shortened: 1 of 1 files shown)


themylogin:
@Mrt134 us-east-1 default is not correct! It should be something like "Leave empty to detect region automatically. This won't work with private regions such as AWS GovCloud"

Mrt134:
Understood, I'm just looking at your comments now and should have a PR update soon.

src/app/helptext/system/cloudcredentials.ts
```diff
@@ -57,9 +57,14 @@ export const helptext_system_cloudcredentials = {
...
     )
   },
+
   region_s3: {
     placeholder: T("Region"),
-    tooltip: T(""),
+    tooltip: T('<a href="https://docs.aws.amazon.com/general/latest/gr/rande-manage.html" \
+ target="_blank">AWS resources in a geographic area</a>. Leave empty to \
+ automatically detect the correct public region for the bucket. Enter a \
+ private region name to allow the credential to find Amazon buckets \
```
themylogin: Credential does not find anything, can it be something like "Enter a private region name to allow using this credential with S3 buckets created in that region."

Mrt134: Yes, I can reword that.

Mrt134:
Ok, I reworded the private region description slightly and added a link and example of configuring with AWS GovCloud

PR #2775, opened 2019-09-18
NAS-103152

dkmullen:
Removes 'disable config dialog' from preferences, cleans up lots of unneeded code, especially in upgrade component

### DIFF
--- a/src/app/pages/system/update/update.component.ts
+++ b/src/app/pages/system/update/update.component.ts
@@ -442,7 +442,7 @@ export class UpdateComponent implements OnInit, OnDestroy {
                 } else {
                   this.ds  = this.dialogService.confirm(
                     T("Download Update"), T("Upgrades both controllers. Files will be downloaded in the Active Controller\
-                      and then transferred to the Standby Controller. Upgrade process will start concurrently on both nodes.\
+                      and then transferred to the Standby Controller. The upgrade process starts concurrently on both TrueNAS Controllers.\
                       Continue with download?"),true).subscribe((res) =>  {
                       if (res) {
                         this.update()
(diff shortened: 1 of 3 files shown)


src/app/pages/system/update/update.component.ts
```diff
@@ -419,92 +411,45 @@ export class UpdateComponent implements OnInit, OnDestroy {
...
+                      this.update();
+                    }
+                  }
+                });
                 } else {
                   this.ds  = this.dialogService.confirm(
                     T("Download Update"), T("Upgrades both controllers. Files will be downloaded in the Active Controller\
-                    and then transferred to the Standby Controller. Upgrade process will start concurrently on both nodes.\
-                    Continue with download?"), true).subscribe((res) =>  {
+                      and then transferred to the Standby Controller. Upgrade process will start concurrently on both nodes.\
```
Mrt134:
```suggestion
                      and then transferred to the Standby Controller. The upgrade process starts concurrently on both TrueNAS Controllers.\
```

src/app/pages/system/update/update.component.ts
```diff
@@ -419,92 +411,45 @@ export class UpdateComponent implements OnInit, OnDestroy {
...
+                      });
+                    }
+                    else{
+                      this.update();
+                    }
+                  }
+                });
                 } else {
                   this.ds  = this.dialogService.confirm(
                     T("Download Update"), T("Upgrades both controllers. Files will be downloaded in the Active Controller\
```
Mrt134:
```suggestion
                    T("Download Update"), T("Upgrades both controllers. Files are downloaded to the Active Controller\
```

PR #2792, opened 2019-09-23
Nas 103190

dkmullen:
Drops maxWidth from first and last displayed cols in User table.
If first and/or last cols have maxWidth set, there is a display glitch (see pic)
This is applied only to user table, b/c this is the only table with maxWidths set


dkmullen:
Screenshot from 2019-09-23 11-44-14

PR #2802, opened 2019-09-24
Nas 101423a

dkmullen:
Adds human-readability to the label for Available memory on VM wizard

### DIFF
--- a/src/app/services/storage.service.ts
+++ b/src/app/services/storage.service.ts
@@ -344,6 +344,7 @@ export class StorageService {
       return num * this.convertUnitToNum(unit);
   };
 
+  // Converts a number from bytes to the most natural human readable format
   convertBytestoHumanReadable(bytes) {
     let i = -1;
     const units = [' KiB', ' MiB', ' GiB', ' TiB', 'PiB'];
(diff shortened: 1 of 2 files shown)


src/app/services/storage.service.ts
```diff
@@ -342,5 +342,17 @@ export class StorageService {
...
+  // Converts a number from bytes to the most natural human readable format
+  convertBytestoHumanReadable(bytes) {
+    let i = -1;
+    const units = [' KiB', ' MiB', ' GiB', ' TiB', 'PiB'];
+    do {
+        bytes = bytes / 1024;
+        i++;
+    } while (bytes > 1024);
+  
+    return Math.max(bytes, 0.1).toFixed(2) + units[i];
```
wblock: Reuse the IECUnits already declared for the routines above.
Needs a spacer between the units and value.  Could have a problem with numbers less than 1024--would the user rather see "256 bytes" or "0.25 KiB"?
Number of decimals should be a parameter.
I also feel like this could be done more efficiently, but I'm not sure.

dkmullen: Latest version reuses IECunits, preserves the space, allows for 'bytes' if # is less than 1024, adds an optional parameter for decimal places which defaults to two.

PR #2810, opened 2019-09-25
NAS-103240

dkmullen:
NAS-103240
Adds an indeterminate progress indicator and a 'Uploading file...' message to the first dialog in manual update process

### DIFF
--- a/src/app/pages/common/entity/entity-job/entity-job.component.html
+++ b/src/app/pages/common/entity/entity-job/entity-job.component.html
@@ -21,7 +21,8 @@
 		</mat-progress-bar>
 	</div>
 	<div *ngIf="!description">
-		<span [innerHTML]="altMessage"></span>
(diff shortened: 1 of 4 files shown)


src/app/pages/common/entity/entity-job/entity-job.component.html
```diff
@@ -10,13 +10,21 @@ <h1 mat-dialog-title>{{title | translate}}</h1>
 		<label *ngIf="!hideProgressValue">
 			{{ progressTotalPercent | number: '1.2-2'}}%
 		</label>
-    </div>
+	</div>
+	<div *ngIf="description === 'Uploading file...'">
```
erinix: perhaps make this more generic so it can be used for other things

dkmullen: Latest commit should allow for any message, with default message set to 'Starting job...'

PR #2829, opened 2019-09-30
NAS-102450: Sync HTTP redirect help text with docs

Mrt134:


src/app/helptext/system/general.ts
```diff
@@ -62,8 +62,14 @@ export const helptext_system_general = {
...
     placeholder: T("WebGUI HTTP -> HTTPS Redirect"),
     tooltip: T(
-      "Check this to redirect <i>HTTP</i> connections to\
- <i>HTTPS</i>. A <i>GUI SSL Certificate</i> must be selected."
+      'Redirect <i>HTTP</i> connections to <i>HTTPS</i>. A \
+ <i>GUI SSL Certificate</i> is required for <i>HTTPS</i>. Activating \
+ this also sets the <a \
+ href="https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security" \
+ target="_blank">HTTP Strict Transport Security (HSTS)</a> maximum age \
+ to <i>31536000</i> seconds /(one year/). This means that after a \
```
wblock: Originally the slashes looked like escapes.  I don't think these are necessary, or need to be backslashes if they are:
```suggestion
 to <i>31536000</i> seconds (one year). This means that after a \
```

PR #2840, opened 2019-10-02
NAS-103197

lola27:


src/app/pages/sharing/iscsi/portal/portal-form/portal-form.component.ts
```diff
@@ -169,7 +169,9 @@ export class PortalFormComponent {
 
     this.iscsiService.getIpChoices().subscribe((res) => {
       for (const item in res) {
-        this.arrayModel.options.push({label : item, value : res[item]});
+        console.log(item);
```
erinix: you left a console.log in there :)

PR #2843, opened 2019-10-02
Zh hans 11.2 stable

dingyifei:
used weblate


erinix:
fyi, there is now a freenas/11.3-stable branch too, any translations you want in 11.3 will need to go in there

erinix:
actually nevermind that last message, we have not branched yet, just keep using master

PR #2847, opened 2019-10-03
NAS-103351

lola27:


src/app/pages/storage/snapshots/snapshot-list/snapshot-list.component.ts
```diff
@@ -16,10 +16,10 @@ export class SnapshotListComponent {
 
   public title = "Snapshots";
   protected queryCall = 'zfs.snapshot.query';
-  protected queryCallOption = [[["pool", "!=", "freenas-boot"]], {"select": ["name"], "order_by": ["name"]}];
+  protected queryCallOption = [[["pool", "!=", "freenas-boot"]], {"select": ["name", "id"], "order_by": ["name"]}];
```
william-gr: Please dont use id, it will slow down the query.

id is the same as name, you can just use that.

william-gr:
Thanks

PR #2874, opened 2019-10-09
Nas 103396

lola27:


src/app/pages/jails/jail-list/jail-list.component.html
```diff
@@ -1,32 +1,22 @@
 <mat-card *ngIf="availablePools != undefined">
-  <div *ngIf="availablePools.length > 0">
     <form (ngSubmit)="activatePool($event)">
-      <label *ngIf="activatedPool">
-        <h4>Activated Pool for Jail Manager: {{activatedPool}}</h4>
-      </label>
-      <div class="dynamic-field form-select has-tooltip" *ngIf="!activatedPool || toActivatePool">
+      <h4>Activated Pool for Jail Manager: {{activatedPool}}</h4>
```
erinix: "Activated Pool for Jail Manager:" should be piped into translation service

src/app/pages/jails/jail-list/jail-list.component.ts
```diff
@@ -178,36 +178,54 @@ export class JailListComponent implements OnInit {
...
 
-  getAvailablePools(){
-    this.ws.call('pool.query').subscribe( (res)=> {
-      this.availablePools = res;
+  noPoolDialog() {
+    const dialogRef = this.dialogService.confirm(
+      T('No Pools'),
+      T('Jails cannot be created or managed until a pool is present for storing them.'),
+      true,
+      T('Create Pool'));
```
wblock: Wait, no cancel button?  What does the user do if they are not ready to create a pool at that moment?

erinix: It still has a cancel button, it just uses the default which says "Cancel" for the text

PR #2875, opened 2019-10-09
Nas 103388

dkmullen:


src/app/pages/common/entity/entity-table/entity-table.component.ts
```diff
@@ -487,6 +488,13 @@ export class EntityTableComponent implements OnInit, AfterViewInit, OnDestroy {
       this.setPaginationInfo();
     }
 
+    if (this.title === 'Boot Environments') {
```
erinix: Try to make this rule more generic, add a boolean to the conf for this

PR #2882, opened 2019-10-10
Nas 103431

dkmullen:
Fixes some step errors, etc, to get error messages working on VM wizard


src/app/pages/vm/vm-wizard/vm-wizard.component.ts
```diff
@@ -191,6 +191,7 @@ export class VMWizardComponent {
         {
           type: 'input',
           name: 'volsize',
+          inputType: 'number',
```
erinix: so rather than using inputType number, I would recommend using a rangeValidator and set the minimum to 0, the rangeValidator validates against anything that isn't numbers and inputType number allows things like negative numbers

PR #2888, opened 2019-10-11
NAS-103452 Fix some labels and ids

dkmullen:


erinix:
These look good to me, what say you @wblock ?

PR #2890, opened 2019-10-12
Add validation for system general IP fields

erinix:
NAS-103423

### DIFF
--- a/src/app/helptext/system/general.ts
+++ b/src/app/helptext/system/general.ts
@@ -188,5 +188,10 @@ export const helptext_system_general = {
   snackbar_download_success: {
     title: T("Download Sucessful"),
     action: T("Success")
+  },
+
+  validation_errors: {
+    ui_address: T("When 0.0.0.0 has been selected, selection of other addresses is not allowed."),
+    ui_v6address: T("When :: has been selected, selection of other addresses is not allowed.")
   }
 };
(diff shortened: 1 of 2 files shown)


src/app/helptext/system/general.ts
```diff
@@ -188,5 +188,10 @@ export const helptext_system_general = {
   snackbar_download_success: {
     title: T("Download Sucessful"),
     action: T("Success")
+  },
+
+  validation_errors: {
+    ui_address: T("When 0.0.0.0 has been selected, selection of other addresses is not allowed."),
```
wblock: I don't know if this works as a general error.  Does selecting 0.0.0.0 or :: always mean "all addresses"?  If it does, we should say so: "When all has been selected by entering <samp>0.0.0.0</samp> or <samp>::</samp>, no other addresses can be entered."

erinix: I believe so, you can either pick specific ip addresses or the 0.0.0.0 or :: wildcards

wblock: Okay, how about:
"When all addresses have been selected by choosing <samp>0.0.0.0</samp> or <samp>::</samp>, additional addresses cannot be chosen."

src/app/helptext/system/general.ts
```diff
@@ -188,5 +188,10 @@ export const helptext_system_general = {
   snackbar_download_success: {
     title: T("Download Sucessful"),
     action: T("Success")
+  },
+
+  validation_errors: {
+    ui_address: T("When all addresses have been selected by choosing 0.0.0.0, additional addresses cannot be chosen."),
```
wblock:
```suggestion
    ui_address: T("Select <samp>0.0.0.0</samp> to include all addresses. When this has been chosen, additional addresses cannot be selected."),
```

PR #2893, opened 2019-10-13
re-order three dot menu so delete does not have priority

q5sys:
This resorts the Three dot menu to the following order:  Activate, Clone, Rename, Delete, Keep.


src/app/pages/system/bootenv/bootenv-list/bootenv-list.component.ts
```diff
@@ -250,7 +250,7 @@ export class BootEnvironmentListComponent {
   updateBootState(): void {
     this.ws.call("boot.get_state").subscribe(wres => {
       if (wres.scan.end_time) {
-        this.scrub_msg = moment(wres.scan.end_time.$date).format("MMMM Do YYYY, h:mm:ss a");
+        this.scrub_msg = moment(wres.scan.endhttps://jira.ixsystems.com/browse/NAS-103467_time.$date).format("MMMM Do YYYY, h:mm:ss a");
```
wblock: Paste error?

erinix: I think this paste error is also causing it to not build properly

q5sys:
yea, must have been an errant Ctrl-V. I'll fix it.

PR #2894, opened 2019-10-13
NAS-103463: Update create tooltip to include allowable chars

q5sys:

### DIFF
--- a/src/app/helptext/system/bootenv.ts
+++ b/src/app/helptext/system/bootenv.ts
@@ -22,8 +22,8 @@ export const helptext_system_bootenv = {
   create_name_placeholder: T("Name"),
   create_name_tooltip: T(
     "Enter the name of the boot entry.\
- Allowed characters are alphanumeric, as well as\
- dashes (*-*), underscores (*_*), and periods (*.*)"
+ Alphanumeric characters, dashes (*-*), underscores (*_*),\
+ and periods (*.*) are allowed.)"
  ),
 
   list_dialog_activate_action: T("Activate"),
--- a/src/app/helptext/system/bootenv.ts
+++ b/src/app/helptext/system/bootenv.ts
@@ -20,7 +20,11 @@ export const helptext_system_bootenv = {
   clone_source_tooltip: T("This is the boot environment to be cloned."),
 
   create_name_placeholder: T("Name"),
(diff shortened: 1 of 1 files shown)


src/app/helptext/system/bootenv.ts
```diff
@@ -20,7 +20,11 @@ export const helptext_system_bootenv = {
   clone_source_tooltip: T("This is the boot environment to be cloned."),
 
   create_name_placeholder: T("Name"),
-  create_name_tooltip: T("Enter the name of the boot entry."),
+  create_name_tooltip: T(
+    "Enter the name of the boot entry.\
+ Names may include alphanumeric characters as well as\
```
wblock: See #2896, but since these are all the same, this would be better as a shared string to not have repeated literal strings to get out of sync.

q5sys:
updated as requested

src/app/helptext/system/bootenv.ts
q5sys: That'd probably make sense.  The clone entry had a different first sentence so I don't know you want to handle that.  Making them all the same would be easier, I assume at some point in the past the tooltip dialog was different

src/app/helptext/system/bootenv.ts
```diff
@@ -20,7 +20,11 @@ export const helptext_system_bootenv = {
   clone_source_tooltip: T("This is the boot environment to be cloned."),
 
   create_name_placeholder: T("Name"),
-  create_name_tooltip: T("Enter the name of the boot entry."),
+  create_name_tooltip: T(
+    "Enter the name of the boot entry.\
+ Allowed characters are alphanumeric, as well as\
```
wblock:
```suggestion
 Alphanumeric characters, dashes (*-*), underscores (*_*),\
```

src/app/helptext/system/bootenv.ts
```diff
@@ -20,7 +20,11 @@ export const helptext_system_bootenv = {
...
+ dashes (*-*), underscores (*_*), and periods (*.*)"
```
wblock:
```suggestion
 and periods (*.*) are allowed."
```

q5sys:
Updated as requested.

PR #2895, opened 2019-10-13
NAS-103464: Update clone tooltip to include allowable chars

q5sys:


q5sys:
Updated as requested

q5sys:
Updated as requested.

PR #2896, opened 2019-10-13
NAS-103465: Update rename tooltip to include allowable chars

q5sys:

### DIFF
--- a/src/app/helptext/system/bootenv.ts
+++ b/src/app/helptext/system/bootenv.ts
@@ -30,8 +30,8 @@ export const helptext_system_bootenv = {
   rename_name_placeholder: T("Name"),
   rename_name_tooltip: T(
     "Rename the existing boot environment.\
-  Allowed characters are alphanumeric, as well as\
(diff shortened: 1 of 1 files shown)


src/app/helptext/system/bootenv.ts
```diff
@@ -28,7 +28,11 @@ export const helptext_system_bootenv = {
   list_dialog_scrub_action: T("Start Scrub"),
 
   rename_name_placeholder: T("Name"),
-  rename_name_tooltip: T("Rename the existing boot environment."),
+  rename_name_tooltip: T(
+    "Rename the existing boot environment.\
+  Names may include alphanumeric characters as well as\
```
wblock: "May" implies permission. 
```suggestion
  Allowed characters are alphanumeric,\
```

q5sys:
Updated as requested

wblock:
See change as in #2894.

q5sys:
Updated as requested.

PR #2898, opened 2019-10-14
Nas 103453

dkmullen:
Adds standard instructions to the doubleConfirm dialog about what to type in, provides a pseudo mask of the name of the dataset, plug-in, etc. to type over.

### DIFF
--- a/src/app/pages/plugins/plugins.component.ts
+++ b/src/app/pages/plugins/plugins.component.ts
@@ -57,7 +57,7 @@ export class PluginsComponent {
       doubleConfirm: (item) => {
         return this.dialogService.doubleConfirm(
           T('Verify Deletion of ') + item.name + T(' Plugin'),
-          T('To delete the <b>') + item.name + T('</b> plugin and all data and snapshots stored with it, please type the name of the plugin to confirm:'),
+          T('To delete the <b>') + item.name + T('</b> plugin and all data and snapshots stored with it, please type the name of the plugin to confirm.'),
           item.name,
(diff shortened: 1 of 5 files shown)


src/app/pages/plugins/plugins.component.ts
```diff
@@ -57,7 +57,7 @@ export class PluginsComponent {
       doubleConfirm: (item) => {
         return this.dialogService.doubleConfirm(
           T('Verify Deletion of ') + item.name + T(' Plugin'),
-          T('To delete the <b>') + item.name + T('</b> plugin and all data and snapshots stored with it, please type the name of the plugin to confirm:'),
+          T('To delete the <b>') + item.name + T('</b> plugin and all data and snapshots stored with it, please type the name of the plugin to confirm.'),
```
wblock: Can we make this a generic, reusable routine?  We already have similar functions here and for deleting datasets. We could reuse it for shares, users, groups, and other things.

dkmullen: That's what this PR does. The above text is the specific instructions for deleting plugins. They probably should be changed, because the reusuable part appears below  this:

`Enter <plug-in name> below to confirm.`)

This works for datasets too, and for any other place we use the double confirm.

dkmullen: PS - I updated the directions specific to plugins. 
Screenshot from 2019-10-14 11-53-53
Screenshot from 2019-10-14 11-54-30

PR #2901, opened 2019-10-14
Populate user and group fields with uid/gid and warn if the name could not be resolved

erinix:
NAS-103433

### DIFF
--- a/src/app/helptext/storage/volumes/datasets/dataset-acl.ts
+++ b/src/app/helptext/storage/volumes/datasets/dataset-acl.ts
@@ -159,6 +159,6 @@ dataset_acl_add_item_btn: T('Add ACL Item'),
 acl_defaults_placeholder: T('Default ACL Options'),
 acl_defaults_tooltip: T('Set default options on this ACL.'),
 
-user_not_found: T('User name could not be resolved.'),
+user_not_found: T('Could not find a user name for this user ID.'),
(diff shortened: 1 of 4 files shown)


src/app/helptext/storage/volumes/datasets/dataset-acl.ts
```diff
@@ -157,5 +157,8 @@ dataset_acl_toplevel_dialog_message: T('Editing top-level datasets can\
 dataset_acl_add_item_btn: T('Add ACL Item'),
 
 acl_defaults_placeholder: T('Default ACL Options'),
-acl_defaults_tooltip: T('Set default options on this ACL.')
+acl_defaults_tooltip: T('Set default options on this ACL.'),
+
+user_not_found: T('User name could not be resolved.'),
```
wblock: After talking with Andrew:
```suggestion
user_not_found: T('Could not find a user name for this user ID.'),
```

src/app/helptext/storage/volumes/datasets/dataset-acl.ts
```diff
@@ -157,5 +157,8 @@ dataset_acl_toplevel_dialog_message: T('Editing top-level datasets can\
...
+group_not_found: T('Group name could not be resolved.')
```
wblock: As above:
```suggestion
group_not_found: T('Could not find a group name for this group ID.')
```

PR #2947, opened 2019-10-22
NAS-103439

dkmullen:
Show admin portal by default, put ip addresses on panel


src/app/pages/plugins/plugins.component.ts
```diff
@@ -271,6 +272,7 @@ export class PluginsComponent {
   }
 
   dataHandler(entityList: any) {
+    console.log(entityList)
```
erinix: remove the console.log :)

PR #2952, opened 2019-10-23
NAS-103584

dkmullen:
Adds validation and required status to the two fields required for saving the email profile, closes loader when an err returns from mw (Middleware accepts an integer of 0+, and provides some basic email validation)

### DIFF
--- a/src/app/helptext/system/email.ts
+++ b/src/app/helptext/system/email.ts
@@ -1,10 +1,11 @@
 import { Validators } from "@angular/forms";
 import { T } from "app/translate-marker";
-import { matchOtherValidator } from "app/pages/common/entity/entity-form/validators/password-validation";
+import { rangeValidator } from 'app/pages/common/entity/entity-form/validators/range-validation'
 
 export const helptext_system_email = {
   fromemail: {
     placeholder: T("From E-mail"),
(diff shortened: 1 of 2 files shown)


src/app/helptext/system/email.ts
```diff
@@ -26,6 +27,7 @@ export const helptext_system_email = {
 
   port: {
     placeholder: T("Mail Server Port"),
+    validation: [Validators.required, rangeValidator(0)],
```
erinix: For validating ports you could use the max range of 65535 along with the min range of 0 (FTP service does this)

src/app/helptext/system/email.ts
```diff
@@ -1,10 +1,11 @@
 import { Validators } from "@angular/forms";
 import { T } from "app/translate-marker";
-import { matchOtherValidator } from "app/pages/common/entity/entity-form/validators/password-validation";
+import { rangeValidator } from 'app/pages/common/entity/entity-form/validators/range-validation'
 
 export const helptext_system_email = {
   fromemail: {
     placeholder: T("From E-mail"),
+    validation: [Validators.required],
```
erinix: You could also use Validators.email here to prevent invalid email addresses being typed in

PR #2957, opened 2019-10-24
Make network checkin text more consistent, move helptext to helptext file

erinix:
NAS-103595

### DIFF
--- a/src/app/helptext/network/interfaces/interfaces-list.ts
+++ b/src/app/helptext/network/interfaces/interfaces-list.ts
@@ -5,13 +5,15 @@ import { rangeValidator } from '../../../pages/common/entity/entity-form/validat
 export default {
 pending_changes_text : T('There are unapplied network changes. Apply them now?\
  Unapplied changes will be lost.'),
-pending_checkin_text : T('Network changes have been applied. Keep changes permanently?\
+pending_checkin_dialog_text : T('Network changes have been applied. Keep changes permanently?\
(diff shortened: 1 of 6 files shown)


src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -31,4 +33,12 @@ ha_enabled_edit_msg: T("Editing interfaces while HA is enabled is not allowed.")
 
 ha_enabled_delete_title: T("Cannot Delete while HA is Enabled"),
 ha_enabled_delete_msg: T("Deleting interfaces while HA is enabled is not allowed."),
+
+checkin_text: T("Once applied, changes will revert after "),
```
wblock: "Once" can be misinterpreted as "one time", prefer "When" or "After".  But also try to avoid the if/then sentence structure with the pause in the middle.
```suggestion
checkin_text: T("Changes that have been applied will revert after "),
```

src/app/helptext/network/interfaces/interfaces-list.ts
```diff
@@ -31,4 +33,12 @@ ha_enabled_edit_msg: T("Editing interfaces while HA is enabled is not allowed.")
...
+checkin_text_2: T(" seconds unless kept permanently. Adjust this amount to allow for testing."),
```
wblock:
```suggestion
checkin_text_2: T(" seconds unless KEEP PERMANENTLY is chosen. Adjust the delay time to allow for testing."),
```

PR #2958, opened 2019-10-24
Nas 103544

dkmullen:
Gets rid of polling for dir services monitor icon in top bar - sends ws call once, then subscribes for status changes.

Also:
Adds spinner to dir services monitor, since the response can take a bit if one is in a faulted state
Makes all the rows in the monitor clickable, linking to each directory services form
Adds a refresh button to the monitor. This might be overkill, since you have to close this monitor to make changes, or at least go to another window. Also, thought about subscribing here too, but Andrew says that the changes in status that aren't initiated by the user are only checked every ten minutes.


dkmullen:
rebuild

PR #2961, opened 2019-10-24
Updated influxDB helptext link from v1.5 to v1.7

ZackaryWelch:


dlavigne:
I think we're better off pointing to the more general https://www.influxdata.com/products/influxdb-overview/. Otherwise someone has to remember to change the tooltip whenever the version changes, which assumes we're tracking version numbers for all of the alert services (which we are not).

ZackaryWelch:
Changed to just https://docs.influxdata.com/influxdb/, will automatically link to the latest stable version.

PR #2962, opened 2019-10-24
Nas 103596

lola27:


lola27:
@freenas/docs-team Could you please provide the tooltips? thanks

erinix:
code looks ok, just needs tooltips

wblock:
I only see helptext.name_tooltip.  That variable name should be a lot more specific.  Are there any others that need to be filled out?

lola27:
> I only see helptext.name_tooltip. That variable name should be a lot more specific. Are there any others that need to be filled out?

yes. They are more fields in the helptext/ssh-connections.ts need to be filled out, especially for the host and url field.

wblock:
I meant that a variable called just helptext.name_tooltip does not give any idea what it is a name *of*.
As far as other helptext, I don't see from this code what is needed.

PR #2975, opened 2019-10-29
Nas 103414

dkmullen:
Gives a blank option in AD and LDAP forms for kerberos realms and principals so user can de-select previously selected values.


erinix:
Looks like it should work, I do not currently have a system to test this with though, what are your thoughts @anodos325 ?

anodos325:
> Looks like it should work, I do not currently have a system to test this with though, what are your thoughts @anodos325 ?

It looks good. Users should be able to de-select this. If they end up with some crazy combination of parameters, then middleware should raise a validation error.

PR #2976, opened 2019-10-29
Nas 103616

dkmullen:
Customizes dialog for when exporting a pool in an unknown state
Do we need the 'Delete config' checkbox?
Screenshot from 2019-10-29 14-23-40
Screenshot from 2019-10-29 14-25-59


erinix:
code looks ok

wblock:
> Do we need the 'Delete config' checkbox?
I'd say yes, unchecked by default. It might be the user forgot to remove share settings after exporting a pool from the command line, and this will do that for them.

PR #2982, opened 2019-10-30
Add comment field for SMB shares

erinix:
NAS-103677

### DIFF
--- a/src/app/helptext/sharing/smb/smb.ts
+++ b/src/app/helptext/sharing/smb/smb.ts
@@ -17,7 +17,7 @@ export const helptext_sharing_smb = {
 
     placeholder_comment: T("Comment"),
     tooltip_comment: T(
-      "Set the share comment."
+      "Description of the share or notes on how it is used."
     ),
 
     placeholder_home: T('Use as home share'),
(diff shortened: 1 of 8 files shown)


src/app/helptext/sharing/smb/smb.ts
```diff
@@ -14,6 +15,11 @@ export const helptext_sharing_smb = {
     placeholder_name: T('Name'),
     tooltip_name: T('Enter a name for the share.'),
 
+    placeholder_comment: T("Comment"),
+    tooltip_comment: T(
+      "Set the share comment."
```
wblock:
```suggestion
      "Description of the share or notes on how it is used."
```

src/app/helptext/sharing/smb/smb.ts
```diff
@@ -4,6 +4,7 @@ import { Validators } from "@angular/forms";
 export const helptext_sharing_smb = {
     column_name: T('Name'),
     column_path: T('Path'),
+    column_comment: T('Comment'),
```
wblock: Network interfaces have a similar field called Description. We should standardize on one name for this type of field, and Description is clearer.

src/app/helptext/sharing/smb/smb.ts
```diff
@@ -14,6 +15,11 @@ export const helptext_sharing_smb = {
     placeholder_name: T('Name'),
     tooltip_name: T('Enter a name for the share.'),
 
+    placeholder_comment: T("Comment"),
```
wblock: As above.

erinix: See my last commit, I have renamed Comment description in SMB as well as AFP, NFS and WebDav (will make tickets for the other places this should probably be done in)

PR #2983, opened 2019-10-30
Warn user about services that will be disrupted by locking the pool

erinix:
NAS-103666


src/app/helptext/storage/volumes/volume-list.ts
```diff
@@ -41,6 +41,10 @@ detachDialog_pool_detach_confim_checkbox_placeholder: T("Confirm Export/Disconne
 
 upgradePoolDialog_warning: T("Proceed with upgrading the pool? WARNING: Upgrading a pool is a\
  one-way operation that might make some features of\
- the pool incompatible with older versions of FreeNAS: ")
+ the pool incompatible with older versions of FreeNAS: "),
+
+pool_lock_warning_paratext_a: T("WARNING: Locking pool <i>"),
+pool_lock_warning_paratext_b: T("</i>.\
+ Data on the pool will not be available after locking.")
```
wblock: This is kind of misleading.
```suggestion
 Data on the pool will not be accessible until the pool is unlocked.")
```

PR #2985, opened 2019-10-31
NAS-103671

dkmullen:
shows netbiosname_b only if ha_mode is LEGACY


erinix:
Looks good, perhaps we should do this to smb service too?

PR #2989, opened 2019-11-01
Nas 103713

dkmullen:
Replaces snackbars with info dialogs in Sharing menu area, moves a little helptext around to git rid of duplication on dialogs


erinix:
code looks good and works, will defer to docs on the doc changes

src/app/helptext/sharing/index.ts
```diff
@@ -6,5 +6,17 @@ export * from "./nfs/nfs";
...
 export * from "./webdav/webdav";
 
-export const delete_share_message = T("The sharing configuration will be removed.\
- Data in the share dataset will not be affected.");
+export const shared = {
+    delete_share_message: T("The sharing configuration will be removed.\
+    Data in the share dataset will not be affected."),
+
+    dialog_title: T("Enable service"),
+    dialog_message: T("Enable this service to start automatically."),
```
wblock: Where is this dialog shown?  In a nightly build, I don't see any snackbar or dialog when clicking the "start automatically" checkbox.  It doesn't need one, the checkbox itself shows whether it is set.

erinix: It only appears if the service is disabled when you save the form.

PR #2990, opened 2019-11-01
Nas 103715

lola27:


erinix:
Code looks good and works, will defer to @freenas/docs-team on the tooltip change

src/app/helptext/task-calendar/cloudsync/cloudsync-form.ts
```diff
@@ -70,7 +70,7 @@ transfer_mode_warning_move: T('<b>MOVE</b>: After files are <i><u>copied</u></i>
 
 snapshot_placeholder: T('Take Snapshot'),
 snapshot_tooltip: T('Set to take a snapshot of the dataset before a\
- <i>PUSH</i> or <i>PULL</i>.'),
+ <i>PUSH</i>.'),
```
wblock: I don't see why snapshots should only be allowed before a push. If I'm receiving into a snapshot with a pull, I might want to snapshot it beforehand as a backup.

erinix: This was fixed in a different PR that was already merged

PR #2997, opened 2019-11-04
NAS-103714

dkmullen:
Snackbars removed from failover


erinix:
Code looks good to me, how does @freenas/docs-team feel about the dialog changes

PR #2998, opened 2019-11-04
NAS-103763

dkmullen:
Remove snackbars from sys adv
Regarding the tooltip fix, I can't think of any other paraTexts with tooltips, but I think they would have had the same problem as this one (the icon was on the next line) due to the tooltip being outside the div that contained the message.


erinix:
code looks ok to me, how is @freenas/docs-team with the wording

PR #3016, opened 2019-11-07
NAS-103823

dkmullen:
Snackbars to menus in topbar - I dropped the response to accepting the EULA - seems unnecessary?


src/app/components/common/topbar/topbar.component.ts
```diff
@@ -352,14 +351,15 @@ export class TopbarComponent extends ViewControllerComponent implements OnInit,
...
-      this.snackBar.open(this.replicationDetails.repl_status.toString(), ok);
+    this.translate.get('Replication Status').subscribe((title: string) => {
+      this.dialogService.Info(title, this.replicationDetails.repl_status.toString());
     });
   }
 
   showResilveringDetails() {
-    this.translate.get('Ok').subscribe((ok: string) => {
-      this.snackBar.open(`Resilvering ${this.resilveringDetails.name} - ${Math.ceil(this.resilveringDetails.scan.percentage)}%`, ok);
+    this.translate.get('Resilvering Status').subscribe((title: string) => {
```
erinix: dialog.info doesn't need a translate.get because the title should be piped to a translate on the component, 'Resilvering Status' does need to be wrapped with a T() marker though

PR #3022, opened 2019-11-07
NAS-103820

dkmullen:
Snackbars into dialogs for update
I'm not sure we are using the save config stuff in manual update.


src/app/pages/system/update/manualupdate/manualupdateconfig-save/manualupdateconfig-save.component.ts
```diff
@@ -19,24 +17,17 @@ export class ManualUpdateConfigSaveComponent {
...
-    });
-  }
+  constructor(protected ws: WebSocketService, protected router: Router, public dialog: DialogService) {}
 
   doSubmit() {
     this.sub = this.ws.call('core.download', ['config.save', [{ 'secretseed': this.secretseed }], 'freenas.db'])
       .subscribe(
         (res) => {
-          this.openSnackBar("Download Sucessful", "Success");
           window.location.href = res[1];
```
erinix: this could be converted into a programattically downloaded thing

PR #3047, opened 2019-11-11
NAS-103711 ix-auto property

dkmullen:
Adds (via a directive) a unique identifier, ix-auto, to as many elements in the UI as possible. Also fixes some broken ids in the tables. This branch should work behind the scenes without making any visible changes in the UI


erinix:
could you please test this @wblock ?

wblock:
This looks very good to me.  The only thing I saw that was a little long was "clickable_overlay".  I'd prefer just "overlay" to make the labels shorter, although it's not a big problem.  This is something we can refine, assuming it doesn't require editing in a lot of places to change.
At present, I don't use the ix-auto-type property, but other automation might find them useful.

wblock:
@ericbsd What do you think--will this cause problems for you?

ericbsd:
> @ericbsd What do you think--will this cause problems for you?

As far as I see, I don't think it will cause much issue.

src/app/pages/common/entity/entity-table/entity-table.component.html
```diff
@@ -245,6 +268,10 @@
             <div class="clickable"
               *ngIf="col.toggle"
               id="{{row.name}}_{{col.name}}-overlay"
+              [ix-auto]="col?.name"
+              ix-auto-type="clickable-overlay"
```
wblock: Shortening this seems fine to me:
```suggestion
              ix-auto-type="overlay"
```

dkmullen: done!

PR #3049, opened 2019-11-12
NAS-103894 

lola27:


erinix:
It built but Jenkins appears to be stuck, merging

PR #3056, opened 2019-11-12
NAS-103869 Translate tags for entity table and related components

dkmullen:
Adds tags to entity table, actions, expandable rows, etc. which will eventually be rolled into the PO files


MDVAurelien:
Thank you for your responsiveness Dennis!

erinix:
This PR is merged but the actual PO files won't have the words in them until we update them again

PR #3060, opened 2019-11-13
Rename snapshot clone to "clone to new dataset"

erinix:
NAS-103882


erinix:
@freenas/docs-team the tooltip could use a little bit of work too, could you provide me with a better one for this?

src/app/helptext/storage/snapshots/snapshots.ts
```diff
@@ -27,7 +27,7 @@ snapshot_clone_name_placeholder: T('Name'),
 snapshot_clone_name_tooltip: T('Enter a name for the cloned snapshot.'),
```
wblock:
```suggestion
snapshot_clone_name_tooltip: T('Name of the new dataset created from the cloned snapshot.'),
```

PR #3062, opened 2019-11-13
NAS-103925 

lola27:


src/app/pages/plugins/available-plugins/available-plugins.component.ts
```diff
@@ -16,7 +16,9 @@ export class AvailablePluginsComponent implements OnInit {
     @Input() parent: any;
 
     protected queryCall = 'plugin.available';
-    protected queryCallOption = {'plugin_repository': 'https://github.com/freenas/iocage-ix-plugins.git'};
+    protected queryCallOption = {
```
sonicaj: I don’t follow. Middleware already returns the correct URI for the ixsystems collection in both FN and TN, why do we have to check this again and hard code the URI ?

lola27: Thanks @sonicaj ! fixed it :)

PR #3069, opened 2019-11-14
Hide allow repeated GIDs field when in edit mode and set it to true

erinix:
NAS-103835


erinix:
@freenas/docs-team the fieldname and tooltip could be better for this one, could you please provide me something better to go with it?

wblock:
Helptext changed.  I think this covers what is needed, but please verify.

PR #3077, opened 2019-11-14
NAS-103883 improve dialog for connection error

dkmullen:
As described in ticket discussion, makes the dialog more friendly, eliminates traceback

### DIFF
--- a/src/app/pages/storage/VMware-snapshot/VMware-snapshot/VMware-snapshot-form.component.ts
+++ b/src/app/pages/storage/VMware-snapshot/VMware-snapshot/VMware-snapshot-form.component.ts
@@ -15,6 +15,7 @@ import { DialogService } from 'app/services/dialog.service';
 import { AppLoaderService } from '../../../../services/app-loader/app-loader.service';
 import helptext from '../../../../helptext/storage/VMware-snapshot/VMware-snapshot';
 import { first } from 'rxjs/operators';
+import { EntityUtils } from '../../../../pages/common/entity/utils';
 
 @Component({
   selector: 'app-vmware-snapshot-form',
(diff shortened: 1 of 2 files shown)


src/app/pages/storage/VMware-snapshot/VMware-snapshot/VMware-snapshot-form.component.ts
```diff
@@ -257,7 +257,11 @@ export class VMwareSnapshotFormComponent {
         (error)=>{
           this.datastore.options.length = 0;
           parent.loader.close();
-          parent.dialogService.errorReport(error.error,error.reason, error.trace.formatted);
+          if (error.reason.includes('[ETIMEDOUT]')) {
+            parent.dialogService.errorReport(helptext.connect_err_dialog.title, helptext.connect_err_dialog.msg, '');
+          } else {
+            parent.dialogService.errorReport(error.error,error.reason, error.trace.formatted);
```
erinix: rather than using the error.error/error.reason thing that generated the weird looking dialog, we should just use the entityUtils standard handleWSError for other cases

src/app/pages/storage/VMware-snapshot/VMware-snapshot/VMware-snapshot-form.component.ts
```diff
@@ -257,7 +257,11 @@ export class VMwareSnapshotFormComponent {
         (error)=>{
           this.datastore.options.length = 0;
           parent.loader.close();
-          parent.dialogService.errorReport(error.error,error.reason, error.trace.formatted);
+          if (error.reason.includes('[ETIMEDOUT]')) {
```
erinix: you might want to do if (error.reason && error.reason.includes('[ETIMEDOUT]')) just in case the reason isn't even there.

PR #3104, opened 2019-11-20
Allow null values for quota/reservation fields

erinix:
NAS-103947


themylogin:
@erinix is this the only PR for this ticket or will there be more?

erinix:
> @erinix is this the only PR for this ticket or will there be more?

what else needs to be done?

PR #3117, opened 2019-11-24
NAS-104093

dszidi:
Only send Authentication event when user logs in. Not during reauthentication via token. Theme service no longer listens for Authentication event directly. Removed some comments from preferences service


src/app/core/services/preferences.service.ts
```diff
@@ -43,6 +43,7 @@ export class PreferencesService {
     this.core.register({observerClass:this, eventName:"Authenticated",sender:this.api}).subscribe((evt:CoreEvent) => {
       // evt.data: boolean = authentication status
       if(evt.data){
+        console.log(evt);
```
erinix: remove your console.log

PR #3124, opened 2019-11-26
Nas 104079

dkmullen:
Adds validation to lldp form for country field


erinix:
looks ok to me, @freenas/docs-team how does the validation message look?

PR #3141, opened 2019-11-27
Nas 104060

lola27:


erinix:
I think critical could probably use this too, that field is supposed to be a positive temperature too

erinix:
Informational too

PR #3148, opened 2019-12-02
NAS-104058: Add help text to fields added in System/Alert Services

Mrt134:
- Includes Mattermost: `Icon URL` field and SNMP Trap: `Hostname` and `Port` fields.
- Yarn build test: no issues.
- Jira issue Fix Version is `11.3-RC2`
- TODO: Port final text into the FreeNAS guide

### DIFF
--- a/src/app/helptext/system/alert-service.ts
+++ b/src/app/helptext/system/alert-service.ts
@@ -84,7 +84,9 @@ export default {
  webhook settings.'),
 
     Mattermost_icon_url_placeholder: T('Icon URL'),
-    Mattermost_icon_url_tooltip: T(''),
+    Mattermost_icon_url_tooltip: T('Icon file to use as the profile \
+ picture for new messages. Example: \
+ <i>https://mattermost.org/wp-content/uploads/2016/04/icon.png</i>.'),
 
     OpsGenie_api_key_placeholder: T('API Key'),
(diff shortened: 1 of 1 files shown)


src/app/helptext/system/alert-service.ts
```diff
@@ -113,10 +115,12 @@ export default {
  this service.'),
 
     SNMPTrap_host_placeholder: T('Hostname'),
-    SNMPTrap_host_tooltip: T(''),
+    SNMPTrap_host_tooltip: T('Hostname or IP address of the system that \
+ will receive SNMP trap notifications.'),
```
wblock: s/that will/to/

src/app/helptext/system/alert-service.ts
```diff
@@ -113,10 +115,12 @@ export default {
...
 
     SNMPTrap_port_placeholder: T('Port'),
-    SNMPTrap_port_tooltip: T(''),
+    SNMPTrap_port_tooltip: T('UDP port number on the system receiving \
+ SNMP trap notifications. Example: <i>162</i>'),
```
wblock: If 162 is the default, that should be mentioned here, rather than using it as an example.
Also missing an ending period.
s/Example: 162/The default is 162./

PR #3155, opened 2019-12-03
NAS-104040 Clarify min and max poll helptext for NTP servers.

a-stjohn:
The common \<sup> tag would not work. When used with integers it returned the calculation of the numbers used (which was an incorrect exponent calculation). When used with a string the tag did nothing.

### DIFF
--- a/src/app/helptext/system/ntpservers.ts
+++ b/src/app/helptext/system/ntpservers.ts
@@ -38,10 +38,9 @@ export const helptext_system_ntpservers = {
     minpoll: {
       placeholder: T("Min Poll"),
       tooltip: T(
-        "Calculated in powers of 2. Enter the exponent to be \
- calculated. For example, if the <i>Min. Poll</i> is 6, the \
- calculation is 2^6. Cannot be lower than 4 or higher than \
- <i>Max. Poll</i>."
+        "The minimum polling interval, in seconds, as a power of 2. \
+ For example, <i>6</i> means 2^6, or 64 seconds. The default is 6, \
+ minimum value is 4."
       ),
       validation: [Validators.min(4), Validators.required]
     },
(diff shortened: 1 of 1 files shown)


src/app/helptext/system/ntpservers.ts
```diff
@@ -38,17 +38,20 @@ export const helptext_system_ntpservers = {
     minpoll: {
       placeholder: T("Min. Poll"),
```
wblock:
```suggestion
      placeholder: T("Min Poll"),
```

src/app/helptext/system/ntpservers.ts
```diff
@@ -38,17 +38,20 @@ export const helptext_system_ntpservers = {
...
       tooltip: T(
-        "Power of 2 in seconds; cannot be lower than 4 or\
- higher than <i>Max. Poll</i>."
+        "Calculated in powers of 2. Enter the exponent to be \
```
wblock: This is still pretty confusing.  Let's fix the label of the field while here, nobody abbreviates "min" or "max" with a period. The description should first tell what this field is, then describe the calculation.  Suggestion:
`The minimum polling interval, in seconds, as a power of 2. For example, <i>6</i> means 2^6, or 64 seconds. The default is 6, minimum value is 4.`

src/app/helptext/system/ntpservers.ts
```diff
@@ -38,17 +38,20 @@ export const helptext_system_ntpservers = {
...
       ),
       validation: [Validators.min(4), Validators.required]
     },
 
     maxpoll: {
       placeholder: T("Max. Poll"),
       tooltip: T(
-        "Power of 2 in seconds; cannot be higher than 17 or\
- lower than <i>Min. Poll</i>."
+        "Calculated in powers of 2. Enter the exponent to be \
```
wblock: `The maximum polling interval, in seconds, as a power of 2. For example, <i>10</i> means 2^10, or 1,024 seconds. The default is 10, maximum value is 17.`

src/app/helptext/system/ntpservers.ts
```diff
@@ -38,17 +38,20 @@ export const helptext_system_ntpservers = {
...
+ calculated. For example, if the <i>Min. Poll</i> is 6, the \
+ calculation is 2^6. Cannot be lower than 4 or higher than \
+ <i>Max. Poll</i>."
       ),
       validation: [Validators.min(4), Validators.required]
     },
 
     maxpoll: {
       placeholder: T("Max. Poll"),
```
wblock:
```suggestion
      placeholder: T("Max Poll"),
```

PR #3162, opened 2019-12-03
NAS-103684

dszidi:
Accounts pages now use web socket API exclusively.


erinix:
this looks good to me, could you review the fieldset title @freenas/docs-team

PR #3163, opened 2019-12-03
NAS-103691 NIS from rest to ws

dkmullen:
Oh I now see that there are notifications to take care of on this one.


erinix:
looks good, we should probably still add a fieldset with a single header for consistency though

erinix:
looks like the build is failing, it is complaining about this:

Cannot find module '../../../services/dir-services.service'

PR #3169, opened 2019-12-04
Nas 103691

dkmullen:
NIS - rest to ws, add new service to be used by other dir services, add simple fieldSet and a label to NIS


erinix:
looks good to me, could you review the fieldsets @freenas/docs-team

PR #3180, opened 2019-12-04
NAS-103696

aervin:
Not sure if were _always_ using fieldsets or not. If so, could use some help on how to lay this form out.


erinix:
looks okay to me, what do you think of the fieldset name @freenas/docs-team ?

wblock:
I'm not sure if adding titles to screens with only one fieldset is really helpful.  Not sure it's a bad thing, either, though.  As far as description, "General Options" is harmless.

PR #3187, opened 2019-12-05
Nas 103752

dkmullen:
Replaces notifiers in VM components


erinix:
looks good, needs to be targeted at 12-devel instead of master though

PR #3188, opened 2019-12-05
NAS-103698

aervin:


erinix:
code looks ok, I'm not sure how I feel about the entire right hand side showing up after basic vs advanced, though it looks kind of weird

PR #3189, opened 2019-12-05
NAS-103700

aervin:


src/app/pages/services/components/service-tftp/service-tftp.component.ts
```diff
@@ -75,29 +98,22 @@ export class ServiceTFTPComponent {
               protected _injector: Injector, protected _appRef: ApplicationRef,
               protected userService: UserService) {}
 
-  resourceTransformIncomingRestData(data: any) {
```
aervin: Please test that permissions are working as expected. Guessing this piece of code was a workaround for an issue that no longer exists.

erinix: as discussed, it appears that we need to invert the values both on load and save with the websocket api

erinix:
looks good to me, @freenas/docs-team could you review the fieldset labels?

PR #3219, opened 2019-12-09
NAS-103746 -- move replication status from rest to ws event

aervin:


src/app/components/common/topbar/topbar.component.ts
```diff
@@ -148,14 +145,25 @@ export class TopbarComponent extends ViewControllerComponent implements OnInit,
...
     });
-    this.continuosStreaming = observableInterval(10000).subscribe(x => {
-      this.showReplicationStatus();
-      if (this.is_ha) {
-        this.getHAStatus();
-      }
-      this.checkNetworkCheckinWaiting();
-      this.checkNetworkChangesPending();
-    });
+    this.replicationStatusSub = this.ws
```
erinix: this should not replace the continous streaming, but should be outside of it, the getHAStatus needs to still be in there until we convert it to use events instead of check every 10 seconds

PR #3220, opened 2019-12-09
Re-remove polling checks for network status from UI

erinix:
NAS-103543


erinix:
New ticket for this one: NAS-104252

PR #3221, opened 2019-12-09
Nas 103688 k-realms rest to ws

dkmullen:
Also displays and accepts fields which accept more than one value as space -separated lists, and filters out commas


erinix:
looks ok to me, could you review the labels and messages @freenas/docs-team

PR #3228, opened 2019-12-10
NAS-103725

dszidi:
Cron pages now use websocket API (no more rest calls)


src/app/pages/task-calendar/cron/cron-list/cron-list.component.ts
```diff
@@ -18,92 +18,105 @@ import { TaskScheduleListComponent } from '../../components/task-schedule-list/t
...
+            onClick: row =>
+            this.dialog.confirm(T("Run Now"), T("Run this job now?"), true)
+              .pipe(
+                filter(run => !!run),
+                switchMap(() =>
+                  this.ws.call('cronjob.run',[row.id])
+                )
+              ).subscribe(
+                res => {
+                  const message = row.enabled == true ? T('This job is scheduled to run again ' + row.next_run + '.') : T('This job will not run again until it is enabled.');
```
wblock: Can't this just be `row.enabled` instead of `row.enabled == true`?

dszidi:
Went with another solution that doesn't touch form-scheduler at all.

https://github.com/freenas/webui/pull/3232

PR #3240, opened 2019-12-11
Nas 104272

lola27:


src/app/pages/common/entity/entity-form/components/form-errors/form-errors.component.html
```diff
@@ -23,4 +23,7 @@
     <mat-error *ngIf="control.hasError('regex')">
         {{"Invalid format or character." | translate}}
     </mat-error>
+    <mat-error *ngIf="control.hasError('ip')">
+        {{"Select " | translate}} {{control.errors.info[1]}} {{" to include all " | translate}} {{control.errors.info[0] | translate}} {{" addresses. When this has been chosen, additional addresses cannot be selected." | translate}}
```
erinix: @freenas/docs-team  could you review this string?

erinix:
looks good

src/app/pages/common/entity/entity-form/components/form-errors/form-errors.component.html
wblock: control.errors.info seems like a misleading variable name.  It is "All IPv4" or "All IPv6" or "All IPv4 and IPv6"?  The rest of the string seems reasonable.

lola27: > control.errors.info seems like a misleading variable name. It is "All IPv4" or "All IPv6" or "All IPv4 and IPv6"? The rest of the string seems reasonable.

control.errors.info include three items like this [ipType === 'ipv4' ? 'IPv4' : 'IPv6', wildcard, thisControl.value[i]], which means the ip type, the value which means all ipv4 or ipv6 (0.0.0.0 or ::), the current field value.

PR #3241, opened 2019-12-11
NAS-103723

dszidi:
Convert tunables to web socket api.
Add fieldsets to form


src/app/pages/system/tunable/tunable-form/tunable-form.component.ts
```diff
@@ -1,67 +1,88 @@
...
+          type: 'textarea',
+          name: 'value',
+          placeholder: helptext.value.placeholder,
+          tooltip: helptext.value.tooltip,
+          required: true,
+          validation : helptext.value.validation
+        },
+        {
+          type: 'select',
+          name: 'type',
```
erinix: type appears to be a required variable, if you try to save without setting type it gives an unhelpful error

PR #3242, opened 2019-12-11
NAS-103721

dszidi:
Converted System Dataset page to websocket calls.
Applied fieldset layout to form.


erinix:
looks good, only one request, could you move the fieldset header to helptext?

PR #3249, opened 2019-12-12
NAS-103702 -- afp share config rest -> ws

aervin:
Also adds a FieldSets class which tries to centralize some of the more common fieldset operations.

### DIFF
--- a/src/app/helptext/sharing/afp/afp.ts
+++ b/src/app/helptext/sharing/afp/afp.ts
@@ -3,8 +3,17 @@ import { T } from "app/translate-marker";
 
 export const helptext_sharing_afp = {
   fieldset_general: T('General Options'),
-  fieldset_access: T('Access'),
+  fieldset_permissions: ('Permissions'),
+  fieldset_allow: T('Allow'),
+  fieldset_deny: T('Deny'),
+  fieldset_ro: T('Read Only'),
+  fieldset_rw: T('Read/Write'),
+  fieldset_hostsallow: T('Hosts Allow'),
+  fieldset_hostsdeny: T('Hosts Deny'),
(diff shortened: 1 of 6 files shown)


src/app/helptext/sharing/afp/afp.ts
```diff
@@ -2,6 +2,19 @@ import { Validators } from "@angular/forms";
 import { T } from "app/translate-marker";
 
 export const helptext_sharing_afp = {
+  fieldset_general: T('General Options'),
+  fieldset_permissions: ('Permissions'),
+  fieldset_allow: T('Allow'),
+  fieldset_deny: T('Deny'),
+  fieldset_ro: T('Read Only'),
+  fieldset_rw: T('Read/Write'),
+  fieldset_hostsallow: T('Hosts Allow'),
```
wblock: I would flip these around to "Allow Hosts" and "Deny Hosts" to make more sense to the user.

src/app/helptext/sharing/afp/afp.ts
```diff
@@ -2,6 +2,19 @@ import { Validators } from "@angular/forms";
...
+  fieldset_hostsdeny: T('Hosts Deny'),
+  fieldset_other: T('Other Options'),
+
+  placeholder_user_or_group: T('User or Group'),
+  placeholder_host_or_ip: T('Hostname or IP address'),
```
wblock: We probably should just standardize on "Hostname" everywhere and just point out in the docs that it means either a DNS hostname or an IP address.  But that's going to be big project.

aervin:
All helptext which has `Comma-, space-, or tab-delimited list of allowed hostnames or IP addresses.` will need to be updated. Entries are added one at a time now.

PR #3258, opened 2019-12-13
Adjust pool unlock to use websocket call and id for services to restart

erinix:
NAS-104167


erinix:
please test pool unlock thoroughly, I converted it to use websocket call since the rest call wasn't working

PR #3264, opened 2019-12-13
Nas 104285

lola27:


src/app/pages/task-calendar/replication/replication-list/replication-list.component.ts
```diff
@@ -105,7 +105,9 @@ export class ReplicationListComponent {
     }
 
     stateButton(row) {
-        if (row.job) {
+        if (row.state.state === 'HOLD') {
+            this.dialog.Info(T('Task is hold'), row.state.reason, '500px', 'info', true);
```
wblock: I don't know what this means.  Is it saying the task has been put on hold by the user, is on hold for some other reason, or something else?  Probably this should say at least "Task is on hold" or "Task is waiting", but I don't have context to know.

PR #3265, opened 2019-12-13
NAS-104311 set graph age and points on submit

dkmullen:
Refreshes comparison values on save, so that data that is immediately reentered will be checked correctly


dkmullen:
The latest version of the PR moves the 'Destroy RRD' warning to a confirm dialog, which seems more consistent with how we present such info elsewhere in the app. User sees the warning each time (rather than having it hid in a tooltip). 

This also takes care of resetting the form (NAS-104288); Reset now restores the age/points to 12/1200, sets the checkbox for percentage to false, leaves the hostname in place

PR #3279, opened 2019-12-16
NAS-104314 Clear selected items when single actions are taken

dkmullen:
Activate and Keep/Unkeep don't navigate to another page, so this PR empties the 'selected' list when these single actions are taken.


erinix:
This does deselect all of the things but the multi actions are still present after doing a select all and then deleting an individual boot environment.

PR #3316, opened 2019-12-18
Nas 103170

aervin:


src/app/pages/storage/volumes/manager/manager.component.ts
```diff
@@ -570,22 +570,20 @@ export class ManagerComponent implements OnInit, OnDestroy, AfterViewInit {
...
         } else {
-          body  = {volume_add: this.name, layout: layout };
+          body  = {name: this.name, topology: layout };
         }
         this.busy =
-          this.rest
-          .post(this.resource_name, {
-            body: JSON.stringify(body)
-          })
+          this.ws
```
erinix: You should use an entityJob to track the progress of pool creation

src/app/pages/storage/volumes/manager/manager.component.ts
```diff
@@ -570,51 +524,61 @@ export class ManagerComponent implements OnInit, OnDestroy, AfterViewInit {
...
+              if (e.code === 409) {
+                this.error = "";
+                for (let i in e.error) {
+                  e.error[i].forEach(error => {
+                    this.error += error + "<br />";
+                  });
                 }
               } else {
-                this.dialog.errorReport(T('Error creating pool'), res.error.error_message, res.error.traceback);
+                this.dialog.errorReport(T("Error creating pool"), res.error.error_message, res.error.traceback);
```
erinix: this error handling code should probably just be another entityUtils().handleWSError now, probably no need for the if else either, websockets are pretty consistent

src/app/pages/storage/volumes/manager/manager.component.ts
```diff
@@ -570,51 +524,50 @@ export class ManagerComponent implements OnInit, OnDestroy, AfterViewInit {
...
-                this.error = '';
-                for (let i in res.error) {
-                  res.error[i].forEach(
-                    (error) => { this.error += error + '<br />'; });
-                }
-              } else {
-                this.dialog.errorReport(T('Error creating pool'), res.error.error_message, res.error.traceback);
+
+        const dialogRef = this.mdDialog.open(EntityJobComponent, {
+          data: { title: T("Creating Pool"), disableClose: true }
```
erinix: if this pool is not new then it is "Extending" a pool not creating a pool

erinix:
I also found that during creation, it does not like "RAIDZ" as a vdev type, upon looking at code it expects RAIDZ1, this should be fixed in both vdev.component.ts and in manager.component.ts (for the auto select when you add 3 disks in)

PR #3323, opened 2019-12-18
Update words that can be translated

MDVAurelien:
Update the words that can be translated but they will have to be injected into the PO file.

### DIFF
--- a/src/app/core/components/widgets/widgetnic/widgetnic.component.html
+++ b/src/app/core/components/widgets/widgetnic/widgetnic.component.html
@@ -24,18 +24,18 @@
 
               <button mat-icon-button class="" (click)="router.navigate(['network/interfaces/edit/'+nicState.name])">
                 <mat-icon role="img"
-                          matTooltip="Configure {{nicState.name}}"
+                          matTooltip="{{Configure | translate}} {{nicState.name}}"
                   matTooltipPosition="above"
-                  aria-label="Configure Interface">
+                  aria-label="{{Configure Interface | translate}}">
                     settings
                 </mat-icon>
               </button>
 
               <button mat-icon-button class="" (click)="router.navigate(['reportsdashboard/network'])">
                 <mat-icon role="img"
-                  matTooltip="Network Reports"
+                  matTooltip="{{Network Reports | translate}}"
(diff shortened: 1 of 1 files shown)


erinix:
This doesn't build, looking at your PR I see that you did things like {{Configure | translate}}, this needs to be {{"Configure" | translate}}, otherwise it thinks you are passing a variable named Configure into the translate pipe

erinix:
or also {{'Configure' | translate}} when inside of double quotes

erinix:
still not quite right

MDVAurelien:
Yes sorry I'm going to correct, I went a little too fast!

MDVAurelien:
I still have a error but I don't see where it is...

src/app/core/components/widgets/widgetnic/widgetnic.component.html
```diff
@@ -24,18 +24,18 @@
 
               <button mat-icon-button class="" (click)="router.navigate(['network/interfaces/edit/'+nicState.name])">
                 <mat-icon role="img"
-                          matTooltip="Configure {{nicState.name}}"
+                          matTooltip="{{'Configure' | translate}} {{nicState.name}}"
                   matTooltipPosition="above"
-                  aria-label="Configure Interface">
+                  aria-label="{{'Configure Interface' | translate}}">
```
erinix: the build is complaining about these aria-labels, try removing the translates from these

src/app/core/components/widgets/widgetnic/widgetnic.component.html
```diff
@@ -24,18 +24,18 @@
...
                 </mat-icon>
               </button>
 
               <button mat-icon-button class="" (click)="router.navigate(['reportsdashboard/network'])">
                 <mat-icon role="img"
-                  matTooltip="Network Reports"
+                  matTooltip="{{'Network Reports' | translate}}"
                   matTooltipPosition="above"
-                  aria-label="Network Reports">
+                  aria-label="{{'Network Reports' | translate}}">
```
erinix: this one too

MDVAurelien:
alléluia :)

PR #3347, opened 2019-12-20
NAS-103851

lola27:


erinix:
code looks ok, could you please review @freenas/docs-team

src/app/helptext/task-calendar/replication/replication.ts
```diff
@@ -86,6 +86,12 @@ export default {
     properties_tooltip: T('Include dataset properties with the replicated \
  snapshots.'),
 
+    replicate_placeholder: T('Full Filesystem Replication'),
+    replicate_tooltip: T('Send a replication stream package, which will replicate \
```
wblock: Sorry, it's not clear at all whether this is saying it sends something different ("replication stream package") than an ordinary replication.
I think this is just saying "do a full replication of this entire thing".  Not clear why it uses "filesystem" rather than "dataset".
Guessing at meaning, I would rewrite this to not be so wordy and to use standard terms:
```
Completely replicate the selected dataset. The target dataset will have all of the properties,
snapshots, child datasets, and clones from the source dataset.
```
Then it gets into other settings which must be made, which indicates this is not a standalone setting and has not been designed with the user in mind.  If the user enables just this option, it doesn't do what the name indicates.  Maybe the name is wrong.  More information is needed to improve this.

themylogin: I agree with Warren.

PR #3351, opened 2019-12-20
Reset MTU to 1500 if MTU is cleared

erinix:
NAS-104441


wblock:
This is a little inconsistent.  The first time an interface is edited, the MTU is blank. If a different MTU value is set and cleared, then it resets to 1500.  So there is no way to get back to an unset state.  It should probably show the actual MTU value from the start, rather than a blank field.

PR #3354, opened 2019-12-20
NAS-104425 Rearrange disk form

dkmullen:


src/app/pages/storage/disks/disk-form/disk-form.component.ts
```diff
@@ -173,9 +185,17 @@ export class DiskFormComponent {
   }
 
   beforeSubmit(value) {
+    if (value.passwd && value.passwd === '') {
```
erinix: This still clears the password out if you save, I believe that value.passwd evaluates to false when it is empty string

PR #3359, opened 2019-12-24
NAS-103755 -- user form rest -> ws + replace notifier

aervin:


src/app/pages/account/users/user-form/user-form.component.ts
```diff
@@ -375,51 +349,76 @@ export class UserFormComponent {
...
-              this.shell.options.push({label : item[1], value : item[0]});
-            }
-          } else {
-            this.shell.options.push({label : item[1], value : item[0]});
-          } 
-        } 
+      this.userService.shellChoices().then(choices => {
+        this.shell = this.fieldSets.config("shell");
+
+        // Leaving this here for reference. Will delete after review/before merge.
```
aervin: Here's where things get confusing @william-gr

william-gr: Ok, I get it know.

netcli only makes sense for builtin users (in fact, only root), so presenting that option to regular users wouldnt make sense, as they dont have privileges to use it.

We could probably handle that in middleware side if we change user.shell_choices to accept an user id as parameter

aervin: Should I remove the filtering then? Just use all choices?

william-gr: Yes, lets filter it on middleware side. Your recent change looks good

erinix:
seems to work ok with the new middleware, could you please fix the merge conflicts?

PR #3364, opened 2019-12-26
DOCS-116: Update iSCSI/Portals/IP Address help text

Mrt134:
- Update to include text about IPv6 wildcard.
- Fix bug that was preventing help text from displaying in Sharing/iSCSI/Portals/Add
- Yarn testing: no issues.

### DIFF
--- a/src/app/helptext/sharing/iscsi/iscsi-wizard.ts
+++ b/src/app/helptext/sharing/iscsi/iscsi-wizard.ts
@@ -50,7 +50,9 @@ export default {
  Mutual CHAP."),
 
     ip_placeholder: T("IP"),
-    ip_tooltip: T("Select the IP address associated with an interface or the wildcard address of 0.0.0.0 (any interface)."),
+    ip_tooltip: T("Select the IP addresses associated with an interface \
+ or the wildcard addresses 0.0.0.0 (all IPv4 addresses) and/or :: (all \
+ IPv6 addresses)."),
 
     port_placeholder: T("Port"),
     port_tooltip: T("TCP port used to access the iSCSI target. Default is 3260."),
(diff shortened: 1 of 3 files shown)


src/app/helptext/sharing/iscsi/iscsi-wizard.ts
```diff
@@ -50,7 +50,9 @@ export default {
  Mutual CHAP."),
 
     ip_placeholder: T("IP"),
-    ip_tooltip: T("Select the IP address associated with an interface or the wildcard address of 0.0.0.0 (any interface)."),
+    ip_tooltip: T("Select the IP addresses associated with an interface \
+ or the wildcard addresses 0.0.0.0 (all IPv4 addresses) and/or :: (all \
```
wblock: Don't use "and/or".  Also, check with Lola on whether 0.0.0.0 and :: can be combined.  I didn't think they could, but there was another value for "everything".  Also, "associated with an interface" is kind of misleading.  With iSCSI, this is to choose allowed or denied IP addresses, or which address the server daemon listens on.

src/app/helptext/sharing/iscsi/iscsi.ts
```diff
@@ -70,9 +70,9 @@ export const helptext_sharing_iscsi = {
   ),
 
   portal_form_placeholder_ip: T("IP Address"),
-  portal_form_tooltip_ip: T(
-    "Select the IP address associated with an interface\
- or the wildcard address of <i>0.0.0.0</i> (any interface)."
+  portal_form_tooltip_ip: T("Select the IP addresses associated with an \
+ interface or the wildcard addresses 0.0.0.0 (all IPv4 addresses) and/or \
```
wblock: As above.

PR #3365, opened 2019-12-26
NAS-104305

dszidi:
- VolumeDataRequest event now sends websocket call instead of REST.
- Pool Widget on dash needed to adapt to the different data structure returned from middleware.

### DIFF
--- a/src/app/core/services/api.service.ts
+++ b/src/app/core/services/api.service.ts
@@ -147,9 +147,6 @@ export class ApiService {
         return redef;
       },
       postProcessor(res,callArgs){
-        //DEBUG: console.warn("POOLDISKS POSTPROCESSOR");
-        //DEBUG: console.log(res);
-        //DEBUG: console.log(callArgs);
         let cloneRes = Object.assign({},res);
         cloneRes = {callArgs:callArgs ,data: res}
         return cloneRes;
--- a/src/app/core/services/api.service.ts
+++ b/src/app/core/services/api.service.ts
@@ -44,25 +44,16 @@ export class ApiService {
         protocol:"websocket",
         version:"2.0",
         namespace:"user.set_attribute",
-        args: [],// eg. [["id", "=", "foo"]]
-        //responseEvent: null
+        args: [],
       },
       preProcessor(def:ApiCall){
-        //console.log("USER DATA PREPROCESSOR");
         let uid:number = 1;
         let redef = Object.assign({}, def);
-        //console.log(def.args)
-        //Do some stuff here
-        // [1,{attributes:{usertheme:theme.name}}]
         redef.args =  [ uid, "preferences",def.args ] ;
         return redef;
       },
       postProcessor(res,callArgs,core){
-        //console.log("USER DATA POSTPROCESSOR");
-        //console.log(res);
-        //console.log(callArgs);
         let cloneRes = Object.assign({},res);
-        //cloneRes = {callArgs:callArgs ,data: res}
         if(res == 1){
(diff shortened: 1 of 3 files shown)


src/app/core/services/api.service.ts
```diff
@@ -71,12 +62,23 @@ export class ApiService {
...
+        let redef = Object.assign({}, def);
+
+        const queryFilters = [
+          ["name","~", "^[^\/]+$" ], // Root datasets only
+        ];
+
+        const params = [queryFilters];
+        redef.args = params;
+        return redef;
+      },
```
aervin: I think this could be shortened without sacrificing readability:
```ts
{
  preProcessor(def: ApiCall) {
    return { args: [[["name", "~", "^[^\/]+$"]]], ...def }; // Root datasets only
  }
}
```

src/app/core/services/api.service.ts
```diff
@@ -722,7 +721,7 @@ export class ApiService {
       }
 
       let call = cloneDef.apiCall;//this.parseEventWs(evt);
-      this.ws.call(call.namespace).subscribe((res) => {
+      this.ws.call(call.namespace, call.args ? call.args : []).subscribe((res) => {
```
aervin: `call.args || []`

src/app/pages/dashboard/dashboard.component.ts
```diff
@@ -291,27 +291,20 @@ export class DashboardComponent implements OnInit, AfterViewInit, OnDestroy {
 
     for(let i in evt.data){
       let avail = null;
+      let used_pct = evt.data[i].used.parsed / (evt.data[i].used.parsed + evt.data[i].available.parsed);
```
aervin: Could this be `const`?

src/app/pages/dashboard/dashboard.component.ts
```diff
@@ -323,7 +316,10 @@ export class DashboardComponent implements OnInit, AfterViewInit, OnDestroy {
     });
 
     this.core.register({observerClass: this, eventName: 'VolumeData'}).subscribe((evt:CoreEvent) => {
-      this.setVolumeData(evt);
+      const nonBootPools = evt.data.filter(v => v.id !== 'boot-pool');
+      let clone = Object.assign({}, evt);
```
aervin: `const`

dszidi: Not here. clone's data property is changed before it is returned

src/app/pages/dashboard/dashboard.component.ts
dszidi: Yes

src/app/core/services/api.service.ts
dszidi: Good call.

src/app/pages/dashboard/dashboard.component.ts
aervin: `clone` itself is never reassigned

src/app/core/services/api.service.ts
dszidi: Yes it could be a little less verbose. I dislike too many nested arrays though. I changed it to the following...

```
      preProcessor(def:ApiCall){
        const queryFilters = [
          ["name","~", "^[^\/]+$" ], // Root datasets only
        ];

        return { args: [queryFilters], ...def };
      },

```

src/app/pages/dashboard/dashboard.component.ts
dszidi: kk. Changed to const

PR #3368, opened 2019-12-26
NAS-104454 -- dyn dns fieldsets

aervin:


src/app/pages/services/components/service-dynamicdns/service-dynamicdns.component.ts
```diff
@@ -5,100 +5,116 @@ import * as _ from 'lodash';
...
+          type: 'input',
+          name: 'custom_ddns_path',
+          placeholder: helptext.custom_ddns_path_placeholder,
+          tooltip: helptext.custom_ddns_path_tooltip,
+          disabled: true,
+          isHidden: true
+        },
+        {
+          type : 'input',
+          name : 'domain',
```
erinix: it complains about domain name being a required field, if this is in fact required now we should probably mark it with required:true so the user cannot save without passing this validation

PR #3371, opened 2019-12-27
DOCS-124: Update Services/Rsync help text

Mrt134:
- Yarn testing: no issues.


src/app/helptext/services/components/service-rsync.ts
```diff
@@ -14,14 +14,16 @@ rsyncd_auxiliary_tooltip: T('Enter any additional parameters from <a\
```
erinix: we should add T() markers to these strings

PR #3379, opened 2019-12-31
Nas 103711 v12

dkmullen:
Adds unique identifier (ix-auto) to entity table, wizard, form and some form elements, etc. This is #1 of three PRs. Should go in first if we want to do this in steps.


src/app/pages/common/entity/entity-table/entity-table-row-details/entity-table-row-details.component.ts
```diff
@@ -18,6 +18,7 @@ export class EntityTableRowDetailsComponent implements OnInit, OnChanges {
   ngOnInit() {
     this.buildColumns();
     this.actions = this.getActions();
+    console.log(this.config)
```
erinix: don't forget to remove your console.log :)

PR #3388, opened 2020-01-02
Update translations readme file 20200102

erinix:


amotin:
If we keep the list sorted, then J should be before R.

erinix:
fixed :)

PR #3408, opened 2020-01-06
Nas 104450

lola27:


src/app/helptext/sharing/iscsi/iscsi.ts
```diff
@@ -276,9 +276,7 @@ export const helptext_sharing_iscsi = {
     "Mutual secret password. Required when Peer User is set. Must be\
  different than the <i>Secret</i>."
   ),
-  authaccess_error_peersecret: T('Must match Peer Secret (Confirm) and be\
- between 12 and 16 characters in length. Cannot be the same as\
- Secret.'),
+  authaccess_error_peersecret: T('Secret and Peer Secret can not be the same.'),
```
a-stjohn:
```suggestion
  authaccess_error_peersecret: T('Mutual secret password. Required when Peer User \
 is set. Must be different than the Secret.'),
```

lola27: Thanks @a-stjohn , this one error message will be displayed under the secret field or the peer secret field, it depends on which one field is the last one been changed. So we should not add other info to this one. And if user set peer user, the peer secret field will be marked as required.

a-stjohn: ahh. Sorry. I misunderstood what was happening in the UI. In that case, it looks good to me!

lola27: thanks @a-stjohn ! The name is a little bit confused, so I pushed a commit to changed it.

PR #3418, opened 2020-01-07
NAS-104584 master

dszidi:
Reports should use server time when calculating API call time frames


dszidi:
Need to rebase on Master

PR #3433, opened 2020-01-09
Silently refresh the app if a missing chunk is found

erinix:
NAS-104277

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -133,6 +133,14 @@ export class AppComponent {
         }
       }
     });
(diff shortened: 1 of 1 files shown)


erinix:
note: the easiest way to test this is to have this same code on an 11.3 branch as well as a master branch, hotpatch the machine with both branches and use a symlink to switch between the two, load the page with one of the hotpatches then switch the symlink but do not refresh and try to click around on the navigation bar until it finds a page that it doesn't like and it will then refresh it

PR #3435, opened 2020-01-09
NAS-102175 / 12.0 / Add fieldset, description field and T markers to initshutdown scripts

erinix:
NAS-102175


src/app/helptext/task-calendar/initshutdown/initshutdown.ts
```diff
@@ -31,4 +31,9 @@ export default {
 
     ini_timeout_placeholder: T('Timeout'),
     ini_timeout_tooltip: T('Automatically stop the script or command after the specified seconds.'),
+
+    ini_description_placeholder: T('Description'),
+    ini_description_tooltip: T('Enter a description of the Init/Shutdown script.'),
```
Mrt134:
```suggestion
    ini_description_tooltip: T('Comments about this script.'),
```
Reworded to describe the field without the imperative (the field doesn't appear to be required) and avoided using the name of the field in the definition of the field.

PR #3444, opened 2020-01-10
NAS-104651 / 11.3 / NAS-104651-stable

lola27:


src/app/pages/task-calendar/cloudsync/cloudsync-form/cloudsync-form.component.ts
```diff
@@ -710,7 +710,15 @@ export class CloudsyncFormComponent implements OnInit {
 
     for (let i = 0; i < bwlimit.length; i++) {
       const sublimitArr = bwlimit[i].split(',');
+      if (sublimitArr.length === 1) {
```
themylogin: This should only be applicable if also `bwlimit.length == 1`. Else it should be a field error like "bandwidth limit list element should be in format 08:00,1M".

PR #3454, opened 2020-01-10
NAS-104457 / 12.0 / Apply fieldsets to snmp service

erinix:
NAS-104457


erinix:
I tweaked the password fields a bit to make them more user friendly, they should not display existing password data but also should not be required because there are cases where they should not be, for instance when the password is already set on the backend or the protocol is not selected, either way the middleware will let the user know if they need a password.  Also removed double confirm password as it doesn't really make much sense for this component.

PR #3455, opened 2020-01-10
NAS-104456 / 12.0 / Apply fieldsets to S3 configuration

erinix:
NAS-104456


erinix:
the requirement for password and doubleconfirm was dropped here so that if the password is already saved the user would not have to type it in again, the middleware should let them know if they need one

PR #3461, opened 2020-01-12
Update fr.po

MDVAurelien:
Update of some translations

### DIFF
--- a/src/assets/i18n/fr.po
+++ b/src/assets/i18n/fr.po
@@ -45,6 +45,8 @@ msgid ""
 " addresses. When this has been chosen, additional addresses cannot be "
 "selected."
 msgstr ""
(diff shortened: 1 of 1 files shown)


erinix:
if you want your changes in 11.3 please also make a PR that targets freenas/11.3-stable

MDVAurelien:
Thank you it's done. In your opinion how I can visualize the modifications on the 11.3 branch? Is it always with nightly builds? I now have to do my translations on the two branches (11.3 and Master)? Thks & Happy new year ;-)

erinix:
at this point the nightlies will be 12, the only way to really see 11.3 will be in releases or alternatively you could try setting up the build environment as per the README file

MDVAurelien:
Thank you Erin, that's clear.

PR #3462, opened 2020-01-13
Update zh-hans.po

ZX-GuiSu:


erinix:
if you want your changes in 11.3 please also make a PR that targets freenas/11.3-stable

ZX-GuiSu:
ok, but i have a question, how do I know what new fields have been updated and where can I see them?

erinix:
I was able to cherry-pick this one in, however the next one probably won't be as simple.  Nightlies will be 12.0 now so any changes that go into that one should be there, if you need to view your changes for the 11.3 the only way you will see them will either be in releases or you can try setting up the dev environment as per the README.

ZX-GuiSu:
Thank you for your answer.  I think there is something wrong with my description.  What I want to express is that if there are new fields in the po file, how can I quickly find it for translation?

PR #3463, opened 2020-01-13
NAS-104676 / 12.0 / Add a process.exit to the script so it can error out on bad PO files

erinix:
NAS-104676


erinix:
see https://github.com/freenas/webui/pull/3464, this should make it where really bad PO files (ones that po2json cannot skip text on) are caught by the build process

PR #3486, opened 2020-01-15
NAS-100806 / 12.0 / Nas 100806

lola27:


erinix:
looks nice, one suggestion, could this page show somewhere on it which disk it is currently viewing the smart results for?  I feel like once I click into it I cannot tell

src/app/pages/storage/disks/smart-results/smart-results.component.ts
```diff
@@ -32,6 +32,7 @@ export class SmartResultsComponent {
     preInit(entityForm: any) {
         this.aroute.params.subscribe(params => {
             this.disk = params['pk'];
+            this.title = T(`S.M.A.R.T Test Results of ${this.disk}`);
```
erinix: this won't work for translations, use translate.get to get the text then append the this.disk to it

PR #3491, opened 2020-01-15
Translation until 14.01.2020

Liliththerebel:
I started with some translation in italian

### DIFF
--- a/src/assets/i18n/it.po
+++ b/src/assets/i18n/it.po
@@ -19,7 +19,7 @@ msgstr ""
 msgid " Plugin"
 msgstr ""
 
-msgid " Service"
+msgid " Servizi"
 msgstr ""
 
 msgid ""
--- a/src/assets/i18n/it.po
+++ b/src/assets/i18n/it.po
@@ -35,8 +35,8 @@ msgid ""
 msgstr ""
 
 msgid ""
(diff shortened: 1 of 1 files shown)


src/assets/i18n/it.po
```diff
@@ -4,10 +4,10 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 
-msgid " Check Notifications for more details."
+msgid " Controlla le notifiche per maggiori dettagli."
```
erinix: these wont work, the msgid needs to be the original english and the msgstr needs to be your translation

erinix:
also if you want to see your changes go into 11.3 please make a PR against the branch freenas/11.3-stable

src/assets/i18n/it.po
```diff
@@ -104,7 +104,7 @@ msgstr ""
 msgid "(Nessuna descruzione)"
 msgstr ""
 
-msgid "(Remove pool from database)"
+msgid "(Rimuovere pool dal database)"
```
erinix: this still isn't correct, it needs to look like this:

msgid "(Remove pool from database)"
msgstr "(Rimuovere pool dal database)"

The translations uses the english string as the id to match to the translated string, it cannot find your translations if it does not know which string to match to

PR #3496, opened 2020-01-15
NAS-104006 / 12.0 / Nas 104006

lola27:


src/app/pages/common/entity/entity-form/entity-form.component.ts
```diff
@@ -198,6 +198,9 @@ export class EntityFormComponent implements OnInit, OnDestroy, OnChanges, AfterV
       if (this.conf.isEntity) {
         this.pk = params['pk'];
         if (this.pk && !this.conf.isNew) {
+          if (this.conf.saveSubmitText === undefined) {
```
erinix: we could also possibly use similar logic to set the button to be "Save", if (!this.conf.isEntity and !this.conf.isNew and !this.pk), I think this would take care of single forms like system general, etc.  I notice you manually set that on many pages but I wonder if it isn't possible to cover most of those with a default like that.

lola27: Thanks @erinix but `if (!this.conf.isEntity and !this.conf.isNew and !this.pk)` won't works. i will try another way to make it more easier.

PR #3511, opened 2020-01-17
Do not display clickable button for LEGACY periodic snapshot tasks

themylogin:
@erinix this is needed for 11.3-RELEASE


william-gr:
@dkmullen do you mind reviewing that, plese? We need to fix go in as soon as we can and @erinix starts late

william-gr:
Thanks!

PR #3512, opened 2020-01-17
NAS-102288 / None / NAS-102288: move dry run dialog to help text file

Mrt134:
- Add a little more text to the dialog to explain what will happen.


Mrt134:
I was able to finally get a 12.0 dev system up and running. Here's what my changes look like after building: 
Screenshot_2020-01-17 TrueNAS - localhost

PR #3513, opened 2020-01-17
NAS-104727 / 11.3 / Nas 104727

dkmullen:
Ensures that source data from replication wizard and form gets converted from string to array; also gets rid of user-entered whitespace. Thanks to Vladimir Vinogradenko for the clean lodash solution


dkmullen:
> Does this affect regular replication add/edit form?

Yes - meaning the form that you get via the Advanced button on the wizard, which is the same form you get when you edit an existing task

PR #3575, opened 2020-01-28
NAS-104796 / 12.0 / Nas 104796 master

dkmullen:
Moves password comparison validation to the Validation service for user form and four other components. We weren't able to confirm that this fixes the problem in the ticket, but it does fix the issue of the validators that stop listening when the user navigates away from the page, then comes back.

Also please check: It seemed like a mistake that the password for dynamic DNS had a minLength validator of 8, since it's the pw for any number of outside services


erinix:
This seems to have a different problem I see on users form, it doesn't match validation at all
screenshot-20200128-145626

PR #3579, opened 2020-01-29
NAS-101577 / 12.0 / Nas 101577

dkmullen:
Adds choice of date and time format in Sys General which applies to Task Manager, Reports, Boot Environments, Snapshot List DETAILS (snapshots often are named by date too, which doesn't change), Notifications (requires refresh to take effect), Date Picker (the list part), Certificates/CAs. 

-Dash widget displays an AM/PM time as part of a string with other info. Would like to ask MW team to supply this as a 24 hour time
- There are known issues (in all branches) with the time in Task Manager which are covered in another ticket; seem to be some timezone issues in a few places which aren't unique to this ticket.


erinix:
could you review the text @freenas/docs-team

PR #3585, opened 2020-01-30
NAS-104855 / 11.3 / Remove excess \ before # that is added by translate service

erinix:
NAS-104855


erinix:
I'm not a huge fan of this solution, it strips off the \ from the #, but it seems to be the translation service that is putting the \ in there in the first place, do you see any concerns with this @freenas/docs-team ?

nugget:
The code as written addresses backslashes but in both of the related Jira tickets the bug reports reference forward slashes.

erinix:
Yea the browser seems to convert the \ into a /

PR #3587, opened 2020-01-30
NAS-104344 / 12.0 / NAS-104344

lola27:


src/app/helptext/jails/storage.ts
```diff
@@ -19,5 +19,7 @@ readonly_placeholder: T('Read-Only'),
 readonly_tooltip: T('Set to prevent users from modifying the \
 <b>Destination</b>.'),
 
+acl_edit_msg: T('Root dataset ACL cannot be edited.'),
```
erinix: we probably want a different tooltip for this one since jails will never be on the root dataset ACL,  could you please assist with a tooltip for this one @freenas/docs-team

Mrt134: Here's a text-only suggestion:
`ACL cannot be changed when the Source is the top-level dataset for a pool.`

PR #3605, opened 2020-02-05
NAS-104940 / 12.0 / Nas 104940

lola27:


src/app/pages/common/entity/entity-table/entity-table.component.html
```diff
@@ -291,6 +291,11 @@
                  ix-auto-type="value"
                  [ix-auto-identifier]="row[conf.rowIdentifier || 'name']">
               {{convertDisplayValue(row[col.prop])}}
+              <button mat-icon-button *ngIf="row[col.icon] && conf.iconAction"
```
erinix: The icon might look better on the left instead of the right of the name

PR #3652, opened 2020-02-14
NAS-103602 / 12.0 / Nas 103602

lola27:


src/app/components/common/layouts/admin-layout/admin-layout.template.html
```diff
@@ -16,7 +16,7 @@
             </ng-template>
           </div>
           <ng-template #lightLogo>
-            <div class="branding mat-bg-primary" *ngIf="is_freenas; else isTruenasLight">
+            <div class="branding mat-bg-primary" *ngIf="product_type == 'CORE'; else isTruenasLight">
```
erinix: This shouldn't be necessary, it should just always show TrueNAS Logos (the branding of TrueNAS will be same between Core and Enterprise)

src/app/core/components/widgets/widgetsysinfo/widgetsysinfo.component.ts
```diff
@@ -168,7 +168,7 @@ export class WidgetSysInfoComponent extends WidgetComponent implements OnInit,On
       } else {
         this.manufacturer = "other";
       }
-      if (this.is_freenas === 'true') {
+      if (this.product_type === 'CORE') {
```
erinix: same here, just show TrueNAS logo

PR #3653, opened 2020-02-14
NAS-104745 / 12.0 / Nas 104745

lola27:


src/app/pages/jails/jail-form/jail-form.component.ts
```diff
@@ -861,18 +861,19 @@ export class JailFormComponent implements OnInit, AfterViewInit {
...
-        (res) => {
-          for (let i in res) {
-            this.interfaces.vnetDisabled.push({ label: res[i].name, value: res[i].name });
-          }
-        },
-        (res) => {
-          new EntityUtils().handleWSError(this, res, this.dialogService);
+    this.jailService.getInterfaceChoice().subscribe(
+      (res) => {
+        for (const i in res) {
```
william-gr: The key of jail.interface_choices should be the value, the label should be the value of the object because it can return values like:

{"igb0": "igb0: description here"}

lola27: get it. thanks @william-gr

PR #3659, opened 2020-02-16
Fix an entity detail height update error

JoeyWNK:
Test with Chrome  79.0.3945.130 (64-bit) / Firefox 73.0 (64-bit)
Patch: The datatable-body class does not receive height update in Chrome
Potential issue: Cannot have more than one ngx-datatable in single page


JoeyWNK:
Testing some better options

PR #3660, opened 2020-02-17
Fix Detail Display Issue

JoeyWNK:
Pre allocate the space to avoid manually changing height which is not proerly handled.
Another solution is to directly set tableHeight value rather than directly edit the style. Set tableHeight value can properly trigger ngx-table componement recalculation.


erinix:
I created a ticket to track this, see https://jira.ixsystems.com/browse/NAS-105282

erinix:
So I am not quite certain that we want to merge this into 11.3 but I wouldn't mind trying it in 12.0/master.  Would you mind adjusting this PR so that it targets master instead of 11.3?

JoeyWNK:
That will be fine. Should I close this request and open another one?

erinix:
Yea go ahead and do that :)

PR #3671, opened 2020-02-19
NAS-105008 / 11.3 / Nas 105008

dkmullen:
Fix timezone issues in form scheduler, reports, snapshots. Thanks to Aaron Ervin for extra help

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "hopscotch": "0.3.1",
     "lodash": "4.17.15",
     "mobx-angular": "3.0.1",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -77,6 +77,7 @@
     "hopscotch": "0.3.1",
     "lodash": "4.17.15",
     "mobx-angular": "3.0.1",
+    "moment-timezone": "^0.5.27",
```
erinix: remove the ^, those things can bite us if an updated version breaks things

PR #3672, opened 2020-02-19
Update pl.po

adam23450:
Translations continued

### DIFF
--- a/src/assets/i18n/pl.po
+++ b/src/assets/i18n/pl.po
@@ -77,19 +77,19 @@ msgid ""
 msgstr ""
 
 msgid " service has been enabled."
-msgstr ""
+msgstr "usługa została wyłączona"
 
 msgid " should be greater than "
-msgstr ""
+msgstr " powinno być większe niż "
 
 msgid " should be less than "
-msgstr ""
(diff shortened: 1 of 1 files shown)


erinix:
you should add an entry into the README.md that lists polish as one of the languages :)

adam23450:
ok

adam23450:
Already :)

adam23450:
When will the language be available on the NAS?

erinix:
master goes into nightlies, you will see these updates in the next nightly and then in 12.0 when we release it, if you want your changes to go into an 11.3 version you should also make translations to the branch freenas/11.3-stable

adam23450:
How do I update it so that I can see the translations?

erinix:
these current changes will go into tonight's nightly, so you could download that and run it in VM (I wouldn't suggest using it on a production box just yet)

adam23450:
I understand. Thank you. Meanwhile, I'm going on with the further translation :)

erinix:
you're welcome, also I had a change request on your README PR, we are trying to keep the languages in that list in alphabetic order :)

adam23450:
ok

PR #3675, opened 2020-02-19
Update pl.po

adam23450:
Translations continued

### DIFF
--- a/src/assets/i18n/pl.po
+++ b/src/assets/i18n/pl.po
@@ -77,19 +77,19 @@ msgid ""
 msgstr ""
 
 msgid " service has been enabled."
-msgstr "usługa została wyłączona"
+msgstr ""
 
 msgid " should be greater than "
-msgstr " powinno być większe niż "
+msgstr ""
 
 msgid " should be less than "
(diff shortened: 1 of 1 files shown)


src/assets/i18n/pl.po
```diff
@@ -575,70 +575,70 @@ msgid ""
...
 
 msgid "AWS Region"
-msgstr "Region AWS"
+msgstr ""
 
 msgid "About"
 msgstr ""
 
 msgid "Access"
-msgstr "Dostęp"
```
erinix: I notice this PR removes some translations, was that intentional?

adam23450:
It was not intended. It must have happened by mistake

src/assets/i18n/pl.po
adam23450: From what I see everything is as it should be. Unless I look wrong? hmm

erinix: if you look at what it adds in in its place all you see is a "", I think your PR is somehow removing these, perhaps you were working off an older version of this file?

PR #3703, opened 2020-02-26
NAS-105155 / 12.0 / Nas 105155

lola27:


src/app/helptext/system/kmip.ts
```diff
@@ -0,0 +1,42 @@
+import { Validators } from "@angular/forms";
+import { T } from "app/translate-marker";
+
+export const helptext_system_kmip = {
+    fieldset_server: T('KMIP Server'),
+    fieldset_certificate: T('Certifiacte'),
```
Mrt134:
```suggestion
    fieldset_certificate: T('Certificate'),
```

src/app/helptext/system/kmip.ts
```diff
@@ -0,0 +1,42 @@
...
+    },
+
+    manage_sed_disks: {
+        placeholder: T('Manage SED Password'),
+        tooltip: T(''),
+    },
+
+    manage_zfs_keys: {
+        placeholder: T('Manage ZFS Keys'),
+        tooltip: T(''),
```
Mrt134: Use the KMIP server to manage ZFS encrypted dataset keys. The key server stores, applies, and destroys encryption keys whenever an encrypted dataset is created, when an existing key is modified, an encrypted dataset is unlocked, or an encrypted dataset is removed. Unsetting this option leaves all encryption key management with the local system.

src/app/helptext/system/kmip.ts
```diff
@@ -0,0 +1,42 @@
...
+    },
+
+    jobDialog: {
+        title: T('Saving KMIP Config'),
```
sonicaj: I am not sure if this is already being handled, we should be showing the job progress as well because during update if user switched the kmip server, we would be performing a migration from the old KMIP server to the new one and we should show what progress the migration is at.

william-gr:
@erinx @lola27 can we please get mockups done for this and be added to the NEP for it to be approved before we merged this? Thank you

src/app/helptext/system/kmip.ts
lola27: yes, the job dialog able to display the progress and description.

src/app/helptext/system/kmip.ts
```diff
@@ -0,0 +1,62 @@
...
+    },
+
+    syncInfoDialog: {
+        title: T('Sync Keys'),
+        info: T('Synced ZFS/SED keys between KMIP Server and TN database.'),
+    },
+
+    clearSyncKeyInfoDialog: {
+        title: T('Clear Sync Keys'),
```
sonicaj: I think it should be worded more appropriately that keys which are pending sync should be cleared, so something like `Clear Pending Sync Keys` ?

src/app/helptext/system/kmip.ts
```diff
@@ -0,0 +1,62 @@
...
+ leaves all encryption key management with the local system.'),
+    },
+
+    jobDialog: {
+        title: T('Saving KMIP Config'),
+    },
+
+    syncInfoDialog: {
+        title: T('Sync Keys'),
+        info: T('Synced ZFS/SED keys between KMIP Server and TN database.'),
```
sonicaj: `Synced` -> `Sync`

src/app/helptext/system/kmip.ts
```diff
@@ -0,0 +1,62 @@
...
+    },
+
+    clearSyncKeyInfoDialog: {
+        title: T('Clear Sync Keys'),
+        info: T('Cleared all keys which are pending to be synced between KMIP server and TN database.'),
```
sonicaj: `Cleared` -> `Clear`

lola27:
yes. the clear pending sync key button will be shown all the time. but if there is no sync pending, the key will be disabled.

PR #3735, opened 2020-03-04
NAS-105089 / 12.0 / NAS-105089

lola27:


erinix:
works in chrome but I can't seem to get it to work in firefox

PR #3753, opened 2020-03-06
NAS-105253 / 11.3 / Nas 105253

dkmullen:
Adds descriptions for presets on Form Scheduler; hides 'Begin' and 'End' settings on Replication form, Periodic Snapshot Tasks form, unless the chosen setting is Hourly or Custom


erinix:
That's actually a pretty interesting solution, hiding the start and end on the settings where it isn't actually relevant, good job.  Do these labels make sense @freenas/docs-team ?

PR #3769, opened 2020-03-10
NAS-100331 / 12.0 / Rename extend pool to "add vdevs"

erinix:
NAS-100331

### DIFF
--- a/src/app/pages/storage/storage.routing.ts
+++ b/src/app/pages/storage/storage.routing.ts
@@ -75,7 +75,7 @@ export const routes: Routes = [
           },
           {
             path: 'manager/:pk', component: ManagerComponent,
-            data: { title: 'Extend Pool', breadcrumb: 'Extend Pool' }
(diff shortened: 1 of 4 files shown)


erinix:
In this PR I rename the menu item and the breadcrumb text to say "Add Vdevs" (unlike the original ticket which says "Add Vdev", you can technically add more than one to a pool at a time so I thought the pluralization made more sense).  There is still "Extend" and "Extend Pool" on the buttons for submitting and confirming when creating, I wasn't sure if we wanted to change it in those places too.  What do you think @freenas/docs-team ?

Mrt134:
Either way should be ok - Add VDEV on the submit / confirm button is more consistent and specific to what is happening, but Extend Pool seems more descriptive of the general action that is being performed and seems to fit better with that kind of a button. I personally tilt towards consistency and just having "Add VDEV", but that's largely preference.

PR #3781, opened 2020-03-11
i18n: es-ar updated

castillofrancodamian:
Signed-off-by: Franco Castillo <castillofrancodamian@gmail.com>


erinix:
I notice that this PR does some weird things with many of the message ids, but it baffles me because there are no merge conflicts so I don't know how they got changed the way they did, I could go ahead and merge and your changes would go in, you don't appear to be using most of the weird message ids for anything so they would just get changed again when the next extract job runs but it still seems weird.  What editor did you use for this?

castillofrancodamian:
Poedit. Debe ser por este editor. Sólo hice modificaciones en las cadenas traducibles. Es mi primera contribución al proyecto.

castillofrancodamian:
Poedit. It must be by this publisher. I only made modifications to the translatable strings. It is my first contribution to the project.

erinix:
Yea it looks like the ones that got changed were the longer things like tooltip descriptions, not sure if they would work in the new form or not

castillofrancodamian:
And how could I test this translation in FreeNAS?

erinix:
other than waiting for it to hit a nightly the only other way is to setup the dev environment like the readme.  I'll manually test it before I push it in just to see if it still works ok

castillofrancodamian:
Would I need to compile? How often are the nightly versions updated? It is my first approach to this project and I do not know how they are handled.

erinix:
there should be a new nightly tomorrow you'll see these in, if you want to compile locally you'll need to install yarn and node and then setup yarn so you can compile locally, there's a readme at the base of https://github.com/freenas/webui that explains how to set this up

castillofrancodamian:
Thanks i prefer to wait 😁

PR #3792, opened 2020-03-16
NAS-100045 / 12.0 / NAS-100045 PCI-E passthru support (webui part)

stattin42:
webui support for PCI-E passthru in vm plugin
DEPENDS: https://github.com/stattin42/freenas/tree/NAS-100045-12.0
ok to test


stattin42:
PR rebased onto current master.

stattin42:
PR rebased onto current master.

stattin42:
@sonicaj, thank you for attending to the PR to the `middlewared` `vm` plugin. I think it would be useful with corresponding support in the webui. If you agree and think time is right, could you suggest a suitable reviewer for this PR?

stattin42:
@erinix, F Y I, I noticed there were a few instances of leading tabs in the PR so I changed those to spaces and rebased the PR on current master while I was at it. Thanks!

PR #3807, opened 2020-03-18
NAS-105395 / 12.0 / Nas 105395

dkmullen:
Removes range low/high from smb form, updates load and save methods to accept aliases as an array, provides custom validation for aliases


erinix:
This looks pretty nice although I am wondering if for the 12 version of this ticket we shouldn't convert the aliases into a mat-chip since it is a space separated thign

PR #3844, opened 2020-03-31
NAS-105566 / 12.0 / NAS-105566 Force topbar to present eula on license enter

dkmullen:
TN Core returns true for 'is_eula_accepted' so this PR forces the eula to be presented when a new license is entered in core. PS - truenas.get_eula is returning null right now on 12

### DIFF
--- a/src/app/components/common/topbar/topbar.component.ts
+++ b/src/app/components/common/topbar/topbar.component.ts
@@ -85,7 +85,7 @@ export class TopbarComponent extends ViewControllerComponent implements OnInit, 
       });
     }
 
-  ngOnInit() {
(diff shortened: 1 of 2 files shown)


src/app/components/common/topbar/topbar.component.ts
```diff
@@ -85,7 +85,7 @@ export class TopbarComponent extends ViewControllerComponent implements OnInit,
       });
     }
 
-  ngOnInit() {
+  ngOnInit() { console.log(window.localStorage.getItem('product_type'))
```
erinix, line 88: you left a console.log in here :)

PR #3853, opened 2020-04-02
NAS-105589 / 12.0 / NAS-105589

dkmullen:
Add checkbox to force HDD standby

### DIFF
--- a/src/app/helptext/storage/disks/disks.ts
+++ b/src/app/helptext/storage/disks/disks.ts
@@ -33,8 +33,8 @@ disk_form_hddstandby_options: [
 
 force_hdd_standby: {
     placeholder: T('Force HDD Standby'),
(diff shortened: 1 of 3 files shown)


src/app/helptext/storage/disks/disks.ts
```diff
@@ -31,6 +31,12 @@ disk_form_hddstandby_options: [
  {label:'330', value: '330'},
 ],
 
+force_hdd_standby: {
+    placeholder: T('Force HDD Standby'),
+    tooltip: T('Performing SMART operations on some drives prevents them from sleeping, \
+ even when the operations involve no physical activity. Enable to keep drive on standby.')
```
Mrt134, line 37:
```suggestion
    tooltip: T('Allows the drive to enter standby, even when non-physical S.M.A.R.T. \
 operations could prevent the drive from sleeping.')
```

PR #3886, opened 2020-04-10
NAS-105459 / 12.0 / Nas 105459

erinix:


erinix:
note: this is still a bit of a work in progress, it still needs a bit of css tweaking and various scenarios tested, I am making a PR so docs can start reviewing the text

src/app/helptext/storage/volumes/datasets/dataset-unlock.ts
```diff
@@ -0,0 +1,38 @@
...
+    unlock_children_placeholder: T('Unlock Children'),
+    unlock_children_tooltip: T('Also unlock any encrypted dataset stored within this dataset.'),
+    restart_services_placeholder: T('Restart attached VMs, Jails and Services'),
+    restart_services_tooltip: T('Restarting system services after unlocking the dataset makes \
+     the information contained within the dataset available to the rest of the system.'),
+    upload_key_file_placeholder: T('Upload Key file'),
+    upload_key_file_tooltip: T('Browse to the exported key file that can be used to unlock this \
+     dataset.'),
+    dataset_key_placeholder: T('Dataset Key'),
+    dataset_key_tooltip: T(''),
```
erinix, line 17: @freenas/docs-team can I get a tooltip suggestion for this one?  I presume it could be something similar to how dataset_passphrase is, it is a text area that the user can copy and paste the key into

Mrt134: I couldn't find this field when build testing. I'm updating my system and will look for it again.

Mrt134, line 17: After investigating, it looks like this field isn't used or is hidden. I can't get it to show up in a yarn build, but from looking at the code, here's my suggestion.
```suggestion
    dataset_key_tooltip: T('Encryption key that can unlock the dataset.'),
```

erinix: It is hidden by default, you have to uncheck inherit, then set the key type instead of passphrase

PR #3901, opened 2020-04-15
NAS-105744 / 12.0 / Nas 105744

dkmullen:
First time you see it (on first boot), tagline at top, instructions at bottom
Screenshot from 2020-04-15 14-00-16

After that...
Screenshot from 2020-04-15 13-59-17


erinix:
text should be translatable and preferably in helptext files, other than that this PR looks good

PR #3920, opened 2020-04-17
NAS-105762 / 12.0 / Add final results dialog to show after datasets have been unlocked

erinix:
NAS-105762

### DIFF
--- a/src/app/helptext/storage/volumes/datasets/dataset-unlock.ts
+++ b/src/app/helptext/storage/volumes/datasets/dataset-unlock.ts
@@ -32,6 +32,11 @@ export default {
         ok_button: T('Continue'),
         cancel_button: T('Cancel')
     },
+    unlock_result_dialog: {
+        errors_message: T('These datasets could not be unlocked.'),
+        unlock_message: T('These datasets were successfully unlocked.'),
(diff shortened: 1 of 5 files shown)


src/app/helptext/storage/volumes/datasets/dataset-unlock.ts
```diff
@@ -32,6 +32,11 @@ export default {
         ok_button: T('Continue'),
         cancel_button: T('Cancel')
     },
+    unlock_result_dialog: {
+        errors_message: T('These datasets could not be unlocked.'),
+        unlock_message: T('These datasets were successfully unlocked.'),
+        skipped_message: T('These datasets were skipped since their parents could not be unlocked.'),
```
Mrt134, line 38: "Skipped" seems a little ambiguous when we could just say if the dataset was locked or unlocked.
```suggestion
        skipped_message: T('These datasets were not unlocked because the parent datasets could not be unlocked.'),
```

erinix: I was tempted to word this this as "not unlocked because the parent datasets could not be unlocked" too so I'll merge that one

PR #3921, opened 2020-04-17
NAS-105758 / 12.0 / Nas 105758

erinix:


src/app/pages/storage/volumes/volumes-list/volumes-list.component.ts
```diff
@@ -1153,7 +1153,8 @@ export class VolumesListTableConfig implements InputTableConf {
           onClick: (row) => {
             // open encryption options dialog
             const can_inherit = (row.parent && row.parent.encrypted);
-            const is_key = !row.is_passphrase;
+            const passphrase_parent = (row.parent && row.parent.key_format.value === 'PASSPHRASE');
```
dkmullen, line 1156: On a top-level dataset (/mnt/dataset) I get a console error and the dialog doesn't open because the parent doesn't have a key_format

image

PR #3931, opened 2020-04-20
NAS-105687 / 12.0 / Nas 105687 master

dkmullen:
When navigating from SMB List or form to ACL editor, check first if pool or dataset is locked

### DIFF
--- a/src/app/pages/sharing/smb/smb-form/smb-form.component.ts
+++ b/src/app/pages/sharing/smb/smb-form/smb-form.component.ts
@@ -386,9 +386,33 @@ export class SMBFormComponent {
         ),
         tap(([doConfigureACL, dataset]) =>
           doConfigureACL
-            ? this.router.navigate(
-                ['/'].concat(ACLRoute)
-              )
+          // Make sure pool or dataset is online
+          ? this.ws.call('pool.dataset.query', [[["name", "=", poolName]]]).subscribe(ds => {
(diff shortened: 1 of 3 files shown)


src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -386,9 +386,33 @@ export class SMBFormComponent {
...
+                  )
+                } else {
+                  this.dialog.errorReport(helptext_sharing_smb.action_edit_acl_dialog.title,
+                    `${helptext_sharing_smb.action_edit_acl_dialog.message1} <i>${datasetId}</i> 
+                      ${helptext_sharing_smb.action_edit_acl_dialog.message2}`);
+                }   
+              })
+            } else {
+              // ZFS Encryption
+              if (ds[0].locked) {
```
erinix, line 406: This only matches the root dataset, if the root dataset is unlocked or not-encrypted but dataset itself is encrypted and locked it would still allow them to go to the acl editor.  You should check to see if the dataset itself is locked (if it is a dataset, if it is a folder figure out its parent and see if the parent dataset is locked).

erinix: Also the error text needs to reflect that it is the dataset that is locked and not the pool in this case.

PR #3947, opened 2020-04-23
NAS-105296 / 11.3 / Nas 105296

dkmullen:
The mw call is in the stable / master repositories but not yet in a build
New line in helptext is from api doc


erinix:
could you review the text @freenas/docs-team

PR #3952, opened 2020-04-23
NAS-105377 / 12.0 / Nas 105377

lola27:


erinix:
could you please review the text @freenas/docs-team

src/app/views/sessions/signin/signin.component.html
```diff
@@ -99,6 +99,13 @@
             </div>
           </div></span>
 
+          <div class="truecommand">
```
william-gr, line 102: Maybe we should only show this if TC is configured/connected?

@kmoore134 @beanpole135 should we have a link to the TC wireguard IP here instead of the static portal link?

beanpole135: The Wireguard endpoint is not useful to show at all. There is nothing there that the user can interact with. Showing the NAS's IP within the Wireguard network will probably be useful though.

william-gr: How so? Isnt the wireguard endpoint IP the IP of the truecommand instance? Its two ways, no?

src/app/components/common/topbar/topbar.component.ts
```diff
@@ -540,4 +560,113 @@ export class TopbarComponent extends ViewControllerComponent implements OnInit,
...
+        },
+        {
+          type: 'checkbox',
+          name: 'enabled',
+          placeholder: helptext.updateDialog.enabled_placeholder,
+          tooltip: helptext.updateDialog.enabled_tooltip,
+          value: true,
+        }
+      ],
+      saveButtonText: self.tcConnected ? helptext.updateDialog.save_btn : helptext.updateDialog.connect_btn,
```
sonicaj, line 617: I am not sure how this shows up in UI, but if the user changes the api key - then essentially we are starting the process afresh by registering the key with the portal and then polling. So I think it might be nice if we handle it like that

src/app/helptext/topbar.ts
```diff
@@ -36,6 +36,24 @@ export default {
...
+        connect_btn: T('CONNECT'),
+    },
+
+    updateDialog: {
+        title_connect: T('Connect to TrueCommand Cloud'),
+        title_update: T('Update TrueCommand Settings'),
+        connect_btn: T('CONNECT'),
+        save_btn: T('SAVE'),
+        api_placeholder: T('API Key'),
+        api_tooltip: T(''),
```
Mrt134, line 54:
```suggestion
        api_tooltip: T('Enter or paste the API key provided from \
 <a href="https://portal.ixsystems.com/portal/login/" target="_blank">iXsystems Account Services</a>. \
 Login or signup is required.'),
```

src/app/helptext/topbar.ts
```diff
@@ -36,6 +36,24 @@ export default {
...
+        enabled_placeholder: T('Enable'),
+        enabled_tooltip: T(''),
```
Mrt134, line 56:
```suggestion
        enabled_tooltip: T('Immediately connect to TrueCommand.'),
```

src/app/views/sessions/signin/signin.component.html
kmoore134: We should be showing the WG internal IP if we show anything, not the public endpoint IP. It should be showing 172.10.x.1 or similar which is the legit TC http/https webUI address.

kmoore134: If we can't get the WG internal IP / network information, then no point in showing anything beyond portal link.

PR #3958, opened 2020-04-24
NAS-105805 / 12.0 / Nas 105805

lola27:


lola27:
@dkmullen yes, I see the difference, only a few lines contains HTML tag. but if use innerHTML for such big text content, it will crash on mac with firefox, and take longer time to load.

PR #3999, opened 2020-04-30
NAS-105963 / 11.3 / Have the application refresh if user does not commit network changes

erinix:
NAS-105963


erinix:
this is actually for 105964

PR #4072, opened 2020-05-11
NAS-106079 / 12.0 / Nas 106079

dkmullen:
Checks the non active field in a custom compare validator to make sure errors get turned off as needed

### DIFF
--- a/src/app/pages/sharing/iscsi/authorizedaccess/authorizedaccess-form/authorizedaccess-form.component.ts
+++ b/src/app/pages/sharing/iscsi/authorizedaccess/authorizedaccess-form/authorizedaccess-form.component.ts
@@ -173,13 +173,14 @@ export class AuthorizedAccessFormComponent {
             delete errors['manualValidateError'];
             delete errors['manualValidateErrorMsg'];
           }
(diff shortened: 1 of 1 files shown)


src/app/pages/sharing/iscsi/authorizedaccess/authorizedaccess-form/authorizedaccess-form.component.ts
```diff
@@ -171,8 +173,20 @@ export class AuthorizedAccessFormComponent {
             delete errors['manualValidateError'];
             delete errors['manualValidateErrorMsg'];
           }
+          // If the error gets cleared in this comparison validator, make sure it is cleared for both fields
+          if (otherCtrl.hasError('manualValidateError')) {
+            delete otherErrors['manualValidateError'];
+            delete otherErrors['manualValidateErrorMsg'];
+            delete otherErrors['matchesOther'];
```
lola27, line 180: do we need to delete the validation errors of otherCtrl here? since after we reset the value in line183, the validator will check it again.

dkmullen: Good point

PR #4103, opened 2020-05-20
Added Nord Theme

Ninja73737:
Hello!

Love the Nord color scheme and I made a custom one for my server, so I figured if you guys are open to it it could be added to the official themes, no worries if it's not quite up to the standard though!

I followed the theme's [official documentation](https://www.nordtheme.com/docs/colors-and-palettes) pretty closely, the only colour that's not part of the official palette is the violet because there's no corresponding colour.


dkmullen:
Nice!
Screenshot from 2020-05-20 09-36-18

Ninja73737:
Thanks, and also thanks for adding the screenshot, probably should've done that in the first place

PR #4118, opened 2020-05-22
NAS-106204 / 12.0 / NAS-106204 Expose loglevel for AFP

dkmullen:
The update to mw doesn't appear to be in the build yet.
The tooltip is the same as for log level in Services > SMB. It probably should be checked in both places to make sure it is up-to-date and provides help.

### DIFF
--- a/src/app/helptext/services/components/service-afp.ts
+++ b/src/app/helptext/services/components/service-afp.ts
@@ -38,9 +38,9 @@ afp_srv_chmod_request_tooltip: T('Indicates how to handle Access Control Lists.\
  <b>Simple:</b> is set to chmod() as requested\
  without any extra steps.'),
 afp_srv_chmod_request_options : [
-  {label : 'Ignore', value : 'IGNORE'},
-  {label : 'Preserve', value : 'PRESERVE'},
-  {label : 'Simple', value : 'SIMPLE'},
+  {label : T('Ignore'), value : 'IGNORE'},
+  {label : T('Preserve'), value : 'PRESERVE'},
+  {label : T('Simple'), value : 'SIMPLE'},
 ],
 
 afp_srv_map_acls_placeholder : T('Map ACLs'),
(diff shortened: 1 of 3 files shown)


src/app/helptext/services/components/service-afp.ts
```diff
@@ -49,11 +49,23 @@ afp_srv_map_acls_tooltip: T('Select mapping of permissions for\
...
-  {label : 'None', value : 'NONE'},
-  {label : 'Mode', value : 'MODE'},
+  {label : T('Rights'), value : 'RIGHTS'},
+  {label : T('None'), value : 'NONE'},
+  {label : T('Mode'), value : 'MODE'},
 ],
 
+loglevel: {
+  placeholder: T('Log Level'),
+  tooltip: T('Choices are <i>Minimum, Normal, or Debug</i>.'),
```
Mrt134, line 59: Suggestion to add more of a usage explanation to the help text:

```suggestion
  tooltip: T('Record AFP service messages up to the specified log level in the system log. By default, severe and warning level messages are logged.'),
```

dkmullen: @Mrt134 , should we apply the same text to log-level in Services > SMB?

Mrt134: I checked with Andrew and the text is generally applicable with one change as SMB has multiple log files:
```
Record SMB service messages up to the specified log level. By default, error and warning level messages are logged.

Samba messages are logged to various files:
<i>/var/log/samba4/log.smbd</i> : smbd process
<i>/var/log/samba4/log.winbindd</i> : winbindd parent process
<i>/var/log/samba4/log.wb-{DOMAIN} : forked winbindd child domain processes
<i>/var/log/samba4/log.nmbd</i> : NetBIOS name server
<i>/var/log/samba4/auth_audit.log</i> : authentication logs
```

PR #4127, opened 2020-05-26
NAS-106249 / 12.0 / NAS-106249

lola27:


src/app/helptext/system/alert-settings.ts
```diff
@@ -0,0 +1,276 @@
+import { T } from '../../translate-marker';
+
+export default {
```
Mrt134, line 3: Does each type of alert need a separate tooltip, or can we create a one help text field for "level" and one for "policy" and have the pages/system/alert.component.ts file use these two tooltips for each "Warning Level" and "Set Frequency" field that is created?

a-stjohn: @Mrt134 I've already discussed with Lola that it's going to be 2 tool tips total in source. They will be applied to each clone in the UI. Working on this now.

PR #4133, opened 2020-05-27
Update fr.po

MDVAurelien:
Add some translations


erinix:
I notice there is still some problem with fr.po, perhaps a msgstr missing, it seems to break the extraction job.  If you have any ideas on how to find it I would be grateful for a fix.

MDVAurelien:
It's fixed, sorry it's a mistake on my part.

PR #4163, opened 2020-06-03
NAS-106348 / 12.0 / NAS-106348

lola27:
empty value


src/app/pages/services/components/service-nfs/service-nfs.component.ts
```diff
@@ -77,18 +77,21 @@ export class ServiceNFSComponent {
       config: [
         {
           type: 'input',
+          inputType: 'number',
```
erinix, line 80: rather than inputType we should probably use a range validator here like we do in other services, there is probably a specific numeric range that these can take, either the full range of ports or some specific port range (also number doesn't work right in Firefox)

PR #4350, opened 2020-07-08
NAS-106665 / 12.1 / Add salt to buildtime url to prevent caching of buildtime file

erinix:
NAS-106665


erinix:
the way I tested, hotpatch a box with two copies of the dist folder, edit assets/buildtime to be different in one directory, swap directories and go back to the login page

lola27:
LGTM, the UI will fresh if build time changed, and able to automatically login if the token is valid.

PR #4398, opened 2020-07-16
NAS-105099 / 12.1 / Improve naming schema tooltip

themylogin:


themylogin:
@erinix can you please verity that my fix fixes the tooltip which is now reading `The default is snap-{0}-{1}-{2}-{3}-{4}. Must include the strings {5}, {6}, {7}, {8}` (`{1}` instead of `%Y`)?

erinix:
> @erinix can you please verity that my fix fixes the tooltip which is now reading `The default is snap-{0}-{1}-{2}-{3}-{4}. Must include the strings {5}, {6}, {7}, {8}` (`{1}` instead of `%Y`)?

screenshot-20200717-124117

themylogin:
@erinix I want them to be `%Y-%m-%d` but if I type this literally I see `{0}-{1}-{2}`. I've tried `%%Y` as most common method to escape % in %-encoded strings, but, as we see, it does not do the job right. Can you please try to find a way to make it say literally `%Y`?

erinix:
I fixed it and pushed to branch, apparently what we needed was \%

PR #4412, opened 2020-07-20
NAS-106859 / 12.1 / Nas 106859

dkmullen:
On SCALE:
- Enables VNC Settings for either bootloader options but...
- Removes Delay VM Boot Until VNC Connects
- Doesn't offer Grub as an option

On Core, Enterprise
- hides VNC options for Grub and Legacy

On all:
- Exposes verbose bootloader descriptions to the user

### DIFF
--- a/src/app/pages/vm/vm-wizard/vm-wizard.component.ts
+++ b/src/app/pages/vm/vm-wizard/vm-wizard.component.ts
@@ -456,8 +456,7 @@ export class VMWizardComponent {
 
 
     ( < FormGroup > entityWizard.formArray.get([0]).get('bootloader')).valueChanges.subscribe((bootloader) => {
-      console.log(bootloader)
-      if(this.productType !== 'SCALE' && bootloader === 'UEFI_CSM'){
+      if(this.productType !== 'SCALE' && bootloader !== 'UEFI'){
         _.find(this.wizardConfig[0].fieldConfig, {name : 'enable_vnc'})['isHidden'] = true;
         _.find(this.wizardConfig[0].fieldConfig, {name : 'wait'})['isHidden'] = true;
       _.find(this.wizardConfig[0].fieldConfig, {name : 'vnc_bind'}).isHidden = true;
@@ -614,7 +613,7 @@ export class VMWizardComponent {
       }
       else {
         if (!grub && this.productType !== 'SCALE') {
-          this.bootloader.options.push({label : 'Grub', value : 'GRUB'});
+          this.bootloader.options.push({label : 'Grub bhyve (specify grub.cfg)', value : 'GRUB'});
         }
         ( < FormGroup > entityWizard.formArray.get([1])).controls['vcpus'].setValue(1);
         ( < FormGroup > entityWizard.formArray.get([1])).controls['cores'].setValue(1);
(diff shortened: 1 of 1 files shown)


src/app/pages/vm/vm-wizard/vm-wizard.component.ts
```diff
@@ -874,9 +880,20 @@ async customSubmit(value) {
...
-    
-    if(value.enable_vnc &&value.bootloader !== "UEFI_CSM"){
-      vm_payload["devices"].push({
+
+    if (value.enable_vnc) {
+      if (this.productType === 'SCALE') {
+        vm_payload["devices"].push({
+          "dtype": "VNC", "attributes": {
+            "vnc_port": String(this.getRndInteger(5900,65535)),
+            "vnc_resolution": "1024x768",
```
sonicaj, line 889: Setting resolution is not supported in SCALE. Can we please remove this ?

src/app/pages/vm/vm-wizard/vm-wizard.component.ts
```diff
@@ -886,7 +886,6 @@ async customSubmit(value) {
         vm_payload["devices"].push({
           "dtype": "VNC", "attributes": {
             "vnc_port": String(this.getRndInteger(5900,65535)),
-            "vnc_resolution": "1024x768",
```
erinix: do we want to keep it in conditionally for 12.1/freebsd? (I notice there is a backport of this too, does it make sense to port any of these changes back to 12?)

dkmullen: I kept it in for 12.1 and 12 ( few lines below the above snippet). The change to middleware affects CORE and ENTERPRISE too - https://github.com/freenas/freenas/pull/5229

erinix: I still don't see where you are keeping the resolution, what line is that on?

src/app/pages/vm/vm-wizard/vm-wizard.component.ts
```diff
@@ -874,9 +880,19 @@ async customSubmit(value) {
...
     }
-    
-    if(value.enable_vnc &&value.bootloader !== "UEFI_CSM"){
-      vm_payload["devices"].push({
+
+    if (value.enable_vnc) {
+      if (this.productType === 'SCALE') {
+        vm_payload["devices"].push({
+          "dtype": "VNC", "attributes": {
+            "vnc_port": String(this.getRndInteger(5900,65535)),
```
sonicaj, line 888: I just noticed, is this a custom function retrieve a random integer between this range ?
If yes, we should not be doing this and instead we should make a call to middleware which will return the next available vnc port `vm.vnc_port_wizard`.

src/app/pages/vm/vm-wizard/vm-wizard.component.ts
dkmullen: It's on line 903. The conditional on 898 runs if the system isn't SCALE, adds in the non-SCALE properties of 'wait' and 'resolution'

erinix: Oh I see now, sorry about that

PR #4420, opened 2020-07-21
NAS-106837 / 12.1 / Fix file download issue by doing map the new rxjs way

erinix:
NAS-106837


erinix:
note: this issue only seems to show up in the prod build itself, it does not have this problem when doing localhost build so the only way to test is with a hotpatch

PR #4432, opened 2020-07-23
NAS-106904 / 12.1 / NAS-106904 Update listing of boot options in VM edit form

dkmullen:
Fixes list of bootloader options on VM edit form. The VNC button issue mentioned in the ticket was addressed in https://github.com/freenas/webui/pull/4427


sonicaj:
@dkmullen does this address the VNC device issue as well where UI does not display VNC option a legacy bios based VM in scale even if the vm has a vnc device ?

sonicaj:
Ah got it, nevermind. Thanks

PR #4440, opened 2020-07-27
NAS-106919 / 12.1 / Use generic error handler for save debug errors

erinix:
NAS-106919


erinix:
Note it is actually difficult to get a real error message to generate from this, I had to hack the middleware to get it to show up for me

PR #4447, opened 2020-07-28
NAS-106824 / 12.1 / Only allow exporting dataset keys if a pool has a key encrypted dataset

erinix:
NAS-106824


dkmullen:
I noticed that, after a switched to a passphrase and the table repainted, I could still get to the 'Download Key' option and could download the empty key {} until I refreshed the page

src/app/pages/storage/volumes/volumes-list/volumes-list.component.ts
```diff
@@ -1731,6 +1731,8 @@ export class VolumesListComponent extends EntityTableComponent implements OnInit
       this.zfsPoolRows.pop();
     }
 
+    this.has_key_dataset = {};
+    this.has_encrypted_root = {};
     this.ws.call('pool.dataset.query_encrypted_roots_keys').subscribe(res => {
       for (const key in res) {
         if (res.hasOwnProperty(key)) {
```
dkmullen, line 1738: This ws.call, when it is made right after I save my password, returns the key I just got rid of, resulting in that {} being available (till page refresh). I noticed that setting a timeout delivers the right response. Must be a race condition somewhere?

PR #4448, opened 2020-07-29
NAS-106884 / 12.1 / Nas 102423

lola27:


src/app/helptext/sharing/nfs/nfs.ts
```diff
@@ -22,6 +22,10 @@ export const helptext_sharing_nfs = {
   ),
   validators_path: [Validators.required],
 
+  placeholder_alias: T("Alias"),
+  tooltip_alias: T(''),
```
Mrt134, line 26:
```suggestion
  tooltip_alias: T('To define a custom shortcut for the <i>Path</i>, \
   enter a memorable name. Example: <i>/nfsshare</i>. NFS clients \
   can connect using the <i>Alias</i> instead of the <i>Path</i> \
   when the share is active. A share with multiple paths requires \
   either not using aliases or defining an alias for each path in \
   the share.'),
```

src/app/helptext/sharing/nfs/nfs.ts
```diff
@@ -22,6 +22,10 @@ export const helptext_sharing_nfs = {
...
+  error_alias: T('Alias can be all empty or required for all pathes.'),
```
Mrt134, line 27:
```suggestion
  error_alias: T('The <i>Alias</i> field can either be left empty or \
   have an alias defined for each path in the share.'),
```

src/app/helptext/sharing/nfs/nfs.ts
```diff
@@ -22,6 +22,16 @@ export const helptext_sharing_nfs = {
...
   validators_path: [Validators.required],
 
+  placeholder_alias: T("Alias"),
+  tooltip_alias: T('To define a custom shortcut for the <i>Path</i>, \
+ enter a memorable name. Example: <i>/nfsshare</i>. NFS clients \
+ can connect using the <i>Alias</i> instead of the <i>Path</i> \
+ when the share is active. A share with multiple paths requires \
+ either not using aliases or defining an alias for each path in \
+ the share.'),
+  error_alias: T('The <i>Alias</i> field can either be left empty or \
```
erinix, line 32: the <i> </i> tags on the error do not translate to actual italics for me (are these errors displayed through innerHTML?)

erinix: screenshot-20200729-113211

PR #4449, opened 2020-07-29
NAS-106920 / 12.1 / Nas 106920

dkmullen:
Displays BOTH a dialog AND paragraph warning when user wants to create a NEW AFP share (but not when editing an existing one). We probably want one or the other - need guidance on which; dialog offers an easy way (via a button) to redirect to the SMB share creation form. Also need guidance on the message

Screenshot from 2020-07-28 18-41-53


william-gr:
Isn't it excessive to have both paragraph and dialog?

Curious on what other people might think.

Mrt134:
I think a confirm dialog is the right option so that users can quickly go to the recommended form if they decide to switch to SMB. Regardless, only a single message is needed; having both is too redundant. I want to reword the text a little though, as it feels a little aggressive towards the user. How about something like:
```
Beginning in 2013, Apple began using the SMB sharing protocol as the default option for file sharing and ceased development of the AFP sharing protocol. It is recommended to use SMB sharing over AFP unless files will be shared with legacy Apple products.
```

PR #4450, opened 2020-07-29
Nas 106868

dkmullen:
Uses only one menu item, 'Edit Permissions' which takes the user to Permissions editor if ACL is trivial, and to the ACL editor if not. Adds a button for switching from Permissions Editor to ACL and (if ACL is trivial) back again. When Strip ACL is selected, forces Recursive to be selected and disabled.


dkmullen:
Replacing with https://github.com/freenas/webui/pull/4451

PR #4451, opened 2020-07-30
NAS-106868 / 12.1 / Nas 106868 

dkmullen:
Uses only one menu item, 'Edit Permissions' which takes the user to Permissions editor if ACL is trivial, and to the ACL editor if not. Adds a button for switching from Permissions Editor to ACL and (if ACL is trivial) back again. Makes 'Strip ACL a button with its own submission method

### DIFF
--- a/src/app/helptext/storage/volumes/datasets/dataset-acl.ts
+++ b/src/app/helptext/storage/volumes/datasets/dataset-acl.ts
@@ -152,6 +152,18 @@ dataset_acl_stripacl_tooltip: T("Set to remove all ACLs from the current\
  dataset. ACLs are also recursively stripped from directories and child\
  datasets when those options are set."),
 
+stripACL_dialog: {
+  title: T('Strip ACLs'),
+  message: T('This action removes all ACLs from the current \
+ dataset. ACLs are also recursively stripped from directories and files of the current dataset. \
+ Stripping the ACL resets dataset permissions. This can make data inaccessible until new \
+ permissions are created.'),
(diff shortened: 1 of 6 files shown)


src/app/helptext/storage/volumes/datasets/dataset-acl.ts
```diff
@@ -152,6 +152,18 @@ dataset_acl_stripacl_tooltip: T("Set to remove all ACLs from the current\
...
 
+stripACL_dialog: {
+  title: T('Strip ACLs'),
+  message: T('This action removes all ACLs from the current \
+ dataset. ACLs are also recursively stripped from directories and files of the current dataset. \
+ Stripping the ACL resets dataset permissions. This can make data inaccessible until new \
+ permissions are created.'),
+  traverse_checkbox: T('Also strip permissions from child datasets of the current dataset'),
+  warning: T('Removes permissions recursively from all child \
+  datasets of the current dataset. This can make data inaccessible untin new permissions are created.')
```
Mrt134, line 163:
```suggestion
  message: T('This action removes all ACLs from the current \
 dataset and any directories or files contained within this \
 dataset. Stripping the ACL resets dataset permissions. This \
 can make data inaccessible until new permissions are created.'),
  traverse_checkbox: T('Remove the ACL and permissions from child datasets of the current dataset'),
  warning: T('Removes the ACL and permissions recursively \
 from all child datasets of the current dataset, including all directories and files contained within those child datasets. This can make data inaccessible until new permissions are created.')
```

PR #4481, opened 2020-08-06
NAS-107093 / 12.1 / Fix relogin after refresh on cached section detection, use window reload

erinix:
NAS-107093


erinix:
To test this, I did a hotpatch on the current build environment, and did git checkout HEAD@ on a commit that happened before the recent VM section commits from last weekand put these fixes in on top of it and hotpatched a previous build environment with that, the expected result is you will reboot from one boot environment to the other and click the vm section and will notice that it will refresh because it detected a missing chunk but then it will log you back in after and the ui should work as expected.

PR #4485, opened 2020-08-06
NAS-107098 / 12.1 / Listen to navigation events for cleanup to prevent memory leaks

erinix:
NAS-107098


erinix:
An observation, this scenario where the OnDestroy doesn't trigger seems to happen more in the production build than it does on a localhost dev build.

PR #4514, opened 2020-08-14
NAS-107228 / 13.0 / Account Form Populate SMB Checkbox, Disable for Builtins

DaAwesomeP:
NOTE: **This is untested.** I am working on that. I have verified however that `smb` is returned in `user.query` for builtins and other users.

https://jira.ixsystems.com/browse/NAS-107228

SCALE TrueNAS-13.0-MASTER-20200813-092913


DaAwesomeP:
Hmmm, ok. I cannot reproduce the error with the latest SCALE nightly in a VM. I think maybe I am missing a key part of how the forms populate or something like that. In my VM, the checkbox is properly populated and the root user has no issue saving changes.

DaAwesomeP:
What's really interesting is that the UI in the VM does seem to send `smb: false` when updating root, but the backend does not complain. In my physical setup, the backend does complain, which started this issue.

DaAwesomeP:
Updated to TrueNAS-SCALE-13.0-MASTER-20200816-172912, issue has gone away.

PR #4523, opened 2020-08-14
Make shark icon smaller

reesericci:
Already PR'd in other branches. @erinix


erinix:
11.1?  This branch is old, we no longer are producing versions earlier than 11.3

PR #4524, opened 2020-08-15
NAS-107121 / 12.0 / Nas 107121

erinix:


src/app/helptext/network/interfaces/interfaces-form.ts
```diff
@@ -112,6 +112,9 @@ failover_virtual_alias_address_tooltip: T('IP address to use for \
  administrative access to the system. The netmask <i>/32</i> is \
  required for this value and cannot be changed.'),
 
+failover_alias_set_error: T('Each alias must have an IP address for both\
+ controllers as well as the virtual interface.'),
```
Mrt134, line 116:
```suggestion
failover_alias_set_error: T('Each alias must have an IP address for each \
 controller and the virtual interface.'),
```

yocalebo: I don't like the suggested verbiage change because the word `Each` is used twice in the same sentence which makes it redundant.

I'd rather use verbiage like `An IP address must be provided for each controller as well as the virtual interface.`

Mrt134: That's fine, the redundancy was meant to reinforce the idea, but no biggie. Still would suggest replacing "as well as" with "and"

erinix: "An IP address must be provided for each controller and the virtual interface."?

yocalebo: @erinix Yep, agree.

Mrt134: Sure!

PR #4546, opened 2020-08-18
NAS-107128 / 11.3 / Add setTimeout to add/remove Vdev to avoid race conditions

erinix:
NAS-107128


erinix:
This was already fixed in master and 12.0, just needed to be backported to 11.3

PR #4589, opened 2020-08-27
NAS-107368 / 20.10 / Nas 107368

dkmullen:
Opens warning dialog when 'Apple style' is changed for the first time on an existing share

### DIFF
--- a/src/app/helptext/sharing/smb/smb.ts
+++ b/src/app/helptext/sharing/smb/smb.ts
@@ -239,7 +239,7 @@ export const helptext_sharing_smb = {
  This parameter alters how file names are read from and written to storage. Changes to \
  this paramater after data has already been written may impact ability to access or delete files containing \
  mangled characters.'),
-   action: T('Proceed with change')
+   action: T('I Understand')
  }
 
 };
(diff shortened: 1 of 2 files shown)


src/app/helptext/sharing/smb/smb.ts
```diff
@@ -231,6 +231,15 @@ export const helptext_sharing_smb = {
     message: T('An ACL is detected on the selected path but <i>Enable ACL</i> is not selected for this share. \
  ACLs must be stripped from the dataset prior to creating an SMB share.'),
     button: T('Close')
+ },
+
+ manglingDialog: {
+   title: T('Warning'),
+   message: T('The value has changed for the field <i>Use Apple-style character encoding</i>. \
+ This parameter alters how file names are read from and written to storage. Changes to \
+ this paramater after data has already been written may impact ability to access or delete files containing \
```
Mrt134, line 240:
```suggestion
   message: T('The <i>Use Apple-style character encoding</i> value has changed. \
 This parameter affects how file names are read from and written to storage. Changes to \
 this parameter after data is written can prevent accessing or deleting files containing \
```

PR #4593, opened 2020-08-28
NAS-107064 / 20.10 / Nas 107064

dkmullen:
Implements slide-in component and three instances of its use: Group Add/Edit form, VM Add wizard and User Add/Edit (2 columns)


dkmullen:
Annotation 2020-08-28 130208a
Annotation 2020-08-28 130208
Annotation 2020-08-28 130208b

src/app/components/common/modal/modal.component.css
```diff
@@ -0,0 +1,110 @@
...
+}
+
+.jw-modal.open {
+    z-index: 1000;
+}
+
+.slide-in-form {
+    width: 400px;
+    position: fixed;
+    top: 50px;
```
lola27, line 26: this one should be `top: 48px` since the height of top menu bar is 48px. otherwise, there will be a white line between it and the side-form (also the black mask, and second level side menu) in light theme.

dkmullen:
Wrong ticket number in the last commit message

PR #4601, opened 2020-08-31
NAS-107401 / 20.10 / NAS-107401 disable autocomplete for one-time pw

dkmullen:


src/app/views/sessions/signin/signin.component.html
```diff
@@ -57,6 +57,7 @@
                       [(ngModel)]="signinData.otp"
                       #otp="ngModel"
                       required
+                      autocomplete="off"
```
btecu, line 60: Chrome ignores `off` value for `autocomplete`. It might be better to set it to something else, `nein`, `rejected`, ... anything else would work, except `off`.

https://stackoverflow.com/questions/12374442

erinix: We have seen that problem with some fields like username and password (which are more aggressivly hit by browsers with these auto complete things) but in our testing it does work for this field

PR #4614, opened 2020-09-03
NAS-107400 / 20.10 / Add generic password confirm dialog for exporting dataset keys

erinix:
NAS-107400

### DIFF
--- a/src/app/helptext/global-helptext.ts
+++ b/src/app/helptext/global-helptext.ts
@@ -76,7 +76,7 @@ rootpw: {
   tooltip: T('Enter the root password to authorize this operation.'),
   error_msg: T('The administrator password is incorrect.'),
   error_title: T('Warning'),
-  dialog_title: T('Action Requires Credential Confirmation')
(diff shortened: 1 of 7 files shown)


src/app/helptext/global-helptext.ts
```diff
@@ -75,7 +75,8 @@ rootpw: {
   placeholder: T('Root Password'),
   tooltip: T('Enter the root password to authorize this operation.'),
   error_msg: T('The administrator password is incorrect.'),
-  error_title: T('Warning')
+  error_title: T('Warning'),
+  dialog_title: T('Action Requires Credential Confirmation')
```
Mrt134, line 79:
```suggestion
  dialog_title: T('Action Requires Administrator Confirmation')
```

Mrt134 (approved):
Looks fine, just a section to replace "Credential" to "Administrator", as that is a little more specific and is used in other dialogs.

dkmullen (approved):

PR #4652, opened 2020-09-15
NAS-107583 / 20.10 / Use relative paths for fonts

erinix:
NAS-107583


dkmullen (approved):
lgtm - tested with hotpatch

PR #4657, opened 2020-09-16
NAS-107393 / 20.10 / NAS-107393-glob

dszidi:
Move CSS from General-Settings to Global file and layouts to template


dkmullen (commented):
Everything seems to be working. I like the looks of the General Preferences Dash overall. It does seem like the view shifts to one column too soon, leaving the GUI and Localization cards very wide. Will it be possible to set such things at a granular level with component-specific rules?

image

erinix (approved):
lgtm

dszidi:
> Everything seems to be working. I like the looks of the General Preferences Dash overall. It does seem like the view shifts to one column too soon, leaving the GUI and Localization cards very wide. Will it be possible to set such things at a granular level with component-specific rules?
> 
> image

Yes, layouts should be done on the component level via FlexLayoutModule. Styling should be accomplished by using provided CSS classes from the Global CSS files

dszidi:
> > Everything seems to be working. I like the looks of the General Preferences Dash overall. It does seem like the view shifts to one column too soon, leaving the GUI and Localization cards very wide. Will it be possible to set such things at a granular level with component-specific rules?
> > image
> 
> Yes, layouts should be done on the component level via FlexLayoutModule. Styling should be accomplished by using provided CSS classes from the Global CSS files

I'll also add that those breakpoints and layouts were chosen because the alternative was to put those two side by side. One of them has an ipv6 address that is currently empty which might be a problem in the side by side scenario. This is why the single column was enforced earlier. This can easily changed later via the template + FlexLayoutModule's attributes

PR #4678, opened 2020-09-21
NAS-107655 / 20.10 / NAS-107655

dszidi:
CPU widget should degrade to Unknown if model is null. WidgetSysInfo now has a case for Scale systems in product detection

### DIFF
--- a/src/app/core/components/widgets/widgetsysinfo/widgetsysinfo.component.ts
+++ b/src/app/core/components/widgets/widgetsysinfo/widgetsysinfo.component.ts
@@ -188,7 +188,7 @@ export class WidgetSysInfoComponent extends WidgetComponent implements OnInit,On
         this.systemLogo = 'logo.svg';
         this.getFreeNASImage(evt.data.system_product);
         this.isFN = true;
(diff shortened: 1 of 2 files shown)


src/app/core/components/widgets/widgetsysinfo/widgetsysinfo.component.ts
```diff
@@ -184,15 +184,19 @@ export class WidgetSysInfoComponent extends WidgetComponent implements OnInit,On
       } else {
         this.manufacturer = "other";
       }
-      if (this.product_type === 'CORE') {
+      if (this.product_type === 'CORE'){ 
         this.systemLogo = 'logo.svg';
         this.getFreeNASImage(evt.data.system_product);
         this.isFN = true;
-      } else {
+      } else if(this.product_type == 'Enterprise' && evt.data.license !== null) {
```
erinix, line 191: should be .includes('ENTERPRISE')

erinix (changes requested):

erinix (approved):

PR #4699, opened 2020-09-25
NAS-107534 / 20.10 / Nas 107534a

dkmullen:
Backup Credentials Dashboard

Wide:
Screenshot 2020-09-25 081736a

Medium:
Screenshot 2020-09-25 082310d

Med with no menu:
Screenshot 2020-09-25 082337e

Narrow:
Screenshot 2020-09-25 082402f

SSH Connections
Screenshot 2020-09-25 083006

SSH Keypairs with actions menu open at bottom
Screenshot 2020-09-25 082238c

Cloud Credentials:
Screenshot 2020-09-25 081920b


Mrt134 (approved):

william-gr:
How do you scan for the remote host key when addint a new ssh connection?

dkmullen:
XD 

https://xd.adobe.com/view/518c3a07-1e8c-4003-8f3c-adfd198e1e47-04df/?fullscreen
--

dkmullen:
Screenshots of current UI...
Screenshot from 2020-09-28 12-40-13
Screenshot from 2020-09-28 12-40-35
Screenshot from 2020-09-28 12-40-47

william-gr:
@themylogin since you're the one responsible for most of these (if not all), do you have any suggestions/ideas/thoughts?

william-gr:
For the Cloud Credentials Add screen, I am thinking Verify Credential should be the blue button since most of the times thats what we want people to do to make sure its working and after its validated turn it into a Save. Thoughts?

themylogin:
I don't like important actions hidden under the menu. Agree with William: "Verify Credentials" should go before "Save" (but should be of less important button color IMO). Also "Generate keypair" can go below private key field and "Discover remote host key" below "Remote host key" field, like "Add" button here https://i.stack.imgur.com/huyq1.png

dkmullen:
Dash layout: Can download both public/private keys with one click on the icon for SSH Keypairs
image

'Log in to provider' now appears in context rather than at the bottom, verify credentials now appears before save - both have the standard colors and are disabled if the form is invalid. It IS still possible to Save without validating due to the case, for example, when validation fails because system clock is off.
image

'Discover Remote Host Key' now appears in context rather than at the bottom
image

'Generate Key Pair' now appears in context under Private Key, and the download choices appear in the 'More' menu (downloading can also happen from the Dashboard view) and each is disabled until the form has a name and a value for the key being downloaded.
image

themylogin (approved):
LGTM!

william-gr:
Generate key pair button seems a bit strange between private and public key. Does anyone agree or is it just my impression?

themylogin:
Can we put it above private key field and make it `width: 100%`?

dkmullen:
Here's the Generate Keypairs button moved and 100% width, and for consistency, all the in-form buttons at full column width to match the form elements:
image
image
image

william-gr (approved):

william-gr:
Looking good to me.

@dszidi can you also please review this for CSS consistency to make sure we are implementing things generic in the design system and not per component?

kmoore134 (approved):
This looks quite nice. Good job @dkmullen !

dszidi (approved):
CSS looks fine. Nice work

PR #4709, opened 2020-09-28
NAS-107741 / 20.10 / Handle alerts better in alert settings

erinix:
NAS-107741


erinix:
to test, SCALE currently traces back but leaves you hanging, 12.1 should continue to work as expected

dkmullen (approved):
looks good - handles error on SCALE before mw patch;

erinix:
backport

PR #4760, opened 2020-10-12
NAS-107876 / 11.3 / Nas 107876 v11.3

dkmullen:
When geli-locked pool fails to decrypt, the lock turns red and a message appears, and the lock icon that opens the unlock dialog is not shown. Create a test case with:
midclt call datastore.update storage.volume 5 '{"vol_encrypt": 1}' 
(5 being the vol number)

The yellow lock is for a pw-protected pool, and has the icon on the right for unlocking. The red lock is a geli-encrypted pool that failed to decrypt.
image


lola27 (approved):
LGTM

PR #4761, opened 2020-10-12
NAS-107876 / 12.0 / Nas 107876 12

dkmullen:
When geli-locked pool fails to decrypt, the lock turns red and a message appears.
 Create a test case with:
midclt call datastore.update storage.volume 5 '{"vol_encrypt": 1}'
(5 being the vol number)

The yellow lock is for a pw-protected pool. The red lock is a geli-encrypted pool that failed to decrypt.

Screenshot from 2020-10-12 12-32-23


lola27 (approved):
LGTM

PR #4767, opened 2020-10-14
NAS-107889 / 12.0 / Nas 107889

dkmullen:
Prevent row resize and auto sort for entity table when handleData is called outside of page load


erinix (approved):

erinix:
backport

PR #4768, opened 2020-10-15
NAS-107951 / 12.0 / NAS-107951 Make room for all menu items

dkmullen:
Allows room for all 22 possible menu items under System, although 21 is probably the max need under current restrictions. Seems to have no ill effect on shorter lists. Should not be an issue with the new menu on SCALE

ALSO: Removes 2FA from the menu if the product type is enterprise


lola27 (approved):
LGTM

PR #4775, opened 2020-10-19
NAS-107988 / 20.10 / NAS-107988 reset users/groups when table refreshes

dkmullen:
Resets user/group lists when when the slide-in menu closes and refreshes the table. This isn't an issue for previous versions because the form navigated to the table causing a complete page reload.


lola27 (approved):
LGTM

PR #4778, opened 2020-10-21
NAS-107706 / 20.12 / NAS-107706

dszidi:
- Add page title component
- Add global actions
- Storage/Pools is now just Storage
- Reporting uses pagetitle and global actions
- Entity-table now uses pagetitle and global actions


dkmullen (changes requested):
Looks like the row references for 'Create Snapshot' in volumes-list.component need to be changed from row.id to rowData.id for snapshots to work.

dkmullen (approved):
lgtm

PR #4786, opened 2020-10-26
NAS-107864 / 20.12 / Nas 107864 scale

dkmullen:
Adds gMail oauth as a way to authenticate from the email component. User can switch to smtp and back. Depends on mw PR https://github.com/freenas/freenas/pull/5919 merged Oct 26


dkmullen:
Will backport the 12.0 PR

PR #4795, opened 2020-10-28
NAS-105082 / 20.12 / NAS-105082

dszidi:
Dashboard global configuration (Widget visibility)


dkmullen (changes requested):
There seems to be a bug in the displaying of widgets. Once I added a pool, I couldn't reliably hide it, although it would disappear when I tried to hide another item, as if the show/hide function was working off an old list of items.

Also, if you want to make the slide-in display as one column, just a simple property addition...see inline comments

dkmullen (commented):
nice!

dkmullen (approved):

PR #4797, opened 2020-10-29
NAS-107751 / 20.12 / Nas 107751a

dkmullen:
Must be merged after https://github.com/freenas/webui/pull/4795
Moves config actions in General Settings dashboard into the new title area (created in the above PR)

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@angular/core": "10.0.3",
     "@angular/flex-layout": "10.0.0-beta.32",
     "@angular/forms": "10.0.3",
-    "@angular/material": "10.0.1",
+    "@angular/material": "10.2.4",
     "@angular/platform-browser": "10.0.3",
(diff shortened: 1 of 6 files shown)


Mrt134 (approved):
Looks good!

dszidi (commented):
The actual code looks fine. I'm wondering if we can come up with a better button label than "Config Files". Maybe "Configure" might be more appropriate? Maybe Tim can chime in?

Mrt134:
The button label "Config Files" does seem a little ambiguous for a label that has multiple actions underneath. What about changing to be a little more descriptive, something like "Manage Config", "Manage System Configuration", "Config File Actions", "Configuration Management"?

dszidi (approved):
I like the suggestions. This probably will require some consideration and I don't want to hold up this PR. I will make a new ticket and include these comments 

LGTM

PR #4805, opened 2020-11-03
NAS-107948 / 20.12 / NAS-107948 Add sys dataset warnings

dkmullen:
Could use more from docs team on a specific warning of any dangers of detaching the sys dataset pool.

### DIFF
--- a/src/app/helptext/storage/volumes/volume-list.ts
+++ b/src/app/helptext/storage/volumes/volume-list.ts
@@ -46,7 +46,11 @@ exportError: T("Error exporting/disconnecting pool."),
 
 exportDialog:  {
     title: T("Export/disconnect pool: '"),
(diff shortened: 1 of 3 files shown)


src/app/helptext/storage/volumes/volume-list.ts
```diff
@@ -46,6 +46,7 @@ exportError: T("Error exporting/disconnecting pool."),
 
 exportDialog:  {
     title: T("Export/disconnect pool: '"),
+    warningSysDataset: T('This pool contains the SYSTEM DATASET.'),
```
Mrt134, line 49: Not sure if additional information is being provided in other strings that are added in, but here's a mostly unformatted suggestion that provides more detail to the warning:
```suggestion
    warningSysDataset: T('This pool contains the system dataset that stores critical data like debugging core files, encryption keys for pools, and Samba 4 metadata such as the user/group cache and share level permissions. Exporting this pool will transfer the system dataset to another available pool. If the only available pool is encrypted, that pool will no longer be able to be locked. When no other pools exist, the system dataset transfers back to the TrueNAS operating system device.'),
```

erinix (approved):

PR #4806, opened 2020-11-03
NAS-108184 / 12.0 / NAS-108184

dszidi:
Drive tray handle color on ES102 is almost the exact same color as the red we use for FAULT status. For this reason the handle asset was converted to greyscale to avoid confusion. It also works better with the PIXI.Sprite.tint method.


dkmullen (approved):
lgtm

PR #4808, opened 2020-11-04
NAS-108211 / 12.0 / NAS-108211

dszidi:
Fixes too many concurrent calls error on enclosure UI. This was caused by the DiskTemperatures subscription living in the base class. This is a feature only on the MINI line but the service was being activated on the rack mount models too.

The rear drive bays on the M50 were referencing the drives that belong to the front. Rear drive bay detection has been reworked and rear drives now correctly reference the disks that belong to that chassis.


dkmullen (approved):
Tested on tn14 with help spotting and inserting drives from Jeff E. & Damian

PR #4809, opened 2020-11-04
NAS-108212 / 12.0 / NAS-108212

dszidi:
Use the more flexible convertToRGB method. It checks the input value type and will not throw an error when given rgba values.


dkmullen (approved):
nice!

PR #4811, opened 2020-11-05
NAS-108210 / 20.12 / NAS-108210 Remove unneeded first step for Import Pool wizard for SCALE

AlexTimDev:


AlexTimDev:
Removed unneeded first step for Import Pool wizard for SCALE.

src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
```diff
@@ -251,27 +252,19 @@ export class VolumeImportWizardComponent {
...
-      this.importIndex = 1;
+      this.wizardConfig.splice(0,2);
+      this.importIndex = 0;
+      this.getImportableDisks();
     }
   }
 
   afterInit(entityWizard: EntityWizardComponent) {
-    const createPoolText = T("Create Pool")
+    const nextText = T("Next ")
```
dkmullen, line 262: Good thinking here; however, the entity wizard (our pattern for making wizards) will default to 'Next', so you can just remove this line (#262) and also #264, which sets the customNext text.  And one more thing: Our boss doesn't like us to leave console logs, so please remove #267 as well.

dkmullen (changes requested):

dkmullen (approved):
LGTM!

PR #4813, opened 2020-11-05
NAS-108124 / 20.12 / NAS-108124 add validation to naming_schema of snapshot-form to exclud…

AlexTimDev:
Added validation to name element of snapshot-form to exclude / character.


dkmullen (changes requested):
To make the error message show, find form-errors.component.html and extend the rule that applies to 'regex' so that is also applies to pattern (ie, use an or condition). 'Regex' is our custom validator. Then, add the backport tag for v. 12

dkmullen (approved):
lgtm

PR #4837, opened 2020-11-12
Update fr.po

MDVAurelien:
Add some translations


erinix (approved):

erinix:
there won't be too many 11.3 releases in the near future, I would suggest focusing more on 12.0 and master at this point :)

MDVAurelien:
Yes absolutely this is my last commit in this branch ;-)

PR #4878, opened 2020-11-18
NAS-108260 / Added confirmation dialog for plugin update. Using 'plug…

ryousaf:
'plugin.update' instead of 'jail.update_to_latest_patch'
added confirmation dialog for jail update as well


erinix (approved):
lgtm

PR #4900, opened 2020-11-24
NAS-108361 / 12.0 / Change icon color for unavail case on pool card / NAS-108361

AlexKarpov98:
Change icon color for unavail case on pool card / NAS-108361


dszidi (approved):
LGTM

PR #4903, opened 2020-11-24
NAS-108348 / 20.12 / Nas 108348

ryousaf:
Added slide in forms instead of new routes for create/edit datasets in pools

NAS-108348


erinix:
this works a lot better but one remaining issue, the pools page does not update after you create a new dataset

erinix (approved):
lgtm

PR #4908, opened 2020-11-27
NAS-108417 / 20.12 / Added Change Password as a dialog instead of a new page

ryousaf:
NAS-108417

### DIFF
--- a/src/app/components/common/topbar/topbar.component.ts
+++ b/src/app/components/common/topbar/topbar.component.ts
@@ -28,6 +28,7 @@ import { FlexLayoutModule, MediaObserver } from '@angular/flex-layout';
 import { DialogFormConfiguration } from '../../../pages/common/entity/entity-dialog/dialog-form-configuration.interface';
 import { TruecommandComponent } from '../dialog/truecommand/truecommand.component';
 import { ResilverProgressDialogComponent } from '../dialog/resilver-progress/resilver-progress.component';
+import { matchOtherValidator } from 'app/pages/common/entity/entity-form/validators/password-validation';
 
 @Component({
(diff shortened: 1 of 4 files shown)


src/app/helptext/topbar.ts
```diff
@@ -93,6 +95,18 @@ Login or signup is required.'),
         title: T('Verify Email Address'),
         message: T('A message with verification instructions has been sent to the new email \
 address. Please verify the email address before continuing.')
+    },
+    changePasswordDialog: {
+        pw_form_title_name: T('Change Administrator Password'),
+        pw_current_pw_placeholder : T('Current Password'),
+        pw_new_pw_placeholder : T('New Password'),
+        pw_new_pw_tooltip : T('Passwords cannot contain a <b>?</b>. Passwords should\
+         be at least eight characters and contain a mix of lower and\
```
erinix, line 104: there should only be one space at the beginning of this text on each line, the translation extractor will pull out multiple spaces in this case

src/app/components/common/topbar/topbar.component.ts
```diff
@@ -793,4 +794,60 @@ export class TopbarComponent extends ViewControllerComponent implements OnInit,
...
+          inputType : 'password',
+          required: true,
+          validation : [ matchOtherValidator('password') ]
+        }
+      ],
+      saveButtonText: T('Save'),
+      custActions: [],
+      parent: this,
+      customSubmit: (entityDialog) => {
+        const pwChange = entityDialog.formValue;
```
erinix, line 832: you should add a loader in here for when it is submitting

dkmullen (approved):
Nice job.

PR #4911, opened 2020-11-30
NAS-107697 / 12.0 / WebUI makes firefox 80.0.1 hang

AlexKarpov98:
WebUI makes firefox 80.0.1 hang

### DIFF
--- a/src/app/core/components/viewchartgauge/viewchartgauge.component.ts
+++ b/src/app/core/components/viewchartgauge/viewchartgauge.component.ts
@@ -56,7 +56,7 @@ export class ViewChartGaugeComponent /*extends DisplayObject*/ implements AfterV
 
   @Input() config: GaugeConfig;
 
-  constructor() { 
+  constructor() {
   }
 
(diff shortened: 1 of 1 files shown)


dszidi (commented):
`if (document.hasFocus()) {...}` will stop the chart from updating when the user is browsing in another window. Users will want to have the charts functioning while browsing other pages.

AlexKarpov98:
> `if (document.hasFocus()) {...}` will stop the chart from updating when the user is browsing in another window. Users will want to have the charts functioning while browsing other pages.

Sure, all data is being updated, just unnecessary drawing is removed during the site is fully invisible by other tabs or by other software

AlexKarpov98:
> `if (document.hasFocus()) {...}` will stop the chart from updating when the user is browsing in another window. Users will want to have the charts functioning while browsing other pages.

Seems you are confused with middleware works.
It is frontend so another tab as well as another window will see different content according to the user action.
In a word they are all independent.

erinix (approved):

PR #4913, opened 2020-11-30
NAS-108353 / 12.0 / Moved estimateSize() method to the end of ngOnInit so proper values a…

ryousaf:
…re set before capacity estimation is performed

NAS-108353


erinix (approved):

erinix:
backport

PR #4918, opened 2020-12-02
NAS-107989 / 20.12 / NAS-107989

dszidi:
Convert Entity-Table to mat-table
Make style consistent with other tables


dkmullen (approved):
Great job!

PR #4921, opened 2020-12-02
NAS-108499 / 20.12 / NAS-108499

dszidi:
Removed red border around cron picker.
Added a close button on top left of popup to keep things consistent


dkmullen (approved):
lgtm

PR #4930, opened 2020-12-03
NAS-108324 / 21.02 / Convert Certificate and CA forms to wizards

AlexKarpov98:
Convert Certificate and CA forms to wizards


dkmullen (changes requested):
Sorry, this task wasn't described very well in the ticket. We need the following:

- When 'Import Certificate/CA' is selected, unneeded steps should be hidden
- The four fields at the bottom of the current forms (Basic Constraints, Key Usage, etc) can be combined into one step
- The last step should probably be a summary of what the form contains. See the VM wizard for one example. We might want to discuss with the team how much should be included in that summary.
- Add `private isLinear = true;` which keeps the user from advancing without filling out required fields 

Nice job on the methods for adding and updating data.

dszidi (approved):
LGTM

PR #4934, opened 2020-12-04
NAS-108500 / 12.0 / fix pCloud hostname is not sent when verifying credentials

AlexKarpov98:
pCloud hostname is not sent when verifying credentials


dszidi (approved):
LGTM

PR #4936, opened 2020-12-07
NAS-100629 / 21.02 / Nas 100629

AlexKarpov98:
The text in the Shell appears to wrap prematurely even on a sufficiently wide display/screen


dszidi:
Row text does not seem to be wrapping at all in Chrome

image

dszidi (approved):
EDIT: After checking changes against the other themes, the font colors don't work on lighter themes

PR #4942, opened 2020-12-08
NAS-108530 / 12.0 / NAS-108530

dszidi:
Fix tooltip on system>support


ryousaf (commented):
We should add tooltip position override for the "Attach Debug" field as well in FnSupportComponent/TnSupportComponent

ryousaf (commented):
We also need to do the same in FnSupportComponent

dszidi:
Replaced by new PR https://github.com/freenas/webui/pull/4954

PR #4958, opened 2020-12-11
NAS-108449 / 12.0 / fix CANCEL/CLOSE Buttons for iX Dark Theme need addressed

AlexKarpov98:
fix CANCEL/CLOSE Buttons for iX Dark Theme need addressed


dszidi (commented):
Since we are using a css custom property for the background, let's also use the corresponding property for the text color. Please use `color: var(--accent-txt) !important`

dszidi (approved):
LGTM

PR #4963, opened 2020-12-14
NAS-108628 / 12.0 / Keep yes/no enabled column for consistency

denysbutenko:
Remove the piece of code that replaces "no" to "false" in the CRON table component


dszidi (approved):
LGTM

PR #4971, opened 2020-12-16
NAS-108660 / 12.0 / make error text pre-formatted

denysbutenko:


src/app/pages/common/info-dialog/info-dialog.component.scss
```diff
@@ -1,5 +1,12 @@
 .info-panel {
     max-height: 115px !important;
```
dszidi, line 2: Set the max-height to `80vh !important`

src/app/pages/common/info-dialog/info-dialog.component.scss
```diff
@@ -1,5 +1,12 @@
...
+
+    span {
+      display: block;
+      unicode-bidi: embed;
+      font-family: monospace;
+      white-space: pre;
```
dszidi, line 8: Set the white-space to normal

dszidi (commented):
Couple of suggestions to avoid unnecessary scroll bars.

src/app/pages/common/info-dialog/info-dialog.component.scss
denysbutenko: The parent block `.mat-dialog-content` has `max-height: 65vh`, it may be an issue whenever the output is very long.

src/app/pages/common/info-dialog/info-dialog.component.scss
denysbutenko: This was changed intentionally to preserve white spaces in a string. Let me show a difference

dszidi (approved):
LGTM

PR #4973, opened 2020-12-17
NAS-108481 / 12.0 / zvol zfs encryption settings missing from UI

AlexKarpov98:


dszidi (commented):
The "Encryption Options" seem to be missing from the table row actions and there is no lock icon

NVM was a live reload problem. Works after recompile.

dszidi (approved):
LGTM

PR #4974, opened 2020-12-17
NAS-107574 / 21.02 / NAS-107574

ryousaf:


dszidi (changes requested):
The zvol field should be populated with the newly create volume once the wizard is submitted. Currently it only shows "New"

src/app/pages/storage/volumes/volumes-list/volumes-list.component.ts
```diff
@@ -1064,6 +1064,7 @@ export class VolumesListTableConfig implements InputTableConf {
           name: T('Add Zvol'),
           label: T("Add Zvol"),
           onClick: (row1) => {
+            console.log('addZvol', rowData.id);
```
william-gr, line 1067: Leftover debug?

william-gr (commented):
@ryousaf would you care to post a few screenshots so we can see what it looks like?

src/app/pages/storage/volumes/volumes-list/volumes-list.component.ts
ryousaf: Yes. After the review by Damian, I was making further changes. Will clean it up.

ryousaf:
> @ryousaf would you care to post a few screenshots so we can see what it looks like?

Here are some screenshots as you go through the process of adding a new Zvol while adding a new disk.

Screenshot 2020-12-23 201318
Screenshot 2020-12-23 201337
Screenshot 2020-12-23 201347
Screenshot 2020-12-23 201402
Screenshot 2020-12-23 201438
Screenshot 2020-12-23 201451
Screenshot 2020-12-23 201500
Screenshot 2020-12-23 201511
Screenshot 2020-12-23 201522
Screenshot 2020-12-23 201719
Screenshot 2020-12-23 201740
Screenshot 2020-12-23 201752

william-gr:
Thanks!

A few notes:
- The padding is looking very odd for the Device Add Form.
- Zvol Directory isn't really a thing. I wonder if we have a selector that let us choose a ZFS Filesystem, no path. Just `tank/foo`, instead of `/mnt/tank/foo`
- The Add New option, how hard is it to keep it sticky at the end of the list? and the let scroll work over the existing items?

dszidi:
Adding the class `mat-card-padded` to `<mat-card>` in the device-add.component.html file will fix your padding issue

dszidi (approved):
LGTM

PR #4982, opened 2020-12-18
NAS-108695 / 21.02 / Hide static IP fields when DHCP is set in launch docker image form

AlexKarpov98:


denysbutenko (approved):
Looks good to me

PR #4984, opened 2020-12-21
NAS-108680 / 21.02 / Hide ix-applications from host path selector

AlexKarpov98:


william-gr (approved):

Ornias1993:
Might want to add the same for iocage, in case people migrated from CORE.
Might cause future issues if people go select those paths in the SCALE in the UI and try to go back.

PR #4985, opened 2020-12-21
NAS-108725 / 21.02 / Added conditional sticky property for the first column in the table

ryousaf:
NAS-108725


dszidi (commented):
This doesn't seem to take into account multiselect where the checkbox is the first column. 
When the table implements multiselect checkboxes, sticky column should be the first two.

dszidi:
We should also implement this approach with entity-tree-table so the column with the tree is always sticky

dszidi (commented):
One last thing. Let's put a border on the last sticky column on the left and the actions column on the right. (td + th)

image


PS. Don't forget to put `border-collapse: separate;` on the table element
PPS. Add a little padding to the actions column. The icon should have equal space on the right and left sides

dszidi (approved):
LGTM

PR #4997, opened 2020-12-25
NAS-108682 / 12.0 / NAS-108682

ryousaf:
NAS-108682


dszidi (commented):
We don't need so many different versions of the same svg. All our icons are imported into the mat-icon registry so we can simply use css to change the color to whatever we want.

Please only import a single svg and make the adjustments to the css color rule like you would any other mat-icon instance

dszidi:
This was fixed in another PR

https://github.com/freenas/webui/pull/5014

PR #5013, opened 2020-12-31
Fix undefined global error

dszidi:


dszidi:
Already fixed in another PR

PR #5016, opened 2021-01-05
NAS-108853 / 21.02 / NAS-108853

dszidi:
Fix hover styles on mat-tables


denysbutenko (commented):
Just one suggestion that comes to my mind is to move the `onHover` and `findRow` functions to a single file and use it from there.

denysbutenko (approved):
LGTM

PR #5020, opened 2021-01-07
NAS-108688 / 21.02 / Allow access to pod console from UI / NAS-108688

AlexKarpov98:


dszidi:
Let's use the same icon in the mat-menu-item that we use for the shell in Core. It's console-line from the mdi set.

image

sonicaj (commented):
If I set `/bin/bash` as command for minio container, UI does not show the error and user is just presented with an empty screen. It's confusing and we should show the error we get from middleware as otherwise user would probably think that he's going to be presented with console of a container whereas in reality the command has errored out.

AlexKarpov98:
Fixed

sonicaj (commented):
I just tried it and it's nice, however we are not refreshing the shell screen on Reconnect, so that can be confusing and we should do that.
Screenshot 2021-01-12 at 10 23 45 PM

AlexKarpov98:
Good point - fixed

sonicaj (commented):
We are not refreshing screen after switching pods/containers, please see attached screenshot. I was testing nextcloud app for this which you can use to reproduce. About the erroring out bit, it's working nicely.
Screenshot 2021-01-14 at 12 48 29 AM

AlexKarpov98:
@sonicaj 
It also should be fixed and now I can't see that issue.
Can you please check once again?
Thank you.

sonicaj (approved):
Thanks

kmoore134:
retest this please

kmoore134 (approved):
Ran through it here, looks great :)

william-gr:
@freenas/ui-team This has been sitting for review for 9 days, can someone in the UI team review to get t his merged? Thanks!

dszidi:
There are some conflicts since dynamic forms were merged in.

kmoore134:
Ran through the container build again just now, seems to work well if anybody is wanting to merge this.

dszidi (approved):
LGTM

william-gr:
@dszidi please let developers merge their own PRs.

PR #5035, opened 2021-01-11
NAS-108855 / 21.02 / Re-initialized dropdown options before populating them with appropria…

ryousaf:
…te data

NAS-108855


dszidi (approved):
LGTM

PR #5036, opened 2021-01-12
NAS-108894 / 21.02 / Added option to add alert with disable alert capability

ryousaf:
…values selected

NAS-108894


dszidi:
Currently this PR creates a validation error. This is not correct since it is a valid choice and does not invalidate the form. 
Let's do a different approach.

- Let's put an alert icon suffix after each high level log option in the dropdown
- When the user selects one of the options with the icon, it will trigger a warning dialog
- To avoid annoying the user with constant dialogs, let's also implement a "Don't show me this again" checkbox in the dialog

dszidi:
We don't need the "Disable all alerts" checkbox

image

dszidi (approved):
LGTM

PR #5039, opened 2021-01-12
NAS-108942 / 21.02 / SCALE UI pages give 'reconnect' button after using the shell

AlexKarpov98:
This commit will fix both of NAS-108942 and NAS-108943


dszidi (changes requested):
We should unregister from the CoreService in the ngOnDestroy life cycle hook.
`this.core.unregister({observerClass: this});`

dszidi (approved):
LGTM

PR #5044, opened 2021-01-12
NAS-108952 / 21.02 / Fixed Pools table default columns, EntityTreeTable sticky columns

ryousaf:
NAS-108952


dszidi:
Looks good. Could we get a little more padding when the horizontal scroll bar kicks in? The cells look awfully cramped.

image

dszidi:
There is also an issue with hover style not being applied to actions column

image

dszidi (approved):
LGTM

PR #5046, opened 2021-01-13
Revert "NAS-108690 / 21.02 / Dynamically create chart release form in UI / Nas 108690"

dszidi:
Reverts freenas/webui#4996


Ornias1993:
Mind explaining, instead of just reverting without any explaination?

PR #5048, opened 2021-01-14
NAS-108950 / 21.02 / Added searchable and reset icons on the left and right for form-input and from-combobox

ryousaf:


dszidi:
Looks like password fields suffix is conflicting with the new close icon. Perhaps the changes created in this PR should not apply to password fields.

image

dszidi (changes requested):
The combobox input field is actually an autofill so does not require the search icon

dszidi (approved):

PR #5049, opened 2021-01-14
NAS-108968 / 21.02 / Update services page

denysbutenko:
- Use entity-table on the Services page
- Update entity-table to show interactive controls: mat-slide-toggle and mat-checkbox
- Update vm-list component to use new controls
- Cleanup files


src/app/pages/services/services.component.ts
```diff
@@ -51,81 +59,72 @@ export class Services implements OnInit {
...
+      })
     }
+    return actions;
   }
-
-  displayAll() {
-    this.cards = this.cache;
+  
+  onSliderChange(service) {
+    console.log('onSliderChange', service);
```
dszidi, line 120: Remove Console logs

dszidi:
A couple of suggestions

- Having the config action hidden in the three dot menu when there is only one item to list is bad.
- The table does not have enough columns to justify the Columns multi-select button on the top right. IMO this should be suppressed. For this table, all columns should be shown by default without option to hide

image

image

denysbutenko:
Good point! Added changes according to your feedback

dszidi:
Getting console errors...

image

denysbutenko:
Good catch! It occurs when a column doesn't have `always_display` property at the beginning and a user clicks on toggle column visibility then the column will be saved to the database as a preferred column to display (`tableDisplayedColumns`) then add `always_display: true` property for this column and on this step happened conflict, one column comes from the component – `alwaysDisplayedColumns` and the second one comes from the database – `tableDisplayedColumns`.
The fix would be ignoring column from the database if it is present in component with `always_display: true`

dszidi (approved):
LGTM

PR #5057, opened 2021-01-18
NAS-108974 / 21.02 / Show SCALE logo for certified devices

denysbutenko:


dszidi (changes requested):
FreeNAS certified systems have their own svg logo icon that we use. Please undo this change and you can add the following line after line 54.

`<mat-icon *ngIf="(product_type === 'SCALE' || product_type == 'SCALE_ENTERPRISE') && certified && retroLogo == 0 && !isHA" class="generic" svgIcon="freenas_certified"></mat-icon>`

The problem seems to be it was assumed that FreeNAS Certified systems would not be running Scale. The line I provided fixes that.

For reference, below you can see what it should look like in the browser
image

dszidi (approved):
LGTM

PR #5061, opened 2021-01-19
NAS-109022 / 21.02 / Properly show if there are no installed apps in UI

AlexKarpov98:


sonicaj (commented):
Do you think the text is perhaps too small ?
Screenshot 2021-01-20 at 12 30 43 PM

sonicaj:
I do like @dszidi suggestion in the ticket, not sure if UI would like to do that - but just my opinion :p

AlexKarpov98:
Implemented Empty State component and applied it on Apps section

src/app/pages/applications/chart-releases/chart-releases.component.ts
```diff
@@ -35,12 +37,13 @@ export class ChartReleasesComponent implements OnInit {
   private refreshTable: Subscription;
   private refreshForm: Subscription;
   
-  public emptyPageConf = {
+  public emptyPageConf: EmptyConfig = {
+    type: EmptyType.loading,
+    large: true,
     title: helptext.message.loading,
```
dszidi, line 43: Lets add a message too.

"Applications you install will automatically appear here. Click below and browse the TrueNAS catalog to get started."

dszidi (changes requested):
Great start to the empty state component! Couple of things need to be tweaked though. The magnify icon is meant to be shown in scenarios where a search has returned no results, so it's not really appropriate here. On this page the "format-list-text" icon would be a better choice IMO. Perhaps better property names could also be used. Let's try these changes:

- Rename `user_cleared`  to  `no_page_data`
- Rename `no_results` to `no_search_results`
- Set the type to `no_page_data` for this page
- Add this message "Applications you install will automatically appear here. Click below and browse the TrueNAS catalog to get started." 

The other thing is that the loader shows the button and the message too. The loader scenario should only show the logo. Also a delay might be good here since the empty component comes up rather quickly.

AlexKarpov98:
Fixed

dszidi:
The message requested in the previous comment was supposed to be the EmptyConfig.message value. It should look like this...

image

dszidi (approved):
LGTM

PR #5070, opened 2021-01-21
NAS-109027 / 12.0 / Removed anchor link's tab selection for navigation panel to avoid mat-nav-list issues

ryousaf:
NAS-109027


AlexKarpov98 (approved):
LGTM

PR #5073, opened 2021-01-22
NAS-109076 / 21.02 / Repopulating entity-tree-table when config is changed by the containing component

ryousaf:
NAS-109076


dszidi (approved):
LGTM

PR #5074, opened 2021-01-22
NAS-108949 / 21.02 / Pushed right hand bulk actions & search field on new line when running out of space for better spacing

ryousaf:
NAS-108949


denysbutenko (approved):
Looks better now

PR #5085, opened 2021-01-28
NAS-109159 / 21.04 / Update rendering for a list / NAS-109159

AlexKarpov98:
Change add button from first item and add remove button on first item in a list.


dszidi:
Getting too many concurrent calls traceback...

image

dszidi:
The button style is inconsistent with the rest of the UI. 
image

Also, if you hit delete on the only item, you are left with nothing. There is no indication that there is anything there other than the vague plus sign.

image


A couple of things will fix this.
1 . Put a text label on the left
2 . Make the add button a regular mat-button with text and no icon.

It should look something like this...

image

dszidi:
Last change. Let's change that delete button to an 'X' icon and put it in the top left. Also for the dynamic label we should use something more generic like 'Configure ' + config.name so we don't over use the word 'Add'

AlexKarpov98:
Updated

dszidi (approved):
LGTM

dszidi:
This seems to break ACL manager. If you select one of the items to be default using the checkbox, the submission process is stuck at 50%.

image

AlexKarpov98:
It is middleware issue and it should be fixed with this PR: https://github.com/freenas/freenas/pull/6356

PR #5110, opened 2021-02-04
NAS-109257 / 21.02 / Moved Guide link to settings list in the topbar from Misc page

ryousaf:
NAS-109257


src/app/pages/credentials/credentials.component.html
```diff
@@ -5,7 +5,6 @@ <h3>Misc</h3>
         <mat-divider></mat-divider>
         <ul>
             <li (click)="systemNav('email')">Email</li>
-            <li (click)="navExternal('https://www.truenas.com/docs/')">Guide</li>
```
denysbutenko, line 8: if there is no usage of `navExternal` method do we need to keep it?

ryousaf: Good idea. Removed.

denysbutenko (approved):

PR #5117, opened 2021-02-04
NAS-109029 / 21.04 / Allow user to configure GPU in Launch Docker image wizard in UI / NAS-109029

AlexKarpov98:


src/app/pages/common/entity/utils.ts
```diff
@@ -265,4 +265,183 @@ export class EntityUtils {
...
+          operator: relation[1],
+          value: relation[2],
+        }]
+      };
+    });
+
+    return result;    
+  }
+
+  parseSchemaFieldConfig(schemaConfig, parentName=null, parentIsList=false) {
```
ryousaf, line 289: We should have classes and types where feasible for cleaner implementation and maintenance

AlexKarpov98: Regarding using class, they are all dynamic so we can't know its structure beforehand.
Also it is related with dynamic chart release and all data is from schema that middleware sends and it has unexpected structure.
That's why we can only use `any `or `object` here.

ryousaf (approved):
LGTM

ryousaf (approved):
LGTM

PR #5120, opened 2021-02-05
Update pl.po

grigoriii:
first translation of TrueNAS, just want to check if works, next will make another translation.
Regards grigoriii


kmoore134 (approved):

kmoore134:
Added, thanks for the contribution!

PR #5123, opened 2021-02-05
NAS-109267 / 21.02 / After upgrading chart release UI does not show updated version of app / NAS-109267

AlexKarpov98:


sonicaj (commented):
We should update the information for the whole app tile basically, it's possible it starts consuming different ports or using different images etc.

sonicaj (commented):
We should be showing protocol as lowercase, api returns them upper case.

How about we have common logic for rendering the app component and then use that one here and the normal rendering one as well when we show apps ? ( just a recommendation :P )

AlexKarpov98:
We are using filter for lowercase in rendering part so don't have to worry about it.

sonicaj (approved):

Ornias1993 (approved):
Looks good, clean code, simple improvement
LGTM

dszidi (approved):
lgtm

PR #5125, opened 2021-02-05
NAS-109261 / 21.04 / Refreshed table upon resetting config

ryousaf:
NAS-109261


denysbutenko (commented):
The table still empty when added CSR

denysbutenko (approved):

PR #5132, opened 2021-02-09
NAS-109042 / 21.04 / Improve checkbox alignment

denysbutenko:


AlexKarpov98 (commented):
Nice fix for this issue.
But seems you didn't apply this new option to all broken pages.
So I see this checkbox issue is fixed on `Dashboard Configuration` page, but not on other pages like `Disk Edit` page and so on.

denysbutenko:
> Nice fix for this issue.
> But seems you didn't apply this new option to all broken pages.
> So I see this checkbox issue is fixed on `Dashboard Configuration` page, but not on other pages like `Disk Edit` page and so on.

That's correct. Initially, the idea was to improve grid alignment and add the ability to handle checkbox position manually. Two checkboxes height matches one input field. Also, there is an option to make checkbox the same height as the input. Updating all forms is out of scope. The `Disk Edit` form requires more rearranges and will be improved as a separate ticket.

AlexKarpov98 (approved):

PR #5135, opened 2021-02-09
NAS-109312 / 21.04 / Chart Release creation should be a wizard like "launch docker image" / Nas 109312

AlexKarpov98:


Ornias1993:
Awesome work Alex, this would significantly improve the UX for Apps :)

Question:
With this implemented, what is the use of the seperate (hardcoded) modal/UI for "launch docker image"?

https://github.com/truenas/webui/blob/fd9542b3213b975253ba033eadbe633a4ae9c8dc/src/app/pages/applications/forms/chart-release-add.component.ts

It seems perfectly possible to load the questions.yaml from the ix-chart and applying these functions, instead of hardcoding...

AlexKarpov98:
> seperate

AlexKarpov98:
@Ornias1993 
That question is reasonable.
AFAIK `ix-chart` is different from `Dynamic Chart Release`, it means `ix-chart` is mostly static. That's why we handle `ix-chart` separately.
Also this is already discussed with middleware team.

Ornias1993:
@AlexKarpov98 Thanks for the response.
ix-chart is mostly static, yes.

But the path from UI -> middleware -> k8s is mostly the same. 
imho it's a bit very-much redundant to do it twice, but if thats what the middleware team wants thats not a biggy. :)

sonicaj (commented):
I have following concerns:

1) We should be showing `Application Name` separately and not in Container Images. That is something which is present for each chart release.
Screenshot 2021-02-22 at 6 52 34 PM

2) Nothing happens when I click on cancel button, i was expecting the form to close.
Screenshot 2021-02-22 at 6 54 33 PM

Ornias1993:
@sonicaj good catch, yes those two nitpicks are indeed relevant...
With the first one users might think it's somehow related to the image name, which it obviously isn't.

AlexKarpov98:
Updated

ryousaf (commented):
A lot of methods don't have type annotations for their parameters. We should also use `any` as a last resort. The PR looks good to me otherwise.

sonicaj (commented):
We should not allow moving on to next section if some required field has not been provided. Like if I donot fill in application name, it will still allow me to move to next section.

AlexKarpov98:
Updated

sonicaj (approved):

src/app/pages/common/entity/entity-wizard/entity-wizard.component.ts
```diff
@@ -234,7 +234,7 @@ export class EntityWizardComponent implements OnInit {
     }
   }
 
-  selectionChange(event) {
+  selectionChange(event: any) {
```
ryousaf, line 237: `event` here should be of type `StepperSelectionEvent`

src/app/pages/common/entity/entity-wizard/entity-wizard.component.ts
```diff
@@ -259,7 +259,7 @@ export class EntityWizardComponent implements OnInit {
     }
   }
 
-  getSummaryValue(fieldConfig, formControl) {
+  getSummaryValue(fieldConfig: FieldConfig, formControl: any) {
```
ryousaf, line 262: formControl should be of type `FormControl`

Ornias1993:
This might actually be a good moment to take note of NAS-109632
As lists take up quite a big portion of every step in the wizzard.

src/app/pages/applications/forms/chart-wizard.component.ts
```diff
@@ -43,7 +44,7 @@ export class ChartWizardComponent implements OnDestroy {
     this.utils = new CommonUtils();
   }
 
-  createRelations(relations, parentName) {
+  createRelations(relations: any[], parentName: string) {
```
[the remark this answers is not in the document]
AlexKarpov98: @ryousaf The value of "relations" is the array of array, like this [["hostPathEnabled", "=", false], ["datasetName", "=", true]], not the array of object. Should we define new class/type for this array and convert this array to new class/type?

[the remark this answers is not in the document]
ryousaf: Ah, Ok. But in general, we should have a clear scope of what actions are supported by relations and how a relation is represented in the form of a class as I had mentioned. I'm not sure I understand the hierarchy here. We are using relations more and more and we should have a better structure of what a relation object looks like and what actions are supported. Reusability and maintenance will improve, I think.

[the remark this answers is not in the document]
AlexKarpov98: @ryousaf Exactly, we need to represent relations in a class and actually we have already have it `FieldRelation` as well as `RelationGroup`. Both are used for represent relations and its group in other places. In this function the params `relations` is just a raw data from middleware which will be converted to `FieldRelation` class later somewhere else.
I couldn't find the very reason to introduce new class/type here, but anyway made a new class for it as it might be definitely helpful if we have a chance to reuse it somewhere else later.

src/app/pages/common/entity/entity-wizard/entity-wizard.component.ts
AlexKarpov98: Updated

src/app/pages/common/entity/entity-wizard/entity-wizard.component.ts
AlexKarpov98: Updated

AlexKarpov98:
Updated

ryousaf (approved):
LGTM

Ornias1993 (approved):

PR #5136, opened 2021-02-10
NAS-109263 / 21.04 / Set overflow-wrap property for title, message in entity-dialog. Fixed width of form-paragraph.

ryousaf:
NAS-109263

### DIFF
--- a/src/app/pages/common/entity/entity-dialog/entity-dialog.component.css
+++ b/src/app/pages/common/entity/entity-dialog/entity-dialog.component.css
@@ -5,3 +5,7 @@ button.action-btn {
 div.entity-dialog-form-message {
     margin: 0 0 13px;
 }
+
+.entity-dialog-form-title, .entity-dialog-form-message {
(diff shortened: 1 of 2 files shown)


denysbutenko (commented):
No way to scroll dialog to make actions

src/app/pages/common/entity/entity-dialog/entity-dialog.component.css
```diff
@@ -4,4 +4,11 @@ button.action-btn {
 
 div.entity-dialog-form-message {
     margin: 0 0 13px;
+}
+
+.entity-dialog-form-content {
+    overflow-y: scroll !important;
```
denysbutenko, line 10: It means scrollbar will be visible always even if the content fits without scrolling, isn't?

ryousaf: Nice catch. Fixed.

denysbutenko (approved):

PR #5138, opened 2021-02-10
NAS-109275 / 21.02 / Fix object detection in getDataType

denysbutenko:
Add back checking for `null` since `typeof null == 'object'`


dszidi (approved):
LGTM

PR #5140, opened 2021-02-10
NAS-108990 / 12.0 / Handled job response case when state is ABORTED for disk import

ryousaf:
NAS-108990

### DIFF
--- a/src/app/pages/storage/import-disk/import-disk.component.ts
+++ b/src/app/pages/storage/import-disk/import-disk.component.ts
@@ -181,7 +181,7 @@ export class ImportDiskComponent implements OnDestroy, Formconfiguration {
     this.dialogRef.componentInstance.success.subscribe((job_res) => {
       this.dialogRef.close();
       this.entityForm.success = true;
-      this.job.showLogs(job_res, T('Disk Imported: Log Summary')), T('Close');
+      this.job.showLogs(job_res, T('Disk Imported: Log Summary'), T('Close'));
       this.custActions = [
         {
           id: 'view_import_log',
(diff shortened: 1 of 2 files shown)


src/app/pages/storage/import-disk/import-disk.component.ts
```diff
@@ -180,8 +180,8 @@ export class ImportDiskComponent implements OnDestroy, Formconfiguration {
     this.dialogRef.componentInstance.submit();
     this.dialogRef.componentInstance.success.subscribe((job_res) => {
```
william-gr, line 181: Why are we reaching this code if the job didn't succeed?

ryousaf: From what I gather, ABORTED is a valid state for the job and not failure. Failure is when there's an error in job execution. ABORTED is a valid state where no error occurred. We could improve the code by make a separate clearer indication like `dialogRef.componentInstance.success` (maybe `dialogRef.componentInstance.aborted`)

william-gr: Yes, lets do that, please. Treating ABORTED as success would be wrong.

ryousaf: The option to abort isn't provided in the UI, however. At least, not in this case. This case occurs when the job is initiated from the UI but aborted from the machine shell directly.

william-gr: Right, but soon enough we will let it be done via UI

src/app/pages/storage/import-disk/import-disk.component.ts
```diff
@@ -181,7 +181,7 @@ export class ImportDiskComponent implements OnDestroy, Formconfiguration {
...
       this.dialogRef.close();
       this.entityForm.success = true;
-      this.job.showLogs(job_res, T('Disk Imported: Log Summary'), T('Close'));
+      this.job.showLogs(job_res, T('Disk Imported: Log Summary')), T('Close');
```
william-gr, line 184: This seems wrong?

ryousaf: Yes. Misplaced parenthesis. Fixed.

william-gr (approved):

PR #5143, opened 2021-02-11
NAS-109326 / 21.04 / Show warning dialog for first time when editing advanced settings

denysbutenko:


AlexKarpov98:
@denysbutenko 
This implementation allows to show warning dialog only when the user try to add `Sysctl` first time other than other assets like `Console`, `Kernel` and `Syslog`.
So it means in the case when you try to add a `Console` first and try to add a `Sysctl` next, you can't see warning dialog at all.
Not sure if this is expected behavior.

AlexKarpov98 (commented):
@denysbutenko 
This implementation allows to show warning dialog only when the user try to add `Sysctl` for first action on this page other than for other assets like `Console`, `Kernel` and `Syslog`.
So it means in the case when you try to add a `Console` first and try to add a `Sysctl` next, you can't see warning dialog at all.
Not sure if this is expected behavior.

denysbutenko:
> @denysbutenko
> This implementation allows to show warning dialog only when the user try to add `Sysctl` for first action on this page other than for other assets like `Console`, `Kernel` and `Syslog`.
> So it means in the case when you try to add a `Console` first and try to add a `Sysctl` next, you can't see warning dialog at all.
> Not sure if this is expected behavior.

Not expected 👍🏻 Good catch! Thanks

denysbutenko:
Updated. The warning will be shown for any slide-out form on the page when opened for the first time.

AlexKarpov98 (approved):
LGTM

PR #5151, opened 2021-02-12
NAS-108957 / 21.04 / Show chart release events in the UI

AlexKarpov98:


sonicaj (commented):
It is very nice. Can we please change the cursor shape when it hovers above the area from where we can actually open up the summary dialog as it's not intuitive otherwise that the user should click and something will happen. It's the same we do for buttons - btw while checking this I have noticed that when i hover over Stop button, sometimes the cursor will change shape and sometimes it won't. Can we also please look into that while we are looking into this ?

sonicaj (commented):
Let's please add `catalog`, `catalog_train` and `chart_metadata.name` attributes as well. They will be incredibly helpful when we have multiple catalogs or using apps from multiple trains in a catalog.

sonicaj (commented):
Also when we don't have any events present for a chart release, can have a text saying that please ? Right now we click on the drop down and nothing shows up if there are no events. The text can be something like `No Recent Events`.

AlexKarpov98:
Updated all the above

src/app/pages/applications/dialogs/chart-events/chart-events-dialog.component.html
```diff
@@ -0,0 +1,78 @@
+<div mat-dialog-content fxLayout="column wrap" fxLayoutAlign="center start" >
+    <div class="logo-container" fxLayout="row" fxLayoutAlign="center center">
+        <div class="logo" style="background-image: url('{{ catalogApp.icon }}');"></div>
+        <div>
+            <strong class="chart-name">{{ catalogApp.name }}</strong><span class="version"> {{ catalogApp.version}}</span>
+        </div>
+    </div>
+    <div class="ports">
+        <span>Used Ports:</span><span class="value" textLimiter threshold="70">{{ catalogApp.used_ports }}</span>
```
denysbutenko, line 9: The most of strings in this file can't be translated to other langs

AlexKarpov98: > The most of strings in this file can't be translated to other langs

Fixed

denysbutenko (approved):

sonicaj (changes requested):
I have following concerns:

1) When a chart release is not using any ports, we should have some text saying that no ports are being used.
Screenshot 2021-02-19 at 8 46 56 PM

2) We should show if there are updates available for the chart or the images it is using - probably somewhere where we are showing the version.

3) For each container image we are showing in the summary, we should have another label next to it probably or some text in the background saying if it's up to date or if there's an update available. Changes are being introduced in https://github.com/truenas/middleware/pull/6456 where `chart.release.query` endpoint will retrieve this information making it easier to display this information.

4) Please rename `Chart Metadata Name` to `Item Name` and `Catalog Train` to `Train` I think.

Apart from this, it's really looking nice :)

AlexKarpov98:
Updated

sonicaj (commented):
It looks very nice now. I have one minor nit, i think having another row just saying Update Available is not very nice and we should use the space next to current version ? How about Having a tag like ( Update Available ) similar to how status is shown, and then if user moves over it, we show the new version which is available. Thoughts ?

Screenshot 2021-02-26 at 7 47 09 AM

AlexKarpov98:
Updated

sonicaj (approved):
Thank you! It looks really nice now :)

PR #5164, opened 2021-02-19
NAS-109431 / 21.04 / UI does not show null selected entry in enum for chart release edit/create form / Nas 109431

AlexKarpov98:
- UI does not show null selected entry in enum for chart release edit/create form
- Allow display text for null valued options in a select control


sonicaj (approved):

dszidi (approved):
LGTM

PR #5170, opened 2021-02-22
NAS-109087 / None / Store boot pool name in Global Cache

dszidi:


dszidi:
Work on global cache of system info will be continued as part of a separate ticket

PR #5171, opened 2021-02-22
NAS-109443 / 21.04 / Dynamically created fields based on schemas returned from middleware for ACME DNS Authenticators

ryousaf:
NAS-109443

### DIFF
--- a/src/app/pages/credentials/certificates-dash/forms/acmedns-form.component.ts
+++ b/src/app/pages/credentials/certificates-dash/forms/acmedns-form.component.ts
@@ -5,6 +5,8 @@ import { ModalService } from 'app/services/modal.service';
 import { FieldConfig } from '../../../common/entity/entity-form/models/field-config.interface';
 import { FieldSet } from 'app/pages/common/entity/entity-form/models/fieldset.interface';
 import { helptext_system_acme as helptext, helptext_system_acme } from 'app/helptext/system/acme';
+import _ from 'lodash';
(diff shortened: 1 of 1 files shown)


src/app/pages/credentials/certificates-dash/forms/acmedns-form.component.ts
```diff
@@ -96,6 +36,57 @@ export class AcmednsFormComponent {
...
+                placeholder : helptext.authenticator_provider_placeholder,
+                tooltip : helptext.authenticator_provider_tooltip,
+                options : [
+                ],
+                parent: this
+              }
+            ]
+          }]
+        for(let schema of schemas) {
+          fieldSet[0].config[1].options.push({label: schema.key, value: schema.key});
```
denysbutenko, line 66: That's not very clear what stand for `fieldSet[0].config[1]`. It may be more verbose if we use `FieldSets` class from https://github.com/truenas/webui/blob/master/src/app/pages/common/entity/entity-form/classes/field-sets.ts 

Then it may be something like:
```
const fieldSet = new FieldSets([...])
const authenticatorOptions = fieldSet.config("authenticator")
authenticatorOptions.options.push(...)
```

Just a thought. What do you think @ryousaf?

ryousaf: Good idea! I'll try and get that done. Thanks!

denysbutenko (approved):
LGTM

PR #5173, opened 2021-02-23
NAS-109446 / 21.04 / Show pod logs in UI / Nas 109446

AlexKarpov98:


sonicaj (commented):
Good job!

I have following concerns:
1) Let's please reword `Download` to `Downlaod Logs`.
2) Let's please have the same configurable menu popup when Download is clicked so user can customize how many lines to download or change container etc.
3) Let's please show timestamp as discussed highlighting it ( we should have that come in from middleware here https://github.com/truenas/middleware/pull/6481/files )
4) I think it would be nice if we could somehow have an arrow/button leading back to Installed Applications page - same is true for console as it's too many steps right now. We need to actually click on Applications -> Installed Applications.
5) Would be nice if we show some metadata / context of what resource the logs are being shown like the App name / pod name / container name as in complex applications, it can be easy to get confused to see whose logs the user is seeing currently.

Ornias1993:
> 1. Let's please reword `App Name` to `Application Name`.


Maybe not, it might be confusing when (for example) there is an App (lets say Nextcloud) with a Redis and Database dependency.
Some users might confuse "Application" with (one of the) three included applications, while "App" more clearly defines what we would define as the "release"

sonicaj:
@Ornias1993 the idea is user not being aware of the internal components of the chart which can be consuming any number of dependencies. I believe this is us being consistent with other parts in the UI like saying Installed Applications or Application Name.

AlexKarpov98:
Updated

sonicaj (commented):
@AlexKarpov98 I still cannot download logs after switching container.
Screenshot 2021-02-27 at 1 37 15 AM

AlexKarpov98:
Updated

sonicaj (commented):
It is working quite nicely now, just one minor nit, can we please name the downloading file with `appname_podname_containername.log` ? Because just using `pods.log` will end up with using same name for multiple container logs.

AlexKarpov98:
Updated

Ornias1993:
Does this also allow for downloading the last logs for k8s jobs, pre-install/update and post-install/update containers?

Might be worthwhile, its hard getting users to submit enough troubleshooting data...

sonicaj:
@Ornias1993 this will basically allow to access any containers logs in a pod which exists right now. K8s job pods remain there after completion depending on `spec. successfulJobsHistoryLimit` value specified, so if it exists, we should be able to retrieve logs.

About pre-install/update, they are jobs but they are jobs created via helm hooks. So that depends on `hook-delete-policy` iirc, based on that the above statement applies if the pod exists, we are able to retrieve logs :)

sonicaj (approved):
LGTM, thank you!

dszidi (approved):
LGTM

PR #5180, opened 2021-02-26
NAS-108914 / 21.04 / Showing updates to apps

AlexKarpov98:


sonicaj (changes requested):
I have following concerns:

1) Not aligned buttons and a scrollbar showing
Screenshot 2021-02-27 at 1 49 35 AM

2) We should try to have some kind of distinction between text `Update Available` and the newer version ( or if you think perhaps showing version with different font/color/size would be nice, that would be good as well - this does not seem very clear to me especially with multiple apps )
Screenshot 2021-02-27 at 1 50 29 AM

AlexKarpov98:
Updated

sonicaj (changes requested):
We are now not using `human_version` key to show the human catalog item version. Let's please use that.

kmoore134:
Just tested this out, showed upgrades properly, and upgrade worked. Only issue was that the 'upgrade' option was still visible from three-dot menu after the upgrade was done. 

image

AlexKarpov98:
Updated

sonicaj (commented):
We have rollback button grayed out even if a chart release has history versions which it can rollback to.

Screenshot 2021-03-02 at 1 34 13 AM

sonicaj:
We can check history via
```
midclt call chart.release.query '[["id", "=", "ipfs-test"]]' '{"extra": {"history": true}}'
```

AlexKarpov98:
It should solved as middleware is updated accordingly.

sonicaj (approved):

dszidi:
Code looks good. Just one small thing. Please add types to the "poolList" and "selectedPool" properties of the CatalogComponent.

AlexKarpov98:
Updated

src/app/pages/applications/chart-releases/chart-releases.component.html
```diff
@@ -52,7 +53,7 @@
           </button>
           <mat-menu #actionsMenu="matMenu">
             <button name="update" mat-menu-item (click)="update(item.name)" 
-              [disabled]="!item.update_available"
+              [disabled]="!item.update_available && !item.container_images_update_available"
               ix-auto ix-auto-type="option" ix-auto-identifier="rollback">
```
dszidi, line 57: ix-auto-identifier should probably equal "update"

dszidi (approved):
LGTM

PR #5181, opened 2021-02-28
Update fr.po

MDVAurelien:
Add some translations


src/assets/i18n/fr.po
```diff
@@ -419,7 +419,7 @@ msgid ""
 msgstr ""
 "Type de<a "
 "href=\"https://www.fastmail.com/help/technical/ssltlsstarttls.html\" "
-"target=\"_blank\"> chiffrement des courriels</a>. Les choix sont <i>Plain "
+"target=\"_blank\"> chiffrement des emails</a>. Les choix sont <i>Plain "
```
denysbutenko, line 422: Below all words have a dash `e-mails`, is it a typo?

MDVAurelien: No in France it is more conventional to read 'E-mail' but both are valid. It is also common to find the term 'courriel' in the Administration. Thanks

kmoore134:
ok to test

denysbutenko (approved):
LGTM

PR #5184, opened 2021-03-01
NAS-109495 / 21.04 / Fixed race condition where system.config.info request was being sent before auth confirmation

ryousaf:
NAS-109495


denysbutenko (approved):
LGTM

PR #5205, opened 2021-03-05
NAS-109632 / 21.04 / Allow not showing first entry of list by default / Nas 109632

AlexKarpov98:


sonicaj (approved):

Ornias1993 (approved):

Ornias1993:
Just to be 200% clear: this is optional for Apps, right? I can either show-1 or show-none?

Because if it isn’t, that’ll cause the same issue but “the other way around”...

AlexKarpov98:
So this is optional, we can configure it as we want.
In general, list form will render one item by default.
But is different for dynamic chart release. List form in dynamic chart release will render nothing by default but will render one item only when it has `default` attribute in its schema. (Detailed logic to show one item by default is a bit complex, will share it if it is needed)

Ornias1993:
@AlexKarpov98 awesome, so when I add `default: []` I should get the current behavior (one item) and if I do not I get none?

AlexKarpov98:
Can you please make it clear? So where is that `default` attribute belongs to ?

Ornias1993:
@AlexKarpov98 I don't know, I ask a general questions... You started talking about the `default: []` attribute.

I simply want to know how I can keep the default behavior on some objects.

AlexKarpov98:
Basically if the schema has `default: []`, it should be visible.

Ornias1993:
@AlexKarpov98 Ahh, so all I need to do to have 1 showing, is adding `default: []` and without `default: []` I would get no-entry showing by default

AlexKarpov98:
To be clear, we check `default` attribute of an item's schema, but not list itself.
`List` includes items, each item might have several variables with its schema respectively.
Here we check `default` attribute for each variable.
So we show first item by default only when both of following conditions are satisfied:
1. There is at least one variable with `default` attribute is defined, even if it is `[]` or else.
2. There is no variable which has `required: true` but `default` attribute is `undefined`.

Ornias1993:
@AlexKarpov98 This whole thing was initiated by me, because the problem was `required: true` on one of the variables + showing the first item by default causing problems.

So this won't actually solve my issues at all, showing the first item wasn't the problem in itself. 

It was showing the first item which as variables with `required: true` was. Because that means having such a list means users cant finish the Installation UI without removing the list item.

AlexKarpov98:
@Ornias1993 
I understand what the problem we have currently is.
Current logic is also made based on your idea - User has to fill out or remove required fields if they are displayed by default but not autofilled.
It is working in that way at this moment.
If you can see unexpected result, please let me know.
Also adding new attribute like `show_by_default` will definitely solve this problem!

sonicaj:
@AlexKarpov98 we should only render a default entry if an actual item is specified in `default: []` and not based on what default attribute an `item` inside `list` has ( if ) specified. 
@Ornias1993 can you please create a suggestion ticket for `show_by_default` ? It would certainly make this easier to process / handle.

Rendering the default entry based on these vague conditionals is not good imho and we should actually respect list schema specified default which by default is `[]` ( JFYI ) as otherwise i think we violate POLA where we render a default entry because the list item has a default specified.

Ornias1993:
> @AlexKarpov98 we should only render a default entry if an actual item is specified in `default: []` and not based on what default attribute an `item` inside `list` has ( if ) specified.
> @Ornias1993 can you please create a suggestion ticket for `show_by_default` ? It would certainly make this easier to process / handle.

Sure, I tried to explain it as clear as possible and linked to this discussion:
https://jira.ixsystems.com/browse/NAS-109761

> Rendering the default entry based on these vague conditionals is not good imho and we should actually respect list schema specified default which by default is `[]` ( JFYI ) as otherwise i think we violate POLA where we render a default entry because the list item has a default specified.

Indeed @sonicaj  thats also an issue, we have a default parameter on lists that isn't actually used...

AlexKarpov98:
Updated!
So this change only makes list form to hide first element by default.
Conditional displaying will be solved on upcoming ticket: https://jira.ixsystems.com/browse/NAS-109761

Ornias1993:
@AlexKarpov98 Perfect, nice AGILE way of spreading it out! :)

ryousaf (approved):
LGTM

PR #5206, opened 2021-03-05
NAS-109511 / 12.0 / Show dialog when smart has no results

denysbutenko:


dszidi:
The smart tests results page is using entity-table. I think entity-empty might be a better choice for this since that is what we use for other empty tables.

PR #5227, opened 2021-03-10
NAS-109682 / 21.04 / Added event to handle table refresh manually inside EntityTableComponent

ryousaf:
NAS-109682


denysbutenko (approved):
LGTM

PR #5230, opened 2021-03-11
NAS-109739 / 12.0 / Add snapshot name to keyprops. 

dszidi:
Also better formatting for multiple keyprops in entity-table delete message


denysbutenko (approved):
Looks good

PR #5237, opened 2021-03-12
NAS-109677 / 21.04 / Update repo info

denysbutenko:

PR #5238, opened 2021-03-12
Update readme with correct header

Ornias1993:
The subtitle already reads:
"This is the project for the Angular.io WebUI for TrueNAS 12 & TrueNAS SCALE."

So the header should reflect this too.


Ornias1993:
Retracted in favor of #5237

PR #5240, opened 2021-03-12
NAS-109682 / 21.04 / Refresh page using getData when new vm has been created

ryousaf:
NAS-109682

### DIFF
--- a/src/app/pages/vm/vm-wizard/vm-wizard.component.ts
+++ b/src/app/pages/vm/vm-wizard/vm-wizard.component.ts
@@ -448,8 +448,6 @@ export class VMWizardComponent {
   }
 
   afterInit(entityWizard: EntityWizardComponent) {
-    console.log(entityWizard)
-    console.log(< FormGroup > entityWizard.formArray.get([0]))
     this.ws.call('vm.query').subscribe((res) => {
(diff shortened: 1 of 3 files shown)


src/app/pages/vm/vm-wizard/vm-wizard.component.ts
```diff
@@ -955,6 +953,7 @@ async customSubmit(value) {
       this.ws.call('vm.create', [vm_payload]).subscribe(vm_res => {
         this.loader.close();
         this.modalService.close('slide-in-form');
+        this.modalService.refreshTable();
```
denysbutenko, line 956: I have a similar issue and was curious if there any objections to call `modalService .refreshTable()` directly in `modalService .close()`?

ryousaf: Maybe instead of `refreshTable` we could call it `onClose`? That way it comes a wider array of scenarios and makes sense to use it inside `modalService.close()`.

denysbutenko: Agreed, makes sense

denysbutenko (approved):
Looks good

PR #5248, opened 2021-03-16
NAS-109202 / 21.04 / Data Protection Dashboard

denysbutenko:


dszidi:
I'm getting a traceback deleting a periodic snapshot task. It seems like the problem lies in how it tries to refresh the table after the task is deleted because once I refresh the page, the the task is gone.

image

dszidi:
The navigation menu item and the page title are both Data Protection but the URL is still "tasks". We should make the URL match the title and navigation.

image

src/app/helptext/system/tunable.ts
```diff
@@ -58,5 +58,10 @@ export const helptext_system_tunable = {
       "Enable this tunable. Unset to disable this tunable\
  without deleting it."
     )
+  },
+
+  tunable: {
+    add: T('Add Tunable'),
```
dszidi, line 64: In Scale these are Sysctl not Tunable.

denysbutenko: Fixed

dszidi (approved):
LGTM

PR #5252, opened 2021-03-17
NAS-109804 / 21.04 / Added column based sorting in EntityTreeTable

ryousaf:
NAS-109804


denysbutenko (approved):
Looks good. One thing that takes my attention is missing curly braces for if-else statements. Apparently, we don't have rules for that yet, so it should be fine.

PR #5265, opened 2021-03-22
NAS-109553 / 21.04 / Add checkbox for "Show Console Messages" to GUI settings

denysbutenko:


AlexKarpov98 (commented):
Nice implementation. Btw I can't see `Show Console Messages` status in GUI card on System General Dashboard page. Not sure this is expected behavior.

denysbutenko:
Good point, @AlexKarpov98. Pushed a fix for it

AlexKarpov98:
Thanks for quick fix, what do you think about this styling issue?
Maybe we should remove divider for last element or maybe we should use some padding for the card. :)
image

denysbutenko:
I think it must be improved at some point but it requires finding a suitable approach. The first case that comes to my mind is where the last item of a list doesn't stick to the card bottom, so we can't remove the bottom border of the last item. That is a good suggestion to resolve separately.

dszidi (approved):
LGTM

PR #5271, opened 2021-03-24
NAS-109713 / 21.04 / Creating devices separately with vm.device.create when creating a new VM through VMWizardComponent

ryousaf:
NAS-109713


dszidi (approved):
LGTM

PR #5272, opened 2021-03-24
NAS-109893 / 21.04 / Handled configure case for OpenVPN Client and Server services

ryousaf:
NAS-109893


denysbutenko (approved):
Looks good to me

PR #5279, opened 2021-03-30
NAS-109689 / 12.0 / Add warning message for Log vdev

denysbutenko:
image

### DIFF
--- a/src/app/helptext/storage/volumes/manager/manager.ts
+++ b/src/app/helptext/storage/volumes/manager/manager.ts
@@ -21,7 +21,8 @@ manager_disknumExtendConfirmMessage : T("It is not recommended to extend a pool 
 
 manager_vdevtypeErrorMessage : T("Adding data vdevs of different types is not supported."),
 
-manager_stripeVdevTypeErrorMessage : T("vdev may result in data loss if it fails combined with a power outage."),
+manager_stripeVdevTypeErrorMessage : T("vdev is highly discouraged and will result in data loss if it fails"),
+manager_logVdevWarningMessage: T("A stripe log vdev may result in data loss if it fails combined with a power outage."),
 
 manager_diskAddWarning : T("The contents of all added disks will be erased."),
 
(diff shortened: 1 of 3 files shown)


william-gr (changes requested):
There are two things to consider.

For data vdev the old text should remain the same with same color.

The new text and should be used for Log vdev only.

src/app/helptext/storage/volumes/manager/manager.ts
```diff
@@ -21,7 +21,7 @@ manager_disknumExtendConfirmMessage : T("It is not recommended to extend a pool
 
 manager_vdevtypeErrorMessage : T("Adding data vdevs of different types is not supported."),
 
-manager_stripeVdevTypeErrorMessage : T("vdev is highly discouraged and will result in data loss if it fails"),
+manager_stripeVdevTypeErrorMessage : T("vdev may result in data loss if it fails combined with a power outage."),
```
Mrt134, line 24: small rework suggestion to put the danger first.
```suggestion
manager_stripeVdevTypeErrorMessage : T("Data loss can occur when the vdev fails and the system encounters a power outage!"),
```

Mrt134 (approved):
added a minor suggestion

src/app/helptext/storage/volumes/manager/manager.ts
denysbutenko: > There are two things to consider.
> 
> For data vdev the old text should remain with the same color.
> 
> The new text and color should be used for Log vdev only.

If we add a new independent warning message that appears on `log vdev` type and can be visible with the existing message at the same time. For example, Data Vdev has 1 disk and we will see the red message then Log Vdev has 1 disk and we will see the orange message like this, does it look good enough?:
Screenshot 2021-04-01 at 14 16 56

denysbutenko: @Mrt134 looks good, thanks for the suggestion. one thing that I wondering it is just a part of the warning message. Full message would be `A stripe log vdev may result in data loss if it fails combined with a power outage.` and `A stripe {type} vdev is highly discouraged and will result in data loss if it fails` like on the screenshot above.

william-gr: LGTM

denysbutenko: @Mrt134 there are two similar messages now. Could you take a look, please?

denysbutenko:
The requested changes were addressed. @william-gr, could you review it again?

william-gr (approved):
Thanks!

PR #5281, opened 2021-03-30
NAS-109467 / 21.04 / Hidden CSR List appropriately, fixed beforeSubmit to always return data

ryousaf:
NAS-109467


denysbutenko (approved):
Looks good

PR #5285, opened 2021-04-01
NAS-109993 / 21.06 / Fix spacing and border color on dark themes

denysbutenko:
How it is currently:
image
How it will be:
image

### DIFF
--- a/src/app/pages/account/groups/members/members.component.css
+++ b/src/app/pages/account/groups/members/members.component.css
@@ -36,4 +36,5 @@
 
 mat-card-title {
   margin-left: 15px;
+  margin-top: 15px !important;
 }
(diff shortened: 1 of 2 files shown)


src/app/pages/account/groups/members/members.component.css
```diff
@@ -36,4 +36,5 @@
 
 mat-card-title {
   margin-left: 15px;
+  margin-top: 15px !important;
```
undsoft, line 39: You don't need `!important` here.
You can just increase [specificity of selector](https://specificity.keegan.st/), for example by using: `mat-card.form-card mat-card-title`.

undsoft (approved):
Looks good.
Снимок экрана 2021-04-26 в 13 05 59

ryousaf (approved):
Functions as intended

PR #5289, opened 2021-04-02
NAS-110005 / 21.04 / Parse snapshot delete job results and show errorReport if any errors

dszidi:


sonicaj (approved):

src/app/pages/storage/snapshots/snapshot-list/snapshot-list.component.ts
```diff
@@ -284,13 +283,36 @@ export class SnapshotListComponent {
...
       dialogRef.close();
       this.entityList.getData();
       this.entityList.selected = [];
 
-      let infoMessage: string = T('Deleted') + ' ' + params[1].length + ' ';
-      infoMessage += params[1].length > 1 ? T('snapshots') : T('snapshot');
-      this.dialogService.Info(infoMessage, '', '320px', 'info', true );
+      
+      if(jobErrors.length > 0){
+        const errorTitle = T('Warning') + ', ' + jobErrors.length + ' of ' + params[1].length + ' ' + T('snapshots could not be deleted.');
```
undsoft, line 303: Do we have support for plurals in translation messages?
Different languages may have different order or subject/object.
Something similar to: https://angular.io/guide/i18n#plurals-alternates

undsoft: After discussing with Damian, sounds like this is not something currently supported, so we can skip it for now.

undsoft (approved):

PR #5291, opened 2021-04-04
NAS-109847 / 21.04 / Cannot create docker containers without some fields / NAS-109847

gunnarpieter:
externalInterfaces can now be left out of the form when adding (i.e. no empty external interface needs to be given when it is not needed).


AlexKarpov98 (approved):
LGTM

PR #5313, opened 2021-04-11
fix German 'State' translation

jrtberlin:
The word "State" is used in the context of a location of an organization in the CA and Certificate creation wizard. "State" can be translated into multiple words in German.
- "Zustand" refers to a state of something
- "Region" refers to a location

### DIFF
--- a/src/assets/i18n/de.po
+++ b/src/assets/i18n/de.po
@@ -13520,7 +13520,7 @@ msgid "Starting job..."
 msgstr "Starte Auftrag..."
 
 msgid "State"
-msgstr "Zustand"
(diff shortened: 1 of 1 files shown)


undsoft:
Hi,

Thank you for the PR.

We also do use State when we mean "Zustand" and this usage appears to be more common than referring to a location.
Unfortunately our translation system currently cannot distinguish between the two, which is something we need to fix.

I have created [a ticket to do that](https://jira.ixsystems.com/browse/NAS-110152). Meanwhile I would have to close this PR. We will return to this, once we have improved our translation system.

I apologise for the inconvenience.

PR #5336, opened 2021-04-15
NAS-109664 / 12.0 / Use global service for disk.query events

dszidi:


undsoft (approved):
Did some superficial regression testing.

PR #5349, opened 2021-04-18
NAS-110044 / 21.06 / When editing iSCSI extent with Extent Type = File, changing dropdown value to "Device" yields in "null_value" 

AlexKarpov98:


ryousaf (approved):
We should rename `i` to something better like `key` or `attr` so it's not confused to be an index.

PR #5350, opened 2021-04-18
NAS-109985 / 21.06 / Added GPU pass through fields to UI

ryousaf:
GPU Pass through related fields added to VM Creation Wizard
Isolated GPU PCI Ids config added to System -> Advanced
NAS-109985


src/app/pages/system/advanced/advanced-settings.component.ts
```diff
@@ -256,6 +257,13 @@ export class AdvancedSettingsComponent implements OnInit, OnDestroy {
   getDataCardData() {
     this.getAdvancedConfig = this.ws.call('system.advanced.config').subscribe((res) => {
       this.configData = res;
+      const isolated_gpu_pci_ids_str = res.isolated_gpu_pci_ids.reduce((acc, id) => {
```
undsoft, line 260: Can you do:
```
const isolatedGpuPciIds = res.isolated_gpu_pci_ids.join(', ');?
```

src/app/pages/system/advanced/advanced-settings.component.ts
```diff
@@ -335,9 +343,14 @@ export class AdvancedSettingsComponent implements OnInit, OnDestroy {
         },
         {
           id: 'sysctl',
-          title: helptext_system_advanced.fieldset_sysctl,
+          title: helptext_system_advanced.fieldset_sysctl + " - rehan",
```
undsoft, line 346: 🏴‍☠️

src/app/pages/vm/vm-wizard/vm-wizard.component.ts
```diff
@@ -934,6 +1003,18 @@ async customSubmit(value) {
       ]
     }
 
+    if(value.gpu) {
+      const gpuIndex = this.gpus.findIndex(gpu => gpu.addr.pci_slot == value.gpu);
+      vm_payload["devices"].push(...this.gpus[gpuIndex].devices.map(d => {
```
denysbutenko, line 1008: is `d` meaning `device` here? can we avoid a single character variable?

src/app/pages/system/advanced/advanced-settings.component.html
```diff
@@ -5,18 +5,21 @@
   gdColumns="50% 50%"
   gdColumns.lt-lg="50% 50%"
   gdColumns.lt-md="auto"
-  gdAreas="data-card1 data-card2 | sysctl-card data-card3 | cron-card initshutdown-card"
-  gdAreas.lt-lg="data-card1 data-card2 | sysctl-card data-card3 | cron-card initshutdown-card"
-  gdAreas.lt-md="data-card1 | data-card2 | sysctl-card | data-card3 | cron-card | initshutdown-card"
+  gdAreas="data-card1 data-card2 | sysctl-card data-card3 | cron-card initshutdown-card | data-card7 ."
```
denysbutenko, line 8: dot at the end is a typo?

src/app/pages/vm/vm-wizard/vm-wizard.component.ts
```diff
@@ -427,6 +454,26 @@ export class VMWizardComponent {
...
     })
+
+    this.ws.call("device.gpu_pci_ids_choices").subscribe((pci_choices: Object) => {
+      const isolated_gpu_pci_ids_conf = _.find(this.wizardConfig[5].fieldConfig, {name : "isolated_gpu_pci_ids"});
+      for(let key in pci_choices) {
+        isolated_gpu_pci_ids_conf.options.push({label: key, value: pci_choices[key]})
+      }
+    })
+
+    this.ws.call("device.get_gpus").subscribe((gpus) => {
```
undsoft, line 465: Could not find docs for `device.get_gpus`.
Can you:
1. Ask middleware dev that was working on this, to add at least some documentation.
2. Yourself add GpuDevice interface with what you see in inspector.

src/app/pages/system/advanced/advanced-settings.component.ts
```diff
@@ -335,9 +343,14 @@ export class AdvancedSettingsComponent implements OnInit, OnDestroy {
...
           tableConf: this.sysctlTableConf,
         },
+        {
+          title: T("Isolated GPU PCI Ids"),
+          id: 'isolated_gpu_pci_ids',
+          items: [{label: "Isolated GPU PCI IDs", value: isolated_gpu_pci_ids_str}]
```
denysbutenko, line 352: this label should be translatable

src/app/pages/vm/vm-wizard/vm-wizard.component.ts
```diff
@@ -427,6 +454,26 @@ export class VMWizardComponent {
...
+      this.gpus = gpus;
+      const gpu = _.find(this.wizardConfig[5].fieldConfig, {name : "gpu"});
+      for(let item of gpus) {
+        gpu.options.push({label: item.description, value: item.addr.pci_slot})
+      }
+    })
+
+    this.systemGeneralService.getAdvancedConfig.subscribe((res) => {
+      ( < FormGroup > this.entityWizard.formArray.get([5])).controls['isolated_gpu_pci_ids'].setValue(res.isolated_gpu_pci_ids)
```
undsoft, line 474: I am worried that if in the future someone makes changes to `formArray` and the element that you expect to be 5 becomes 6, it would not be easily noticeable. 
Is there anything else we can do here, i.e. query by some other condition and/or move to a separate getter?

undsoft (changes requested):

denysbutenko:
noticed three different ways of writing labels: 

1. "Isolated GPU PCI IDs" – `IDs`
2. "Isolated GPU PCI Ids" – `Ids`
3. "Isolated PCI ID's" – `ID's`

src/app/pages/system/advanced/advanced-settings.component.html
ryousaf: It's for CSS grid. To specify empty cell in the CSS grid with a dot. The row wasn't showing up otherwise.

denysbutenko: Thanks for the clarification 👍🏻

src/app/pages/vm/vm-wizard/vm-wizard.component.ts
ryousaf: `device.get_gpus` is not a public method. Private methods don't have their documentation in `api/docs`. An alternative is `device.get_info` which has the same purpose but less specific according to the middleware dev.

william-gr: @sonicaj UI should not be using private methods. Why get_info is less specific?

ryousaf: `get_info `is less specific in the sense that it addresses more than just GPUs whereas `get_gpus `only addresses GPUs. Both return the same information when `get_info `is used to get gpu devices info.

src/app/interfaces/AdvancedConfig.ts
```diff
@@ -0,0 +1,29 @@
...
+    serialport?: string;
+    serialspeed?: string;
+    swapondrive?: number;
+    syslog_tls_certificate?: number;
+    syslog_transport?: string;
+    sysloglevel?: string;
+    syslogserver?: string;
+    traceback?: boolean;
+    uploadcrash?: boolean;
+    sed_passwd?: string;
```
undsoft, line 28: Is everything really optional?

undsoft (approved):
Thanks for all the fixes!

src/app/interfaces/AdvancedConfig.ts
ryousaf: Everything has `required: false` in `api/docs` which is why I kept it optional. Not sure if that means it won't be in the response or it will be null. For reference, there are fields with `type` `number` or `null`. Those might be nullable. The rest might not exist. At least that's my understanding.

ryousaf: An example is `overprovision` for `system.advanced.update` middleware. According to `api/docs`

`
"overprovision": {
   "type": [
    "integer",
    "null"
   ],
   "_name_": "overprovision",
   "title": "overprovision",
   "_required_": false
  }
`

william-gr: Right, for update methods most of the times the attributes are optional, meaning you only need to submit the attribute you're changing

undsoft: Okay, sounds good.

william-gr: The attributes are returned at all times

ryousaf: @undsoft Ideally, we'd like to use the same interface for both `get` and `update` calls. It might make sense to adhere to the most relaxed definition that we can get away with.

sonicaj (changes requested):
I have following concerns:
1) We should only show 1 field here preferably the GPU one only and when user selects the GPU's isolated from the system, we update the endpoint accordingly by sending PCI ids as this can be quite confusing otherwise having 2 fields doing the same thing.
Screenshot 2021-04-21 at 9 47 38 PM

2) GPU field should allow selecting more then 1 option which is not the case right now. This also co-relates to point 1, we discussed that we would be creating PCI devices for the GPU(s) user selects but right now we are only allowing selecting one GPU. To clarify, what we should be doing is
a) We show 1 field with to the user with a list of available GPUs
b) User selects one or more GPUs
c) User moves on the next item in the wizard
d) We in the background will now add the GPUs user selected to `system.advanced` as required and also create PCI devices for the GPU(s) selected.

For step (a), we do not show which GPU(s) have already been isolated to the user as that's not what we are doing in the wizard. Our objective is to allow user to select GPU(s) which he/she would like to use with the VM and even if that's already isolated, that's fine - we still need the user to specify which GPU(s) he/she wants and we just create PCI devices then. This makes the process easy for the user to understand/go through.

It would be nice if we also allowed editing this from VM settings perhaps, so user can add/remove GPU's from there as well.

Please let me know if there are concerns. Thank you

src/app/interfaces/AdvancedConfig.ts
undsoft: I hear ya, but there's another thing we can do.
We can define this interface properly without question marks, and then add:
```
export type AdvancedConfigUpdate = Partial<AdvancedConfig>;
```

This way we can be sure where we to expect fields or not.

More info: https://www.typescriptlang.org/docs/handbook/utility-types.html

ryousaf: > The attributes are returned at all times
@undsoft let me know if I should remove the optional feature of those attributes

PR #5354, opened 2021-04-19
NAS-110265 / 21.06 / Unset Pool on SCALE Apps screen should ask for confirmation before starting task

AlexKarpov98:


undsoft (approved):
Confirmation works correctly. 
Снимок экрана 2021-04-26 в 13 53 35

PR #5361, opened 2021-04-21
NAS-110195 / None / NAS-110195 - Preparing for --no-implicit-any. Part 7

undsoft:


src/app/pages/reportsdashboard/reports.service.ts
```diff
@@ -22,15 +22,15 @@ export class ReportsService implements OnDestroy {
   public dataEvents: Subject<CoreEvent> = new Subject<CoreEvent>();
   private reportsUtils: Worker;
 
-  constructor(private ws: WebSocketService, private core:CoreService) { 
+  constructor(private ws: WebSocketService, private core:CoreService) {
     //@ts-ignore
```
denysbutenko, line 26: Is it still necessary?

denysbutenko (approved):

src/app/pages/reportsdashboard/reports.service.ts
undsoft: Probably not, will remove in future PRs.

PR #5370, opened 2021-04-24
NAS-110329 / 21.06 / Remove 'Sort language by' option from localization form

ryousaf:
NAS-110329

### DIFF
--- a/src/app/pages/system/general-settings/localization-form/localization-form.component.ts
+++ b/src/app/pages/system/general-settings/localization-form/localization-form.component.ts
@@ -40,24 +40,6 @@ export class LocalizationFormComponent implements OnDestroy{
           options: [],
         },
         {
-          type: "radio",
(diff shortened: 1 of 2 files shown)


src/app/pages/system/general-settings/localization-form/localization-form.component.ts
```diff
@@ -39,24 +39,6 @@ export class LocalizationFormComponent implements OnDestroy{
           tooltip: helptext.stg_language.tooltip,
           options: [],
         },
-        {
-          type: "radio",
-          name: "language_sort",
-          placeholder: helptext.stg_language_sort_label,
```
undsoft, line 45: helptext strings need to be checked for usage and removed if they are not used.

undsoft (approved):
👍🏼
Снимок экрана 2021-04-26 в 10 27 08

PR #5371, opened 2021-04-24
NAS-110293 / 21.06 / Fixed user count. Added dialog with user names when checkbox clicked

ryousaf:
NAS-110293

### DIFF
--- a/src/app/interfaces/user.ts
+++ b/src/app/interfaces/user.ts
@@ -1,31 +1,28 @@
-interface Attributes {
-    preferences: object;
-}
+import { Preferences } from './preferences';
+import { Group } from './group';
+
 export interface User {
+    id: number;
     uid: number;
     username: string;
-    group: object;
-    group_create: boolean;
+    unixhash: string;
+    smbhash: string;
     home: string;
-    home_mode: string;
     shell: string;
-    email: string;
-    password: string;
+    full_name: string;
+    builtin: boolean;
+    smb: boolean;
     password_disabled: boolean;
     locked: boolean;
-    microsoft_account: boolean;
-    smb: boolean;
     sudo: boolean;
     sudo_nopasswd: boolean;
-    sudo_commands: string[];
+    sudo_commands: { [property: string]: any }[]
+    microsoft_account: boolean;
+    attributes: { preferences: Preferences }
+    email: string;
+    group: Group;
+    groups: { [property: string]: any }[]
     sshpubkey: string;
-    groups: number[];
-    attributes: Attributes;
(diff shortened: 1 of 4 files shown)


src/app/interfaces/user.ts
```diff
@@ -0,0 +1,31 @@
+interface Attributes {
+    preferences: object;
```
undsoft, line 2: `Object` is very generic, for example any function is an `object`, array is an `object`.
Can you use { [attribute: string]: someType }?

src/app/interfaces/user.ts
```diff
@@ -0,0 +1,31 @@
...
+    groups: number[];
+    attributes: Attributes;
+    builtin: boolean;
+    full_name: string;
+    id: number;
+    id_type_both: boolean;
+    local: boolean;
+    smbhash: string;
+    unixhash: string;
+}
```
undsoft, line 31: 👍🏼

src/app/pages/account/groups/group-list/group-list.component.ts
```diff
@@ -137,36 +139,58 @@ export class GroupListComponent implements OnDestroy {
...
+                        }
+                      }
+                    });
+                    
+                  }
+                },
+                customSubmit: function (entityDialog) {
+                  entityDialog.dialogRef.close(true);
+                  self.loader.open();
+                  self.ws.call(self.wsDelete, [members_delete.id, entityDialog.formValue]).subscribe((res) => {
```
undsoft, line 177: res is not used.

src/app/pages/account/groups/group-list/group-list.component.ts
```diff
@@ -137,36 +139,58 @@ export class GroupListComponent implements OnDestroy {
...
+              this.loader.close();
+
+              const conf: DialogFormConfiguration = {
+                title: helptext.deleteDialog.title,
+                message: helptext.deleteDialog.message + `<i>${members_delete.group}</i>?`,
+                fieldConfig: [],
+                confirmCheckbox: true,
+                saveButtonText: helptext.deleteDialog.saveButtonText,
+                preInit: function () {
+                  if (usersInGroup.length > 0) {
```
undsoft, line 154: Let's change it to [return early](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement), i.e.

```typescript
if (!usersInGroup.length) {
  return
}

...
everything else
```

undsoft (approved):
The count is correct: 
Снимок экрана 2021-04-26 в 11 20 44

PR #5372, opened 2021-04-24
NAS-110246 / 12.0 / Removed queryCall to avoid race condition. Data fetched in preinit instead

ryousaf:
NAS-110246


undsoft (approved):
The error described in the ticket does not happen anymore.
Do we need to forwardport it?

ryousaf:
Issue is not present in Scale. Port not needed.

flotpg:
Same here:
CleanShot 2021-06-21 at 12 56 42@2x

PR #5389, opened 2021-04-29
NAS-110331 / 21.06 / NAS-110331 - Guard to wait for translations to simplify translation code

undsoft:
I was thinking about what @ryousaf said about simplifying translation system.

Here's the best I can do:

```javascript
const translated = this.translate.instant('Not translated');
```

instead of:
```javascript
this.translate.get('Not translated').subscribe((translated) => {
  
});
```

Things to test during review: correct language change in settings, correct loading of translations.


denysbutenko (approved):
Looks good 👍🏻

PR #5391, opened 2021-04-29
NAS-110293 / 21.06 / Fixed user count. Added dialog with user names when checkbox clicked

ryousaf:
NAS-110293


undsoft (approved):
Approving, testing was done in previous PR: https://github.com/truenas/webui/pull/5371

PR #5392, opened 2021-04-29
NAS-109985 / 21.06 / Added GPU pass through fields to UI

ryousaf:
NAS-109985


sonicaj (commented):
LGTM, just one minor nit - when no GPU is selected, we should have some visual indication that is indeed the case ( like saying No GPU is selected or dash line perhaps ? )

Screenshot 2021-05-03 at 6 21 02 AM


Also I was not able to access System -> Advanced because of perhaps another issue already present on master where we can't access Advanced settings.

sonicaj (commented):
LGTM, one minor nit

Can we show names of GPU's instead of their ids when displaying the isolated GPUs in system.advanced page ?

Screenshot 2021-05-10 at 6 08 44 AM


Also to confirm did we make a ticket to tackle the empty drop down indicator ?

sonicaj (changes requested):
Changing GPU selection in System->Advanced has no effect. Please see attached video.

https://user-images.githubusercontent.com/17968138/118472614-7420a680-b722-11eb-9ddd-fec52f48b2ea.mov

sonicaj (approved):

PR #5394, opened 2021-04-30
NAS-109996 / 21.06 / Fixed display device issues for VMs

ryousaf:
NAS-109996


ryousaf:
Integrating NAS-110377 into the same PR

sonicaj (commented):
I have following concerns:

1) If user has 1 display device and it's password configured, why do we ask the user to select which display device to use ?

Screenshot 2021-05-03 at 6 57 02 AM

2) If user specifies an incorrect password, we should display the error above the password field instead of changing the dialog as that would mean user would have to repeat many clicks to get back to the password window.

Screenshot 2021-05-03 at 6 57 26 AM

3) I think we are repeating information and not utilizing the space effectively with the following dialog

Screenshot 2021-05-03 at 6 58 19 AM

Why do we repeat "Display Device" and why have the empty space between the line and the options ?

ryousaf:
Regarding the spacing under the dialog, that's how dialogs are configured in the UI at the moment. There's a "title", a "message" and then the dialog with the form inputs starts a little bit below. Message is optional. But spacing is configured that way for all dialogs. @sonicaj

ryousaf:
What I can do is remove the extra "Display Device" over the radio buttons. Let's see if that looks better

ryousaf:
image
Here's how that looks

sonicaj:
Handling the spacing is a UI team decision, so if that works with you - great. Removing the duplicate text has made it better (Y)

sonicaj (changes requested):
I have following concerns:

1) We should allow users to see what the password text is ( in plain text like we do at other places )

Screenshot 2021-05-05 at 7 38 32 AM

2) When editing a DISPLAY device, please compare the borders of the box where the fields are with the creation of a device window - while editing they are really close, something changed.

Screenshot 2021-05-05 at 7 37 41 AM
Screenshot 2021-05-05 at 7 41 04 AM

This is true for any device being edited ( also please the same applies for password field here - point 1 )

sonicaj (changes requested):
And `Edit DISPLAY 43` is not really very user friendly, we should say `Edit Display Device` - the user does not need to be aware of the id of the VM device ( we don't do that anywhere else )

sonicaj (approved):
Thanks!

PR #5396, opened 2021-04-30
NAS-110364 / 21.06 / Do not show EC curve field type when selected key type is RSA

AlexKarpov98:


undsoft (approved):
Looks good.
Снимок экрана 2021-05-03 в 11 33 08

PR #5399, opened 2021-05-03
NAS-110471 / 21.06 / NAS-110471 – Making it easier to type WS calls

undsoft:
Okay, so this is not fantastic, but an intermediate improvement, that we can make better in the future.

File `src/app/interfaces/api-directory.interface.ts` will contain typings for all API calls and you will only have to specify types there.
When making ws calls with `this.ws.call` the types will be picked up and you won't have to specify them locally.

Disadvantage is that we need to add awkward types like:
```
userQuery: 'user.query' = 'user.query';
```


denysbutenko (approved):
Tested branch against the latest nightly. Looks good.

PR #5403, opened 2021-05-04
NAS-110381 / 21.06 / NAS-110381: Enable more linter rules

undsoft:


denysbutenko (approved):
Looking much cleaner now!

PR #5406, opened 2021-05-04
rename "Create" in storage overview to "Create Pool"

Ornias1993:
The "Create" button might not be clear enough on what people are actually creating.

As it's on the big "Storage" overview, it's not 100% clear it means "Create a Pool" unless people take the time to view the tooltip or read the manual.

In other windows, the "Big Blue Button", is expected to be read as an addition to the page title.
For example:
Sharing AFP - Add

But for storage this causes confusion because it's read like:
Storage - Create

But you're not "Creating" storage, but creating a new pool.

An example of such a confused user can be found here:
https://www.truenas.com/community/threads/can%E2%80%99t-add-l2arc-or-slog-in-anglefish.92968/#post-643677

(and to be honest: I understand why he is confused looking at the screenshot)


undsoft (approved):

PR #5407, opened 2021-05-04
NAS-110409 / 21.06 / NAS-110409: Hide gluster from services list and code cleanup

undsoft:
Testing: make sure gluster is not shown on Services page + general regression for Services page.


denysbutenko (approved):
Gluster service is not present in the list anymore. Looks good
image

PR #5410, opened 2021-05-05
NAS-109476 / 12.0 / Fix Full Filesystem Replication tooltip

themylogin:
@Mrt134 fix the wording if necessary please.

@undsoft can you then please make sure that i18n files are updated properly?

Master backport is not needed.


undsoft (approved):
i18n will be adjusted in the future

PR #5412, opened 2021-05-06
NAS-110048 / 21.06 / NAS-110048 - Adds support for ICU message format in translations

undsoft:
Notable changes:
* You can now author strings in ICU format. Click "Learn More" here to learn about format: http://format-message.github.io/icu-message-format-for-translators/index.html
* PO files replaced with JSON files. Harder to read for contributors, but something that our tooling supports.
* Strings will be extracted during commit, not as a separate job => slower commits, faster builds. 
This is mostly done because there is a probability of writing a broken string and we want to catch it early.
* Additional changes to make all this work.

To test:
- Check that message "Deleted x snapshots" renders correctly.


denysbutenko (approved):
image
👍🏻

PR #5413, opened 2021-05-06
NAS-110048 / 12.0 / NAS-110048 - Fix for translations taking too long to load

undsoft:
This fixes the issue when translations would take too long to load in some cases.


denysbutenko (approved):
No white screen anymore. Loading instantly using remote host. 👍🏻

PR #5414, opened 2021-05-06
NAS-110434 / 21.06 / Cant Use upercase letters for name of new app

AlexKarpov98:


undsoft:
Can we also show an error as user types in the field or even prevent user from typing uppercase characters in the first place?

Ornias1993:
@AlexKarpov98 I hope you're aware that the app name is used in quite a few places in Helmcharts, some of which do NOT allow uppercase?

AlexKarpov98:
> Can we also show an error as user types in the field or even prevent user from typing uppercase characters in the first place?

Exactly that's the normal behavior of validation in UI. But in this case, the app install is based on dynamic schema and the validation is also performed on middleware side. The validation logic might be updated anytime on middleware as well.
So don't think we can add UI validation at this moment.

Ornias1993:
@AlexKarpov98 Ahh shot I was mistaken, this is actually the error for when users try it, not trying to allow users to do this. Sorry.!

Ornias1993 (approved):

undsoft:
We need to add the same behaviour when user presses blue Launch Docker Image on top.

AlexKarpov98:
> > We need to add the same behaviour when user presses blue Launch Docker Image on top.
> 
> I read that the "Big Blue Button" is also moving to dynamic UI config, shouldn't that already consume this codepath?

Exactly, we are going to move ix-chart to dynamic way. So hopefully this behavior will be applied to ix-chart as well.

undsoft (approved):
Okay, then it looks good.

PR #5415, opened 2021-05-06
Removed extra dot

aidik:
I removed the dot which was causing the final About window to have two dots on the end of the sentence: "Does your business need Enterprise level support and services? Contact iXsystems for more information.."


undsoft (changes requested):
Hey,

I would prefer to keep all strings in one place and remove the extra dot from the template.
There's one on:
src/app/components/common/dialog/about/about-dialog.component.html:38
and seems like there's another one on line 46.

Thanks!

undsoft:
Closing in favour of https://github.com/truenas/webui/pull/5524

PR #5417, opened 2021-05-06
NAS-110474 / 12.0 / Dashboard CSS bugs in Firefox 88

AlexKarpov98:


undsoft:
I couldn't reproduce this issue even before the fix, so @dszidi, this is for you.

PR #5419, opened 2021-05-07
NAS-105480 / 12.0 / NAS-105480: Make Jail list and other tables occupy more space

undsoft:


denysbutenko (approved):
I couldn't find an example with many entries to test `autoFillWindowHeight`. The table with empty state and when it has a few rows look as before. The code looks good 👍🏻

PR #5421, opened 2021-05-07
NAS-110503 / 12.0 / NAS-110503: Preselect drives when navigating to reporting from dashboard

undsoft:


denysbutenko (approved):
Works as described

PR #5425, opened 2021-05-08
[Apps] Change default catalog branch to main

Ornias1993:
"main" is the current default for github, as people make more community catalogs, it's likely most are made on github.

Note:
Someone confirmed thats also the default for gitlab.


Ornias1993:
On a side note:
For TrueCharts i've pushed an template repository, trying to stick with both iX, Github and Gitlab defaults... Which (in that regard) also defaults to "main":

https://github.com/truecharts/template

sonicaj (approved):

PR #5427, opened 2021-05-09
NAS-110505 / 12.0 / Fixed method call for data loading, improved types

RehanY147:
NAS-110505


undsoft (approved):
Can confirm that SMART results are there. Nice work.

Снимок экрана 2021-05-10 в 10 07 14

PR #5443, opened 2021-05-13
NAS-110553 / None / NAS-110553: Do not fail if catalog app does not have versions

undsoft:


denysbutenko (approved):
`Cannot read property 'human_version' of undefined` doesn't appear anymore. List of apps loaded successfully.
image

PR #5453, opened 2021-05-14
NAS-110628 / None / NAS-110628: Adding return types 8

undsoft:


src/app/core/classes/hardware/vdev-labels-svg.ts
```diff
@@ -269,33 +266,33 @@ export class VDevLabelsSVG {
...
     if (targetEl) {
-      targetEl.style.opacity = 0;
+      targetEl.style.opacity = '0';
     }
   }
 
-  hideAllTiles(tiles: any[], exceptions?: string[]) {
+  hideAllTiles(tiles: NodeListOf<any>, exceptions?: string[]): void {
     tiles.forEach((item) => {
       item.style.opacity = 0;
```
denysbutenko, line 285: why opacity value has string type in `hideTile()` and number in `hideAllTiles()`?

denysbutenko (approved):
👍🏻

src/app/core/classes/hardware/vdev-labels-svg.ts
undsoft: Because it's incorrect in `hideAllTitles` because NodeListOf<any> is not fully typed :)

PR #5454, opened 2021-05-14
NAS-110595 / 21.06 / Have some text/placeholder when icon for an application cannot be retrieved

AlexKarpov98:


RehanY147 (approved):
Text overflow for `latest_version` field. See example
Screenshot 2021-05-17 145617

AlexKarpov98:
> Text overflow for `latest_version` field. See example
> Screenshot 2021-05-17 145617

I can't reproduce it on my end, and it should show like this.
image

Also that issue is not related with this ticket.
Can you please confirm once again and create a new ticket if you can still see the issue?
Thank you.

PR #5459, opened 2021-05-17
NAS-110670 / 21.06 / Move SMB form to side panel

denysbutenko:
Work in progress


undsoft:
I know this is a draft, but I couldn't help myself and tested it a little bit :)
It looks good, the issues that I found were: 
- Cancel button doesn't work (think we can remove it)
- List of shares is not reloaded after a share is added.

denysbutenko:
Thanks for the feedback! It should be fixed now

undsoft (approved):
Looks good, thanks! 
Снимок экрана 2021-05-18 в 16 29 34

PR #5466, opened 2021-05-18
NAS-110590 / 21.06 / show default entry of "charts" for "Preferred Trains" attribute when creating a catalog as otherwise it's not intuitive if any option has been selected

AlexKarpov98:


undsoft (approved):
It works.
Снимок экрана 2021-05-19 в 09 04 19

PR #5468, opened 2021-05-19
NAS-110355 / 21.06 / NAS-110355: Fixes link on CPU widget in dashboard

undsoft:
To test:
- Cpu widget on dahsboard works
- Link to report on cpu widget actually goes to reports.


denysbutenko (approved):
image
Works fine!

PR #5473, opened 2021-05-20
NAS-110415 / 21.06 / NAS-110415: Removing Automatic Acoustic Management

undsoft:
Testing: 
- Go to Disks (Storage -> Cog) and test that edit form doesn't have Acoustic Management and works correctly.
- Select multiple disks and check that Acoustic Management is not there and form works.


dszidi (approved):
Tested functionality as per PR description. All seems to work as intended.

image

PR #5477, opened 2021-05-20
NAS-110459 / 21.06 / NAS-110459: Fixing DNS Authenticator form in some scenarios

undsoft:
How to test:
Credentials -> Certificates -> ACME DNS-Authenticators
When cloudflare is selected, either 'email' + 'API key' or 'API token' need to be added, but not both.
There was a bug when user touched email field, but then cleaned it, and actually filled in API token.
Plus, would be good to do general regression testing on the form.

### DIFF
--- a/src/app/pages/credentials/certificates-dash/forms/certificate-acme-add.component.ts
+++ b/src/app/pages/credentials/certificates-dash/forms/certificate-acme-add.component.ts
@@ -139,7 +139,7 @@ export class CertificateAcmeAddComponent implements FormConfiguration {
   }
 
   preInit(entityForm: EntityFormComponent): void {
-    this.ws.call('acme.dns.authenticator.query').subscribe((authenticators: any[]) => {
+    this.ws.call('acme.dns.authenticator.query').subscribe((authenticators) => {
       this.dns_map = _.find(this.fieldSets[2].config[0].templateListField, { name: 'authenticators' });
       authenticators.forEach((item) => {
         this.dns_map.options.push({ label: item.name, value: item.id });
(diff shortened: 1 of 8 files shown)


src/app/pages/credentials/certificates-dash/forms/certificate-acme-add.component.ts
```diff
@@ -139,7 +139,7 @@ export class CertificateAcmeAddComponent implements FormConfiguration {
   }
 
   preInit(entityForm: EntityFormComponent): void {
-    this.ws.call('acme.dns.authenticator.query').subscribe((authenticators: any[]) => {
+    this.ws.call('acme.dns.authenticator.query').subscribe((authenticators) => {
```
RehanY147, line 142: Why was the type removed?

RehanY147: I get that it's redundant but specific `any[]` is better than `any` I think

undsoft: When we specify types in `api-directory.interface.ts`, it will correctly pick up types in places where this API call is used.
In this case, I've typed `acme.dns.authenticator.query`, so `any[]` is no longer necessary.

RehanY147: Yep. My bad. Forgot about that.

RehanY147 (approved):
Tested form submission.
Typing something in a field and then erasing it fully causes validation issues in master that are resolved here

PR #5478, opened 2021-05-20
NAS-104984 / 21.06 / Data Protection improvements

denysbutenko:
Changes:

- Added `Frequency` and `When` columns on the `Periodic Snapshot Tasks` page
- Replace `Recursive` column with `Frequency` for `Periodic Snapshot Tasks` card on the Dashboard
- Added `Frequency` column (human-readable cron) on `Data Protection Dashboard` for `Scrub Tasks`, `Cloud Sync Tasks`, `Rsync Tasks`, `S.M.A.R.T. Tests`
- Added interfaces for each card
- Added enums
- Improved i18n
- Fixed deprecated warnings (dialog.confirm)


src/app/interfaces/periodic-snapshot-task.interface.ts
```diff
@@ -1,7 +1,12 @@
 import { LifetimeUnit } from 'app/enums/lifetime-unit.enum';
 import { Schedule } from 'app/interfaces/schedule.interface';
 
-export interface PeriodicSnapshotTask {
+interface PeriodicSnapshotTaskState {
+  state: string;
+  error: string;
+}
+
+export interface PeriodicSnapshotTaskRaw {
```
undsoft, line 9: What does PeriodicSnapshotTaskRaw do? Doesn't look like it's used anywhere else?

src/app/interfaces/cloud-sync-task.interface.ts
```diff
@@ -0,0 +1,52 @@
...
+  provider: string;
+  attributes: { [key: string]: string };
+}
+
+export interface BwLimit {
+  time: string;
+  bandwidth: number;
+}
+
+export interface CloudSyncTask {
```
undsoft, line 19: If any of the new interfaces are coming from API, you can update api-directory file.

undsoft (commented):
Can we update frequency text to be shorter? Doesn't fit on some displays.
I.e. "At 00 minutes past every hour, every hour, every day" to "At 00 every hour, every day"

Снимок экрана 2021-05-21 в 15 51 32

denysbutenko:
> Can we update frequency text to be shorter? Doesn't fit on some displays.
> I.e. "At 00 minutes past every hour, every hour, every day" to "At 00 every hour, every day"
> 
> Снимок экрана 2021-05-21 в 15 51 32

Since we use a third-party library to parse cron schedules, there are two options for output `verbose: true or false`. 
When **verbose: true**:
`At 12:00 AM, every day`
`At 12:00 AM, only on Sunday`
When **verbose: false**:
`At 12:00 AM`
`At 12:00 AM, only on Sunday`
It also has a `title` attribute that shows the full text on mouseover.

undsoft (approved):
Looks good, thank you.

PR #5479, opened 2021-05-20
NAS-110628 / None / NAS-110628 - Adding return types 14 - final 

undsoft:


src/app/pages/data-protection/replication/replication-wizard/replication-wizard.component.ts
```diff
@@ -1027,7 +1027,7 @@ export class ReplicationWizardComponent {
     }
   }
 
-  parsePickerTime(picker: string) {
+  parsePickerTime(picker: string): { minute: string; hour: string; dom: string; month: string; dow: string } {
```
RehanY147, line 1030: This could be a localized interface at this point

src/app/pages/reportsdashboard/reports-utils.worker.ts
```diff
@@ -53,7 +53,7 @@ function inferUnits(label: string) {
   return units;
 }
 
-function convertKMGT(input: number, units: string, fixed?: number) {
+function convertKMGT(input: number, units: string, fixed?: number): { value: number; prefix: string; shortName: string } {
```
RehanY147, line 56: Could be a localized interface as it's being reused too

src/app/pages/storage/disks/disk-list/disk-list.component.ts
```diff
@@ -127,7 +128,7 @@ export class DiskListComponent {
     this.ws.call('smart.test.disk_choices').subscribe((res) => this.SMARTdiskChoices = res, (err) => new EntityUtils().handleWSError(this, err));
   }
 
-  getActions(parentRow: any) {
+  getActions(parentRow: any): EntityTableAction[] {
     const actions = [{
```
RehanY147, line 132: Wouldn't it be better to assign the type here in the first place instead of type-casting on return?

undsoft: It wouldn't work, declaring actions as EntityTableAction[] will require all fields within EntityTableAction to be implemented. Casting will silence this issue. 
The underlying problem is that EntityTableAction is not correctly defined, but this is outside of the scope of this PR.

src/app/pages/vm/devices/device-list/device-list.component.ts
```diff
@@ -64,7 +65,7 @@ export class DeviceListComponent {
     return !(actionId === 'delete' && row.id === true);
   }
 
-  getActions(row: any) {
+  getActions(row: any): EntityTableAction[] {
     const self = this;
     const actions = [];
```
RehanY147, line 70: Type should be assigned here rather than type casting on return

src/app/pages/reportsdashboard/reports-utils.worker.ts
undsoft: Could be, but I am not too familiar with the code here to give it a proper name. Will do another time.

src/app/pages/data-protection/replication/replication-wizard/replication-wizard.component.ts
undsoft: Fixed.

RehanY147 (approved):
Concerns removed. LGTM.

PR #5481, opened 2021-05-21
NAS-110735 / 21.06 / Fix advanced settings

denysbutenko:
Fix "this.parent.doAdd is undefined"


undsoft (approved):
Oh, shi. Thanks, it works.

PR #5482, opened 2021-05-21
Docs 1981 SCALE Root SSH Test

q5sys:


tests/bdd/conftest.py
```diff
@@ -76,8 +76,8 @@ def driver():
 
 
 # Close firefox after all tests are completed
-def pytest_sessionfinish(session, exitstatus):
-    web_driver.quit()
+# def pytest_sessionfinish(session, exitstatus):
```
ericbsd, line 79: Please uncomment lines 79 and 80 to ensure that firefox gets closed after tests are done on bob.

ericbsd (changes requested):

tests/bdd/conftest.py
q5sys: good catch, I forgot to do that.  Just pushed a new commit to uncomment those.

q5sys: wait, sorry this is a different file. I'll fix that in a second

ericbsd (approved):
LGTM

PR #5484, opened 2021-05-21
UI allows opening launch docker image wizard even if pool is not selected

AlexKarpov98:


undsoft:
Launch wizard cannot be opened anymore, which is good, but we probably need to do the same treatment for "Install" buttons next to apps.

undsoft (approved):
Looks good, thank you.

Снимок экрана 2021-05-25 в 14 05 26

PR #5485, opened 2021-05-22
NAS-110736 / 21.06 / Auto-generate MAC address

undsoft:
To test:
- When adding a VM, GPU is no longer a required field
- When adding a NIC device in your VM, MAC address will be autogenerated.


denysbutenko (approved):
The changes look are good! Confirm that GPU is not required anymore.
image

PR #5490, opened 2021-05-24
NAS-110585 / 21.06 / NAS-110585: Race condition when clicking on Launch Docker wizard.

undsoft:
Previous issue was reproducible if user had slow internet connection (can be emulated in Chrome Inspector), went to Apps and quickly pressed Launch Docker Image.


undsoft:
@AlexKarpov98 Not sure if the "fix" I did is the best solution. Please review.

denysbutenko (approved):
Works as described.

https://user-images.githubusercontent.com/351613/119856105-24f52580-bf1b-11eb-8f55-c45235f20587.mp4

PR #5496, opened 2021-05-25
NAS-110778 / 21.06 / Downgrade pixi to 4.7 and remove pixi-projection dependency

dszidi:
This should eliminate any errors complaining about pixi types being out of sync between pixi.js and pixi-projection.


undsoft (approved):
The error appears to be gone.

PR #5498, opened 2021-05-25
NAS-110733 / 21.06 / Shares Dashboard Improvements

RehanY147:
NAS-110733

Improvements to look out for
- In case none of the tables are populated, you should see an empty entity state
- If you change the size of the screen, table rows should reduce/increase accordingly to fit the max rows possible
- Adding/Editing/Deleting items updates the tables right away


src/app/pages/common/entity/table/expandable-table/expandable-table.component.ts
```diff
@@ -22,10 +25,25 @@ export class ExpandableTableComponent {
   disabled = false;
   isEmpty = true;
   isExpanded = false;
+  _tableConf: InputExpandableTableConf;
```
undsoft, line 28: 1. Underscores [violate our code style.](https://github.com/airbnb/javascript#naming--leading-underscore)
2. The underlying variable is still public. It looks like you could benefit from using ngOnChanges more than a setter.

src/app/pages/common/entity/table/expandable-table/expandable-table.component.ts
```diff
@@ -50,11 +68,37 @@ export class ExpandableTableComponent {
     };
   }
 
+  ngAfterViewChecked(): void {
+    if (this._tableConf.limitRows
+      && this.appTable.nativeElement.children[0].children[1].children[0].children[0].children[0].children[0].children[0].children[0]) {
+      const tableHeaderHeight = this.appTable.nativeElement.children[0].children[1].children[0].children[0].children[0].children[0].children[0].children[0].children[0].offsetHeight;
```
undsoft, line 74: source-2

undsoft: Okay,
1. This is hard to read.
2. This will break if DOM structure changes.

Can you at least do something like querySelector()? 
I will also be okay with this being a separate directive.

src/app/pages/sharing/iscsi/target/target-form/target-form.component.ts
```diff
@@ -222,12 +228,12 @@ export class TargetFormComponent implements FormConfiguration {
...
-        this.pk = params['pk'];
-        this.customFilter[0][0].push(parseInt(params['pk'], 10));
-      }
-    });
+    // this.aroute.params.subscribe((params) => {
+    //   if (params['pk']) {
+    //     this.pk = params['pk'];
+    //     this.customFilter[0][0].push(parseInt(params['pk'], 10));
+    //   }
+    // });
```
undsoft, line 236: If you don't need it, it needs to be removed.

src/app/pages/common/entity/table/expandable-table/expandable-table.component.ts
```diff
@@ -50,11 +68,37 @@ export class ExpandableTableComponent {
...
+      const expandableHeaderHeight = this.appTable.nativeElement.children[0].children[0].offsetHeight;
+      const detailsFooterHeight = this.appTable.nativeElement.children[1].offsetHeight;
+      const totalHeight = this.appTable.nativeElement.offsetHeight;
+      const maxRowsHeight = totalHeight - expandableHeaderHeight - tableHeaderHeight - detailsFooterHeight;
+      if (this._tableConf.limitRowsByMaxHeight) {
+        const prevRowsLimit = this._tableConf.limitRows;
+        this._tableConf.limitRows = Math.floor(maxRowsHeight / 48);
```
undsoft, line 81: 48 needs to be a variable.

undsoft (changes requested):
1. This is not the empty page I had in mind. I was thinking about just showing 4 rows horizontally as we've discussed, to sort of train user to our current layout. 
2. When I added my first share, I got an empty modal and had to reload a page.
3. I get "ERROR TypeError: Cannot read property 'offsetHeight' of undefined" when I am looking at the page with 1 share. I cannot add or delete any shares anymore.

RehanY147:
>     1. This is not the empty page I had in mind. I was thinking about just showing 4 rows horizontally as we've discussed, to sort of train user to our current layout.

I went with the designs that had been approved for the empty state. But I get your point. I'll try and implement that.

src/app/pages/common/entity/table/expandable-table/expandable-table.component.ts
```diff
@@ -39,22 +52,36 @@ export class ExpandableTableComponent {
...
         return data.splice(0, this.tableConf.limitRows);
       }
       return data;
     };
   }
 
-  shouldBeCollapsed(): boolean {
-    return this.tableConf.alwaysCollapsed || (this.isEmpty && this.tableConf.collapsedIfEmpty);
+  ngAfterViewChecked(): void {
+    if (this.isExpanded) {
```
undsoft, line 64: Please use return early to avoid nesting.
I.e. 
```
if (!this.isExpanded) {
  return;
}
...
everything else
```

src/app/pages/common/entity/table/expandable-table/expandable-table.component.ts
```diff
@@ -39,22 +52,36 @@ export class ExpandableTableComponent {
...
+      const detailsRow = this.appTable.nativeElement.querySelector('#actions-row');
+      const expandableHeader = this.appTable.nativeElement.querySelector('mat-expansion-panel-header');
+      const tableHeaderHeight = tableHeader ? tableHeader.offsetHeight : 0;
+      const expandableHeaderHeight = expandableHeader ? expandableHeader.offsetHeight : 0;
+      const detailsFooterHeight = detailsRow ? detailsRow.offsetHeight : 0;
+      const totalHeight = this.appTable.nativeElement.offsetHeight;
+      const maxRowsHeight = totalHeight - expandableHeaderHeight - tableHeaderHeight - detailsFooterHeight;
+      if (this.tableConf.limitRowsByMaxHeight) {
+        const prevRowsLimit = this.tableConf.limitRows;
+        const TABLE_ROW_SIZE = 48;
```
undsoft, line 75: All caps doesn't match [the code style](https://github.com/airbnb/javascript#naming--uppercase) in this case.

undsoft (approved):
Functionality-wise I think it looks pretty good.
But to keep it on the safe side, let's keep it as a secondary menu item for this release.
Снимок экрана 2021-05-27 в 16 03 47

PR #5502, opened 2021-05-26
NAS-110200 / 21.06 / Add checkbox support for app-table

denysbutenko:
Changes:
- support multiple checkboxes for the same row
- columns with `checkbox: true` shows as `<mat-checkbox>`
- when `onChange()` method is missing checkbox will be read-only (disabled state)

To test:

- Go to `Data Protection Dashboard` and try to play with checkboxes
- Go to `Shares Dashboard` and try to play with checkboxes


undsoft (approved):
Looks good 👍🏼

PR #5517, opened 2021-05-28
NAS-110582 / 21.08 / Properly show summary of selected options in application's wizard

AlexKarpov98:


src/app/pages/common/entity/entity-wizard/entity-wizard.component.ts
```diff
@@ -250,41 +250,107 @@ export class EntityWizardComponent implements OnInit {
...
+      if (fieldConfig.type === 'select') {
+        const selectedOption = fieldConfig.options.find((option) => option.value == formControl.value);
+        if (selectedOption) {
+          value = selectedOption.label;
+        }
+      }
+    }
+
+    if (value) {
+      result = { [key]: value };
```
undsoft, line 353: How about making a separate component to show summary, instead of forming html in a method?

Ornias1993:
@AlexKarpov98 If I read this correctly this should fix the [object Object]  references in the summary, correct?

src/app/pages/common/entity/entity-wizard/entity-wizard.component.ts
AlexKarpov98: Fixed

AlexKarpov98:
> @AlexKarpov98 If I read this correctly this should fix the [object Object] references in the summary, correct?

Exactly

src/app/pages/common/entity/entity-wizard/components/wizard-summary/wizard-summary.component.html
```diff
@@ -0,0 +1,16 @@
+<ul [ngClass]="{'wizard-ul':!isRoot}">
```
undsoft, line 1: Prefer semantic names:
```
[class.wizard-list]="!isRoot"
```

src/app/pages/common/entity/entity-wizard/components/wizard-summary/wizard-summary.component.html
```diff
@@ -0,0 +1,16 @@
...
+  <li *ngFor="let item of data | keyvalue : originalOrder">
+    <div *ngIf="isArray(item.value)==true; else notArray">
```
undsoft, line 3: == true is unnecessary

src/app/pages/common/entity/entity-wizard/components/wizard-summary/wizard-summary.component.ts
```diff
@@ -0,0 +1,21 @@
...
+  @Input('data') data: any;
+  @Input('isRoot') isRoot: boolean;
+
+  isArray(value: any): boolean {
+    return Array.isArray(value);
+  }
+
+  originalOrder(): void {
+
+  }
```
undsoft, line 20: Does any empty function do anything?

src/app/pages/common/entity/entity-wizard/entity-wizard.component.ts
undsoft: Not exactly what I had in mind.
I was thinking about making a component, whose solely responsibility would be showing summary. And then taking all the code here (`updateSummary`, `getSummaryValue`) and moving it to this new component.
Also would like to avoid forming html in components - `generateSummaryHtml`.
Just so that we can separate concerns a little bit here.

undsoft (changes requested):

AlexKarpov98:
@undsoft Fixed above issues.

undsoft (approved):
Looks good, thanks for the fixes.
Снимок экрана 2021-06-07 в 23 01 49

PR #5526, opened 2021-06-01
NAS-110075 / 21.08 / Standardize subscriptions with UntilDestroy

denysbutenko:
Changes:
- Added eslint-rxjs-plugin (enable `rxjs/no-unsafe-takeuntil` rule)
- Added eslint-rxjs-angular-plugin (enable `rxjs-angular/prefer-takeuntil` rule)
- Added until-destroy package (use `@UntilDestroy()` decorator and `untilDestroyed()` pipe)
- Update all components with subscriptions and non-singleton services
- Remove manual unsubscribing where it's safe

To test:
- Check it out for regressions


undsoft:
Thanks for all the updates.
I have a couple of questions:
1. If developer forgets class decorator this will result in nasty runtime error, so we need to somehow make sure that people do not forget to do it.
2. I've removed UntilDestroy() on one of the classes and run `yarn lint` and didn't get any errors.
3. Enabling `createDefaultProgram` and related changes are unfortunate. They will make linting time = compilation time = several minutes. Which rules use type information?

undsoft:
Since this is a large PR, it's going to attract merge conflicts. If it'll take additional time to figure out linter, you can merge the rest and fix linter in next PR.

denysbutenko:
> Thanks for all the updates.
> I have a couple of questions:
> 
> 1. If developer forgets class decorator this will result in nasty runtime error, so we need to somehow make sure that people do not forget to do it.
> 2. I've removed UntilDestroy() on one of the classes and run `yarn lint` and didn't get any errors.
> 3. Enabling `createDefaultProgram` and related changes are unfortunate. They will make linting time = compilation time = several minutes. Which rules use type information?

1. Yeah, that's a good point. We have a lot of examples for guiding.
2. Unfortunately, it is. The `@UntilDestroy()` is necessary only when `untilDestroyed(this)` is used and it can be linted. Perhaps in the future, we could avoid using the decorator - https://github.com/ngneat/until-destroy/issues/122
3. I keep `parserOptions.project` and `parserOptions.tsconfigRootDir` only, it is needed for `eslint-rxjs-plugin`. `createDefaultProgram` already enabled in the `master` branch.

undsoft (approved):
Great work, thanks!

PR #5535, opened 2021-06-03
Fix slot order on r20a

dszidi:
New slot order on front chassis should be as follows:

01 - 06 - 07 - 12
02 - 05 - 08 - 11
03 - 04 - 09 - 10


undsoft (approved):
Tested first 4 slots in UI and rest by testing generatePosition function.
Снимок экрана 2021-06-04 в 09 45 49

PR #5536, opened 2021-06-03
NAS-110860 / 12.0 / Tweak settings for new temperature_celsius stat

dszidi:
A new temperature_celsius stat was added to reporting.realtime and it was breaking the cpu widget.
To test this PR, please make sure the cpu widget on dashboard does not throw console errors in the following scenarios.

- CPUs with hyperthreading
- CPUs without hyperthreading
- Machines that don't report CPU temperatures (VMs)


undsoft (approved):
Looks good.
Снимок экрана 2021-06-04 в 10 07 21

undsoft:
Forwardport?

dszidi:
> Forwardport?

I don't think this should affect scale. In stable branch the UI uses keys to figure out core count. In scale we should be using the new `physical_cores` from `system.info` endpoint to help with that.

PR #5539, opened 2021-06-04
NAS-110836 / 21.08 / Working towards removing momentjs

RehanY147:
NAS-110836

Should be tested
- FormSchedulerComponent
- Time zone irregularities throughout the system


src/app/pages/storage/volumes/volumes-list/volumes-list.component.ts
```diff
@@ -1701,7 +1701,7 @@ export class VolumesListTableConfig {
 
   getTimestamp(): string {
     const dateTime = new Date();
-    return moment(dateTime).format('YYYY-MM-DD_HH-mm');
+    return format(dateTime, 'YYYY-MM-DD_HH-mm');
```
undsoft, line 1704: This line doesn't work locally. 
`ERROR RangeError: Use `yyyy` instead of `YYYY` for formatting years`

src/app/services/locale.service.ts
```diff
@@ -1,12 +1,12 @@
 import { Injectable } from '@angular/core';
```
undsoft, line 1: Error when going to /system/general

undsoft (changes requested):

src/app/services/locale.service.ts
```diff
@@ -1,12 +1,12 @@
...
 import { CoreEvent } from 'app/interfaces/events';
 import { Option } from 'app/interfaces/option.interface';
```
undsoft, line 3: 1. Selecting time format options with 'am' in settings does not format time correctly (for example in Task Manager).
2. Selecting 'am' options, saving and then reopening localization form will not populate Time Format with current value.

undsoft (changes requested):

PR #5540, opened 2021-06-04
NAS-110837 / 21.08 / Organize imports and enable linter rules

denysbutenko:
Changes:
* Most of the relative paths was replaced with absolute one
* Enabled eslint rules to sort and group imports
* Apply rules to the entire codebase (auto-generated)


undsoft (approved):
Nice work!
My only concern is that it seems inconsistent when new lines are inserted within groups.
For example, here `rxjs`, `rxjs/operators`, `moment` and `angular` are separate.

Снимок экрана 2021-06-06 в 11 26 34

And in here `q`, `lodash` and `rxjs` are together.
Снимок экрана 2021-06-06 в 11 28 14

Maybe we should just not have new lines within groups.

PR #5542, opened 2021-06-04
NAS-110583 / 21.08 / Render ix-chart dynamically in the UI

AlexKarpov98:


denysbutenko:
How to test this change?

AlexKarpov98:
You can add a new catalog to test this - URI to use will be "https://github.com/truenas/charts" and branch would be "test-ix-chart" and the application name would be "test-ix-chart" as well.

AlexKarpov98:
> You can add a new catalog to test this - URI to use will be "https://github.com/truenas/charts" and branch would be "test-ix-chart" and the application name would be "test-ix-chart" as well.

So once you followed this steps, you should see `ix-chart` application like below.
Then you can test if you can successfully install it.
image

RehanY147 (approved):
LGTM. Chart is loading.

sonicaj:
@AlexKarpov98 we are not rendering the wizard being shown by "Launch Docker Image" dynamically - it is still static. I am not sure i follow the change ?

PR #5543, opened 2021-06-05
NAS-110919 / 21.08 / Fix broken storage page

dszidi:


undsoft (approved):
Thanks for jumping on this one, Damian.
This is probably not the right fix, as `VolumesListTableConfig` is not an Angular component. I'd imagine that there is no destroy hook for `UntilDestroy` to latch on.
But we can merge this PR to unblock developers and I believe @denysbutenko is working on a proper fix.

RehanY147 (approved):
LGTM. Storage is working now.

PR #5544, opened 2021-06-07
NAS-109165 / 21.08 / Consistent spacing of topbar icons.

dszidi:
Not all the topbar icons had the same class attribute value so margins were not consistently applied.

To test simply inspect each icon in the right side of the topbar component and make sure they all have the same margins applied.


denysbutenko (approved):
Looks good
image

PR #5549, opened 2021-06-07
Encryption options are not shown/set correctly after canceling "Add dataset" with passphrase encryption

AlexKarpov98:


RehanY147 (approved):
Issue is resolved

PR #5557, opened 2021-06-09
NAS-110981 / None / Fix text alignment on volume list

denysbutenko:
The problem is entity-form adds global styles when opening a wizard. It affects `.mat-content` class in the volume list so that styles should be isolated into components.

How to test?
- Open the Storage page, click the Import button, and then check the volume header it shouldn't be sticky to the top.
- Please check the video in JIRA ticket


undsoft (approved):
Well, the issue doesn't happen, but I am confused as to why `:host` is needed. Aren't all styles encapsulated to component scope by default?
Also, you can target `master` - this is something we can include in next release.

dszidi:
I'm still seeing the behavior

denysbutenko:
Closes due to https://github.com/truenas/webui/pull/5562

PR #5560, opened 2021-06-09
NAS-110905 / None / Docker Wizard has a black overlay with no actionable items after clicking Next from Ports Screen

AlexKarpov98:


undsoft (changes requested):
It works, but this is a bit hardcore.

If the goal is to hide collapsed steps content, how about instead updating entity-wizard and using (selectionChange) to put an `active` class on `mat-step`?
Also if you need to pierce view encapsulation, it's better to move specific style to `src/assets/styles` or even use ::ng-deep. `ViewEncapsulation.None` will just make all styles in the file global.

undsoft (approved):
Thanks!

PR #5562, opened 2021-06-10
NAS-110981 / 21.08 / Fix text alignment on volume list

denysbutenko:
[Original PR](https://github.com/truenas/webui/pull/5557) was based on `release/21.06-BETA.1` branch. This one is on top of `master` branch.

> The problem is entity-form adds global styles when opening a wizard. It affects .mat-content class in the volume list so that styles should be isolated into components.
> 
> How to test?
> * Open the Storage page, click the Import button, and then check the volume header it shouldn't be sticky to the top.
> * Please check the video in JIRA ticket


undsoft (approved):
The pool header bar doesn't change height anymore.

PR #5564, opened 2021-06-10
NAS-110785 / 12.0 / Inherit option overriding encryption on dataset creation

AlexKarpov98:


undsoft (changes requested):
Something isn't right.
Start adding a dataset in an encrypted dataset.
1. Tick and untick Encryption. A disabled form appears. Is it supposed to be there?
2. Or untick Inherit and then untick Encryption. The encryption form is still visible.

AlexKarpov98:
> Something isn't right.
> Start adding a dataset in an encrypted dataset.
> 
> 1. Tick and untick Encryption. A disabled form appears. Is it supposed to be there?
> 2. Or untick Inherit and then untick Encryption. The encryption form is still visible.

Fixed

undsoft:
Trying to add dataset under non-encrypted dataset.
If I until both checkboxes, I see the form.
BTW, these are my assumptions about how this should work. It may be worth checking with release engineering on how they expect this part of the system to work.

Снимок экрана 2021-06-15 в 10 24 25

AlexKarpov98:
Fixed

undsoft (approved):
It's better, but there is still a confusing scenario.
Have non-encrypted parent dataset and start creating a child dataset. The checkbox will say encrypted, even though next one will say "Inherit (non-encrypted)" and dataset will not be encrypted.
I've actually looked into our docs and there it's more correct: https://www.truenas.com/docs/core/storage/pools/storageencryption/#encrypting-a-new-dataset
The first checkbox is "inherit" and the second checkbox is conditional and will enable / disable encryption and will show / hide encryption options.
Let's change to this approach.

undsoft (changes requested):
Accidentally approved.

AlexKarpov98:
@undsoft Fixed

undsoft (approved):
Looks good now, thanks.

PR #5566, opened 2021-06-11
NAS-110907 / 21.06-BETA.1 / Small fixes for Permission Editor

denysbutenko:
Changes:
- Fix empty field for dataset path for ACL, POSIX ACL, Permission Editor forms
- Added types for updateCall
- Restrict subscriptions to take first result and unsubscribe while filtering in combobox
- Add debounce for keyup event for combobox
- Fix empty "User" and "Group" fields in combobox for ACL items

Testing
Open "Storage" page and choose "Edit Permissions" on dataset then switch to "ACL Manager" and make sure:
- Change "User" or "Group" and save it then refresh page and check if it's still valid
- Add a new ACL item with tag "Mask" and save it then refresh page and check if "Who" field is populated with "Mask"
- Add a new ACL item with tag "User", choose any user for "User" field save it then refresh page and check if "User" field is populated with username.
- Add a new ACL item with tag "Group", choose any group for "Group" field save it then refresh page and check if "Group" field is populated with groupname.
- "Dataset path" field shouldn't be empty


undsoft (changes requested):
The part about being able to change the group works, but I was still able to get empty Who fields.
I added a new dataset, then added new ace with Mask type. Doing so completely messed up the list, not only the new ace appears incorrectly, but it corrupted how another ace looks.

anodos325:
I saw an error in backend "filesystem.getacl" output for case where MASK is present. NAS-111038. 1-line diff should be pretty simple to hot-patch. I will backport to 21.06 shortly.

denysbutenko:
> The part about being able to change the group works, but I was still able to get empty Who fields.
> I added a new dataset, then added new ace with Mask type. Doing so completely messed up the list, not only the new ace appears incorrectly, but it corrupted how another ace looks.

that issue has been fixed by https://github.com/truenas/middleware/pull/7032

undsoft (approved):
Seems to be working correctly. Tested some scenarios of adding/removing aces.
Found what appears to be a backend issue: https://jira.ixsystems.com/browse/NAS-111050

PR #5567, opened 2021-06-11
NAS-111009 / 21.06-BETA.1 / Enable aclmode in SCALE

denysbutenko:


undsoft (approved):
It generally works, but there is a nuisance. I've created a new pool and I'm adding a new dataset. If I just submit the form, it'll complain:
```
Must be set to DISCARD when acltype is POSIX or OFF
```

@anodos325 Do you think we can default to DISCARD in all cases?

anodos325:
> It generally works, but there is a nuisance. I've created a new pool and I'm adding a new dataset. If I just submit the form, it'll complain:
> 
> ```
> Must be set to DISCARD when acltype is POSIX or OFF
> ```
> 
> @anodos325 Do you think we can default to DISCARD in all cases?

No.

undsoft:
Added a ticket to fix defaults https://jira.ixsystems.com/browse/NAS-111036, will merge this PR.

PR #5568, opened 2021-06-11
NAS-111005 / 21.06-BETA.1 / Add AFP to advanced options

dszidi:


src/app/helptext/sharing/smb/smb.ts
```diff
@@ -44,6 +44,12 @@ export const helptext_sharing_smb = {
   placeholder_timemachine: T('Time Machine'),
   tooltip_timemachine: T('Enable Time Machine backups on this share.'),
 
+  placeholder_afp: T('AFP'),
```
william-gr, line 47: @anodos325 @Mrt134 what do you think about "Legacy AFP", or "AFP (Legacy)", just to put more emphasis into it?

Mrt134: Only problem that comes to mind with that is it looks a little odd in comparison with TrueNAS CORE, but that's probably not a big deal. What about "Legacy AFP Behavior", Legacy AFP Compatability", or just "AFP Compatability" ?

william-gr: I would vote for Legacy AFP Compatibility, if I will let @anodos325 pick since he is the expert, and if he agrees with a change here.

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -180,6 +180,13 @@ export class SMBFormComponent implements FormConfiguration, OnDestroy {
           tooltip: helptext_sharing_smb.tooltip_timemachine,
           isHidden: true,
         },
+        {
+          type: 'checkbox',
+          name: 'afp',
+          placeholder: helptext_sharing_smb.placeholder_afp,
+          tooltip: helptext_sharing_smb.tooltip_afp,
+          isHidden: true,
```
denysbutenko, line 188: If it is hidden by default. What is the condition to show?

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -596,6 +603,18 @@ export class SMBFormComponent implements FormConfiguration, OnDestroy {
         }
       }
     });
+
+    entityForm.formGroup.controls['afp'].valueChanges.pipe(debounceTime(100)).subscribe((res) => {
+      if (res) {
```
denysbutenko, line 608: Let's imagine a case when it is already enabled then user toggles the checkbox (unchecked state). Should we display the dialog?

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
william-gr: We probably should.

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
william-gr: We probably should

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
dszidi: Line 333. UpdateForm method toggles hidden based on Basic vs Advanced mode
This is why the entire 'other' fieldset has every field's isHidden set the same way.

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
dszidi: What if the user checks the box, gets the warning. Decides against it and unchecks it. They should get the warning again?

denysbutenko: I'd imagine something like:
1. initially **Unchecked** -> Click -> Show the warning -> **Confirm** - `checked` or **Cancel** - `unchecked`
2. initially **Checked** -> Click -> Show the warning -> **Confirm** - `unchecked` or **Cancel** - `checked`

william-gr: Agreed with Denys

src/app/helptext/sharing/smb/smb.ts
dszidi: @william-gr Any final decision on the input label text?

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -598,6 +606,22 @@ export class SMBFormComponent implements FormConfiguration, OnDestroy {
     });
   }
 
+  afpConfirm(): void {
+    const afpControl: any = this.entityForm.formGroup.controls['afp'];
```
undsoft, line 610: You can remove `:any`.

undsoft (commented):
Let's go with _Legacy AFP Compatibility_.

We should also update warning message to incorporate Andrew's feedback and make it more clear what may get corrupted:
> ... pure SMB share or corruption of extended attributes on file can occur which may lead to undefined behavior from applications using the file depending on how the extended attributes were used.

dszidi:
>... We should also update warning message to incorporate Andrew's feedback and make it more clear what may get corrupted:
> 
> > ... pure SMB share or corruption of extended attributes on file can occur which may lead to undefined behavior from applications using the file depending on how the extended attributes were used.

The PR currently contains Docs team's suggestion that was posted in the ticket comments (which was an embellishment of Andrew's text from the ticket description). Do you still want me to revert it to the text in the ticket description?

dszidi:
Adjusted tooltips so they don't get clipped by parent.

image

undsoft (approved):
Looks good. Nice work!

PR #5569, opened 2021-06-11
NAS-110739 / 12.0 / Don't try to upload an iso file for a vm if one hasn't been selected

AlexKarpov98:


undsoft (changes requested):
1. I would imagine that updating a root component like this will make all file inputs in all forms that use form-upload required. I don't think this is always the case.
It should probably be based on a 'required' field in config.
Ideally it would show an inline validation message similar to how other inputs work, but this can be skipped if it's too hard.

2. Also, please update the message to "Please make sure to select a file".

AlexKarpov98:
> 1. I would imagine that updating a root component like this will make all file inputs in all forms that use form-upload required. I don't think this is always the case.
>    It should probably be based on a 'required' field in config.
>    Ideally it would show an inline validation message similar to how other inputs work, but this can be skipped if it's too hard.
> 2. Also, please update the message to "Please make sure to select a file".

Regarding 1st opinion, I don't think it's the right way. I think that workflow is for submitting scope, but we are currently talking about only an uploading section. You know the terms of `required` stand for submit, nothing else. We are talking about the workflow of the upload button itself. Why do we need to follow upload processing even though there is no file selected? I am sure you confused something. Can you please consider this once again? Thank you.

undsoft (approved):
You are right. I made an incorrect assumption about how this form control works.
Thanks.
Do you think we need to port this to master?

AlexKarpov98:
@undsoft Yeah, I think we need to port this to master.

undsoft:
backport

undsoft:
backport

PR #5570, opened 2021-06-11
NAS-110927 / 12.0 / No UI option to change the readonly state of a ZVOL.

AlexKarpov98:


undsoft (changes requested):
When comparing to datasets, datasets also support "inherit" option for read only field.
Looks like readonly on zvols can also be set to inherit, so we need to support this value.
Example from table:
Снимок экрана 2021-06-14 в 16 26 11

AlexKarpov98:
> When comparing to datasets, datasets also support "inherit" option for read only field.
> Looks like readonly on zvols can also be set to inherit, so we need to support this value.
> Example from table:
> Снимок экрана 2021-06-14 в 16 26 11

Fixed

undsoft (approved):
Looks good, thank you. 
Снимок экрана 2021-06-16 в 12 22 54

undsoft:
@AlexKarpov98 
Can you check if need to do the same for Scale?

AlexKarpov98:
@undsoft Yeah, we need to add it on Scale as well.

undsoft:
Don't know if this can be backported, but if it can, you can add 'backport-21.08' label and add 'backport' comment.

AlexKarpov98:
I don't think it can't be backported, it needs to be updated manually.

PR #5576, opened 2021-06-14
NAS-110998 / 21.08 / Fixed issue with ACL Manager controls not loading properly

RehanY147:
NAS-110998


src/app/pages/common/entity/entity-form/components/form-list/form-list.component.ts
```diff
@@ -21,12 +21,13 @@ export class FormListComponent implements Field, OnInit {
...
 
   ngOnInit(): void {
-    setTimeout(() => {
-      this.listsFromArray = this.group.controls[this.config.name] as FormArray;
-      if (this.config.addInitialList && this.listsFromArray.length === 0) {
-        this.add();
-      }
-    }, 0);
+    this.listsFromArray = this.group.get(this.config.name) as FormArray;
+    if (this.listsFromArray.constructor.name === 'Object' || !this.listsFromArray.controls.length) {
```
undsoft, line 25: Can you explain what the issue was and what's going on here with checking for 'Object'?

RehanY147: The `FormArray.controls` property is supposed to be an array. That's how Angular's code treats it in the `FormArray` class. But, when we fetch the `FormArray` here, the `FormArray.controls` property is an object for some reason. I think it's because of how our `entity-foo` components are setup and how the component factory builds components in our project. But, if `FormArray.controls` is an object, that means you can't push new FormGroups (with n number of form controls as per the config) to it without having to worry about naming each one. And FormArray is supposed to have an an array for `FormArray.controls` property or it might have trouble functioning properly. That's why here, I check that if for whatever reason, it is an object, I set it to an empty array to fix that issue. That's how `FormArray` class expects it and that's how it should be.
Without it, you get the error `this.controls.push is not a function` in the `FormArray.push()` method because there, this.controls is now an object instead of an array and `push` method doesn't mean anything for an object.
Let me know if there are more questions.

RehanY147: In `FormListComponent`, `FormArray.controls` is actually and correctly an array. But not when the same `FormGroup` is fetched in teh acl manager components. So, we can't just choose to treat it like an object in those components because it's supposed to be an array and that's how it is in the `FormListComponent`

undsoft: Hm, bizarre. Perhaps, it's a `FormGroup`, not a `FormArray` and it was incorrectly casted. Can you check this in a debugger by comparing properties?
As for check for, can we use `!Array.isArray(this.listsFromArray.controls)`?

RehanY147: > 
> 
> Hm, bizarre. Perhaps, it's a `FormGroup`, not a `FormArray` and it was incorrectly casted. Can you check this in a debugger by comparing properties?

`FormArray` is what is typically used when building dynamic lists like this are built. So, that's the assumption that I worked off of. In the original component, `FormListComponent`, `FormArray.controls` is an array and the code written by the developer also assumes that it is an array. How the `push` method is called on the variable instance. So, I think the intention was also for it to be `FormArray`

undsoft: I think I broke it on this line: https://github.com/truenas/webui/commit/6e29382bd6cd86edd6aafeb67676129c0afbeae8#diff-378db9798d446b7bc50d6ccdca05f75b26f4ec6f96bd808cb744d65dc0a5e610R523
by assuming that it's a FormGroup, when it should have been a FormArray. Perhaps correcting that will resolve this issue.

RehanY147: Oh, dang! I think you might be right. I'll fix this and test. This will most probably solve this issue. Let's see.

undsoft (approved):
I was able to access ACL Manager. The dataset path is somehow empty again, but it's empty on master as well, I'll create another ticket for it.

PR #5577, opened 2021-06-14
NAS-110839 / 21.08 / Jest and Spectator installation and initial configuration

dszidi:
Jest and Spectator is installed and jest is configured.
One sample test is included with app.component.spec.ts. 
The test will run and complete even though the test itself actually fails

To test this PR simply run `yarn run test` and make sure the test runs and completes.


package.json
```diff
@@ -14,7 +14,10 @@
     "build": "yarn run generate_licenses && yarn run clean:dist && node --max_old_space_size=8192 node_modules/@angular/cli/bin/ng build",
     "build:prod": "node scripts/setup_prod.js && yarn run build -- --prod",
     "build:prod:aot": "yarn run build:prod --source-map --base-href /ui/ && node scripts/verify_build.js",
-    "test": "ng test",
+    "jest": "./node_modules/.bin/jest",
+    "test": "jest",
```
undsoft, line 18: Doesn't it work without having 'jest' command?

src/setupJest.ts
```diff
@@ -0,0 +1 @@
+import 'jest-preset-angular';
```
undsoft, line 1: the file should probably be name setup-jest.ts

tsconfig.spec.json
```diff
@@ -0,0 +1,22 @@
...
+    ]
+  },
+  "files": [
+    "polyfills.ts"
+  ],
+  "include": [
+    "**/*.spec.ts",
+    "**/*.d.ts"
+  ]
+}
```
undsoft, line 22: Looks like one of the `tsconfig.spec.json` is extra.

package.json
dszidi: ng and jest require global installation for those commands to work. I added those aliases for convenience.

tsconfig.spec.json
dszidi: Ah, the one in src was probably from when I merged changes from master. I'll remove it.

undsoft (approved):
Nice work!

Two things:
1. `"jest": "./node_modules/.bin/jest"` is in fact unnecessary. `test` command works without it, and having it doesn't let me just type `jest` - I still need to install it globally.
2. app.component.spec.ts failed for me, but I'm assuming this is going to be fixed in future PRs.
If you are interested in the error, here it is:

```
Test suite failed to run

    src/app/app.component.ts:9:32 - error TS2307: Cannot find module 'app/core/classes/custom-icons' or its corresponding type declarations.

    9 import { customSvgIcons } from 'app/core/classes/custom-icons';
                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    src/app/app.component.ts:10:29 - error TS2307: Cannot find module 'app/core/services/data.service' or its corresponding type declarations.

    10 import { DataService } from 'app/core/services/data.service';
                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    src/app/app.component.ts:11:30 - error TS2307: Cannot find module 'app/services/theme/theme.service' or its corresponding type declarations.

    11 import { ThemeService } from 'app/services/theme/theme.service';
                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    src/app/app.component.ts:46:94 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'string'.

    46       this.matIconRegistry.addSvgIcon(name, this.domSanitizer.bypassSecurityTrustResourceUrl(path));

```

PR #5579, opened 2021-06-15
NAS-110285 / 21.08 / Added interfaces for several APIs

RehanY147:
NAS-110285


src/app/enums/acl-type.enum.ts
```diff
@@ -24,3 +24,49 @@ export enum AclMode {
...
+  Read = 'READ',
+  Traverse = 'TRAVERSE',
+}
+
+export enum DAclFlagsBasic {
+  Inherit = 'INHERIT',
+  NoInherit = 'NOINHERIT',
+}
+
+export enum DAclPosix1eTag {
```
denysbutenko, line 65: Looks same with AclItemTag https://github.com/truenas/webui/blob/e29b9adbd134e3ca203b486d7c3838ddcdade08d/src/app/enums/acl-type.enum.ts#L7
is that intended to have two different?

denysbutenko (approved):
Looks good 👍🏻

PR #5583, opened 2021-06-15
NAS-111006 / 21.08 / Submit Button for Charts Broken in 20210610 builds

AlexKarpov98:


undsoft (approved):
Was able to install an application. 
Снимок экрана 2021-06-16 в 10 30 42

PR #5587, opened 2021-06-17
NAS-110543 / 21.08 / Added 'name_regex' to ReplicationWizard and ReplicationForm

RehanY147:
NAS-110543

To test

- Add a new replication task with remote source dataset
- You should be able to either edit 'naming_schema' (& 'also_include_naming_schema') or 'name_regex'. They shouldn't be editable at the same time. Either or.
- Check that the form submission is sending the correct payload to the API and replication task is created successfully.


undsoft (changes requested):
Issues found during testing:
- When I open the form `Naming Schema` input is not visible. I have to toggle radiobuttons back and forth for it to appear.
- If I use Regular Expression, Radio button is not set when I edit replication task.
- Would it be possible to talk to Vladimir and come up with better name for `Naming Schema or Snapshot Name Regular Expression` label. Something like `Filter snapshots`?

AlexKarpov98 (changes requested):
I see 2 issues when I try to edit existing one.
1) When I open edit form, it only shows one item (`Also Include Naming Schema `).
But once I clicked `Snapshot Name Regular Expression` and back to `Naming Schema`, then `Naming Schema ` appears. (First gif)
2) The radio button is not auto selected correctly for those `Snapshot Name Regular Expression` is set. (Second gif)
CPT2106191358-1680x853
CPT2106191408-1680x792

RehanY147:
Waiting on feedback from @themylogin. Will likely be able to update on Monday.

PR #5595, opened 2021-06-18
NAS-111089 / 21.08 / NAS-111089 - Improving type safety

undsoft:
Refactoring `resourceTransformIncomingRestData` to not mutate data in some places to be able to add typing.
Probably makes sense to test some of the affected files, for example in Certificates -> Edit "Signed By" field was not showing correctly.

### DIFF
--- a/src/app/pages/credentials/backup-credentials/forms/ssh-connections-form.component.ts
+++ b/src/app/pages/credentials/backup-credentials/forms/ssh-connections-form.component.ts
@@ -7,6 +7,7 @@ import { Subscription } from 'rxjs';
 import { KeychainCredentialType } from 'app/enums/keychain-credential-type.enum';
 import helptext from 'app/helptext/system/ssh-connections';
 import { FormConfiguration } from 'app/interfaces/entity-form.interface';
+import { KeychainCredential } from 'app/interfaces/keychain-credential.interface';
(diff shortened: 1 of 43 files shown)


src/app/pages/credentials/backup-credentials/forms/ssh-connections-form.component.ts
```diff
@@ -356,11 +357,12 @@ export class SshConnectionsFormComponent implements FormConfiguration {
     );
   }
 
-  resourceTransformIncomingRestData(wsResponse: any): any {
+  resourceTransformIncomingRestData(wsResponse: KeychainCredential): any {
```
RehanY147, line 360: When creating a new SSH Connection, can't submit the form because the `Private Key` field gets the error `Not an Integer` even though it's a dropdown and has `Generate New` option selected and no other option is available. This bug might not be related to this ticket though. Just mentioning here in case it is.

RehanY147 (approved):
Aside from the aforementioned bug about creating SSH Connections, the rest looks good.

undsoft:
> Aside from the aforementioned bug about creating SSH Connections, the rest looks good.

Seems to be happening under specific conditions. Added a ticket: https://jira.ixsystems.com/browse/NAS-111123

PR #5596, opened 2021-06-18
NAS-110989 / 21.08 / UI should consume catalog.create as a job

AlexKarpov98:


undsoft (approved):
I don't know if Task Manager updates the progress when you're looking at it, but the rest looks good.

AlexKarpov98:
> I don't know if Task Manager updates the progress when you're looking at it, but the rest looks good.

Yeah, Task Manager updates the progress.

PR #5597, opened 2021-06-19
NAS-105663 / 12.0 / The dataset will be expanded after coming back from 'Edit Options' or 'Edit Permission' pages

RehanY147:
NAS-105663

How/What to test:

- Create nested datasets in a pool
- Go to any level of choice in datasets and open 'Edit Options' or 'Edit Permission' or 'Acl Manager' pages on that dataset
- When you come back after cancelling or successfully saving the settings, the exact same dataset should be expanded automatically
e.g.,

If you have the following datasets in your pool

`pool`

- `ds` -> `ds-ds` -> `ds-ds-ds`

- `dss` -> `dss2`

Currently, when you go to `Edit Options` or `Edit Permissions` or `Acl Manager` page for dataset `ds-ds` and come back by cancelling or saving your changes, all the datasets are automatically collapsed.

With these changes, when you go to edit any aforementioned settings on `ds-ds` dataset and come back to the pools page by cancelling or saving those settings, the `EntityTreeTable` should be expanded up to that dataset automatically. i.e., 
`pool` -> `ds` -> `ds-ds`

Read ticket description and comments for more explanation or drop me a message

### DIFF
--- a/src/app/pages/common/entity/entity-form/entity-form.component.ts
+++ b/src/app/pages/common/entity/entity-form/entity-form.component.ts
@@ -405,6 +405,9 @@ export class EntityFormComponent implements OnInit, OnDestroy, OnChanges, AfterV
   }
 
   goBack() {
+    if (this.conf.goBack) {
+      return this.conf.goBack();
+    }
     let route = this.conf.route_cancel;
     if (!route) {
       route = this.conf.route_success;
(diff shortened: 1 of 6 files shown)


AlexKarpov98 (approved):
Nice work, thank you. Just one thing I think is that the same needs to be done for ZvolFormComponent.

PR #5598, opened 2021-06-21
Update fr.json

MDVAurelien:
Add some translations


undsoft (changes requested):
Thanks for your contribution.
Thought I'm somewhat confused about what it actually changes. 
I don't see any new translations being added. Strings can be kept empty if original version (i.e. KiB) is already okay.

MDVAurelien:
Yes it is not necessary but I used to translate everything. You can cancel the PR if you want. ++

undsoft:
Closing PR as default strings will already be acceptable.

PR #5599, opened 2021-06-21
NAS-111089 / None / Improving type safety 2

undsoft:
Adds proper typing to ws.job and ws.subscribe among other things. No testing necessary.


src/app/pages/applications/catalog/catalog.component.ts
```diff
@@ -311,7 +311,7 @@ export class CatalogComponent implements OnInit {
           '500px', 'info', true);
       });
     });
-    dialogRef.componentInstance.failure.pipe(untilDestroyed(this)).subscribe((err: string) => {
+    dialogRef.componentInstance.failure.pipe(untilDestroyed(this)).subscribe((err) => {
```
dszidi, line 314: Shouldn't err also be typed?

dszidi (approved):
Looks good

src/app/pages/applications/catalog/catalog.component.ts
undsoft: It's typed. It just gets its type from:
```
src/app/pages/common/entity/entity-job/entity-job.component.ts:44
```
Our ts settings don't allow for not having typing. It won't compile and require at least 'any' put it if it cannot infer the type. In this case it can, because `failure` is typed now.

PR #5600, opened 2021-06-21
NAS-110760 / 21.08 / Making multiselect respect pagination and filter

RehanY147:
NAS-110760

What to test
- Selecting values on the table should trigger `indeterminate` state on the multi-select checkbox in the table header
- Selecting all values one by one or clicking the multi-select checkbox should select all the rows currently visible on the screen and only those rows. i.e., rows from other pages or filtered out rows shouldn't be selected
- De-selecting one row should put the multi-select checkbox back into `indeterminate` state
- Check if changing the `rowIdentifier` for `SnapshotsListComponent` has had any effect on that page
- Multi-select now depends on the `rowIdentifier` or `'name'` field in either/or relationship. If `rowIdentifier` is not set `'name'` field will be used
- Other components that use multi-select should have the unique `rowIdentifier` too i.e., something that's unique across all rows

### DIFF
--- a/src/app/pages/common/entity/entity-table/entity-table.component.html
+++ b/src/app/pages/common/entity/entity-table/entity-table.component.html
@@ -111,10 +111,11 @@
         <!-- Multiselect Checkboxes -->
         <ng-container *ngIf="column == 'multiselect'" matColumnDef="{{column}}" sticky>
           <th class="multiselect-column" mat-header-cell *matHeaderCellDef>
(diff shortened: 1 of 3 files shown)


AlexKarpov98 (changes requested):
Checkbox on the header for all-select doesn't update all the items checked.
snapshot_select_all

AlexKarpov98 (approved):
It works good. Thank you.

PR #5604, opened 2021-06-21
Example unit tests.

dszidi:
This PR provides example unit tests for a generic typescript class and an angular service.

**To run all tests:** `yarn test`
**To test only the typescript class:** `yarn test -i src/app/core/classes/theme-utils/theme-utils.spec.ts`
**To test only the angular service:** `yarn test -i src/app/core/services/disk-state/disk-state.service.spec.ts`


package.json
```diff
@@ -158,6 +158,16 @@
   },
   "jest": {
     "preset": "jest-preset-angular",
-    "setupFilesAfterEnv": ["<rootDir>/src/setup-jest.ts"]
+    "setupFilesAfterEnv": ["<rootDir>/src/setup-jest.ts"],
+        "testPathIgnorePatterns": [
```
undsoft, line 162: Something happened with indentation here.

src/app/core/services/base.service.ts
```diff
@@ -1,20 +1,17 @@
...
 import { CoreService } from './core.service';
 
 @Injectable({
   providedIn: 'root',
 })
 export class BaseService {
-  protected websocket: WebSocketService;
-  protected core: CoreService;
+  // protected websocket: WebSocketService;
+  // protected core: CoreService;
```
undsoft, line 10: Someone needs to start reviewing their code when creating a PR 😛

src/app/core/services/disk-state/disk-state.service.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+   * Test Methods
+   * */
+
+  beforeEach(() => {
+    spectator = createService();
+  });
+
+  it('should instantiate', () => {
+    expect(spectator).toBeTruthy();
+  });
```
undsoft, line 41: This block is probably unnecessary. I think createService() throws if something is not right.

src/app/core/services/disk-state/disk-state.service.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  test('should forward DisksChanged notifications', (done) => {
```
undsoft, line 43: What's the difference between `it` and `test`?

undsoft (approved):
Very nice!

PR #5608, opened 2021-06-22
NAS-110839 / 21.08 / Add unit tests to github actions

dszidi:


undsoft (approved):
👍🏼

dszidi:
It seems there is a problem. The environment setup script is not run in the ci environment so the test fails. We need to mock this somehow in the unit tests.

undsoft:
@dszidi I think environment.ts is created during the build. You can try swapping steps.

dszidi:
Putting it last didn't work. Going to try running tests before the build but using setup_env.js to create an environment file. The build in the next step should overwrite that environment file.

PR #5620, opened 2021-06-25
NAS-111170 / 21.08 / NAS-111170 - Improving type safety

undsoft:
The only testing that you can do is confirm that slider is not broken by going to System Settings -> Shell -> Set font size. (Spoiler: it's not broken :P)


AlexKarpov98 (approved):
Nice work, thanks.

PR #5621, opened 2021-06-25
NAS-111170 / None / NAS-111170 - Improving type safety 2

undsoft:
Don't think any testing is necessary.


RehanY147 (approved):
Looks good.

PR #5624, opened 2021-06-25
NAS-110964 / 21.08 / Storage Widget for Dashboard

AlexKarpov98:


RehanY147:
The pool data should be a bit more centered or maybe a bit more left margin as compared to the top and right margins
image

RehanY147:
When you have 3 pools, the icon looks a bit off center for the 4th box "Create a new pool"
image

RehanY147:
Also, maybe the text should just say "Create Pool" or "New"

src/app/core/components/widgets/widgetstorage/widgetstorage.component.ts
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  getDiskWithErrorsItemInfo(pool: Pool): ItemInfo {
+    let level = 'warn';
+    let icon = 'mdi-alert-circle';
+    let value = T('Unknown');
+
+    if (pool && pool.topology) {
+      const unhealthy: any[] = []; // Disks with errors
+      pool.topology.data.forEach((item: any) => {
```
RehanY147, line 173: Can we use `VDevStats` here instead of `any`?

src/app/core/components/widgets/widgetstorage/widgetstorage.component.ts
```diff
@@ -0,0 +1,271 @@
...
+        if (item.type == VDevType.Disk) {
+          const diskErrors = item.read_errors + item.write_errors + item.checksum_errors;
+
+          if (diskErrors > 0) {
+            unhealthy.push(item.disk);
+          }
+        } else {
+          item.children.forEach((device: any) => {
```
RehanY147, line 181: `VDevStats` here as well

src/app/core/components/widgets/widgetstorage/widgetstorage.component.ts
```diff
@@ -0,0 +1,271 @@
...
+    };
+  }
+
+  getDiskWithErrorsItemInfo(pool: Pool): ItemInfo {
+    let level = 'warn';
+    let icon = 'mdi-alert-circle';
+    let value = T('Unknown');
+
+    if (pool && pool.topology) {
+      const unhealthy: any[] = []; // Disks with errors
```
RehanY147, line 172: If possible, we should replace this `any` too.

RehanY147 (changes requested):

AlexKarpov98:
@RehanY147 Fixed.

undsoft (changes requested):
It works, but would be cooler to get closer to designs. 
Doesn't have to be pixel perfect, but there are some easy to spot differences.
Снимок экрана 2021-06-29 в 11 54 40

[This plugin](https://chrome.google.com/webstore/detail/perfectpixel-by-welldonec/dkaagdgjmgdmbnecmcefdhjekcoceebi?hl=en) is pretty cool to be able to compare things side by side.

undsoft:
Also, before I forget, there is a Configure button on Dashboard, which sets which widgets are shown by default. Would be good to show Storage widget by default, but hide individual Pool widgets, so that users can only enable them if needed.

AlexKarpov98:
@undsoft Fixed

undsoft (approved):
It looks pretty good.

Some minor things:
1. I think the name of the pool is intended to be all uppercase. We can make the font a bit smaller.
2. When pool name is too long, let's try to use text-overflow with ....
3. Can we make the color of warning icons more orange. Currently they are somewhat similar to green okay icons.
4. 
Снимок экрана 2021-07-01 в 12 26 57

RehanY147 (approved):
LGTM

PR #5628, opened 2021-06-29
NAS-111148 / 12.0 / Enclosure UI should handle case of no sas expanders

dszidi:
Testing:

- Test against R10 (tn27) for no SAS Expanders case
- Test against any available TrueNAS controller that does have SAS Expanders


darkfiberiru:
@dszidi  Shouldn't this be head and 12?

undsoft (approved):
Tested on tn29, looks good.

PR #5630, opened 2021-06-29
NAS-111222 / 12.0 / Don't assume REMOVED status will be present

dszidi:
To test/reproduce:

- Create a pool and add a vdev of type SPARE
- physically remove one of the non SPARE disks
- Make sure Enclosure UI still displays the pools once resilver has started

### DIFF
--- a/src/app/core/classes/system-profiler.ts
+++ b/src/app/core/classes/system-profiler.ts
@@ -182,9 +182,7 @@ export class SystemProfiler {
         v.disks[name] = -1; // no children so we use this as placeholder
       } else if (vdev.children.length > 0) {
         vdev.children.forEach((disk, dIndex) => {
-          if (!disk.device && disk.status == 'REMOVED') {
-
-          } else {
+          if (disk.device && disk.status != 'REMOVED') {
             const spl = disk.disk.split('p'); // was disk.device
             const name = spl[0];
             v.disks[name] = dIndex;


src/app/core/classes/system-profiler.ts
```diff
@@ -182,9 +182,7 @@ export class SystemProfiler {
         v.disks[name] = -1; // no children so we use this as placeholder
       } else if (vdev.children.length > 0) {
         vdev.children.forEach((disk, dIndex) => {
-          if (!disk.device && disk.status == 'REMOVED') {
-
-          } else {
+          if (disk.device && disk.status != 'REMOVED') {
```
RehanY147, line 185: The new condition would be true for a modal
`disk.device = true;
disk.status = 'REMOVED'`

The condition isn't exactly matched. With these values, the code would run before in the else block but won't now

RehanY147 (changes requested):

RehanY147 (approved):
LGTM

dszidi:
@undsoft I posted a video into the ticket comments to demo the problem as well as the fix

https://jira.ixsystems.com/browse/NAS-111222?focusedCommentId=134655&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-134655

darkfiberiru:
image
Fix looks good.

undsoft (approved):
Approving based on Damian's video.

darkfiberiru (commented):
Lgtm tested and shared screenshot of fixed system after drive pull

PR #5631, opened 2021-06-30
NAS-110967 / 21.08 / View Permissions Sidebar

undsoft:
https://xd.adobe.com/view/285bc415-335a-4aff-b632-3e9c6ab7e179-7d08/screen/f2264dc8-6552-48db-b89c-d0ff31e86341

**New code patterns:**
- Simple store with ngrx component store.
- Extracted enum labels closer to enums as a map to make them reusable and easier to work with.
- Linter will allow unused _ in argument list, for cases like filter([_, something]) => {}

**How to test:**
1. Ensure that you are using fresh middleware. June 30th+.
2. Create a dataset with POSIX acl type.
2. Click three dots -> View permisions.
3. Inspect the sidebar.
4. When dataset is created, permissions are trivial. Edit them in an ACL editor to make them non-trivial.
5. Inspect view permisions sidebar again.
6. Repeat for another dataset with NFS acl type.

**Known things:**
1. Layout is somewhat different from design. In design sidebar is within a dataset, in implementation it's outside. This change has been confirmed with Damian.
2. Because of this, there are some stylistic changes that needed to happen.
3. Instead of just showing `root` in some cases you will see `user - root` or `owner@ - root` to make it more clear what the permission is.
4. Selected dataset is not highlighted and you cannot view permissions for another dataset just by clicking on a row. This will be tackled in another ticket.


denysbutenko (approved):
image
Everything as described. Nice work 👍🏻

PR #5632, opened 2021-06-30
NAS-111236 / 21.08 / Add unit tests to password-validation.ts

denysbutenko:


src/app/pages/common/entity/entity-form/validators/password-validation/password-validation.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+        other: ['', []],
+      });
+
+      thisControl = form.get('this');
+      otherControl = form.get('other');
+    });
+
+    it('should have parent', () => {
+      expect(thisControl.parent).toBeTruthy();
+    });
```
undsoft, line 33: Doesn't look like this case actually tests anything in password-validation.
You can change it to 'should return null when control does not have a parent' or you can just remove this case.

undsoft (approved):
Noice!

PR #5634, opened 2021-07-01
NAS-111094 / 21.08 / Add load more on scroll for users and groups comboboxes

denysbutenko:
**To test**

- Create a new user
- Open Shares dashboard
- Click `Add` NFS share
- Switch to 'Advanced Options'
- Click on [`Maproot User`,`Maproot Group`,`Mapall User`,`Mapall Group`] dropdown button on the right
- Scroll down to find user/group


RehanY147 (approved):
Upon scroll down, new entries load fine

PR #5635, opened 2021-07-01
NAS-111261 / 21.08 / NAS-111261 - Improving type safety

undsoft:
No testing is necessary.


src/app/pages/system/general-settings/gui-form/gui-form.component.ts
```diff
@@ -296,7 +296,7 @@ export class GuiFormComponent implements FormConfiguration {
 
             this.loader.open();
             this.ws.shuttingdown = true; // not really shutting down, just stop websocket detection temporarily
-            this.ws.call('service.restart', ['http']).pipe(untilDestroyed(this)).subscribe(() => {
+            this.ws.call('service.restart', ['http' as any]).pipe(untilDestroyed(this)).subscribe(() => {
```
AlexKarpov98, line 299: I think it is better to add a new 'http' to 'ServiceName' and then use it here instead of ['http' as any].

undsoft: Fair enough. Fixed.

dszidi (approved):
Looks good!

PR #5636, opened 2021-07-02
NAS-110126 / 21.08 / Selecting multiple directories for source path for push and pull

RehanY147:
We need to allow selecting multiple source directories for cloud sync (local directories for PUSH, remote directories for PULL). All source directories must have the same parent (be on the same level).

When only one directory is selected, the behavior is the same. However, when multiple directories are selected:

1. `path` (for PUSH) or `attributes.folder` (for PULL) is set to the parent directory path.

2. New `include` field is populated with list of the names of the selected directories in the following form: ["/Directory 1/**", "/Directory 2/**"].

For example in the following tree:
image

We should be able to select `sub1.1` and `sub1.2` (but not `sub1.1` and `sub2`). If we select them, we'll get:
`
{
    "path": "/mn/pool/sub1",
    "include": ["/sub1.1/**", "/sub1.2/**"]
}
`

To test,

Select multiple directories when creating a cloud sync task. Submit the form. The payload should be according to the description.

NAS-110126


src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
```diff
@@ -571,17 +598,124 @@ export class CloudsyncFormComponent implements FormConfiguration {
...
+      let prevErrors = folder_source_fc.errors;
+      if (prevErrors === null) {
+        prevErrors = {};
+      }
+      if (!allMatch) {
+        folder_source_fc.setErrors({
+          ...prevErrors,
+          misMatchDirectories: true,
+        });
+        folder_source_field.warnings = 'All selected entries must be at the same level i.e., must have the same parent directory.';
```
denysbutenko, line 633: it should support i18n

src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
```diff
@@ -571,17 +598,124 @@ export class CloudsyncFormComponent implements FormConfiguration {
...
+      let prevErrors = path_source_fc.errors;
+      if (prevErrors === null) {
+        prevErrors = {};
+      }
+      if (!allMatch) {
+        path_source_fc.setErrors({
+          ...prevErrors,
+          misMatchDirectories: true,
+        });
+        path_source_field.warnings = 'All selected entries must be at the same level i.e., must have the same parent directory.';
```
denysbutenko, line 677: same here

denysbutenko:
I created task successfully then trying to edit it but `path` values not filled in explorer and its produce an error on submit:
```cloud_sync_update.path: The path must reside within a pool mount point``` it seems like an issue

RehanY147:
> 
> 
> I created task successfully then trying to edit it but `path` values not filled in explorer and its produce an error on submit:
> `cloud_sync_update.path: The path must reside within a pool mount point` it seems like an issue

You shouldn't be able to select the `root` directory `/mnt` as the path so that error is correctly shown. Fixed the issue where the values aren't loading for the explorer. Once you expand to the correct level, the value are correctly selected automatically. Why `entity-explorer` doesn't automatically try to expand to the selected directories is a separate issue that should be handled by NAS-111275

denysbutenko (approved):
image
Works as described 👍🏻

PR #5639, opened 2021-07-03
NAS-110391 / 21.08 / NAS-110391 - Fixes action menu on Boot Env

undsoft:
To test:
1. System Settings -> Boot
2. Actions -> Boot Pool Status
3. Make sure three dot menu is there and working.

### DIFF
--- a/src/app/pages/system/bootenv/bootenv-status/bootenv-status.component.ts
+++ b/src/app/pages/system/bootenv/bootenv-status/bootenv-status.component.ts
@@ -2,6 +2,8 @@ import { Component, OnInit } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { TreeNode } from 'primeng/api';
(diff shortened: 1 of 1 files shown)


src/app/pages/system/bootenv/bootenv-status/bootenv-status.component.ts
```diff
@@ -101,96 +104,105 @@ export class BootStatusListComponent implements OnInit {
     if (data.stats) {
       stats = data.stats;
     }
+
+    let name: string;
```
AlexKarpov98, line 108: @undsoft 
image
we need to set the initial value for 'name', e.g. let name = data.name

AlexKarpov98 (changes requested):

src/app/pages/system/bootenv/bootenv-status/bootenv-status.component.ts
undsoft: Good catch. Fixed.

AlexKarpov98 (approved):

PR #5640, opened 2021-07-04
NAS-111091 / 21.10 / Refactoring a small form

RehanY147:
NAS-111091


src/app/pages/system/general-settings/localization-form2/components/value-accessor-parent.class.ts
```diff
@@ -0,0 +1,27 @@
+import { ControlValueAccessor } from '@angular/forms';
+
+export class ValueAccessorParent implements ControlValueAccessor {
```
undsoft, line 3: I would prefer not to use inheritance and just have this code in each component.
value will not always be string, you'd have to reimplement this with generics, there may also be different functionality needed for example on `writeValue` and in general inheritance will make it harder to navigate.

src/app/pages/system/general-settings/localization-form2/components/ix-form-errors/ix-form-errors.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component, Input } from '@angular/core';
+import { FormControl } from '@angular/forms';
+
+@Component({
+  selector: 'ix-form-errors',
+  templateUrl: './ix-form-errors.component.html',
+})
+export class IXFormErrorsComponent {
```
undsoft, line 8: Should be IxFormErrorsComponent.

src/app/pages/system/general-settings/localization-form2/components/ix-input/ix-input.component.html
```diff
@@ -0,0 +1,71 @@
...
+        id="{{name}}_form-input-spinner"
+        *ngIf="isLoading">
+      </mat-spinner>
+      <div *ngIf="maskValue">
+        <p>{{'Enter' | translate}} <strong>{{maskValue}}</strong> {{'below to confirm.' | translate}}</p>
+        <p [ngStyle]="{'position': 'relative', 'top':'15px', 'opacity': '.38'}">{{maskValue}}</p>
+      </div>
+      <div class="input-container">
+        <span *ngIf="searchable" class="searchable"><mat-icon>search</mat-icon></span>
+        <input [(ngModel)]="value" [ngStyle]="{'padding-left': searchable ? '2rem' : '', 'padding-right': shouldShowResetInput() ? '2rem': ''}" [type]="inputType ? inputType : 'text'"
```
undsoft, line 29: Two approaches are being mixed here – ngModel and reactive forms. We should just use reactive forms.

src/app/pages/system/general-settings/localization-form2/localization-form2.component.html
```diff
@@ -0,0 +1,11 @@
+<mat-card class="form-card">
+  <div class="mat-content">
+    <form #f="ngForm" (ngSubmit)="onSubmit(f)" class="form-wrap">
+      <ix-form-field-set>
+        <ix-input name="myInput" id="myInput" [placeholder]="'My Name'" [required]="true" ngModel></ix-input>
+        <ix-select [options]="selectOptions" [placeholder]="'My Sibling'" name="mySelect" id="mySelect" ngModel></ix-select>
```
undsoft, line 6: This looks pretty good, I only have two complaints:
1. We should use reactive forms.
2. The resulting components may still be too fat. For example, ix-input has `maskValue` functionality, error handling, show/hide password toggle, etc. This is definitely better than what we currently have, but will also create problems in the future, where if you want different functionality you will have to open base component and add another @Input().
Take a look at https://material.angular.io/components/input/examples and see if we can use a similar approach of having more composable components. We don't have to do it now, if it's too complicated though.

PR #5642, opened 2021-07-05
NAS-110938 / 21.08 / WIP: ViewChartAreaComponent

dszidi:
ViewChartArea is a thin wrapper for chart.js to create area charts.

- It simply takes a ChartData object for configuration and renders a chart.js instance.
- It takes in a max of 8 data sources since the theme palette only has 8 colors and line charts become less useful when overloaded with lines/areas


AlexKarpov98 (approved):

src/app/core/components/viewchartarea/viewchartarea.component.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+        previousValue: null,
+        firstChange: true,
+        isFirstChange: () => true,
+      },
+    });
+
+    // Make sure expected values are present after input is set
+    expect(spectator.component.data).toBeTruthy();
+    expect(spectator.component.canvas).toBeTruthy();
+    expect(spectator.component.chart).toBeTruthy();
```
undsoft, line 91: Would be good to test that not only _some_ canvas is there, but also that the canvas/chart has the same data that was passed in component on line 76.

undsoft (approved):

src/app/core/components/viewchartarea/viewchartarea.component.ts
```diff
@@ -0,0 +1,51 @@
...
+      type: 'line',
+      data,
+      options: {
+        responsive: true,
+      },
+    };
+  }
+
+  render(data: ChartData): Chart {
+    if (data.datasets.length > this.maxSources) throw 'ERROR: Maximum Sources Exceeded. Line/Area charts have a hard limit of 8 data sources';
```
undsoft, line 29: Please wrap in `new Error()` for reasons described here: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-throw-literal.md
Will add linter rule for this one.

src/app/core/components/viewchartarea/viewchartarea.component.html
```diff
@@ -0,0 +1,3 @@
+<div class="wrapper">
```
undsoft, line 1: Also, please [rename to kebab-style](https://angular.io/guide/styleguide#style-02-02), i.e. view-chart-area/view-chart-area.component.html, ts, etc.
I'll rename other files.

PR #5644, opened 2021-07-06
NAS-111312 / 21.08 / Removing barrel files

undsoft:
Adds a couple of linter rules and removing index.ts files.
No testing is necessary.


dszidi (approved):
Looks great!

PR #5645, opened 2021-07-06
Added tests for ValidationService.greaterThan method

RehanY147:
NAS-111235


src/app/services/validation.service.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  it('should have greatherThan error when value1 is smaller', (done) => {
+    formGroup.get(formControlName).setValue(0);
+    formGroup.get(otherFormControlName).setValue(1);
+    formGroup.updateValueAndValidity();
+    expect(formGroup.valid).toBeFalsy();
+    expect(formGroup.get(formControlName).errors.greaterThan).toBeTruthy();
+    expect(formGroup.get(formControlName).errors.fields[0]).toBe(formControlPlaceholder);
+    done();
```
undsoft, line 26: `done` argument and calling `done()` is only really needed when you're testing async code. Not needed here.

src/app/services/validation.service.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    let customError: Error = null;
+    try {
+      formGroup.get(formControlName).updateValueAndValidity();
+    } catch (error) {
+      customError = error;
+    }
+    expect(() => {
+      throw customError;
+    }).toThrow('greaterThanValidator(): other control is not found in parent group');
```
undsoft, line 71: I think it should be possible to just move:
```
formGroup.get(formControlName).updateValueAndValidity();
```
inside `expect` without having explicit try-catch block.

undsoft (approved):
Looks pretty good. 👍🏼

src/app/services/validation.service.spec.ts
RehanY147: The `toThrow` expects the argument to `expect` to be reference to a method instead of an expression/value. The error here is thrown by the service at a deeper level and without this try/catch work around, the error doesn't get picked just by linking `updateValueAndValidity` to `toThrow`. That's why this work around is here.

PR #5647, opened 2021-07-07
NAS-110665 / 12.0 / NAS-110665 - Fixing cloudcredential form + OpenStack Swift

undsoft:
**Setup:**
1. `docker pull openstackswift/saio`
2. `docker run -d -p 8080:8080 openstackswift/saio`
2. If you are using local middleware, you can skip this test. Otherwise you need to expose local openstackswift to the world. You can do this by setting up a tunnel with https://ngrok.com. Something like `ngrok tcp 8080`
3. Go to System -> Cloud Credentials -> Add.
4. Add a new OpenStack Swift credential, where 
UserName: `test:tester`
Password: `testing`
Auth URL: If local: `http://127.0.0.1/auth/v1.0`, otherwise `http://ngrokurlandport/auth/v1.0`, something similar to `http://8.tcp.ngrok.io:16358/auth/v1.0`
Auth Version: `v1`
Tenant Name: `test`
5. Verify Credentials.

**To test:**
- That Advanced options change when Auth Version is changed.
- That when form is saved and then edited, Auth Version is populated and form still works correctly.

Also approve: https://github.com/truenas/webui/pull/5648

### DIFF
--- a/src/app/pages/system/CloudCredentials/cloudcredentials-form/cloudcredentials-form.component.ts
+++ b/src/app/pages/system/CloudCredentials/cloudcredentials-form/cloudcredentials-form.component.ts
@@ -602,7 +602,7 @@ export class CloudCredentialsFormComponent {
               value: 3,
             },
           ],
-          value: '0',
+          value: 0,
           relation: [
             {
               action: 'SHOW',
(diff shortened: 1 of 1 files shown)


denysbutenko (approved):
image
Works as described. Looks good to me

PR #5650, opened 2021-07-07
NAS-111234 / 21.08 / Add unit tests for EntityUtils.parseSchemaFieldConfig

AlexKarpov98:


src/app/pages/common/entity/utils.spec.ts
```diff
@@ -0,0 +1,866 @@
...
+      const result = utils.parseSchemaFieldConfig(item.schema);
+      const isEqual = _.isEqual(item.expectedFieldConfigs, result);
+      expect(isEqual).toBe(true);
+    });
+  });
+
+  describe('parse unexpected type', () => {
+    it('should return empty and show error on console', () => {
+      const originalError = console.error;
+      console.error = jest.fn();
```
dszidi, line 857: A more stream lined way to deal with errors might be to use expect(fn).toThrow()

https://jestjs.io/docs/expect#tothrowerror

AlexKarpov98: @dszidi this is for testing console.error, `console.error('Unsupported type=', schemaConfig);`.
I think `expect(fn).toThrow()` can not test this case properly, because console.error does not throw an error.

dszidi: Ok, sounds good

dszidi (approved):
Looks good!

PR #5652, opened 2021-07-10
NAS-110813 / 21.08 / Added localStorage item to check if dialog's been shown

RehanY147:
NAS-110813

### DIFF
--- a/src/app/components/common/topbar/topbar.component.ts
+++ b/src/app/components/common/topbar/topbar.component.ts
@@ -263,8 +263,9 @@ export class TopbarComponent extends ViewControllerComponent implements OnInit, 
       this.target.next({ name: 'SubmitComplete', sender: this });
       this.isWaiting = false;
     }
+    const setupComplete = localStorage.getItem('setupComplete') as unknown as boolean;
     this.showWelcome = evt.data.showWelcomeDialog;
-    if (this.showWelcome) {
+    if (this.showWelcome && !setupComplete) {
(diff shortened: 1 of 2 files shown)


src/app/components/common/topbar/topbar.component.ts
```diff
@@ -263,8 +263,9 @@ export class TopbarComponent extends ViewControllerComponent implements OnInit,
       this.target.next({ name: 'SubmitComplete', sender: this });
       this.isWaiting = false;
     }
+    const setupComplete = localStorage.getItem('setupComplete') as unknown as boolean;
```
undsoft, line 266: The reason you got a type error is because localstorage actually contains a string.
The check below will evaluate to true for any string, even `'false'`.

undsoft (changes requested):
1. The part about 'message about built in users' is not addressed.
2. Do you know what is the root cause of the issue? Current solution feels like a workaround. Would be better instead to try to solve the underlying issue – why user preferences about these messages are not correctly set or correctly read.

src/app/components/common/topbar/topbar.component.ts
RehanY147: Could not reproduce this but I have seen it happen. The code is pretty straight forward so adding another check was the only solution I could think of.

undsoft:
Will close this PR for now, since we'd want to take another approach at solving this.

PR #5657, opened 2021-07-11
NAS-111123 / 21.08 / Renamed the keypair with a counter to avoid same names when resubmitting the form

RehanY147:
NAS-111123

You can read the ticket description for reproduction steps and explanation


undsoft (changes requested):
This kinda works, but creates another issue – additional keys being created unexpectedly to the user.
I think it would be better to extract code that queries ssh keys (lines 262-271) to a separate method, and query it on line 383, so that we only create a new key once.

PR #5658, opened 2021-07-11
NAS-111015 / 21.08 / Fixed method that takes in fixed constant so it takes the latest value

RehanY147:
NAS-111015


denysbutenko (approved):
Works properly now

PR #5660, opened 2021-07-12
update the term threads to cores for temp values (Core Version)

q5sys:
Cores are physical. Threads are logical.  Since this is regarding about physical temp we are talking about cores not threads.
I'll have another PR for the SCALE UI.


dszidi (approved):
Looks good!

PR #5661, opened 2021-07-12
update the term threads to cores for temp values (SCALE version)

q5sys:
Cores are physical. Threads are logical.  Since this is regarding about physical temp we are talking about cores not threads.
I think master is the scale branch, but if its not, let me know and I'll change the base branch.

### DIFF
--- a/src/app/core/components/widgets/widget-cpu/widget-cpu.component.html
+++ b/src/app/core/components/widgets/widget-cpu/widget-cpu.component.html
@@ -64,7 +64,7 @@
                     <span class="label"><strong>{{"Hottest" | translate}}:</strong></span> &nbsp;&nbsp;
                     <ng-container *ngIf="tempAvailable; else notemps">
                       <span>{{tempMax}}°C &nbsp;</span>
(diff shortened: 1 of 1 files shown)


AlexKarpov98 (commented):
@q5sys LGTM. 
you changed the "thread" to "core" only for temp, but do we have no need to change all "thread" to "core"? for instance, `Highest Usage: 1%  (3 threads at 1%)`, `Stats Per Thread`

q5sys:
> @q5sys LGTM.
> you changed the "thread" to "core" only for temp, but do we have no need to change all "thread" to "core"? for instance, `Highest Usage: 1% (3 threads at 1%)`, `Stats Per Thread`

It's complicated because in that regard sometimes cores/threads are used interchangeably. I guess it depends on the middleware code of what it's actually reporting. If its an hyperthreaded 8 core cpu (thus 16 threads), can the UI ever report >8 for that value.  If so then I'd say we keep it at threads.  If its actually only reporting physical cores, we could change that one as well.  My guess is that the OS is reporting threads there instead of physical cores, but I could be wrong on that.  But it's certainly meaning physical cores when its talking about temps.

q5sys:
Just checked on one of my servers.  I think threads is correct in the highest usage spot because it does show a number > than the core count.  
image

AlexKarpov98 (approved):

PR #5662, opened 2021-07-13
NAS-110966 / 21.08 / Network Widget for Dashboard

AlexKarpov98:


src/app/core/components/view-chart-area/view-chart-area.component.ts
```diff
@@ -19,31 +20,29 @@ export class ViewChartAreaComponent implements OnDestroy, OnChanges {
...
+      options: this.options,
     };
   }
 
-  render(data: ChartData): Chart {
-    if (data.datasets.length > this.maxSources) {
-      throw new Error('ERROR: Maximum Sources Exceeded. Line/Area charts have a hard limit of 8 data sources');
-    }
+  render(): void {
+    if (this.data) {
```
undsoft, line 28: Best to do something like:
```
if (!this.data) {
  return;
}
```
to avoid nesting.

src/app/core/components/widgets/widget-network/widget-network.component.html
```diff
@@ -0,0 +1,81 @@
...
+                          matTooltipPosition="above"
+                          aria-label="Network Reports">
+                            insert_chart
+                        </mat-icon>
+                      </button>
+                    </div>
+                    <div class="info-body" fxLayout="row" fxFlex fxLayoutAlign="space-between stretch">
+                      <ul>
+                        <li fxLayout="row" fxLayoutAlign="start center">
+                          <span class="icon up" *ngIf="nic.state.link_state === 'LINK_STATE_UP'">
```
undsoft, line 40: Please introduce a enum for LINK_STAT_UP/DOWN.

src/app/core/components/widgets/widget-network/widget-network.component.ts
```diff
@@ -0,0 +1,349 @@
...
+          const nicInfo = this.nicInfoMap[nicName];
+          nicInfo.in = received.value + ' ' + received.units + '/s';
+          nicInfo.out = sent.value + ' ' + sent.units + '/s';
+
+          if (evt.data.sent_bytes - nicInfo.lastSent > 1024) {
+            nicInfo.lastSent = evt.data.sent_bytes;
+            this.tableService.updateStateInfoIcon(nicName, 'sent');
+          }
+
+          if (evt.data.received_bytes - nicInfo.lastReceived > 1024) {
```
undsoft, line 159: What is 1024? Some minimal limit? If so, let's introduce a variable to make it clear what's happening.

src/app/core/components/widgets/widget-network/widget-network.component.ts
```diff
@@ -0,0 +1,349 @@
...
+
+    return ip;
+  }
+
+  getLinkState(nic: BaseNetworkInterface): string {
+    if (!nic.state.aliases) { return ''; }
+    return nic.state.link_state.replace(/_/g, ' ');
+  }
+
+  getServerTime(): Date {
```
undsoft, line 263: 1. Since the same code now exists in two places (here and in report.component), it's best to move this to a service.
2. This can be done with Angular's HttpClient, see [{ observe: 'response' }](https://angular.io/guide/http#reading-the-full-response)

src/app/core/components/widgets/widget-network/widget-network.component.ts
```diff
@@ -0,0 +1,349 @@
...
+            pointRadius: 0.1,
+          },
+        ],
+      };
+
+      this.nicInfoMap[nic.name].chartData = chartData;
+    });
+  }
+
+  convertKMGT(value: number): string {
```
undsoft, line 323: Duplicated code that can be extracted in utils.

src/app/core/components/widgets/widget-network/widget-network.component.ts
```diff
@@ -0,0 +1,349 @@
...
+    this.rowHeight = (this.contentHeight - space) / this.rows;
+
+    const newChartOptions = { ...this.chartOptions };
+    newChartOptions.scales.yAxes[0].ticks.maxTicksLimit = maxTicksLimit;
+    newChartOptions.aspectRatio = this.aspectRatio;
+    this.chartOptions = newChartOptions;
+  }
+
+  getIpAddress(nic: BaseNetworkInterface): string {
+    let ip = T('Unknown');
```
denysbutenko, line 246: I was confused by "Unknown" string. Maybe we could specify that it belongs to IP Address?

denysbutenko: image

undsoft (changes requested):
Looks good, I like how up and down icons indicate traffic.

Additional testing results:
- The chart does not seem to update. I downloaded an ubuntu ISO from shell and the chart only updated after refresh.
- Similarly to storage widget, this widget is to replace individual nic widgets. We can hide them by default now.

src/app/core/components/widgets/widget-network/widget-network.component.html
```diff
@@ -0,0 +1,81 @@
...
+                          matTooltipPosition="above"
+                          aria-label="Network Reports">
+                            insert_chart
+                        </mat-icon>
+                      </button>
+                    </div>
+                    <div class="info-body" fxLayout="row" fxFlex fxLayoutAlign="space-between stretch">
+                      <ul>
+                        <li fxLayout="row" fxLayoutAlign="start center">
+                          <span class="icon up" *ngIf="nic.state.link_state === linkState.Up">
```
undsoft, line 40: You can just keep LinkState with first L being capitalized.

src/app/core/components/widgets/widget-network/widget-network.component.ts
```diff
@@ -103,7 +112,8 @@ export class WidgetNetworkComponent extends WidgetComponent implements AfterView
   };
 
   constructor(
-    public router: Router, private ws: WebSocketService, private locale: LocaleService,
+    public router: Router, private ws: WebSocketService,
+    private locale: LocaleService, private rs: ReportsService,
```
undsoft, line 116: It's better not to abbreviate ReportsService to `rs`, it's not a common service and you won't be able to tell what it is if you were to see 'rs' in code.

src/app/enums/network-interface.enum.ts
```diff
@@ -24,3 +24,8 @@ export enum LinkAggregationProtocol {
   RoundRobin = 'ROUNDROBIN',
   None = 'NONE',
 }
+
+export enum LinkState {
```
undsoft, line 28: Also you can type `link_state` in `NetworkInterfaceState`.

undsoft (changes requested):
The chart updates, but I kinda have to wait for it to update. Anything we can do about it? Update faster or if this is about scale, reduce scale to 1h.

undsoft (approved):
Let's try 5 or 10 seconds refresh time, otherwise looks good. Nice work!

PR #5668, opened 2021-07-16
NAS-111291 / 21.08 / Add Jobs List

denysbutenko:
Added MVP of Jobs List page.

**To test:**

- Jobs List page available through History button (when jobs manager has tasks or has empty state)
- Jobs List displays job state properly
- Page matches the design
- Running jobs can be aborted successfully
- The `View Logs` button has appeared and details are available
- Toggle filters are working properly
- Search filter is working properly


src/app/pages/jobs/jobs-list/jobs-list.component.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  ngOnInit(): void {
+    console.info('Job Log ngOnInit');
+
+    this.modalService.refreshTable$.pipe(untilDestroyed(this)).subscribe(() => {
+      this.refreshTab();
+    });
+  }
+
+  dataHandler(entityList: EntityTableComponent): void {
```
undsoft, line 61: There's another thing, called `resourceTransform` or something like that, better to use that, because than you can update rows immutably.

denysbutenko: Fixed

src/app/pages/common/entity/entity-table/entity-table.component.html
```diff
@@ -178,6 +178,48 @@
...
+                For some reason it doesn't work
+                TODO: Try to reuse app-job-item components
+
+                <ng-container [ngSwitch]="currentColumns[i].widget.component">
+                  <app-job-item *ngSwitchCase="'JobItemComponent'" [ngClass]="'test'" [job]="element"></app-job-item>
+                </ng-container>
+            -->
+            <div class="entity-job-item" fxLayout="row" fxLayoutAlign="space-between center" fxLayoutGap="8px">
+              <div fxFlex="100" fxLayout="row" fxLayoutGap="8px" fxLayoutAlign="start center">
+                <mat-icon *ngIf="element.state === EntityJobState.Success" class="state-success">check_circle_outline</mat-icon>
```
undsoft, line 195: This looks like a job for ngSwitch

src/app/pages/common/entity/entity-table/entity-table.component.html
```diff
@@ -178,6 +178,48 @@
...
+                  <mat-progress-bar
+                    *ngIf="element.state === EntityJobState.Running"
+                    [value]="element.progress.percent"
+                    mode="buffer"
+                    class="entity-job-progress"
+                  ></mat-progress-bar>
+                </div>
+              </div>
+
+              <button mat-icon-button type="button" (click)="this.conf.onAborted(element)" *ngIf="element.state === EntityJobState.Running && element.abortable">
```
undsoft, line 216: I actually think we can just remove abortable column. Don't see how it's going to be useful in history.

src/app/pages/jobs/jobs-list/jobs-list.component.ts
```diff
@@ -0,0 +1,77 @@
...
+    private translate: TranslateService,
+    private localeService: LocaleService,
+    private store: JobsManagerStore,
+  ) {}
+
+  preInit(entityList: EntityTableComponent): void {
+    this.entityList = entityList;
+  }
+
+  resourceTransformIncomingRestData(data: JobRow[]): JobRow[] {
```
undsoft, line 58: Incoming data is whatever came from API, i.e. `Job`.

src/app/pages/jobs/jobs-list/jobs-list.component.ts
```diff
@@ -0,0 +1,77 @@
...
+    return data.map((item) => this.transformJob(item));
+  }
+
+  transformJob(job: JobRow): JobRow {
```
undsoft, line 62: Same here, first argument is Job.
You'll get type errors, you can:
- cast `{ ... job } as JobRow`
- or have single return statement:
```
return {
   ...job,
   date_started: ...
}
```

src/app/pages/common/entity/entity-table/entity-table.component.ts
```diff
@@ -1188,4 +1188,14 @@ export class EntityTableComponent<Row = any> implements OnInit, AfterViewInit, O
...
+
+  isBasicColumnTemplate(column: string): boolean {
+    return !['expandedDetail', 'action', 'multiselect', 'expansion-chevrons'].includes(column) && !this.isInteractive(column);
+  }
+
+  isCustomColumnComponent(column: string): boolean {
+    const item = this.currentColumns.find((obj) => obj.prop === column);
+
+    return Boolean(!this.isBasicColumnTemplate(column) && item?.widget?.component && column == 'method');
+  }
```
undsoft, line 1200: I wonder if just not using entity-table at all and using plain table would have been easier. Up to you though.

undsoft (changes requested):
1. Empty state looks funny with two History buttons.
2. Sometimes when filtering and removing search query, table remembers 0 results.

https://user-images.githubusercontent.com/1481270/126779274-705c74e3-76c7-40eb-9492-be71085daf38.mov

3. Three dot menu with Edit/Delete options needs removal.
4. @dszidi Looking at design, I don't remember why we needed both Result and State columns.

src/app/pages/common/entity/entity-table/entity-table.component.ts
denysbutenko: That's a good idea. A lot of headaches are caused using entity-table on this page.

src/app/pages/common/entity/entity-table/entity-table.component.html
denysbutenko: Makes sense. If a job is abortable the button will appear to stop it in case.

denysbutenko:
ready for another round of testing

dszidi:
The toggle buttons and search field should be in the page title component to match the rest of the UI

image

undsoft (changes requested):
The code looks good. 
In terms of testing some additional nitpicks:
- Would be good to have some padding in dates, it's possible to get page in weird state. Also on screenshot there is plenty of space in `Name` column.
Снимок экрана 2021-07-28 в 10 39 20

- When first searching for something and then filtering via buttons, the search query is not used.
- Only some columns are sortable, but all of them show indication that they can be sorted. Can we remove the indicator or actually allow user to sort these columns?
- When hovering over red failed icon, would be good to see default mouse pointer or the one with small question mark, not select text.
- Spinners are not centered when job is running.

https://user-images.githubusercontent.com/1481270/127284014-74d0ffef-5647-4e2c-bae1-eafa9d10876a.mov

denysbutenko:
Thanks for the feedback! All items are resolved

dszidi:
The magnifying glass icon in the search field is off.
I found that if you remove the font-size rule from the scss file the problem goes away.

image


image

dszidi:
Adding the top and position rules to the reset button should also fix it's vertical position...

image

undsoft (approved):
Don't see any other issues. Nice work!

denysbutenko:
> Adding the top and position rules to the reset button should also fix it's vertical position...
> 
> image
> 
> EDIT: Alternatively, instead of building the search field from scratch, you could use entity-toolbar and reuse the filter component.
> We're using this for search fields elsewhere in the UI already.
> 
> Check out lines 72-109 from the applications component for a good example...
> 
> https://github.com/truenas/webui/blob/master/src/app/pages/applications/applications.component.ts

Good catch! Fixed to reuse entity-toolbar for search field 👍🏻

PR #5671, opened 2021-07-17
NAS-111433 / 21.08 / Adding no-implicit this to introduce stricter checks.

undsoft:
+ no-impilicit-this for stricter code
+ appLet directive https://medium.com/@AustinMatherne/angular-let-directive-a168d4248138

No testing is necessary.


RehanY147 (approved):
Looks good

PR #5675, opened 2021-07-18
NAS-111118 / 21.08 / Fixed tooltip

RehanY147:
NAS-111118


undsoft (approved):
The text says brX now.

PR #5677, opened 2021-07-19
NAS-111442 / 21.08 / Added icon to clickable card titles, added run now action to cron jobs table

RehanY147:
NAS-111442


dszidi (approved):
Code looks good and new links work as expected.

PR #5680, opened 2021-07-20
NAS-111476 / 12.0 / Passive widgets also need to listen for HA_Status

dszidi:
To test this, simply make sure that System Information widget for standby controllers consistently populate with content.


undsoft (approved):
The widget is there.
Снимок экрана 2021-07-21 в 11 08 06

@dszidi Does this need to be forwardported?

PR #5685, opened 2021-07-21
NAS-111460 / 21.08 / Fixed bug with wrong reassignment of variable

RehanY147:
NAS-111460

To Test
1. Open the GUI Settings form in the System -> General Settings page
2. On the form, previous data should load Ok and you shouldn't see empty fields


AlexKarpov98 (approved):
LGTM

PR #5686, opened 2021-07-22
NAS-111287 / 21.08 / Fixed clone config and value set

RehanY147:
NAS-111287

To test
1. Go to System -> Boot page
2. Click on the three dots button for a boot env
3. Click on the clone option
4. The clone form should show up and the `source` field should be filled in


dszidi (approved):
Looks good and works as expected

image

PR #5689, opened 2021-07-23
NAS-111478 / 12.0 / NAS-111478 - Fixing redundant error modal

undsoft:
Testing instructions are a bit complicated. I suggest just approving based on ticket description and code, it's pretty clear what the intention was.


AlexKarpov98 (approved):
@undsoft LGTM

PR #5690, opened 2021-07-23
NAS-111428 / 21.08 / NAS-111428 - Confirmation when changing some of the Kubernetes settings

undsoft:
Go to Apps -> Settings -> Advanced Settings, observe warning in form and confirmation modal if some of the settings are changed.


RehanY147 (approved):
LGTM. Although, something to consider would be changing the warning text color. As it is, it doesn't draw user's attention as a "warning". Just looks like more sub-text that's just a description. Not something that NEEDS the user's attention.

undsoft:
This should be okay, as there is a extra confirmation modal if you change one of the settings.

PR #5693, opened 2021-07-27
NAS-111203 / 21.08 / NAS-111203 - Handle app upgrade errors

undsoft:
The error cannot really be reproduced besides simulating in code, so here's a video of it working.

https://user-images.githubusercontent.com/1481270/127135479-e5a58e1e-c5ba-4045-8cb7-26af0b3811d3.mov


RehanY147 (approved):
Looks good

PR #5697, opened 2021-07-28
NAS-111443 / 21.08 / NAS-111443 - Fixing incorrect width of the terminal

undsoft:
* systemprocessed module removed as there are links leading to it.
To test: 
1. Open Shell in settings or shell in one of the Apps.
2. Type `mc` and make sure everything fits the screen. Previously there were a couple of characters on the left that wouldn't fit. If no `mc` is available, just type random stuff in terminal and see where it wraps.


AlexKarpov98 (approved):
@undsoft Perfect!

PR #5702, opened 2021-07-28
NAS-111202 / 21.08 / Properly show which image(s) would be upgraded in UI on app upgrade

AlexKarpov98:


RehanY147:
@AlexKarpov98 How to test this?

RehanY147:
The code looks good

src/app/pages/applications/chart-releases/chart-releases.component.ts
```diff
@@ -330,12 +330,29 @@ export class ChartReleasesComponent implements OnInit, OnDestroy {
...
+      msg += `<div><strong>${helptext.charts.upgrade_dialog.changeLog}</strong></div>`;
+      if (res.changelog) {
+        msg += res.changelog;
+      } else {
+        msg += helptext.charts.upgrade_dialog.noChangeLogMsg;
+      }
       this.dialogService.confirm({
-        title: helptext.charts.upgrade_dialog.title + res.latest_human_version,
-        message: res.changelog,
+        title: `${helptext.charts.upgrade_dialog.title} ${name}`,
```
undsoft, line 354: Joining strings like on this line won't let other people change order of words when translating.
Better to translate like this:
```
title: this.translate.instant('Update {appName}', { appName: name })
```

src/app/pages/applications/chart-releases/chart-releases.component.ts
```diff
@@ -330,12 +330,29 @@ export class ChartReleasesComponent implements OnInit, OnDestroy {
...
+        }
+      } else {
+        msg += `${helptext.charts.upgrade_dialog.noImageUpdate}`;
+      }
+      msg += `<div><strong>${helptext.charts.upgrade_dialog.changeLog}</strong></div>`;
+      if (res.changelog) {
+        msg += res.changelog;
+      } else {
+        msg += helptext.charts.upgrade_dialog.noChangeLogMsg;
+      }
```
undsoft, line 352: I'm on the fence about joining strings like this in the component code.
I think it may be better to just add another component for this.
You can create something like UpdateDialogComponent, open it with `this.matDialog.open`, receive `updateSummary` in `@Inject(MAT_DIALOG_DATA)` and return values in `dialogRef.close(true/false)`.

Then you can construct the message cleanly in an html file.

For examples, see `SelectPresetModalComponent` and how it's being called in `DatasetAclEditorComponent.onUsePresetPressed`

undsoft (changes requested):

undsoft (commented):
The functionality appears to be working correctly, however I could not find an app that would actually have anything in images to update or changelog.
Waqar had offered some help, maybe he can test more fully.

src/app/pages/applications/dialogs/chart-upgrade/chart-upgrade-dialog.component.scss
```diff
@@ -0,0 +1,94 @@
+::ng-deep chart-upgrade-dialog {
```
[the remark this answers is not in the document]
undsoft: Also, before I forget can you try overriding those styles with more specificity and not by using `!important`?

sonicaj (changes requested):
We are not handling the 3 cases I pointed out in my last comment, let's please do that.

sonicaj (commented):
It looks good to me, i have some cosmetic changes:

1) We are not showing the version name appropriately ( on the test system you shared with me i am testing the latest changes )

Screenshot 2021-08-05 at 12 25 48 AM

2) Let's please remove the colon at the end of `Only image(s) will be updated:` sentence

sonicaj (approved):
Thanks!

AlexKarpov98:
@undsoft @RehanY147 This PR is not easy to test, and @sonicaj have already tested the functionality of this PR.
So I think you just need to review the code. Thanks

RehanY147 (approved):
LGTM

src/app/pages/applications/dialogs/chart-upgrade/chart-upgrade-dialog.component.scss
```diff
@@ -0,0 +1,96 @@
+:host {
+  font-size: 14px !important;
+}
+
+::ng-deep .mat-dialog-content {
```
undsoft, line 5: This will affect any `mat-dialog-content` in the system.
Can we have some way of narrowing this down, i.e. by having some class on elements?

src/app/pages/applications/dialogs/chart-upgrade/chart-upgrade-dialog.component.scss
```diff
@@ -0,0 +1,96 @@
+:host {
+  font-size: 14px !important;
```
undsoft, line 2: Please do your best to try using more specificity to avoid using `!important` here or below.

undsoft (approved):
Two minor things, otherwise I'm happy with the code as long as Waqar is happy with functionality.

PR #5704, opened 2021-07-29
NAS-111391 / 21.08 / NAS-111391 - Various smaller fixes for permissions

undsoft:
1. View permissions sidebar now appears correctly on long lists.
2. Create dataset, click View Permissions, Set ACL, pressed Use Permissions Editor. The form should appear correctly.
3. Updated how user string is shown, you can just check that user aces appear correctly.


denysbutenko (approved):
image

Work as described 👍🏻

PR #5705, opened 2021-07-29
NAS-111575 / 21.08 / Fixed IP Address label and title formatting

RehanY147:
NAS-111575


src/app/core/components/widgets/widget-network/widget-network.component.ts
```diff
@@ -251,7 +251,7 @@ export class WidgetNetworkComponent extends WidgetComponent implements AfterView
   }
 
   getIpAddress(nic: BaseNetworkInterface): string {
-    let ip = T('IP Addresses: 0');
+    let ip = T('0');
```
undsoft, line 254: Don't think we need to translate zero :)

undsoft (approved):
Works as intended. 
Снимок экрана 2021-07-30 в 10 24 32

PR #5706, opened 2021-07-29
NAS-111469 / 21.08 / Add inline options for form-radio.

dszidi:
Testing:
- ACL editor should have inline radio buttons for the "ACL Type", "Permissions Type" and "Flags Type" fields.
- Make sure other forms with radios keep column layout by default. (Import disk form is a good one to test)
- When browser window gets too narrow, inline radio buttons should revert to column layout


RehanY147 (approved):
Looks good. See inline radio groups.

PR #5707, opened 2021-07-30
NAS-111205 / 21.08 / NAS-111205 - Treating start/stop app as a job

undsoft:
Starting and stopping an app is now a job with corresponding modal.


dszidi (approved):
Code looks good and app start/stop behaves as expected.

PR #5710, opened 2021-07-30
NAS-111565 / 21.08 / Handled new response from vm.query subscription

RehanY147:
NAS-111565


undsoft (approved):
One thing that I've noticed is that sometimes when I expand and press Start, the state toggle doesn't change.
Don't know if this is a regression or not.

sonicaj (commented):
@RehanY147 when we start the VM, the state status does not toggle for me as well. I think what happens is that when UI calls `vm.start`, before the call returns vm starts and we have an event emitted but as UI is still at the job dialog, we miss that event and the state is not toggled. If the VM is stopped/powered off, i see the state reflecting the current status nicely. Let's please handle this case as well. Thanks!

sonicaj (approved):

undsoft (approved):

PR #5714, opened 2021-08-01
Remove cancel button from replication wizard

dszidi:
This PR affects the replication wizard found on the data-protection dashboard.
All the other forms on the dashboard omit cancel buttons and instead allow users to either click outside of the side panel or click the "X" close button at the top right of the panel.

To test:
- Make sure there is no cancel button on the replication wizard.
- The wizard should close when the user clicks the close button or clicks outside the form.


denysbutenko (approved):
image

Looks good!

PR #5715, opened 2021-08-01
NAS-111470 / 21.08 / Added FormSelectionListComponent, checkbox groups for NFS and POSIX ACL Editors

RehanY147:
NAS-111470

To test, go to Storage -> Dataset three dots menu -> View Permission -> Edit Icon -> Set ACL button

Test the Advanced permissions and flags for NFS and Permissions for POSIX forms


undsoft (changes requested):
1. The checkboxes are yellow for some reason.
2. We have this `default` checkbox in posix acl, which kinda collates with other checkboxes now. Can you add a label called "Flags" above `default`?
Снимок экрана 2021-08-02 в 10 35 31
3. While you are there, can you also change basic flags to radio?

undsoft (approved):
Looks good, thanks!

PR #5717, opened 2021-08-02
NAS-111392 / 21.08 / NAS-111392 - Fix SMART tests page

undsoft:
Testing: 
1. Find an instance that supports smart testing (https://10.238.15.92)
2. Go to Storage -> Disks
3. Ensure that existing smart tests appear correctly for disks that have them.
4. Ensure that there is no broken Add test result in list of results.


dszidi:
Test checks out. One UX nitpick though is that the disk name is not shown in the results anywhere.
Technically it's in the URL but still, it would be better if it was listed in the result

image

undsoft:
Updated, I had to introduce a new service to be able to change the title.

dszidi (approved):
Code looks good and test results display correctly. Disk name is integrated into the test page title.

PR #5719, opened 2021-08-02
NAS-111617 / 21.08 / Apply flex layout to radio button and tooltip

dszidi:
Testing: When present, tooltip component (?) icon should be inline with radio buttons. eg. `system/email` form


denysbutenko (approved):
image

PR #5726, opened 2021-08-03
NAS-111414 / 21.08 / Treat removing vdev in the UI as a job

denysbutenko:
How to test:

- Go to the Storage page
- Create a new pool with 2 disks and pick `STRIPE`
- Click on pool options and choose `Status`
- Click the three dots menu on disk and choose `Remove`

You'll see something like:
image


undsoft (approved):
I can see the job dialog.

PR #5727, opened 2021-08-03
NAS-111623 / 21.08 / Implement entity-empty on network widget

dszidi:
To Test:
- errors should show in place of charts when reporting.get_data returns errors
- There should be an instance of entity-empty for every nic that doesn't have chart data.
- Other pages with entity-empty that aren't configured with compact to be true, should behave as normal.


AlexKarpov98 (approved):
@dszidi LGTM

PR #5728, opened 2021-08-04
NAS-111284 / 21.08 / NAS-111284 - Do not allow root dataset to be edited

undsoft:
Two things to test:
1. Root dataset in every pool should not have pencil edit icon in view permissions sidebar.
2. If you create dataset with acltype = OFF, and then go to edit permissions, you should not see Set ACL button. You should still see this button on datasets with other acltype.


denysbutenko (approved):
I was confused why I couldn't edit permissions for root datasets. It's clear now :)

image
image

Works as described. Well done 👍🏻

PR #5730, opened 2021-08-04
NAS-111227 / 21.08 / NAS-111227 - Refresh app list after app update

undsoft:
1. Find an app that has an update.
2. Remember current version.
3. Update, make sure that list is updated.
4. Rollback, make sure that list is updated.


RehanY147 (approved):
The code looks good. But couldn't find any upgrade-able app

undsoft:
There are some on our cloud vm instance.

RehanY147 (approved):
Looks good. On upgrade and rollback versions update automatically without refresh.

PR #5731, opened 2021-08-04
NAS-111668 / 21.08 / Fix free space on storage widget

denysbutenko:
Free space value is quite different on Dashboard and Storage pages

### DIFF
--- a/src/app/core/components/widgets/widget-storage/widget-storage.component.ts
+++ b/src/app/core/components/widgets/widget-storage/widget-storage.component.ts
@@ -251,8 +251,8 @@ export class WidgetStorageComponent extends WidgetComponent implements OnChanges
         usedValue = filesize(vol.used, { exponent: 3 });
       }
 
-      if (usedValue != 'Locked') {
-        displayValue = this.getSizeString(vol.avail - vol.used);
+      if (usedValue !== 'Locked') {
+        displayValue = this.getSizeString(vol.avail);
       }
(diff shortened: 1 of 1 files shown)


undsoft:
Eh, can you confirm with someone that this is a proper way to get free space?
The numbers do match in webui now, but if I do `zpool list` I will get a completely different number.

denysbutenko:
There is a difference due to reserved space for critical ZFS operations in `zpool list` results. You may want to use `zfs list` in order to get the correct values of usable space. The WebUI uses the same endpoint `pool.dataset.query` on both pages (Storage and Dashboard) to get volume data.



An example on my usage:
```
> zpool list
NAME        SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
LIBRARY    7.27T  2.62T  4.65T        -         -     0%    35%  1.00x    ONLINE  /mnt
```

```
> zfs list
NAME        USED  AVAIL     REFER  MOUNTPOINT
LIBRARY    2.62T  4.53T       96K  /mnt/LIBRARY
```


More details https://serverfault.com/a/796798 and https://unix.stackexchange.com/questions/582005/zpool-list-vs-zfs-list-why-free-space-is-10x-different/582383

undsoft (approved):
Works correctly then, thanks for the explanation.

PR #5733, opened 2021-08-05
NAS-111653 / 21.08 / Can't create a user in SCALE Confirm Password * Field was not expected

AlexKarpov98:


undsoft (changes requested):
Okay, so I got curious about why such a basic thing would break.
It turns out there is some dynamic calls to `clean_uid` (now `cleanUid`) that broke when I renamed some methods.
Basically, you need to update:
```
src/app/pages/common/entity/entity-form/entity-form.component.ts:469
src/app/pages/common/entity/entity-form/entity-form-embedded.component.ts:280
```
to make them call `cleanUid`.

When doing so, also please add a TODO note there for us to remove this functionality.

denysbutenko (approved):
It works as described 👍🏻

undsoft (approved):
Thanks!

PR #5736, opened 2021-08-06
NAS-111662 / None / NAS-111662: Improving type safety - 2

undsoft:
No testing is necessary.


src/app/interfaces/user.interface.ts
```diff
@@ -36,3 +36,8 @@ export interface UserGroup {
   bsdgrp_sudo_commands: { [property: string]: any }[];
   bsdgrp_smb: boolean;
 }
+
+export type DeleteUserParams = [
+  /* id */ number,
+  /* params */ { delete_group: true },
```
AlexKarpov98, line 42:
```suggestion
  /* params */ { delete_group: boolean},
```

src/app/interfaces/api-directory.interface.ts
```diff
@@ -793,12 +790,12 @@ export type ApiDirectory = {
   'user.update': { params: any; response: any };
   'user.create': { params: any; response: any };
   'user.query': { params: QueryParams<User>; response: User[] };
-  'user.set_root_password': { params: any; response: any };
-  'user.delete': { params: any; response: any };
+  'user.set_root_password': { params: [/* password */ string]; response: void };
+  'user.delete': { params: DeleteUserParams; response: void };
```
AlexKarpov98, line 794:
```suggestion
  'user.delete': { params: DeleteUserParams; response: number};
```

src/app/core/services/api.service.ts
```diff
@@ -35,7 +36,7 @@ export class ApiService {
         clone.args = [[['id', '=', 1]]];
         return clone;
       },
-      postProcessor(res: any) {
+      postProcessor(res: User[]) {
```
AlexKarpov98, line 39: I am not sure it is correct. Maybe `any` is proper rather than `User[]`.
I guess the below api docs tell us it may have any of `object,` `array` and `integer`.
image

AlexKarpov98 (changes requested):

src/app/interfaces/api-directory.interface.ts
```diff
@@ -290,11 +297,11 @@ export type ApiDirectory = {
   'cloudsync.list_directory': { params: any; response: any };
   'cloudsync.update': { params: any; response: any };
   'cloudsync.create': { params: any; response: any };
-  'cloudsync.sync': { params: any; response: any };
-  'cloudsync.abort': { params: any; response: any };
+  'cloudsync.sync': { params: [/* id */ number, { dry_run: boolean }?]; response: any };
```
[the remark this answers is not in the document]
undsoft: Whenever api call returns a number this is a job id, for actual result you'd need to look into job result, that's why it's any for now.

src/app/interfaces/api-directory.interface.ts
```diff
@@ -290,11 +297,11 @@ export type ApiDirectory = {
...
+  'cloudsync.abort': { params: [/* id */ number]; response: boolean };
```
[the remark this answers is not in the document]
undsoft: Why void though? The api returns boolean.

src/app/core/services/api.service.ts
undsoft: Okay, so I talked to Waqar, it seems you can specify additional parameters in query api and get a single object. It's not really used on frontend and not currently supported in typings. I'll improve them in future PRs.

src/app/interfaces/api-directory.interface.ts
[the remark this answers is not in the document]
undsoft: Okay, I see the source of confusion. Actual api returns boolean, api documentation is not correct. 
Снимок экрана 2021-08-07 в 11 55 04

undsoft:
Fixed some of the issues.

AlexKarpov98 (approved):
@undsoft LGTM

PR #5737, opened 2021-08-06
NAS-111181 / 12.0 / Added Boolean filter to check if password check was successful

RehanY147:
NAS-111181


undsoft (approved):
I can confirm that pool password is verified on webui before letting user continue.
However as discussed, this will need to be solved better in the future.
Make sure to create a ticket for middleware before merging.

RehanY147:
NAS-111750 is the ticket number on Jira for the middleware issue

PR #5738, opened 2021-08-06
NAS-110863 / 21.08 / Adjusted height

RehanY147:
NAS-110863


denysbutenko:
Looks like a magic number. Can we try to find better approach to solve this? For example something like "$header-height: n; height: calc(100% - $header-height)"?

denysbutenko:
image
In this case, it would be:
```
$page-title-height: 97px;
height: calc(100% - $page-title-height);
```

denysbutenko (approved):

RehanY147:
@denysbutenko Made the suggested changes. Thanks! Please review again

denysbutenko (approved):
image
Looks good 👍🏻

PR #5743, opened 2021-08-08
NAS-111742 / 21.08 / Removed invisible EntityEmpty for empty shares tables

RehanY147:
NAS-111742

Ticket description has a demonstration of the issue. The issue was when ui attempted to calculate the dynamic rows limit for the tables, it overestimated the occupied table height due to EmptyEntity showing unexpectedly and figured no space left for any rows.


undsoft (approved):
The table appears correctly when iscsi is added.

PR #5745, opened 2021-08-09
NAS-111734 / 21.08 / Show resources that depend on the dataset before deleting

denysbutenko:
How to test?

- Create a new dataset
- Create share using the dataset
- Choose 'Delete Dataset' on the Storage page
- Make sure that share is listed in the confirmation dialog

Example:
image


dszidi (approved):
Code looks good. Dataset deletion behaves as expected.

PR #5747, opened 2021-08-09
NAS-111431 / 12.0 / Enclosures should listen for global resilver events

dszidi:
Testing:
- Remove a disk from a pool that has a spare
- Put the disk back
- Make sure the resilver icon shows up on the topbar
- The disk should reappear and should be presented as part of the pool again AFTER the resilver process completes


jervin777 (commented):
Tested this on 8/6.  Pool with spares used.  Drive pulled from Pool, Enclosure Management showed drive removed, pool degraded and spare activated.  Reinserted drive, drive showed attached to pool.

jervin777 (approved):

PR #5751, opened 2021-08-09
NAS-111576 / 21.08 / [scale/apps] Specific field in charts are not remembered on edit

AlexKarpov98:


dszidi (approved):
Code looks good. Tested branch and hostname field is present and populated as per ticket description.

PR #5755, opened 2021-08-10
NAS-111755 / 21.08 / Job should call error method when job is FAILED

dszidi:
Steps to reproduce.
1) Create VM with two pools
2) SSH into server and `cd /var/db/system/samba4`
3) Move system dataset to a different pool via System/Advanced page
4) Form should show dialog with traceback message (as opposed to silently failing like it does currently)

### DIFF
--- a/src/app/services/ws.service.ts
+++ b/src/app/services/ws.service.ts
@@ -246,9 +246,8 @@ export class WebSocketService {
         this.subscribe('core.get_jobs').pipe(untilDestroyed(this)).subscribe((event) => {
           if (event.id == job_id) {
             observer.next(event.fields);
-            if (event.fields.state === JobState.Success || event.fields.state == JobState.Failed) {
-              observer.complete();
-            }
+            if (event.fields.state === JobState.Success) observer.complete();
+            if (event.fields.state === JobState.Failed) observer.error(event.fields);
           }
         });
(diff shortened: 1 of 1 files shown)


undsoft (approved):
As they say on the internet: task failed successfully.

denysbutenko:
I just wondering if it's required for other places like that? https://github.com/truenas/webui/blob/c3bd8656613a6a27f777aa804a8093e94b42fc0f/src/app/services/job.service.ts#L34

dszidi:
> I just wondering if it's required for other places like that?
> 
> https://github.com/truenas/webui/blob/c3bd8656613a6a27f777aa804a8093e94b42fc0f/src/app/services/job.service.ts#L34

I was wondering that as well. 
However, I asked William if there was some distinction between failed and error and he pointed me to the job.py file. Unlike our JobState enum, middleware states only include FAILED (there is no ERROR) so any errors would come in as FAILED.
Based on that I would assume that any FAILED job should be treated as an error.  If there are cases where FAILED is not an error case, then something should probably be changed elsewhere. 

Link to mw file: https://github.com/truenas/middleware/blob/d006ad5e7a2fe0d552cf6979caac7b7c54db39a3/src/middlewared/middlewared/job.py

PR #5757, opened 2021-08-11
NAS-111431 / 21.08 / Enclosure should listen for Resilver events

dszidi:
FYI: @denysbutenko you don't need to test this.
Jeff will do the testing once I get an environment set up for him.

Testing:

Remove a disk from a pool that has a spare
Put the disk back
Make sure the resilver icon shows up on the topbar
The disk should reappear and should be presented as part of the pool again AFTER the resilver process completes

Original PR: #5747
Jira URL: https://jira.ixsystems.com/browse/NAS-111431


src/app/interfaces/resilver-job.interface.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface ResilverScan {
+  bytes_issued: number;
+  bytes_processed: number;
+  bytes_to_process: number;
+  end_time: ResilverTime;
+  errors: number;
+  function: string;
+  pause: number | null;
+  percentage: number;
+  start_time: ResilverTime;
```
denysbutenko, line 14: Looks like `ResilverTime` is redundant and can be replaced with `ApiTimestamp`
https://github.com/truenas/webui/blob/647edd15eceb53a1cac5edaf3118610d15f0ad6d/src/app/interfaces/api-date.interface.ts#L1-L3

src/app/pages/system/view-enclosure/view-enclosure.component.ts
```diff
@@ -148,6 +149,10 @@ export class ViewEnclosureComponent implements OnDestroy {
       this.system.sensorData = evt.data;
     });
 
+    core.register({ observerClass: this, eventName: 'Resilvering' }).pipe(untilDestroyed(this)).subscribe((evt: ResilverEvent) => {
+      if (evt.data.scan.state == 'FINISHED') this.fetchData();
```
denysbutenko, line 153: It would be better to use an enum here.
Probably, `PoolScanState` is most suitable in this case.
https://github.com/truenas/webui/blob/443a13760275f4f78b07aeedf96baf59459b3690/src/app/enums/pool-scan-state.enum.ts#L1-L4

We also have used `JobState.Finished` for some unknown reason here:
https://github.com/truenas/webui/blob/db68e1b4172d2880dc81769a6f21765503fd70a2/src/app/components/common/topbar/topbar.component.ts#L232-L237

src/app/interfaces/resilver-job.interface.ts
```diff
@@ -0,0 +1,27 @@
...
+  state: string;
```
denysbutenko, line 15: update type to `PoolScanState` accordingly if it is suitable

src/app/interfaces/resilver-job.interface.ts
dszidi: Nice catch. Will do

src/app/pages/system/view-enclosure/view-enclosure.component.ts
dszidi: Will replace string with enum

PR #5765, opened 2021-08-13
NAS-111803 / 21.08 / NAS-111803 - Allow ownership information to be changed in an ACL Editor.

undsoft:
Ownership information can now be updated in an acl editor.

Go to Acl Editor on datasets with NFS and POSIX entries and play around with ACL editor, especially with user and group fields.


src/app/pages/storage/volumes/permissions/stores/dataset-acl-editor.store.ts
```diff
@@ -312,54 +313,102 @@ export class DatasetAclEditorStore extends ComponentStore<DatasetAclEditorState>
...
+            return { ...aceAttributes, id };
+          }
+          if ([NfsAclTag.UserGroup, PosixAclTag.Group].includes(ace.tag)) {
+            const id = groupWhoToIds.has(ace.who) ? groupWhoToIds.get(ace.who) : -1;
+            return { ...aceAttributes, id };
+          }
+
+          return {
+            ...aceAttributes,
+            id: -1, // -1 is effectively null for middleware
```
RehanY147, line 396: Is there a specific reason why we're sending `-1` instead of `null` here? If not, it just looks like it'll add unnecessary confusion. If we mean to send `null`, we should just send `null`

dszidi:
The layout could use a little improvement.
If we kept the new fields inline it would save space a look a little tighter too.
Also I was thinking perhaps we should be using "Access Control Entry" instead of "List" on the right.
Here's a screenshot for an example.

image



Otherwise everything seems to work just fine in testing so far

src/app/pages/storage/volumes/permissions/stores/dataset-acl-editor.store.ts
undsoft: Unfortunately `null` is not supported :)

undsoft:
@dszidi 
Updated text, thanks.
As for layout, while I agree that current layout is not ideal, I can't say that I like the proposed change more. It does utilize space more efficiently, but it kinda breaks the flow of the page. 
The left side is currently for things that relate to acl as a whole: buttons, checkboxes that affect saving, while the right side is for narrowing down to an individual ace. 
That's the reason I want to keep ownership dropdowns somewhere on the left, as they feel more 'global'.

I've also tried putting them in the sidebar, but that has its own problems.
Снимок экрана 2021-08-16 в 10 58 36

So far I've adjusted margins a little bit.

dszidi:
LGTM!

dszidi (approved):

RehanY147 (approved):
Looks good

PR #5768, opened 2021-08-13
fix translation

Sagit-chu:
fix and add translation


denysbutenko (approved):
Looks good. Thanks for the contribution!

PR #5769, opened 2021-08-13
NAS-111717 / 21.10 / Layout improvements for Shares dashboard

denysbutenko:
Changes:

- Allow toggling boolean values directly on the table
- Allow hiding empty column (enabled for description column only)
- Improve type safety

Preview:

| Before  | After |
| ------------- | ------------- |
|image|image|


undsoft:
I don't know if allowing users to quickly change `Change User & Group Ownership` is a good idea. It sounds like an invasive operation that you don't want to be doing a) accidentally b) often.

denysbutenko:
> I don't know if allowing users to quickly change `Change User & Group Ownership` is a good idea. It sounds like an invasive operation that you don't want to be doing a) accidentally b) often.

That's a good point. What if we display the confirmation dialog before changes? For example, enable/disable keep as is and show dialog for fields like `Change User & Group Ownership`. Alternatively, we can show read-only checkbox for certain fields.

undsoft:
I just don't know what the use case of changing this information on a dashboard is. We can have a read only checkbox, as long as it's visually distinct from other checkboxes and it's clear that it's not clickable.

denysbutenko:
image
Move read-only checkboxes to the first column. The read-only checkbox has 0.75 opacity of original

RehanY147:
It still doesn't look different enough to my eyes. Maybe we could try a different color. More greyed out?

denysbutenko:
Sure, I'll take a look 👍

denysbutenko:
@RehanY147 It may be tricky to change color for the checkbox it depends on the selected theme. so I reduced opacity from 0.75 to 0.4. How does it look now?
iX Dark:
image
iX Light:
image
Dracula:
image

dszidi:
IMO a better alternative to semi opaque checkbox is to use a check mark with no square border around it. 
Take a look at the checks list on this PR for example. It is a green check mark and there is no confusion as to whether or not it is clickable

image

RehanY147:
I like @dszidi's idea as well.

src/app/helptext/sharing/webdav/webdav.ts
```diff
@@ -26,12 +26,14 @@ export const helptext_sharing_webdav = {
...
-  column_name: T('Share Name'),
+  column_name: T('Name'),
   column_comment: T('Description'),
   column_path: T('Path'),
   column_enabled: T('Enabled'),
   column_ro: T('Read Only'),
-  column_perm: T('Change User and Group Ownership'),
+  column_perm: T('Ownership'),
+
+  column_perm_tooltip: T('Change User and Group Ownership'),
```
RehanY147, line 36: Now that the checkbox is disabled, the wording here should change from `Change User and Group Ownership` to just `User and Group ownership`

src/app/pages/common/entity/table/table.component.ts
```diff
@@ -21,10 +21,30 @@ export interface AppTableHeaderAction {
...
+  hidden?: boolean;
+  listview?: boolean;
+  getIcon?(element: any, prop: string): void;
+}
+
 export interface AppTableConfig<P = any> {
   title?: string;
   titleHref?: string;
-  columns: any[];
+  columns: AppTableColumn[];
```
RehanY147, line 47: The could just be called `TableColumn`

src/app/pages/sharing/components/shares-dashboard/shares-dashboard.component.ts
```diff
@@ -605,4 +621,24 @@ export class SharesDashboardComponent implements AfterViewInit {
         return 'fn-theme-orange';
     }
   }
+
+  hideColumnIfEmpty(column: string, type: ShareType, data: SmbShare[] | WebDavShare[] | NfsShare[]): void {
+    switch (type) {
+      case ShareType.SMB:
+        this.smbTableConf.columns.find((item) => item.prop === column).hidden = !(data as SmbShare[]).some((item) =>
```
RehanY147, line 628: We should break this checks down for better readability and understanding.

src/app/pages/sharing/components/shares-dashboard/shares-dashboard.component.ts
```diff
@@ -605,4 +621,24 @@ export class SharesDashboardComponent implements AfterViewInit {
...
+          item[column as keyof SmbShare]?.toString()?.trim());
+        break;
+      case ShareType.WebDAV:
+        this.webdavTableConf.columns.find((item) => item.prop === column).hidden = !(data as WebDavShare[]).some(
```
RehanY147, line 632: Here too

src/app/pages/sharing/components/shares-dashboard/shares-dashboard.component.ts
```diff
@@ -605,4 +621,24 @@ export class SharesDashboardComponent implements AfterViewInit {
...
+          (item) => item[column as keyof WebDavShare]?.toString()?.trim(),
+        );
+        break;
+      case ShareType.NFS:
+        this.nfsTableConf.columns.find((item) => item.prop === column).hidden = !(data as NfsShare[]).some((item) =>
```
RehanY147, line 637: Same here

src/app/pages/common/entity/table/table.component.ts
denysbutenko: It may be confusing when it is associated with `TableComponent` or `EntityTableComponent`. The other interfaces in this file also have the prefix.

src/app/pages/sharing/components/shares-dashboard/shares-dashboard.component.ts
denysbutenko: That's a good point. I moved this feature to the `TableComponent` itself in the next PR. Instead of implementing for each dashboard, we'll need to pass `hiddenIfEmpty: true` on column config, and the table component will handle it.

denysbutenko:
> IMO a better alternative to semi opaque checkbox is to use a check mark with no square border around it.
> Take a look at the checks list on this PR for example. It is a green check mark and there is no confusion as to whether or not it is clickable
> 
> image

I agree with you. Created an improvement for this: https://jira.ixsystems.com/browse/NAS-111919

src/app/helptext/sharing/webdav/webdav.ts
denysbutenko: Nice catch. Updated

dszidi (approved):
Code looks fine. Behaves as expected. LGTM

PR #5771, opened 2021-08-15
Replaced TooltipComponent with PopperTooltipComponent

RehanY147:
NAS-111064

Tooltips in different places should be tested (e.g., dialogs, side bars, forms, wizards). Tooltips should now auto adjust their placement depending on the space available.
Tooltip position and TooltipDocReplacePipe should also be testsed.


src/app/pages/common/entity/entity-form/components/popper-tooltip/popper-tooltip.component.ts
```diff
@@ -0,0 +1,11 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'popper-tooltip',
```
undsoft, line 4: I am also okay with us just keep calling it `tooltip`.

undsoft (changes requested):
1. Can we adjust tooltip styles to make them more aligned with old tooltips? The top margin is too big, and border rounding is too small.
2. The font of the tooltip seem to be dependent on parent element, i.e. if you open a tooltip next to a label, the text will be small.
3. One of the reason of using this library is to prevent situations like this: 
Снимок экрана 2021-08-16 в 12 08 25. There is a setting in ngx-popper that allows tooltip to be attached to body. I think it's appendTo. Let's try it out, either individually or as a default module setting.
4. You may want to search for scss files for 'tooltip', for example here the left margin disappeared   
Снимок экрана 2021-08-16 в 12 08 25

AlexKarpov98:
@RehanY147 It does not work well. 
tooltip

PR #5772, opened 2021-08-15
NAS-111123 / 21.08 / Creating SSH Connections using the new middleware

RehanY147:
NAS-111123

The ticket description has all the bug details and reproduction steps. Using the new middleware end-point to create SSH Connections.

To test
- Following the reproduction steps, a new unused keypair should not have been created (rollback should be working in case of an error)
- On successful submission of the form, new ssh connection and new keypair should be created
- Edit should be tested for regressions


src/app/interfaces/api-directory.interface.ts
```diff
@@ -388,6 +388,7 @@ export type ApiDirectory = {
   'keychaincredential.remote_ssh_host_key_scan': { params: any; response: any };
   'keychaincredential.delete': { params: any; response: any };
   'keychaincredential.remote_ssh_semiautomatic_setup': { params: any; response: any };
+  'keychaincredential.setup_ssh_connection': { params: any; response: any };
```
undsoft, line 391: Did you mean to add `SshConnectionSetup` or `[SshConnectionSetup]` in params here?

undsoft (approved):
Works well, don't see any issues with how form behaves.

PR #5774, opened 2021-08-16
NAS-111465 / 21.08 / NAS-111465 - Redirect to an ACL Editor after home share is created

undsoft:
1. Use cloud vm for testing. I've hot-pached middleware there with latest changes.
2. Create datasets with POSIX and NFS acl type.
3. Create SMB share in Sharing dashboard pointing at NFS dataset. Use Purpose that will allow you to enable Home Share in advanced settings.
4. You should get redirected to an acl editor with some aces already preset.
5. Delete share and create new pointing to POSIX dataset.
6. Check that Use Preset button acl editor works.


RehanY147 (approved):
Looks good. Redirect happens on share creation. Acl presets are loaded. The button works as well.

PR #5781, opened 2021-08-17
NAS-110406 / 21.10 / Improved yarn.lock flow

RehanY147:
NAS-110406


undsoft (approved):
So far looks good, but first item in the ticket also needs to be addressed. See `.github/workflows/main.yml`

RehanY147:
> 
> 
> So far looks good, but first item in the ticket also needs to be addressed. See `.github/workflows/main.yml`

Yes, still working on that. The image wasn't building on the CI for some reason. Figuring that out.

undsoft (approved):
Looks good!

package.json
```diff
@@ -42,88 +42,88 @@
...
+    "q": "~1.5.1",
+    "rxjs": "~6.6.0",
+    "stylefire": "~7.0.2",
+    "what-is-my-ip-address": "~1.0.3",
+    "xterm": "~4.9.0",
+    "xterm-addon-fit": "~0.4.0",
+    "zone.js": "~0.10.3"
   },
   "devDependencies": {
     "@angular-devkit/build-angular": "0.1000.1",
```
undsoft, line 129: We can add ~ to devDependencies too.

RehanY147: That causes the CI build image to fail for some reason

PR #5784, opened 2021-08-17
NAS-111842 / 21.10 / Improvements for table component and data protection

denysbutenko:
Changes:
- Adapt changes from #5769
- Fix typo
- Improve type safety for table component and related pages

Testing:
- Check for any kind of regressions on pages that using the `TableComponent` (Sharing, Data Protection, Advanced Settings, Certificates, Network, etc...)


src/app/pages/common/entity/table/table.component.ts
```diff
@@ -199,6 +216,24 @@ export class TableComponent implements OnInit, AfterViewInit, AfterViewChecked {
     this.tableService.getData(this);
   }
 
+  updateColumns(): void {
+    this.displayedColumns = this._tableConf.columns
+      .map((column) => {
+        if (this.dataSource && column?.hiddenIfEmpty) {
+          const hasSomeData = this.dataSource.some((row) => row[column.prop]?.toString().trim());
+          column.hidden = !hasSomeData;
```
RehanY147, line 224: This will reset `hidden` to `false` if it was set to `true` originally if `hasSomeData` is false (meaning the column is empty). So, if the intention was to hide this column regardless if it was empty or not, it will get set to `false` if it's not empty.
It should be `column.hidden = column.hidden || !hasSomeData` or something like that

denysbutenko: Oh sh... I'll fix it 👍🏻

RehanY147 (changes requested):
The `Backup Credentials` page tables get these huge headers and rows if one of the tables doesn't have anything in it.
image
image

denysbutenko:
> The `Backup Credentials` page tables get these huge headers and rows if one of the tables doesn't have anything in it.
> image
> image

Seems like it happened randomly on the master branch. I'll create a ticket for that to investigate it.

src/app/pages/common/entity/table/table.component.ts
denysbutenko: Updated

denysbutenko:
All things are done. Including the fix for table rows height

RehanY147 (approved):
Looks good. Normal table functions are being performed same as before.

PR #5789, opened 2021-08-19
NAS-111922 / 21.10 / Fix displaying percent value

denysbutenko:
Pretty simple change. No testing is necessary.

image


RehanY147 (approved):
Looks good

PR #5790, opened 2021-08-19
NAS-111924 / 21.10 / Fix table rows height

denysbutenko:
How to test?

- Go to the Network page
- Check if table rows have the same height (48px)
- Go to the Backup Credentials page
- Check if table rows have the same height (48px)


dszidi (approved):
image


LGTM

PR #5794, opened 2021-08-20
NAS-110232 / 12.0 / Inappropriate triangle warning icon shown upon success

AlexKarpov98:


src/app/pages/directoryservice/activedirectory/activedirectory.component.ts
```diff
@@ -55,7 +55,7 @@ export class ActiveDirectoryComponent {
       function: async () => {
         this.systemGeneralService.refreshDirServicesCache().subscribe((cache_status) => {
           this.dialogservice.Info(helptext.activedirectory_custactions_clearcache_dialog_title,
```
RehanY147, line 57: We should also address the fact that the method name in `DialogService` is `Info(...)` however the default icon is `report_problem`. It should be the other way around. The default icon should be `info` or the method name should be changed to something else like just `dialogReport(...)` or `report(...)` or something more generic.

PR #5795, opened 2021-08-20
NAS-111638 / 21.10 / NAS-111638 - Upgrading Angular

undsoft:
Needs general regression testing.


denysbutenko (approved):
Code looks good. While testing found some issues but they are not related to the upgrade.

PR #5797, opened 2021-08-20
NAS-111928 / 21.08-BETA.1 / NAS-111928 - Do not reload page when additional columns are shown

undsoft:


denysbutenko (approved):
Kapture 2021-08-22 at 14 14 04

Looks good

PR #5798, opened 2021-08-20
NAS-111638 / None / Fix bug and mobile layout improvements

denysbutenko:
Items are found:
- The update dialog is broken due to untilDestroyed pipe (need to be ported to release/21.08 branch)
- Dashboard tables uses two columns on the mobile layout

### DIFF
--- a/src/app/interfaces/dialog.interface.ts
+++ b/src/app/interfaces/dialog.interface.ts
@@ -14,7 +14,7 @@ export interface ConfirmOptions {
 }
 
 export interface ConfirmOptionsWithSecondaryCheckbox extends ConfirmOptions {
-  secondaryCheckBox: boolean;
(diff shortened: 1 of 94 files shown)


src/app/interfaces/dialog.interface.ts
```diff
@@ -14,7 +14,7 @@ export interface ConfirmOptions {
 }
 
 export interface ConfirmOptionsWithSecondaryCheckbox extends ConfirmOptions {
-  secondaryCheckBox: true;
+  secondaryCheckBox: boolean;
```
undsoft, line 17: This was not a mistake. This was to specifically indicate that this interface and its fields kick in when secondaryCheckBox is true.

undsoft (approved):
Looks good.

PR #5800, opened 2021-08-21
NAS-111940 / 21.08-BETA.1 / NAS-111940: Apply Permissions Recursively checkbox is not getting che…

undsoft:
…cked & applied


RehanY147 (approved):
Simple enough change. Looks good.

PR #5802, opened 2021-08-21
NAS-0: Hide Cloud Sync encryption key & salt

Jonty:
This PR hides the encryption password and salt on Cloud Sync pages, in the same way as all other secret fields elsewhere in the UI.


undsoft (approved):
Looks good.

PR #5803, opened 2021-08-21
NAS-0: Hide Backup Credential tokens & keys

Jonty:
This hides all keys and tokens that are visible in the Backup Credentials panels to avoid inadvertently exposing them.


undsoft (approved):
Nice! Thank you for your contribution.

PR #5805, opened 2021-08-22
NAS-111877 / 21.10 / NAS-111877: Closing loader after navigating to Rsync Module form.

undsoft:
Basically in every place where we have tabs and navigate away from entity-form to something else the loader would not close.

Multiple places to check:
1. Services -> Rsync -> Rsync Modules
2. Sharing -> iSCSI Configure -> Targets


dszidi (approved):
Code looks good. 
No more endless spinners on the rsync/modules or iscsi/targets

william-gr:
backport

PR #5806, opened 2021-08-22
NAS-111537 / 21.10 / Allow job dialog to collapsed

denysbutenko:


undsoft (commented):
I like how how it shows where it collapses and that I can bring the window back.
I would prefer a different collapse button, maybe something similar to minimize button in Windows?
Also, note that "Close" button in dialog does similar thing, I think we can basically replace "Close" with this everywhere.

denysbutenko:
I tried to use the `minimize` icon, but it's tricky to align this icon:
image
image

We can keep it or find an alternative option

denysbutenko:
`indeterminate_check_box`:
image

undsoft:
Can we push it to top right and make custom styles to make it basically appears as Windows minimize icon?

undsoft (approved):
Very nice 👍

PR #5807, opened 2021-08-22
NAS-111711 / 21.10 / Status button should convey warnings to users even if status is success/finished

dszidi:


undsoft (changes requested):
Any tips on how I can test this?

dszidi:
> Any tips on how I can test this?

I've consulted with middleware team and there is unfortunately no easy way to trigger warnings on a completed replication task.

During development I faked it by pushing some strings to the warnings array of the API response.

image

undsoft (approved):
Appears to be working correctly, assuming that warnings will come as an array of strings.

PR #5813, opened 2021-08-25
Fix git hooks

denysbutenko:


undsoft:
> if there are no changes after execution of eslint --fix then commit would fail
Why thought? What if I didn't make any style mistakes?

denysbutenko:
> > if there are no changes after execution of eslint --fix then commit would fail
> 
> Why thought? What if I didn't make any style mistakes?

I meant if staged files have auto-fixable errors only then lint-staged prevents from creating the empty commit.

Example output:
```
❯ git commit -m "Test 2"
yarn run v1.22.10
$ ./node_modules/.bin/lint-staged
✔ Preparing...
✔ Running tasks...
✖ Prevented an empty git commit!
✔ Reverting to original state because of errors...
✔ Cleaning up...

  ⚠ lint-staged prevented an empty git commit.
  Use the --allow-empty option to continue, or check your task configuration

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
husky - pre-commit hook exited with code 1 (error)
```

denysbutenko:
This is the case when files after lint-staged execution match the original on the disk.

PR #5821, opened 2021-08-26
NAS-111933 / 21.08-BETA.1 / Fix failed jobs not showing up in list

denysbutenko:
My assumption is that the bug is related to wrong state handling in stores. The bug wasn't reproduced but those fixes are useful for the release.

Pick up fixes from a draft PR #5806

Changes:
- Fix missing job progress in Task Manager
- Fix handling job updates in JobsListStore and JobsManagerStore


denysbutenko:
Since 21.08 is frozen, these changes will be addressed in RP #5806

PR #5829, opened 2021-08-30
NAS-112003 / 21.10 / NAS-112003 - Migrate apps checkbox

undsoft:
This one is a little weird, because sometimes migration process may take a bit of time (longer than session duration).


undsoft:
Also, it doesn't really work for me on our cloud scale instance, but this may be due to performance reasons, not sure.

sonicaj (commented):
We should show migration option when the pool is actually being changed and it shouldn't be there otherwise.

We are not exposing the option in Advanced Settings of Apps as well, let's please expose it there as well as users can switch pools from there too. Also do you think it would be perhaps better to just have 1 settings form instead of 3 options under settings ?

The simple pool form can be displayed when no pool is configured for apps, but once there is, users only use the advanced settings form to make changes ? Right now we have 3 options:
1. Choose Pool
2. Advanced Settings
3. Unset Pool

It just seems redundant :)

undsoft:
True enough, but I'd rather remove pool settings out of Advanced settings.

sonicaj:
@undsoft i am torn on that. If user wants to change pool and some other settings like CIDR, this would be a 2 step process then with user changing the pool from a different location/form in the UI and then opening advanced and changing the CIDR next which would btw also wipe out the apps datasets and re-instantiate them, so to conclude - in such cases we would be creating/destroying and then creating apps datasets again.

undsoft:
@sonicaj CIDR settings are kinda the reason not to have Pool dropdown on that form. Otherwise, it's going to be even more confusing - there is a migrate checkbox, but it only works sometimes and sometimes everything is deleted.
Hopefully users don't need to both move a pool and change CIDR settings too often and we can keep them as different operations.
TLDR: I've removed the Pool dropdown from Advanced settings form.

sonicaj:
@undsoft can you please clarify if you are referring to the migration working sometimes and also can you please elaborate what is being deleted ?

I think when a user is changing pool, that is the ideal time to change CIDR settings as they are going to re-initialize apps - however having both migration + changing CIDR settings is weird as changing those settings is going to re-initialize the apps making the migration a moot point. To conclude, changing pool ( not migrating ) and configuring CIDR settings should be allowed together imho :)

undsoft:
@sonicaj Deleted as to not have the same functionality in two places and to keep it a separate task.
The second paragraph is pretty much my point. If we were to have Migrate apps in Advanced settings, I would have to explain in UI how this checkbox doesn't really work when CIDR settings are changed. 
Now that you can only migrate to another pool by pressing a separate menu item yes it will take you more time if you need to change CIDR settings as well, but there is less room for confusion.

sonicaj (commented):
LGTM. A minor nit, we are not rendering the validation errors raised on the pool selection form.

https://user-images.githubusercontent.com/17968138/131697355-367cb72a-72b8-429a-b882-dee8bdf05010.mov

RehanY147 (approved):
Looks good to me

PR #5830, opened 2021-08-30
NAS-111774 / 12.0 / Built-in group Dispaly Note is randomely showing when editing users in CORE

AlexKarpov98:


undsoft (approved):
The same code also exists on group-list page.

AlexKarpov98:
> The same code also exists on group-list page.

But this isn't included in the ticket, still do we need it?

undsoft:
Yes, it basically the same issue.

undsoft (approved):
👍🏻

PR #5833, opened 2021-08-30
NAS-111493 / 21.10 / user / group quotas not properly displayed

AlexKarpov98:


dszidi:
On the user quotas page, when I hit the "Add user quota" button in the entity-empty message, I get the following run time error in the browser console.

image

dszidi:
The data is now being stored as expected but there seems to be a routing problem in both user and group quota forms. 
Even though you've seemingly set the route correctly, the router doesn't seem to like the route when we're creating a new quota. 
Oddly, it works as expected when editing an existing quota. The problem is only there when creating a new quota.

dszidi (commented):
Adding this comment in an attempt to silence notifications. It seems the previous comment above was not recognized.

AlexKarpov98:
@dszidi Please try to check using latest version. This bug was already fixed.

undsoft (changes requested):
1. Actions menu now has `Add` and `Set Quotas (Bulk)` items that lead to same page. Need to remove one of those.
2. If I add a new quota by filling in `Object Quota` and not filling in `Data Quota`, it breaks the page completely. Could be a middleware issue, not sure.

AlexKarpov98:
@sonicaj Could you help me check if this api call is correct or not?
image

This api call does not occur error, but we get api error on calling get `pool.dataset.get_quota`.
image

sonicaj:
@AlexKarpov98 can you please share the complete traceback in the response ? I ran get_quota command and it seems to work fine but obviously isn't as it wasn't working for you - i think might be related to what you are sending or what state the dataset is in currently.

AlexKarpov98:
@sonicaj 
You can find the dataset(test_quota) on http://tn27.lab.ixsystems.net/ui/dashboard.
image
[error.txt](https://github.com/truenas/webui/files/7134191/error.txt)

sonicaj:
@AlexKarpov98 the fix for that has been merged and should be available in the next nightlies https://github.com/truenas/middleware/pull/7497.

undsoft (approved):
Works well now. Thanks.

PR #5834, opened 2021-08-30
NAS-111493 / 12.0 / user / group quotas not properly displayed

AlexKarpov98:


RehanY147:
How should this ticket affect the `Action` dropdown for the `Quota` tables? As it is now, looks like the column's `Toggle` actions just switched which users it should be showing. On the VM, you can see `user A` in the default view and both users in the toggle view. On this branch, `user B` is seen in the default view and both users in the toggle view.

AlexKarpov98:
@RehanY147 I am not sure I am correct, but I think "looks like the column's Toggle actions just switched which users it should be showing" is correct. I asked the reporter(Andrew Walker), but no response so far.

RehanY147:
@AlexKarpov98 Any updates on this? Were you able to confirm this behavior with Andrew?

AlexKarpov98:
@RehanY147 Unfortunately, not yet.

undsoft (changes requested):
Let's change the text in users and groups to:
> Show all available users, including those that do not have quotas set.
> Show only those users who have quotas. This is the default view.

If we have the same toggle in scale, need to port these changes too.

RehanY147 (approved):
The functionality is fine. Messaging seems better now.

PR #5837, opened 2021-08-31
NAS-111931 / 21.10 / NAS-111931: Adapt disk name retrieval to Linux naming conventions

dszidi:
PR testing needs iX rackmount hardware with enclosure support and a pool with spares. (tn27 should be preconfigured)
To test, simply make sure the spares are being presented as members of the pool.


AlexKarpov98 (approved):
@dszidi LGTM

PR #5842, opened 2021-09-01
NAS-112118 / 21.10 / Fix dialog for Report CPU usage in percent

denysbutenko:


dszidi (approved):
Code looks good and new confirmation dialog behaves as expected.

PR #5844, opened 2021-09-01
NAS-110196 / 21.10 / NAS-110196 - HTML Linter

undsoft:
Linter is not super great and doesn't catch all indentation issues, but better than nothing.
Some places were reformatted manually.


denysbutenko (approved):
I wonder if [eslint-plugin-markup](https://www.npmjs.com/package/eslint-plugin-markup) is applicable here?
Looks much better with consistent alignment 👍🏻

undsoft:
eslint-plugin-markup seems to be a markuplint plugin for eslint, don't see what advantages would that give us.

PR #5846, opened 2021-09-02
NAS-112128 / 21.10 / NAS-112128: Move some html code in separate template files

undsoft:
NAS-112128: Update some dialog calls for newer syntax

For testing maybe test a couple of the dialogs to make sure I didn't mess up anything.


RehanY147 (approved):
Looks good. Dialogs seem to be working OK.

PR #5849, opened 2021-09-02
NAS-111625 / 21.10 / Add ensure_display_device field in the UI for vms

denysbutenko:


undsoft (changes requested):
The checkbox is there, but once I've added a first VM search field and Add button disappeared.

undsoft (changes requested):
Doesn't look like the fix fixed it. The search/Add button are still not there.

undsoft (approved):
It looks good now.

denysbutenko:
Unfortunately, it shows the global actions toolbar when a system does not support virtualization. WIP

denysbutenko:
`disableActionsConfig` must be `true` by default to prevent toolbar initialization. Once the toolbar gets initialized, we can't hide it.
There are two cases when `disableActionsConfig` should be `false`

1. the system has virtualization support
2. the system does not have `vm.virtualization_details` endpoint (old backend + new UI)

PR #5851, opened 2021-09-02
NAS-112011 / 21.10 / NAS-112011: Removing EntityTableService

undsoft:
It's unclear what EntityTableService does, looks like nothing, because exactly the same code exists in ReportingService. Testing may include general regression testing.


AlexKarpov98 (approved):
@undsoft LGTM

PR #5855, opened 2021-09-02
NAS-111091 / 21.10 / New forms

RehanY147:
To see the new forms in action go to `System -> General Settings`

NAS-111091


src/app/pages/system/general-settings/localization-form3/components/ix-input/ix-input.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  @Input() suffixIcon: string;
+  @Output() suffixIconClick: EventEmitter<MouseEvent> = new EventEmitter();
+  @Input() hint: string;
+  @Input() tooltip: string;
+  @Input() required: boolean;
+
+  value: string | number = '';
+
+  onChange: any = (): void => {};
+  onTouch: any = (): void => {};
```
undsoft, line 33: The type of onChange is (value: string) => void, not any.

src/app/pages/system/general-settings/localization-form3/components/ix-input/ix-input.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  writeValue(value: string | number): void {
+    this.value = value;
+  }
+
+  registerOnChange(onChange: any): void {
+    this.onChange = onChange;
+  }
+
+  registerOnTouched(onTouched: any): void {
+    this.onTouch = onTouched;
```
undsoft, line 48: Need to call onTouch whenever input is touched.

src/app/pages/system/general-settings/localization-form3/components/ix-input/ix-input.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  @Input() label: string;
+  @Input() placeholder: string;
+  @Input() prefixText: string;
+  @Input() suffixIcon: string;
+  @Output() suffixIconClick: EventEmitter<MouseEvent> = new EventEmitter();
+  @Input() hint: string;
+  @Input() tooltip: string;
+  @Input() required: boolean;
+
+  value: string | number = '';
```
undsoft, line 30: Is value really a number?

src/app/pages/system/general-settings/localization-form3/localization-form3.component.html
```diff
@@ -0,0 +1,19 @@
+<ix-form [formGroup]="form" (cancel)="cancelled()" [title]="formTitle" (formSubmit)="submit($event)">
+  <ix-input
+    label="Name"
+    name="name"
+    required="true"
+    tooltip="nameTooltip"
+    placeholder="namePlaceholder"
+    [ngModel]="name"
```
undsoft, line 8: Don't understand why ngModel is needed here, since we are using formControlName.

undsoft (commented):
Looks pretty good.
Something to consider in the future (perhaps for combobox) are all the weird things we do to just show a list of users with `updater` and `loadMoreOptions`.

src/app/pages/system/general-settings/localization-form3/components/ix-input/ix-input.component.ts
RehanY147: Sometimes, it can be. I've seen instances where value is an id which is an integer. That can be easily casted to a string either way but I was going for a bit more flexibility.

src/app/pages/system/general-settings/localization-form3/localization-form3.component.html
RehanY147: Setting a default value (e.g., on edit forms), was not getting reflected on the UI. The form control would not pick up the value in the form group either. Have to use this `[ngModel]` to set default values for form controls.
On second that, it might not be needed to set the value here in the actual form component. But `ngModel` was needed to update the UI with the default value because otherwise, it wouldn't pick up the value. Still working on replacing this with just `FormControl` apis

denysbutenko:
Kapture 2021-09-13 at 12 00 07
Works well at first sight. I just wonder if we need to have a consistent loading state for form fields?

src/app/pages/common/ix/ix-combobox/ix-combobox.component.ts
```diff
@@ -0,0 +1,107 @@
...
+    this.onChange = onChange;
+  }
+
+  registerOnTouched(onTouched: any): void {
+    this.onTouched = onTouched;
+  }
+
+  optionSelected(option: Option): void {
+    if (!this.selectedOption) {
+      this.selectedOption = new Option();
```
undsoft, line 76: Why not null?

src/app/pages/common/ix/ix-form/ix-form.component.ts
```diff
@@ -0,0 +1,23 @@
+import {
+  Component, EventEmitter, Input, Output,
+} from '@angular/core';
+import { FormGroup } from '@angular/forms';
+import { UntilDestroy } from '@ngneat/until-destroy';
+
+@UntilDestroy()
+@Component({
+  selector: 'ix-form',
```
undsoft, line 9: As per discussion IxForm and IxFieldset may create the same issues we had with old form implementation where if I want to add something else (like a third button) I may need to start adding more inputs to the component.
Let's see if we can use a different approach by extracting common styling to a css class / directive.

src/app/pages/system/general-settings/localization-form2/localization-form2.component.ts
```diff
@@ -0,0 +1,179 @@
...
+import {
+  LanguageService, ModalService, SystemGeneralService, WebSocketService,
+} from 'app/services';
+import { LocaleService } from 'app/services/locale.service';
+
+@UntilDestroy()
+@Component({
+  selector: 'localization-form2',
+  templateUrl: './localization-form2.component.html',
+  styleUrls: ['./localization-form2.component.scss'],
```
undsoft, line 20: Empty file.

undsoft (changes requested):
👍🏻

src/app/pages/common/ix/ix-combobox/ix-combobox.component.ts
```diff
@@ -0,0 +1,107 @@
...
+    } else {
+      this.filteredOptions = of(this.syncOptions);
+    }
+  }
+
+  registerOnChange(onChange: any): void {
+    this.onChange = onChange;
+  }
+
+  registerOnTouched(onTouched: any): void {
```
undsoft, line 70: The type of the argument here and in `registerOnChange` are the same as in `onChange` and `onTouched` class fields.

undsoft (changes requested):

src/app/pages/system/general-settings/localization-form2/localization-form2.component.ts
```diff
@@ -0,0 +1,179 @@
...
+        this.sortLanguagesByName ? 'label' : 'value',
+      ));
+      this.formGroup.get('language').setValue(this.configData?.language);
+    });
+  }
+
+  makeKbdMapOptions(): void {
+    this.sysGeneralService.kbdMapChoices().pipe(untilDestroyed(this)).subscribe((mapChoices) => {
+      this.kbdMap.options = of(mapChoices);
+      this.formGroup.get('kbdmap').setValue(this.configData?.kbdmap);
```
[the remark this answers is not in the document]
RehanY147: This form is complicated. There are two cases for `configData` to get assigned a value. One is that the value gets passed to it from the parent component. Which will be almost instant since it involves no middleware call. In which case, this line will be used to assign value and the other place might be too soon since the form hasn't initialized yet.
The other case is where the `configData` is request from middleware and takes a while. In which case, this will be too soon since `configData` isn't available yet.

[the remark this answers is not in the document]
undsoft: How is it being passed from the parent component?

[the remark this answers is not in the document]
RehanY147: Basically, for now `getGeneralConfig$` observer is used which gets values from middleware. But when a slide-in panel is implemented, the form will be receiving config through its parent component which will be through `sendGeneralConfig$` and in that case this will be the correct code.

src/app/pages/system/general-settings/localization-form2/localization-form2.component.ts
```diff
@@ -0,0 +1,142 @@
...
+        this.formGroup = this.fb.group({
+          language: [this.configData?.language],
+          kbdmap: [this.configData?.kbdmap],
+          timezone: [this.configData?.timezone],
+          date_format: [this.localeService.getPreferredDateFormat()],
+          time_format: [this.localeService.getPreferredTimeFormat()],
+        });
+        this.formGroup?.get('kbdmap').setValue(this.configData.kbdmap);
+        this.formGroup?.get('language').setValue(this.configData.language);
+        this.formGroup?.get('timezone').setValue(this.configData.timezone);
```
undsoft, line 114: These lines should be the same as passing values during form init.

src/app/pages/common/ix/ix-combobox/ix-combobox.component.ts
```diff
@@ -0,0 +1,97 @@
...
+  registerOnChange(onChange: (value: string | number) => void): void {
+    this.onChange = onChange;
+  }
+
+  registerOnTouched(onTouched: () => void): void {
+    this.onTouch = onTouched;
+  }
+
+  optionSelected(option: Option): void {
+    this.selectedOption = _.cloneDeep(option);
```
undsoft, line 76: What's the reason to do deep cloning here?

src/app/pages/common/ix/ix-input/ix-input.component.ts
```diff
@@ -0,0 +1,55 @@
...
+  @Input() tooltip: string;
+  @Input() required: boolean;
+
+  formControl = new FormControl(this).value as FormControl;
+
+  value: string | number = '';
+  touched = false;
+
+  onChange: any = (): void => {};
+  onTouch: any = (): void => {};
```
undsoft, line 36: any bad

src/app/pages/common/ix/ix-select/ix-select.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  @Input() options: Observable<Option[]>;
+  @Input() required: boolean;
+  @Input() tooltip: string;
+
+  formControl = new FormControl(this).value as FormControl;
+
+  touched = false;
+
+  onChange: any = (): void => {};
+  onTouch: any = (): void => {};
```
undsoft, line 35: :'(

undsoft (approved):
Minor changes, otherwise nice work.

Selected option in Chrome looks funny.

Снимок экрана 2021-09-17 в 11 07 17

src/app/pages/common/ix/ix-combobox/ix-combobox.component.ts
RehanY147: Just to create a new clean assignment in one easy line here. Could be replaced with `{...option}` or something similar.

undsoft: It can be, yes, but why the new assignment?

PR #5858, opened 2021-09-03
NAS-111626 / 21.10 / NAS-111626: Have ability to refresh Application events in app summary dialog

AlexKarpov98:


src/app/pages/applications/dialogs/chart-events/chart-events-dialog.component.ts
```diff
@@ -35,21 +36,19 @@ export class ChartEventsDialog implements OnInit {
...
-        this.loader.close();
-        if (charts) {
-          this.catalogApp = charts[0];
-        }
-        if (events) {
-          this.chartEvents = events;
-        }
-      },
-    );
+    forkJoin([
```
undsoft, line 40: Nice!

undsoft (changes requested):
It works well, but I'd like to make some UX improvements:
1. Have Application Events uncollapsed when Refresh Events is pressed. There may be some users (like me) confused as to why nothing is going on when they press the Refresh button.
2. Let's change the order of events to show new events on top, so that it's easier to see new events after refreshing them.

undsoft (approved):
Thanks for the fixes.

PR #5867, opened 2021-09-07
NAS-112205 / 21.10 / NAS-112205: Simplify ModalService

undsoft:
This replaces manual construction of components in ModalService with Angular's DI mechanism.

To test: test some of the modals to make sure there are no regressions.


denysbutenko (approved):
Animations are smooth now. Looks nice 👍🏻

PR #5869, opened 2021-09-08
NAS-111733 / 21.10 / NAS-111733: Add locked icon to share path

denysbutenko:
On Shares dashboard added lock icon to path columns when a dataset is locked

For testing, you'll be needed a dataset with encryption. Use existing or create a new one with enabled encryption then switch the encryption method from key to passphrase and lock it. Create a new share using locked dataset to test it


denysbutenko:
Preview:
image

The icon position like on the Storage page

undsoft (changes requested):
Shares Dashboard looks good.
Two things though:
1. Can we have a pointer or question mark cursor when hovering over lock icon?
2. There are also separate pages for shares that you can get to by clicking on the card title. These also need updates.

denysbutenko:
Updated

undsoft (changes requested):
Only see lock indication on full NFS list, don't see any locks on full SMB or Webdav list.

denysbutenko:
> Only see lock indication on full NFS list, don't see any locks on full SMB or Webdav list.
Perhaps that is cached, I see something similar when did change for `enabled` checkbox and it needs to toggle column visibility and it gets fixed

denysbutenko:
Tested branch against cloud VM and it is displayed:
image

In case if you still see old values, try to toggle column visibility:
Kapture 2021-09-13 at 12 35 07

undsoft (approved):
Weird, but yes I can see the lock icons now, even though I didn't change anything.

PR #5870, opened 2021-09-09
NAS-112160 / 12.0 / NAS-112160: Fixing enclosure page after using dashboard widget

undsoft:
And this is why, kids, you don't mutate variables.


RehanY147 (approved):
Looks good

PR #5872, opened 2021-09-09
NAS-111834 / 12.0 / Can't remove iSCSI targets

AlexKarpov98:


undsoft (approved):
I was able to successfully remove an iSCSI target.

PR #5874, opened 2021-09-10
NAS-112015 / 21.10 / NAS-112015: Fixed width on list items to create more space for elements

RehanY147:
NAS-112015

To test
Check out the ISCSI targets form on the Shares Dashboard. The list items should be well spaced out and one per line so they're easily usable.


denysbutenko (approved):
image

Looks good

PR #5881, opened 2021-09-12
NAS-112228 / 21.10 / NAS-112228: Fixed view responsiveness for phone screens

RehanY147:
NAS-112228

Test forms on different modules to make sure the changes don't cause any regressions. Also test slide-in forms and edit slide-in forms from different modules on smaller phone screens.


src/app/components/common/modal/modal.component.scss
```diff
@@ -112,3 +112,10 @@ body.jw-modal-open .slide-in-form .jw-modal-body {
   /* body overflow is hidden to hide main scrollbar when modal window is open */
   overflow: hidden;
 }
+
+@media screen and (max-width: 768px) {
```
undsoft, line 116: We now have both media queries and breakpoint values as variables, see `cssvars.scss`.
The breakpoints there [are based on material numbers.](https://material-ui.com/ru/customization/breakpoints/)

src/assets/styles/other/_material-reduction.scss
```diff
@@ -803,6 +803,12 @@ mat-card.mat-card .mat-list-base {
   margin-left: 8px;
 }
 
+@media screen and (max-width: 768px) {
+  .entity-table-controls > div {
+    margin-top: 8px !important;
```
undsoft, line 808: Can you do the same without important?

undsoft (changes requested):
Functionality wise looks good, couldn't find a slide-in form with issues.

src/assets/styles/other/_material-reduction.scss
RehanY147: The important here is needed for the search input. Other views take the margin without important but not the search input.

undsoft: If you are having a problem with `#row-filter` id, I'd prefer to have another css declaration for that rather than introducing `!important`.

src/app/pages/common/entity/entity-form/entity-form.component.scss
```diff
@@ -166,3 +166,19 @@ span.form-message.warning {
     padding: 16px !important;
   }
 }
+
+@import 'cssvars';
```
undsoft, line 170: Best to put imports on top of files.

undsoft (changes requested):

undsoft (approved):
Nice!

PR #5893, opened 2021-09-14
NAS-112329 / 21.10 / NAS-112329: Added value assignment for both cases

RehanY147:
NAS-112329


undsoft (changes requested):
This is not a good fix, this will introduce issues in other places.
Need to find a proper solution and for quick fix for the welcome screen let's use that PR you had for storing stuff in localstorage.

undsoft (approved):
Seems to be working correctly now.

PR #5894, opened 2021-09-15
NAS-112012 / 21.10 / Don't show [Manual Test] button as well when disk does not support S.M.A.R.T. test

AlexKarpov98:


RehanY147 (approved):
Looks good

undsoft (changes requested):
Not sure if this PR breaks it, but I get an error every time I try to open the row.
Снимок экрана 2021-09-15 в 12 23 02

AlexKarpov98:
> Not sure if this PR breaks it, but I get an error every time I try to open the row.
> Снимок экрана 2021-09-15 в 12 23 02

I could see also this bug and is not from my commit.

undsoft (approved):
Looks like a regression that I've introduced. Fixed.

PR #5895, opened 2021-09-15
NAS-112012 / 12.0 / Don't show [Manual Test] button as well when disk does not support S.M.A.R.T. test

AlexKarpov98:


undsoft (approved):
Testing on scale api. Core api doesn't seem to be working correctly, but frontend code is okay.

PR #5896, opened 2021-09-15
NAS-112316 / 21.10 / Cleanup Apps version references

Ornias1993:
**General overview:**
This should improve the version references for Apps in the bigger overview portions of the UI, by splitting the long weird string with `_` in the middle into two parts:
"Container Version" which translates to chart.yaml->appVersion
"App Version" which translates to chart.yaml->Version

**Why Container version?**
In almost any cases the appVersion is used to signify the version of the included application. However, we needed something to differentiate it from `App version`. Hence "Container version" to signify it has nothing to do with the App, but rather with the container inside.

**What does not change with this PR:**
- The shorts-style version: containerversion_appversion  (One might want to replace the `_` with `/` though), it still has a place to give a quick overview.
- The Available Apps overview: It doesn't have the space to apply the same changes and needs additional consideration, so it keeps using short-style version
- The information dialogs on-hover when an update is available: These still use the short-style version.

### DIFF
--- a/src/app/pages/applications/chart-releases/chart-releases.component.html
+++ b/src/app/pages/applications/chart-releases/chart-releases.component.html
@@ -22,8 +22,8 @@
             <span class="update-label" matTooltip="Available version: {{ item.human_latest_version }}">{{ 'Update available' | translate}}</span>
           </div>
           <ng-template #elseBlock><div class="update-label">{{ 'Up to date' | translate }}</div></ng-template>
-          <div class="version-label">{{ 'Container Version' | translate }}:<span class="version" textLimiter threshold="50" content="{{ item.chart_metadata.appVersion}}"></span></div>
-          <div class="version-label">{{ 'App Version' | translate }}:<span class="version" textLimiter threshold="50" content="{{ item.version}}"></span></div>
+          <div class="version-label">{{ 'Container Version' | translate }}:<span class="version-value" textLimiter threshold="50" content="{{ item.chart_metadata.appVersion}}"></span></div>
+          <div class="version-label">{{ 'App Version' | translate }}:<span class="version-value" textLimiter threshold="50" content="{{ item.version}}"></span></div>
         </div>
 
         <div class="chart-button-row" fxLayout="row" fxLayoutAlign="end end" (click)="$event.stopPropagation()">
--- a/src/app/pages/applications/chart-releases/chart-releases.component.html
+++ b/src/app/pages/applications/chart-releases/chart-releases.component.html
@@ -18,11 +18,12 @@
       <div class="content-box"  (click)="showChartEvents(item.name)">
         <div class="content">
           <strong class="chart-name" textLimiter threshold="20" content="{{ item.name }}"></strong>
-          <div><span class="version" textLimiter threshold="50" content="{{ item.human_version}}"></span></div>
           <div *ngIf="item.update_available || item.container_images_update_available; else elseBlock">
             <span class="update-label" matTooltip="Available version: {{ item.human_latest_version }}">{{ 'Update available' | translate}}</span>
           </div>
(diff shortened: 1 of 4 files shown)


src/app/pages/applications/dialogs/chart-events/chart-events-dialog.component.html
```diff
@@ -3,11 +3,22 @@
...
+        <span class="top-label">{{ appStatus() | translate }}</span>
+        <mat-icon *ngIf="getUpdateVersionTooltip()" class="version-tooltip" matTooltip="{{ getUpdateVersionTooltip() }}">info</mat-icon>
+      </div> 
+      <div fxLayout="row" fxLayoutAlign="flex-start center">
+        <span class="top-label">{{ 'Container Version' | translate }}:</span>
+        <span class="top-value"> {{ catalogApp.chart_metadata.appVersion }}</span>
+      </div>  
+      <div fxLayout="row" fxLayoutAlign="flex-start flex-start">
+        <span class="top-label">{{ 'App Version' | translate }}:</span>
+        <span class="top-value"> {{ catalogApp.version }}</span>
```
undsoft, line 20: Old string: 1.24.1.4931_1.6.8
New string: Container Version:1.24.1.4931 App Version:1
Shouldn't app version be 1.6.8?

src/app/pages/applications/chart-releases/chart-releases.component.html
```diff
@@ -18,11 +18,12 @@
       <div class="content-box"  (click)="showChartEvents(item.name)">
         <div class="content">
           <strong class="chart-name" textLimiter threshold="20" content="{{ item.name }}"></strong>
-          <div><span class="version" textLimiter threshold="50" content="{{ item.human_version}}"></span></div>
           <div *ngIf="item.update_available || item.container_images_update_available; else elseBlock">
             <span class="update-label" matTooltip="Available version: {{ item.human_latest_version }}">{{ 'Update available' | translate}}</span>
           </div>
           <ng-template #elseBlock><div class="update-label">{{ 'Up to date' | translate }}</div></ng-template>
+          <div class="version-label">{{ 'Container Version' | translate }}:<span class="version-value" textLimiter threshold="50" content="{{ item.chart_metadata.appVersion}}"></span></div>
+          <div class="version-label">{{ 'App Version' | translate }}:<span class="version-value" textLimiter threshold="50" content="{{ item.version}}"></span></div>
```
undsoft, line 26: The space is kinda limited here, so:
1. Will user be actually interesting in seeing container version right here or would just showing the app version suffice?
2. If not, how about something like this with more muted color and less space used.
Снимок экрана 2021-09-16 в 12 27 15

undsoft (changes requested):

src/app/pages/applications/chart-releases/chart-releases.component.html
Ornias1993: I actually Agree, I was under the impression that iX really wanted to show the container version everywhere...

I'm open for both options, so if you pick one, i'll get it cleaned up and finished :)

src/app/pages/applications/dialogs/chart-events/chart-events-dialog.component.html
Ornias1993: Shoozle, thats indeed wrong. Picked the wrong version var here it seems, thanks for the headsup!

Ornias1993:
I'm not 100% happy with this still.
`Container version` possibly kinda contradicts with the actual container versions (AppVersion isn't guaranteed to be the main container version!).

So in short some TODO's:
- Remove container version from the tiles completely
- Rename container version to something else (software version or program version?)

Ornias1993:
I'm going to close this:
I feel the underscore version numbers fit relatively great in terms of width, the added confusion in splitting it in different version isn't worth it.

PR #5897, opened 2021-09-15
NAS-111393 / 12.0 / Only show icons based on status instead of health

dszidi:


denysbutenko:
@dszidi, the code looks good. What is the way to reproduce this scenario?

dszidi:
@denysbutenko Unfortunately forcing specific zpool status scenarios in ZFS is not straight forward. 
Simplest way to test is to inject the status string via ngOnChanges in the widgetpool.component.ts. 
Master branch has an enum with all the possible cases

https://github.com/truenas/webui/blob/master/src/app/enums/pool-status.enum.ts

denysbutenko:
image
it is not clear what the pool status icon means. also, it has some offset from others

denysbutenko:
> image
> it is not clear what the pool status icon means. also, it has some offset from others

My bad. I assigned the wrong variable above.

image
image
image
image

dszidi:
Looks like I need to hard code the error case in the html. The yellow close icon the existing poolHealth.level value generates should really be red.

denysbutenko (approved):
image

image


Looks good!

PR #5885, opened 2021-09-14
NAS-112109 / 21.10 / NAS-112109: Handling broken reporting database

denysbutenko:
How to test?

1. Open `System Settings -> Shell`
2. Update system time to future, type `date --set='+14 day +1 hour'` for example
3. Reboot the system, type `reboot`
4. Wait until the system is loaded
5. Open Dashboard and check the Network widget
6. Fix reporting database

Repeat steps 1-4
1. Go to Reporting 
2. Fix reporting database

PR #5898, opened 2021-09-15
NAS-112109 / 12.0 / Handling broken reporting database

denysbutenko:
Manual port of #5885

Preview:
image
image

How to test?

- Go to `Shell`
- Update system time to future, type `date 2110011307` for example
- Clear reporting database `midclt call reporting.clear`
- Go to Reporting, wait a couple of minutes to collect some data
- Reboot the system, type `reboot`
- Wait until the system is loaded
- Go to Reporting
- Fix reporting database

FreeBSD behaves differently from Debian, you may notice a warning in shell:
```
Warning: settings changed through the CLI are not written to the configuration database and will be reset on reboot.
```
It means if you change the date and make a reboot it will revert date changes.
So, the idea is to set the future time, then reset reporting database, collect some data, and then make a reboot. The date will be reverted and you'll get `Error: 206` from API.

### DIFF
--- a/src/app/pages/reportsdashboard/components/report/report.component.css
+++ b/src/app/pages/reportsdashboard/components/report/report.component.css
@@ -33,7 +33,7 @@
 }
 
 .legend {
-  display:table; 
+  display:table;
 }
 
 .legend-group-wrapper {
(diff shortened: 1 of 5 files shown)


denysbutenko:
Note: The CORE version should be 2 September or newer for `reporting.clear`

dszidi (approved):
Code looks good. Behaves as advertised. Nice job

PR #5899, opened 2021-09-15
NAS-112057 / 21.10 / Apps: Make "Installed Applications" the default tab

Ornias1993:
This changes the default tab for the Applications interface when clicking "Apps" from the Menu to "Installed Applications".

**Confirmed working when used with pre-configured pool:**
- Clicking "Apps in the Nav menu" -> lands on "Installed Applications"
- Switching between all possible tab combinations
- Using the bulk tools in  "Installed Applications"
- Referesh button on "Available Applications" 
- setup modal
- poolselector pop-up

Might need (considerable?) extra work for the pool selection popup on new systems though, which isn't my strongsuit, Feel free to fork and finish if so :)

### DIFF
--- a/src/app/pages/applications/applications.component.ts
+++ b/src/app/pages/applications/applications.component.ts
@@ -29,8 +29,8 @@ import { ManageCatalogsComponent } from './manage-catalogs/manage-catalogs.compo
   encapsulation: ViewEncapsulation.None,
 })
 export class ApplicationsComponent implements OnInit, AfterViewInit {
-  @ViewChild(CatalogComponent, { static: false }) private catalogTab: CatalogComponent;
   @ViewChild(ChartReleasesComponent, { static: false }) private chartTab: ChartReleasesComponent;
+  @ViewChild(CatalogComponent, { static: false }) private catalogTab: CatalogComponent;
   @ViewChild(ManageCatalogsComponent, { static: false }) private manageCatalogTab: ManageCatalogsComponent;
   @ViewChild(DockerImagesComponent, { static: false }) private dockerImagesTab: DockerImagesComponent;
   selectedIndex = 0;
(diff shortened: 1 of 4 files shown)


src/app/pages/applications/applications.component.ts
```diff
@@ -78,12 +72,12 @@ export class ApplicationsComponent implements OnInit, AfterViewInit {
         this.filterString = evt.data.filter;
       }
 
-      if (evt.data.event_control == 'catalogs') {
+      if (evt.data.event_control == 'catalog') {
```
AlexKarpov98, line 75: I am not sure why this is changed to 'Catalog' from 'Catalogs'.
Also I think we may remove this one as we overwrite 'selectedCatalogOptions' later in several places.
Actually we never use the value assigned here.

Ornias1993: Good catch! :)

Ornias1993: fixed, removed and tested :)

AlexKarpov98 (approved):
LGTM

undsoft (commented):
In case when user is entering Apps section for the first time, I would prefer to take user to Available applications after the "select pool" popup.

Ornias1993:
@undsoft Maybe best to move to "Available Apps" on init, if no Apps are installed every time.
That would still enable users to view "Installed Apps" with the tab, but would skip it if it's useless to the user when selecting "Apps" in the menu, regardless of weither it's first-init or not.

undsoft:
Sounds good too, whatever is easier to implement.

Ornias1993:
@undsoft I've gone your way and just limited myself to showing the "Available Apps" section when the "first start"pool selector is shown (aka: keeping the current behavior on first-launch)

Some testing notes and screenshots, as I know you prefer those:


Pool selection on first clicking "Apps":
image


Finishing the first pool selection shows the available Apps:
image


Going back to dashboard without installing Apps and returning shows the installed Apps section:
image

Clicking the "select pool" settings option when already having a pool selected, does correctly not change tabs:
image

undsoft:
Did you mean to push any new commits?

Ornias1993:
> Did you mean to push any new commits?

Whoops sorry, checkbox was ticked off without me noticing. >.<

Rebased and squashed while I was at it.

undsoft (approved):
Nice!

PR #5900, opened 2021-09-15
NAS-112345 / 21.10 / Dynamic Portals

Ornias1993:
This moves the portal buttons to a dynamic system, where the name of the portal, defines the label of the portal button.
It also adapts the listed ports to dynamically resize depending on the width of the portals and hide if it's not going to fit at all.

**Migration and breakage:**
- In case users do not have the latest version of an app combined with this change in 21.10 those would be named "Web Portal", instead of the current "portal".
- In case users have the newest version of the Apps but an older version of TrueNAS, they will not get any portal button if it's not named "web_portal".


undsoft (changes requested):
So, I'm not too familiar how this works on middleware side. 
Is it up to every app catalog to define names for portals or is it driven by truenas middleware?
If we add support for admin portals now, how would it work with an official app catalog?

In either case, I'd prefer to see changes made elsewhere before seeing them in webui purely so that I could test new behaviour. Having backward compatibility would also be nice, but it depends on answers to previous two questions.

Ornias1993:
Thinking about this some more, and lets not rename web_portal for compatibility at all, would make it much easier and less breaking to implement... I'll add that later.

src/app/pages/applications/chart-releases/chart-releases.component.html
```diff
@@ -26,9 +26,13 @@
         </div>
 
         <div class="chart-button-row" fxLayout="row" fxLayoutAlign="end end" (click)="$event.stopPropagation()">
-          <button mat-button (click)="portal(item)" *ngIf="item.status === ChartReleaseStatus.Active && item.portals && item.portals.web_portal && item.portals.web_portal[0]"
-              ix-auto ix-auto-type="button" ix-auto-identifier="{{ item.name}}-portal">
-            {{ 'Portal' | translate }}
+          <button mat-button (click)="userPortal(item)" *ngIf="item.status === ChartReleaseStatus.Active && item.portals && item.portals.user_portal && item.portals.user_portal[0]"
```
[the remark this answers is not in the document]
Ornias1993: > Is it up to every app catalog to define names for portals or is it driven by truenas middleware?

It's set in questions.yaml, where portals gets loaded into item.portals.

Let me show a snippet:
```
portals:
  web_portal:
    protocols:
      - "$kubernetes-resource_configmap_portal_protocol"
    host:
      - "$kubernetes-resource_configmap_portal_host"
    ports:
      - "$kubernetes-resource_configmap_portal_port"
```

> If we add support for admin portals now, how would it work with an official app catalog?

Currently buttons are not shown if either portals as a whole or the specific portal are not set.
So in the case of the official catalog it would just not show the admin portal button :)

> In either case, I'd prefer to see changes made elsewhere before seeing them in webui purely so that I could test new behaviour. 

I'll work on setting up a demo to test and showcase this :)

> Having backward compatibility would also be nice, but it depends on answers to previous two questions.

Thinking about this some more, and lets not rename web_portal for compatibility at all, would make it much easier and less breaking to implement... I'll add that later.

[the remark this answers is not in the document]
undsoft: Would it make sense to somehow add support for portal label in `questions.yaml` and let developers have whatever portal buttons they want?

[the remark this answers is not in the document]
Ornias1993: I think that would be better suited for a seperate PR, as this is mostly focussed on splitting user-facing portal(s) from admin-only portal(s), which is a quite frequent occurance.

As those portals are lists, it's quite easy to, in the future, add customizable labels. Though i'm not sure it would require a middleware change to do so. Might be more suited for Bluefin tbh.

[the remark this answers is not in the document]
Ornias1993: Demo is working:
image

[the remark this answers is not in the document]
Ornias1993: I'll now take a look at making dynamic labels :)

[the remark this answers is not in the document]
Ornias1993: Okey, i've been able to implement dynamic labels based on the name of the actual portals inside questions.yaml :)
In the TrueCharts catalog, in the "incubator" train there is an example App based on pihole available to test :)

Known downsides:
Needs work oin the ports, possibly move those and give Catalog creators that whole bare to work with.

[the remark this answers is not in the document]
Ornias1993: Okey @undsoft :
Dynamic naming and the number of buttons depending on the questions.yaml file is done.
It also dynamically scales the ports now and hides the port references completely if they become to small.

It think that's all you feedback done :)

src/app/pages/applications/chart-releases/chart-releases.component.html
```diff
@@ -26,10 +26,15 @@
...
         <div class="chart-button-row" fxLayout="row" fxLayoutAlign="end end" (click)="$event.stopPropagation()">
-          <button mat-button (click)="portal(item)" *ngIf="item.status === ChartReleaseStatus.Active && item.portals && item.portals.web_portal && item.portals.web_portal[0]"
+          <div class="ports" >
+		  <span class="ports-text"> {{ appService.getPorts(item) | lowercase }}</span>
+          </div>
+		  <div *ngFor="let portal of item.portals | keyvalue">
+          <button mat-button (click)="portalLink(item, portal.key )" *ngIf="item.status === ChartReleaseStatus.Active"
               ix-auto ix-auto-type="button" ix-auto-identifier="{{ item.name}}-portal">
-            {{ 'Portal' | translate }}
+            {{ portal.key | translate }}
```
undsoft, line 35: `| translate` is unnecessary, as translation system won't be able to pick up dynamic text coming from api.

src/app/pages/applications/applications.component.scss
```diff
@@ -171,6 +171,8 @@ mat-tab-group {
 
     .chart-button-row {
       bottom: 5px;
+	  width: 315px;
+	  height: 45px;
```
undsoft, line 175: How did you manage to circumvent our linter? :)
It should have ran during pre-commit git hooks. Please run manually via `yarn lint:fix`

src/app/pages/applications/chart-releases/chart-releases.component.html
```diff
@@ -26,10 +26,15 @@
         </div>
 
         <div class="chart-button-row" fxLayout="row" fxLayoutAlign="end end" (click)="$event.stopPropagation()">
-          <button mat-button (click)="portal(item)" *ngIf="item.status === ChartReleaseStatus.Active && item.portals && item.portals.web_portal && item.portals.web_portal[0]"
+          <div class="ports" >
+		  <span class="ports-text"> {{ appService.getPorts(item) | lowercase }}</span>
+          </div>
+		  <div *ngFor="let portal of item.portals | keyvalue">
```
undsoft, line 32: Indentation is off.
Also let's move ngIf from down below higher, you can wrap this div with <ng-container *ngIf="item.status === ChartReleaseStatus.Active">

src/app/pages/applications/chart-releases/chart-releases.component.html
undsoft: There is a `snakeToPascal` function in `EntityUtils` that you can use to have something `Web Portal` instead of `Web_portal` for now.

src/app/pages/applications/applications.component.scss
```diff
@@ -212,11 +214,22 @@ mat-tab-group {
...
+    text-align: left;
+    vertical-align: middle;
+    padding-right: 5px;
+    overflow: hidden;
+	display: flex;
   }
-
+  .ports-text {
+    width: 100%;
+    max-width: max(0px, calc((100% - 50px)*999));
```
undsoft, line 227: Can you explain/add about a comment about what this does? I don't see a difference if I remove this line.

undsoft (changes requested):

src/app/pages/applications/chart-releases/chart-releases.component.html
Ornias1993: I wasn't sure if it would translate or not, will change :)

That snakeToPascal function is cool though, thanks for the tip :)

src/app/pages/applications/applications.component.scss
Ornias1993: Deleting it, without staging the deletion. ;-)
That damn linter prevented me from pushing my previous PR's without any feedback what was wrong and added an absolute shit tone of translations into the PR that I was previously told to not-include in PR's.

This indent though >.< 
Will fix!

src/app/pages/applications/chart-releases/chart-releases.component.html
Ornias1993: > Indentation is off.

Shoot, yes... I'm used to my tab-to-spaces pre-commit hook running, will fix!
And yes indeed, it shouldn't show if it isn't active, just like the buttons...

src/app/pages/applications/applications.component.scss
Ornias1993: Will add a comment.

- When there is a slight overlap, the text just get shortned with dots `53/TCP, 53/UD...` 
- However in some cases the buttons don't leave enough room for the ports to show, which ends up with `53`. In those cases (less than 50px width) it removes the text completely. When text is more than 50, it is set to an "endless" max-width, which is limited by the 100% width in practice.

Ornias1993: Without:

image


With:

image


Simply put:
It complete hides all the text if "just a sliver" would remain, instead of showing just 1, 2 or 3 numbers...

src/app/pages/applications/chart-releases/chart-releases.component.html
Ornias1993: Worked like a charm, though snakeToPascal just removes _, I added a function snakeToHuman, that replaces _ with space. As I think that might be usefull in more places :)

Ornias1993:
Known issue:
It does not have the nice popup on-hover when shortened. As it's shortened based on the pixed width of the div, which (afaik) is not supported by `textlimiter`.

undsoft (approved):
Looks good. Nice work.

PR #5901, opened 2021-09-15
NAS-112059 / 21.10 / Use iX Blue + Dark Blue for official Apps

Ornias1993:
As explained in NAS-112059 people are too easily confused between official and community Apps at the moment.

This PR gives two color options:
- Official in a combination of iX Blue and Dark blue
- Community in Yellow and the original disgusting dark yellow.

For Bluefin this, most likely, would need to be redesigned to deal with more than just two color options, but for Anglefish this solves the problem without too much code redundency imho.

**Before:**
image

**After:**
image

### DIFF
--- a/src/app/pages/applications/catalog/catalog.component.html
+++ b/src/app/pages/applications/catalog/catalog.component.html
@@ -24,7 +24,7 @@
               {{ item.catalog.train }}
             </div>
           </div>
-        <div class="badge-area" *ngIf="item.catalog.label != 'OFFICIAL'">
+          <div class="badge-area" *ngIf="item.catalog.label != 'OFFICIAL'">
             <div class="capitalized-text">
               {{ item.catalog.label }}
             </div>
--- a/src/app/pages/applications/catalog/catalog.component.html
+++ b/src/app/pages/applications/catalog/catalog.component.html
@@ -16,8 +16,16 @@
           <div class="logo">
             <img [src]="item.icon_url" [src-fallback]="imagePlaceholder"/>
           </div>
-          <div class="badge-area">
-            <div class="partner-badge capitalized-text">
+          <div class="badge-area" style="background-color: var(--primary)" *ngIf="item.catalog.label == 'OFFICIAL'">
+            <div class="capitalized-text" >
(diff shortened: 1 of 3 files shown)


RehanY147 (changes requested):
Maybe we should try white text on top the blue background?

Ornias1993:
> Maybe we should try white text on top the blue background?

That would become extremely flashy, which means making the blue darker, which in turn makes it stand out extremely from the dark yellow :(

i'll try tweaking the colors some more to look if I can somehow compensate the black-on-blue a bit :)

Ornias1993:
> Maybe we should try white text on top the blue background?

I've changed the PR a bit:
- Color of the train-label is now standardised for both the blue and yellow labels (it uses the same black overlay color)
- Reduced the amount of duplicate CSS code
- Also implemented the new scheme in the "Installed Apps" section

RehanY147:
@Ornias1993 Please provide the latest screenshots with this

RehanY147 (commented):
.

Ornias1993:
> @Ornias1993 Please provide the latest screenshots with this

Understandable, with my latest changes i've edited the PR description and screenshots right away accordingly :)

src/app/pages/applications/catalog/catalog.component.html
```diff
@@ -16,8 +16,16 @@
           <div class="logo">
             <img [src]="item.icon_url" [src-fallback]="imagePlaceholder"/>
           </div>
-          <div class="badge-area">
-            <div class="partner-badge capitalized-text">
+          <div class="badge-area" style="background-color: var(--primary)" *ngIf="item.catalog.label == 'OFFICIAL'">
```
RehanY147, line 19: We should put the styles in CSS classes

RehanY147 (approved):
Looks good. Thank you!

src/app/pages/applications/catalog/catalog.component.html
Ornias1993: I've moved it to an additional class that only overrides the color.

RehanY147 (approved):
Thank you!

PR #5902, opened 2021-09-15
NAS-112068 / 21.10 / Dynamic width of chart update/status and eventlog.

Ornias1993:
With this PR the width of the Chart-Status and Upgrade modals are made more dynamic.
The minimum width is set to the current fixed-width, while the maximum is set a few hunderd pixels higher.
In between the size will be based on 50vh.

This also ports the recent patch that shows more of the Release notes, to the chart summary, primarily the app eventlog.

It basically solves the remaining issues regarding sizing of those modals as stated in NAS-112068

**Example showing the combined effect of these measures on a 4K display:**
image


denysbutenko (approved):
Looks good! Thanks for the contribution.

PR #5905, opened 2021-09-16
NAS-112080 / 21.10 / FieldConfig Part 2

dszidi:
Moved more things out of BaseFieldConfig and into separate child interfaces.


src/app/pages/common/entity/entity-dialog/entity-dialog.component.ts
```diff
@@ -80,7 +86,21 @@ export class EntityDialogComponent<P = any> implements OnInit {
...
+          case 'dict':
+            const dictConfig: FormDictConfig = conf;
+            return dictConfig;
+          case 'selectionlist':
+            const selectionListConfig: FormSelectionListConfig = conf;
+            return selectionListConfig;
+        }
+      };
+
+      const config = configType(this.fieldConfig[i]);
```
undsoft, line 103: Same as just doing:
```
const config = this.fieldConfig[i] as RelationConfig;
```

src/app/pages/common/entity/entity-form/entity-form-embedded.component.ts
```diff
@@ -426,4 +435,20 @@ export class EntityFormEmbeddedComponent implements OnInit, OnDestroy, AfterView
...
+        return listConfig;
+      case 'dict':
+        const dictConfig: FormDictConfig = config;
+        return dictConfig;
+      case 'selectionlist':
+        const selectionListConfig: FormSelectionListConfig = config;
+        return selectionListConfig;
+      default:
+        return null;
+    }
```
undsoft, line 452: Same here. Additional type information that you are providing inside a method doesn't actually leave it, and you get `RelationConfig` out.

src/app/pages/common/entity/entity-form/models/field-config.interface.ts
```diff
@@ -98,6 +91,7 @@ export interface FormComboboxOption {
 
 export interface FormDictConfig<P = any> extends BaseFieldConfig<P> {
   label?: string;
+  relation?: RelationGroup[];
```
undsoft, line 94: Isn't relation a base field though? `createFormGroup` in `EntityFormService` previously didn't care about the type of field config, when dealing with relations.
I suggest you test previous code to see if it supported relations on any type of field.

undsoft (changes requested):
Not sure about `relation` changes. Also this would be a good opportunity to add `type: 'select'`, `type: checkbox` to all the individual field configs.

src/app/pages/common/entity/entity-form/models/field-config.interface.ts
dszidi: From what I saw, the only things using those are the list, selectionlist and dict types which is why I moved it out of the base config.

src/app/pages/common/entity/entity-dialog/entity-dialog.component.ts
dszidi: Fixed

src/app/pages/common/entity/entity-form/models/field-config.interface.ts
undsoft: Please verify it by creating a simple form configuration somewhere with text inputs and relations.

dszidi: I put relation back into baseconfig and removed RelationConfig

src/app/pages/common/entity/entity-form/models/field-config.interface.ts
```diff
@@ -156,19 +165,22 @@ export interface FormParagraphConfig<P = any> extends BaseFieldConfig<P> {
   paragraphIconSize?: string;
   paraText?: string;
   inputType?: string;
+  // type: 'paragraph';
```
undsoft, line 168: //

src/app/pages/common/entity/entity-form/services/field-relation.service.ts
```diff
@@ -22,13 +26,14 @@ export class FieldRelationService {
     });
   }
 
-  getRelatedFormControls(model: FieldConfig,
+  getRelatedFormControls(config: FieldConfig,
     controlGroup: FormGroup): FormControl[] {
     const controls: FormControl[] = [];
+    // const model = this.getConfigType(config);
```
undsoft, line 32: //

src/app/pages/common/entity/entity-form/models/field-config.interface.ts
```diff
@@ -23,19 +23,13 @@ export interface InputUnitConfig {
 
 export interface BaseFieldConfig<P = any> {
```
undsoft, line 24: We can update `FieldConfig<P = any>` definition not to include `BaseFieldConfig<P>` anymore, since people should be using narrowed down definitions.

undsoft (approved):
Looks good, minor changes only.

PR #5906, opened 2021-09-16
NAS-112378 / 21.10 / NAS-112378: Fix Services page

denysbutenko:


undsoft (approved):
It works, but do you have any idea why it broke all of a sudden?

denysbutenko:
> It works, but do you have any idea why it broke all of a sudden?

Nope, found it while testing PR#5903. Probably it was broken in the last two weeks since Cloud VM uses the version from 1 Sept and it is not affected.

PR #5912, opened 2021-09-18
NAS-112424 / 21.10 / NAS-112424: Fix delete message on Shares

denysbutenko:
Before:
image
After:
image


undsoft (changes requested):
When trying to delete an NFS share from sharing dashboard, I get:
> Delete NFS Share undefined?

undsoft (approved):
Thanks, works well now.

PR #5916, opened 2021-09-20
NAS-112435 / 21.10 / NAS-112435: Avoid button overlap

dszidi:
To test:

- Create some replication task (ticket says cloud sync but can be any kind really)
- Misconfigure it so it fails
- Click on the "FAILED" in the state table cell
- In the error dialog click on "More info" button to show the error message.
- Make sure the white text-area does not extend so far that the action buttons cover it.

### DIFF
--- a/src/app/pages/common/error-dialog/error-dialog.component.html
+++ b/src/app/pages/common/error-dialog/error-dialog.component.html
@@ -21,8 +21,6 @@
   </div>
 </div>
 <div mat-dialog-actions>
-  &nbsp;
-  <span fxFlex></span>
   <button mat-button class="mat-button mat-accent" (click)="dialogRef.close(true)"
     ix-auto
     ix-auto-type="button"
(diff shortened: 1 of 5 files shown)


dszidi:
This is what it should look like. In addition to the overlapping issue mentioned in the ticket, I also fixed the "More info" text alignment with the icon. Also made the message area use the theme colors instead of defaulting to white which was really loud on dark themes

image

denysbutenko:
image
`pre` area looking cut off (should take all width) and the buttons get overlap with `textarea`

dszidi:
> `pre` area looking cut off (should take all width) and the buttons get overlap with `textarea`

Yep. I caught that too and am already working on it. Only happens when window height drops below a certain number of pixels.

dszidi:
@denysbutenko Try it now

image

denysbutenko (approved):
Looks nice! Two small things, cursor on "More info" hover looking weird and textarea resize may unnecessary

dszidi:
Yes I agree on the textarea. I'm not really sure why that element is being used here.

PR #5924, opened 2021-09-22
NAS-112282 / 21.10 / NAS-112282: Fixing recursive checkbox when taking snapshot

undsoft:


dszidi (approved):
Code looks good and works as expected

image

PR #5925, opened 2021-09-22
NAS-112363 / 21.10 / NAS-112363: Sending IP type when editing an interface

undsoft:
Use a form in Network -> Interfaces and check that correct `type` is now sent to middleware for every IP Address (alias).


AlexKarpov98 (approved):
LGTM

PR #5932, opened 2021-09-23
NAS-112510 / 21.10 / Filter svclb from pod selector

Ornias1993:
This PR removes all "Service Loadbalancer" pods from the log and shell selector.
These pods are created by the K3S ServiceLoadbalancer and their logs have no value for the user.

They also lead to significant confusion, with users submitting the wrong logs.

**Before:**
---
image

image

**After:**
---
image

image

### DIFF
--- a/src/app/pages/applications/chart-releases/chart-releases.component.ts
+++ b/src/app/pages/applications/chart-releases/chart-releases.component.ts
@@ -501,6 +501,15 @@ export class ChartReleasesComponent implements OnInit {
     this.refreshToolbarMenus();
   }
 
+  filterPodNames(podList: string[]): string[] {
+    for (const entry of podList) {
+      if (entry.includes('svclb')) {
+        podList = podList.filter((name) => name !== entry);
+      }
+    }
+    return podList;
+  }
+
   openShell(name: string): void {
     this.podList = [];
     this.podDetails = {};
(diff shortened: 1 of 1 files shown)


src/app/pages/applications/chart-releases/chart-releases.component.ts
```diff
@@ -501,6 +501,15 @@ export class ChartReleasesComponent implements OnInit {
     this.refreshToolbarMenus();
   }
 
+  filterPodNames(podList: string[]): string[] {
+    for (const entry of podList) {
+      if (entry.includes('svclb')) {
```
undsoft, line 506: Let's add a comment explaining what this does.

src/app/pages/applications/chart-releases/chart-releases.component.ts
```diff
@@ -501,6 +501,15 @@ export class ChartReleasesComponent implements OnInit {
     this.refreshToolbarMenus();
   }
 
+  filterPodNames(podList: string[]): string[] {
+    for (const entry of podList) {
```
undsoft, line 505: You can probably just do
```
return podList.filter((pod) => pod.includes('svclb'));
```

undsoft (changes requested):
It looks good, but let's also wait for Waqar's approval.

Ornias1993:
> It looks good, but let's also wait for Waqar's approval.

Thanks for your feedback!
I've removed the function and directly filtered the podnames like you showed. Also added a comment to make clear why it's done.

I've passed this past Waqar the 06-sept  and got no objection (while discussing https://github.com/truenas/middleware/pull/7481 ) :)

sonicaj:
@Ornias1993 this is not required as middleware is not showing the pods spawned from daemonsets.

PR #5933, opened 2021-09-23
DOCS-2367 / 12.0 / added core/test_NAS_T1222.py with test case steps and code

ericbsd:
to verify core file alert works

here are the results https://builds.ixsystems.com/job/TrueNAS%2012.0%20Internal%20CORE%20UI%20Tests%20from%20Branch/8/


q5sys (approved):
LGTM

PR #5934, opened 2021-09-23
NAS-112431 / 21.10 / NAS-112431: Enabling @typescript-eslint/no-this-alias

undsoft:
Mostly replaces `let that = this;` and some `entityDialog.parent` with arrow functions.
Accidentally fixes a bug where in General -> Configuration cannot be downloaded.

Testing: review the code and do spot testing in a couple of places that look risky.


dszidi (approved):
LGTM

PR #5939, opened 2021-09-24
NAS-112269 / 21.10 / NAS-112269: Maybe fixing GPU field in VM form

undsoft:
I was not able to reproduce the issue as described in the ticket.
This is essentially a blind fix.


RehanY147 (approved):
The screenshots on the ticket show older UI with another issue that's been solved as well. But, the proposed solution looks simple enough so approving.

PR #5945, opened 2021-09-26
NAS-112524 / 21.10 / Do not allow user to navigate to next step if current has errors in Apps

AlexKarpov98:


denysbutenko (approved):
image

Works as described 👍🏻

PR #5948, opened 2021-09-27
NAS-111467 / 21.10 / Add tests for Task Manager

denysbutenko:


denysbutenko:
Some tests are failed due to timezone differences. Looking for a solution...

undsoft (approved):
So far looking pretty good.

PR #5950, opened 2021-09-28
NAS-112429 / 21.10 / NAS-112517: Update entity table when row is deleted when filter is used

undsoft:
Main change is in `entity-table.component.ts`. This change is somewhat risky, because a) system-wide and b) I don't know why it was there in the first place.

Some additional cleanup included.


denysbutenko:
The table row is still present after successful deleting

denysbutenko (approved):
LGTM

PR #5951, opened 2021-09-28
NAS-112572 / 21.10 / NAS-112572: Enable additional linter rules

undsoft:
No testing is necessary.

Enables additional linter rules:
- Force assertions to be `variable as type` instead of `<type>variable`. There is no difference in our codebase, but former is recommended for react codebases, so it's a good habit to have.
- Force `if(variable === true)` to be `if(variable)` when variable is boolean.
- Force type to be specified in catch block


AlexKarpov98 (approved):
Nice work!

PR #5955, opened 2021-09-29
IxCombobox fixes and added IxUserCombobox

RehanY147:
NAS-112309

### DIFF
--- a/src/app/pages/common/ix/components/ix-combobox/ix-combobox.component.ts
+++ b/src/app/pages/common/ix/components/ix-combobox/ix-combobox.component.ts
@@ -1,11 +1,15 @@
 import {
-  Component, ElementRef, EventEmitter, forwardRef, Input, OnChanges, Output, SimpleChanges, ViewChild,
+  Component, ElementRef, EventEmitter, forwardRef, Input, OnChanges, OnInit, Output, SimpleChanges, ViewChild,
 } from '@angular/core';
 import { ControlValueAccessor, FormControl, NG_VALUE_ACCESSOR } from '@angular/forms';
 import { MatAutocomplete, MatAutocompleteTrigger } from '@angular/material/autocomplete';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
-import { fromEvent, Observable, of } from 'rxjs';
-import { map, takeUntil } from 'rxjs/operators';
+import {
+  fromEvent, Observable, of, Subject,
+} from 'rxjs';
+import {
+  debounceTime, distinctUntilChanged, map, takeUntil,
+} from 'rxjs/operators';
 import { Option } from 'app/interfaces/option.interface';
 
 @UntilDestroy()
@@ -21,7 +25,7 @@ import { Option } from 'app/interfaces/option.interface';
     },
   ],
 })
-export class IxComboboxComponent implements ControlValueAccessor, OnChanges {
+export class IxComboboxComponent implements ControlValueAccessor, OnChanges, OnInit {
   @Input() label: string;
   @Input() hint: string;
   @Input() required: boolean;
@@ -42,7 +46,7 @@ export class IxComboboxComponent implements ControlValueAccessor, OnChanges {
   };
 
   filteredOptions: Observable<Option[]>;
-
+  filterChanged$ = new Subject<string>();
   formControl = new FormControl(this);
   value: string | number = '';
   filterValue = '';
@@ -61,6 +65,21 @@ export class IxComboboxComponent implements ControlValueAccessor, OnChanges {
     if (this.selectedOption) {
       this.filterValue = this.selectedOption.label;
     }
+  }
+
+  ngOnInit(): void {
+    this.filterChanged$.pipe(
+      debounceTime(300),
+      distinctUntilChanged(),
+      untilDestroyed(this),
+    ).subscribe((changedValue: string) => {
+      this.filterValue = changedValue;
+      if (changedValue) {
+        this.filteredOptions = this.filter(this.syncOptions, changedValue);
+      } else {
+        this.filteredOptions = of(this.syncOptions);
+      }
+    });
   }
 
   onOpenDropdown(): void {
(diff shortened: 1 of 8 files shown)


denysbutenko:
https://user-images.githubusercontent.com/351613/135223074-bbf8db42-aac3-4d33-b65e-7ffddfc229f8.mp4

Seems like a bug on scrolling

denysbutenko:
image
selected option has the wrong bg color

also, would be nice to support keyboard arrows to select an option from autocomplete. possibly in a separate ticket

src/app/pages/common/ix/components/ix-user-combobox/ix-user-combobox.component.html
```diff
@@ -0,0 +1,32 @@
...
+      <span *ngIf="required">*</span>
+    </label>
+    <tooltip *ngIf="tooltip" [header]="label" class="tooltip" [message]="tooltip"></tooltip>
+  </div>
+
+
+  <div class="outlined input-container">
+    <input #ixInput matInput
+      [value]="selectedOption?.label"
+      [placeholder]="'Search'"
```
denysbutenko, line 15: missing translate pipe

src/app/pages/common/ix/components/ix-combobox/ix-combobox.component.ts
```diff
@@ -58,13 +67,52 @@ export class IxComboboxComponent implements ControlValueAccessor, OnChanges {
...
+            takeUntil(this.autocompleteTrigger.panelClosingActions),
+            untilDestroyed(this),
+          )
+          .subscribe(() => {
+            const scrollTop = this.autoCompleteRef.panel.nativeElement
+              .scrollTop;
+            const scrollHeight = this.autoCompleteRef.panel.nativeElement
+              .scrollHeight;
+            const elementHeight = this.autoCompleteRef.panel.nativeElement
+              .clientHeight;
```
denysbutenko, line 104: nit: these lines can be optimized using object destructuring
```
const { scrollTop, scrollHeight, clientHeight } = this.autoCompleteRef.panel.nativeElement;
// or with renaming
const { scrollTop, scrollHeight, clientHeight: elementHeight } = this.autoCompleteRef.panel.nativeElement;
```

denysbutenko (commented):
image
missing dropdown caret on the combobox

https://user-images.githubusercontent.com/351613/135909019-51e6a090-f58e-482b-b6ed-951554f52ddf.mp4

there are some issues:
- after input is cleared it is not possible to choose the currently selected language
- list of options has last filtered values after selecting the language

denysbutenko:
regarding user input:

https://user-images.githubusercontent.com/351613/135910410-1ff18e92-e162-4360-82e9-2e7c43b19f45.mp4

there is some delay before getting results on filtering (that's okay) and then options get duplicated in the list.
maybe it is worth adding a small spinner on the right to notify users or find different way to indicate about loading

denysbutenko:
Another small issue:
https://user-images.githubusercontent.com/351613/135911583-8e5f4fbb-8274-4dfc-855c-e977e8df0f3c.mp4

src/app/pages/common/ix/components/ix-combobox/ix-combobox.component.ts
```diff
@@ -19,13 +26,17 @@ import { Option } from 'app/interfaces/option.interface';
...
   @Input() label: string;
   @Input() hint: string;
   @Input() required: boolean;
   @Input() tooltip: string;
   @Input() options: Observable<Option[]>;
   @ViewChild('ixInput') inputElementRef: ElementRef;
+  @ViewChild('auto') autoCompleteRef: MatAutocomplete;
+  @ViewChild(MatAutocompleteTrigger) autocompleteTrigger: MatAutocompleteTrigger;
+  @Output() scrollEnd: EventEmitter<string> = new EventEmitter<string>();
+  placeholder = T('Search');
```
denysbutenko, line 39: also, it probably makes sense to add `@Input()` decorator to make it customizable

RehanY147:
> missing dropdown caret on the combobox

As discussed with @undsoft It's visually a better choice to have a placeholder like 'Search' instead of a dropdown caret. Up for discussion.

RehanY147:
> there is some delay before getting results on filtering (that's okay) and then options get duplicated in the list. maybe it is worth adding a small spinner on the right to notify users or find different way to indicate about loading

Adding a loading icon that appears and disappears every second on fast load times might just make the input seem buggy. Other examples of auto-complete like Google search don't have a loading icon for example. Just my opinion. We can have a discussion on this if you'd like. let me know please.

denysbutenko:
> > missing dropdown caret on the combobox
> 
> As discussed with @undsoft It's visually a better choice to have a placeholder like 'Search' instead of a dropdown caret. Up for discussion.

Oh, that's fine. I thought it is a bug because we have a caret before.

> > there is some delay before getting results on filtering (that's okay) and then options get duplicated in the list. maybe it is worth adding a small spinner on the right to notify users or find different way to indicate about loading
> 
> Adding a loading icon that appears and disappears every second on fast load times might just make the input seem buggy. Other examples of auto-complete like Google search don't have a loading icon for example. Just my opinion. We can have a discussion on this if you'd like. let me know please.

I'm using Cloud VM which is working very slowly that's why asked for that. I'm okay to keep it as is. 👍🏻 

Thanks for the clarification!

src/app/pages/common/ix/components/ix-combobox/ix-combobox.component.ts
RehanY147: The idea is that keeping the placeholder fixed to `Search` would indicate that the field is searchable. Otherwise, it might seem like just another text input.

RehanY147:
@denysbutenko Thank you for the detailed testing. Much appreciated :)

src/app/pages/common/ix/components/ix-user-combobox/ix-user-combobox.component.scss
```diff
@@ -58,3 +58,7 @@
 .ix-selected {
   background: var(--primary) !important;
 }
+
+.ix-unselected {
```
denysbutenko, line 62: I thinking of reusing the base styles for custom implementation of the combobox. To minimize the chance of error and simplify maintenance. Maybe it would make sense to  `@import ix-combobox.component.scss` in `ix-user-combobox.component.scss`?

denysbutenko:
Regarding translation files, there are changes in the master branch that requires updating dependencies. I'd suggest the following scenario to fix it:

- Execute `yarn install`
- Modify `.husky/pre-commit` to comment last 3 lines (but don't include changes to commit)
- Edit existing commits to exclude `*.json` files (because it breaks translated strings)
- When no translation files are touched in this branch then execute `yarn extract` and put them in a separate commit
- So, we can make sure that all strings are in place
- Discard changes for `.husky/pre-commit`

PR #5956, opened 2021-09-29
NAS-112570 / 21.10 / NAS-112570: Adjustments for notification component

denysbutenko:
Changes:

- Add i18n support
- Reformat template
- Style improvements
- Make visible the `Dismiss All Alerts` button


undsoft (approved):
Looks good, too bad about `!important`, but I can see that they are needed.

PR #5957, opened 2021-09-29
NAS-112587 / 21.10 / NAS-112587: Remove support for legacy encryption

undsoft:
Some places with `encrypt` in code remain.

To test: check if storage page works correctly, specially with regards to encryption.


AlexKarpov98 (commented):
Not sure how to test it.
Is encryption deprecated on dataset or vdev, zvol as well as pool?
Also still I see encryption option during adding a pool.
And what happens with existing encrypted pool?
image

undsoft:
Sorry, more clarification.
What I removed was an old type of encryption not available on scale. This encryption happened on pool level. This encryption shouldn't be supported by middleware, so there should be no existing pools with this legacy encryption.

Here's a screenshot from FreeNAS 11.3 with encryption. Notice additional indication and menu on the pool itself.
Снимок экрана 2021-09-30 в 10 00 15

Modern encryption will show up on datasets.
So, there should be no change from what you're used to in scale - you should still be able to tick the encryption checkbox when creating a pool, it will appear as an encrypted root dataset. You will also be able to create additional encrypted child datasets.

PR #5959, opened 2021-09-29
NAS-112289 / 21.10 / NAS-112289: Fixed actions addition to include Extend

RehanY147:
NAS-112289

To test
1. Create a pool
2. Have an extra unassigned disk attached with the VM
3. Go to "Status" page from the dropdown menu from the cog icon on the pool table header
4. Check if the `Extend` option is present on the three dot menu of the pool children


dszidi (approved):
LGTM

image

PR #5962, opened 2021-09-30
NAS-112612 / 21.10 / NAS-112612: Enable @typescript-eslint/prefer-for-of

undsoft:
This bans `for (let i = 0; i < something.length; i++)` when you don't actually use an index.
You can use `items.forEach` or `for (const item of items)`.

Airbnb codestyle generally prefers former, so most of the code was converted to `forEach`. `forof` was used in some places where I was unsure about type of the variable or for other reasons.

No testing is necessary.


AlexKarpov98 (approved):
Nice work!

PR #5965, opened 2021-09-30
NAS-112597 / 21.10 / NAS-112597: The system returned the following error - [object Object]

AlexKarpov98:


src/app/pages/common/entity/entity-table/entity-table.component.ts
```diff
@@ -592,7 +596,7 @@ export class EntityTableComponent<Row = any> implements OnInit, AfterViewChecked
       },
       (res: any) => {
         this.isTableEmpty = true;
-        this.configureEmptyTable(EmptyType.Errors, res);
+        this.configureEmptyTable(EmptyType.Errors, res.reason);
```
undsoft, line 599: I don't think res is guaranteed to have `reason` as evidenced by code several lines below.
Let's do `res.reason || res.error` or something similar.

undsoft: Also please change `error: any` in `configureEmptyTable` to `error: string`.

undsoft (changes requested):

undsoft (approved):
I could not actually reproduce in UI, but the code look okay.

PR #5966, opened 2021-10-01
NAS-112631 / 21.10 / NAS-112631: Sync badge colors

denysbutenko:
For testing, go to Applications and check if the badge has correct colors (blue for official charts and yellow for others) on Installed and Available tabs.


AlexKarpov98 (approved):
LGTM

PR #5967, opened 2021-10-01
NAS-112166 / 21.10 / NAS-112166: Replace setTimeouts w/Observable stream and use OnPush strategy

dszidi:
To test:
- create 500+ snapshots.
- go to storage/snapshots
- make sure the table loads in a reasonable amount of time. 

For context, my machine was originally loading 595 snapshots in 5+ seconds and now it is down to 1 second or less)

### DIFF
--- a/src/app/pages/common/entity/entity-table/entity-table.component.html
+++ b/src/app/pages/common/entity/entity-table/entity-table.component.html
@@ -1,6 +1,6 @@
 <div fxLayout="column" id="entity-table-component" class="material mat-card mat-card-table">
 
-  <ng-container *ngIf="!isTableEmpty">
+  <ng-container>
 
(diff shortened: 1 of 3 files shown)


denysbutenko (approved):
table header and pagination visible while loading
image

noticed this error but am not sure which action invoked it
image

https://user-images.githubusercontent.com/351613/135708380-8d88b715-3789-47fc-8331-5f07046ccd45.mp4


https://user-images.githubusercontent.com/351613/135708384-bc3b8b1b-e080-4eb8-87b8-6b3dd470294b.mp4


Looks good and works x2 faster on Cloud VM than before.
The page with 900+ snapshots loaded in 13 seconds with these changes.

undsoft (approved):
Change look arcane :), but it does work faster.

PR #5969, opened 2021-10-04
NAS-112425 / 12.0 / Cloud Sync Tasks should show Disabled instead of a Next Run time if task is disabled

AlexKarpov98:


undsoft (approved):
Looks good.
Снимок экрана 2021-10-04 в 11 56 47

PR #5970, opened 2021-10-04
NAS-110194 / 21.10 / Upgrade ngx-translate-extract and fix i18n strings

denysbutenko:
Changes:

- `ngx-translate-extract` upgraded from `2.3.4` to the latest version `7.0.4`
- Our translate marker `T()` is replaced with marker `ngx-translate-extract-marker`. To use it just `import { marker as T } from '@biesbjerg/ngx-translate-extract-marker';` but `this.translate.instant()` is preferred way.
- fixed strings with variables
- some strings were merged
- backported translated strings (almost all)
- `translate.get()` was finally replaced with `translate.instant()`

Testing: general regression test on pages


undsoft:
1. I suppose reexporing `marker` as `T` didn't work?
2. Previously untranslated strings were shown alphabetically on top of the json files to make it easier for contributors to see what needs to be translated. This does not seem to be the case anymore.
3. Some translations dissapeared for some reason. Check `Destination Location` in `fr.json`. It got replaced with space.
4. Some items that previously didn't have translations now have English strings as translations. Check `has been unlocked.` in same `fr.json`.

denysbutenko:
> 1. I suppose reexporing `marker` as `T` didn't work?
> 2. Previously untranslated strings were shown alphabetically on top of the json files to make it easier for contributors to see what needs to be translated. This does not seem to be the case anymore.
> 3. Some translations dissapeared for some reason. Check `Destination Location` in `fr.json`. It got replaced with space.
> 4. Some items that previously didn't have translations now have English strings as translations. Check `has been unlocked.` in same `fr.json`.

1. Unfortunately, yes. I'd like to keep a local marker wrapper if possible
Thanks, I'll check other issues.

The benefit would be reducing the execution time of `extract_string.js` by about x2 times on each commit.

src/app/helptext/vm/vm-wizard/vm-wizard.ts
```diff
@@ -64,8 +64,7 @@ export default {
  address can be chosen.'),
 
   vcpus_label: T('CPU and Memory'),
-  vcpus_warning: T('The product of vCPUs, cores and threads must not exceed'),
-  vcpus_warning_b: T('on this system.'),
+  vcpus_warning: T('The product of vCPUs, cores and threads must not exceed {maxVCPUs} on this system.'),
```
AlexKarpov98, line 67: IMHO, we can use the same way for globalHelptext as you did here. If I am wrong, please let me know.

undsoft (approved):
Looks good to me, nice work 👍🏻

src/app/helptext/directory-service/active-directory.ts
```diff
@@ -137,7 +136,7 @@ export default {
   activedirectory_netbiosname_a_validation: [Validators.required, Validators.maxLength(15)],
 
   activedirectory_netbiosname_b_name: 'netbiosname_b',
-  activedirectory_netbiosname_b_placeholder: T(`Netbios Name (${globalHelptext.Ctrlr} 2)`),
+  activedirectory_netbiosname_b_placeholder: T('Netbios Name (TrueNAS Controller 2)'),
```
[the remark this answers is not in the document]
denysbutenko: As per ticket description:
```
Strings like:

T(`Netbios Name (${globalHelptext.Ctrlr} 2)`)
will not be picked by ngx-translate-extract.
```

src/app/helptext/vm/vm-wizard/vm-wizard.ts
denysbutenko: The `globalHelptext.Ctrlr` is a static string, where `maxVCPUs` is a variable. We could do this way but it doesn't make sense, since it'd require additional code in components to pass the same string.
Something like:
```diff
- activedirectory_netbiosname_b_placeholder: T(`Netbios Name (${globalHelptext.Ctrlr} 2)`)
+ activedirectory_netbiosname_b_placeholder: T('Netbios Name ({controller} 2')
...
- placeholder: helptext.activedirectory_netbiosname_b_placeholder
+ placeholder: this.translate.instant(helptext.activedirectory_netbiosname_b_placeholder, { controller: globalHelptext.Ctrlr });
...

PR #5975, opened 2021-10-06
NAS-112684: Fix entity table

denysbutenko:
The table header and pagination should be hidden while loading or the table is empty


RehanY147 (changes requested):
Edit: I don't think it's relevant to this PR. I'll create a new ticket if this issue persists. You can ignore the following comment.
Not sure if this is relevant to this PR but I'm seeing this option to add VMs when the system doesn't support VMs and we should be seeing that message instead.
image

RehanY147 (changes requested):
If there's no information to show, could we try and hide the empty row in the expanded row state?
image

RehanY147 (approved):
Looks good

PR #5979, opened 2021-10-06
NAS-112629 / 21.10 / TrueNAS SCALE Enterprise login window text issue

AlexKarpov98:


src/app/enums/product-type.enum.ts
```diff
@@ -4,3 +4,10 @@ export enum ProductType {
   Core = 'CORE',
   Enterprise = 'ENTERPRISE',
 }
+
+export const ProducTypetReadableText = {
```
undsoft, line 8: Typo in name.
Prefer `Map` for this with typing, see something like `nfsAclTagLabels` (translation is not necessary).

src/app/views/sessions/signin/signin.component.ts
```diff
@@ -43,6 +43,7 @@ export class SigninComponent implements OnInit, OnDestroy, AfterViewInit {
 
   failed = false;
   product_type: ProductType;
+  product_type_readable_label: string;
```
undsoft, line 46: productTypeReadableLabel

PR #5983, opened 2021-10-07
NAS-111863: Fix IPMI identify light

undsoft:
If scaley is not available use anything that has something in ipmi column, even if it's core based.
Make sure to disable light after testing is done.


RehanY147:
Otherwise seems to be working OK

PR #6000, opened 2021-10-12
Dynamic titles with @ngrx

RehanY147:
NAS-112663


src/app/store/reducers/index.ts
```diff
@@ -0,0 +1,10 @@
+import { routerReducer, RouterReducerState } from '@ngrx/router-store';
+import { ActionReducerMap } from '@ngrx/store';
+
+export interface State {
```
undsoft, line 4: Let's call it `AppState` to indicate that it's system wide.

src/app/store/effects/router-effects.ts
```diff
@@ -0,0 +1,23 @@
+import { Injectable } from '@angular/core';
```
undsoft, line 1: Prefer `router.effects.ts` for naming.

src/app/store/effects/router-effects.ts
```diff
@@ -0,0 +1,23 @@
...
+@Injectable()
+export class RouterEffects {
+  constructor(private actions$: Actions, private titleService: Title, private translateService: TranslateService) {}
+
+  navigate$ = createEffect(
+    () =>
+      this.actions$.pipe(
+        ofType(ROUTER_NAVIGATION),
+        tap((data: RouterNavigationAction<CustomRouterState>) => {
+          this.titleService.setTitle(this.translateService.instant(data.payload.routerState.title) + ' - ' + window.location.hostname);
```
undsoft, line 18: Translating is good, but then we need translation markers on all route titles :)

src/app/store/reducers/index.ts
```diff
@@ -0,0 +1,10 @@
...
+  router: RouterReducerState<any>;
```
undsoft, line 5: `<any>` should be `CustomRouterState`

src/app/store/effects/router-effects.ts
```diff
@@ -0,0 +1,23 @@
+import { Injectable } from '@angular/core';
+import { Title } from '@angular/platform-browser';
+import { Actions, createEffect, ofType } from '@ngrx/effects';
+import { RouterNavigationAction, ROUTER_NAVIGATION } from '@ngrx/router-store';
```
undsoft, line 4: I think the `ALL_CAPS` syntax is now deprecated. Try `routerNavigationAction` - you may need to cast payload though.

PR #6001, opened 2021-10-12
NAS-112786 / 22.02-RC.1 / NAS-112786: Remove alias field in paths in NFS Sharing form

undsoft:
To test: check that NFS share form still works correctly.


reitermarkus:
Given this broke my NFS shares, any more information on the reason for removing this? The JIRA ticket only mentions “Because of upstream problems” without going into detail. Is this going to be re-introduced once those upstream problems are fixed?

undsoft:
Yes, it may get re-introduced in the future.

PR #6003, opened 2021-10-12
Remove "Octets " from network interface graphs

themylogin:


undsoft (changes requested):
I don't know what changed because octets label is still in the table
Снимок экрана 2021-10-13 в 11 10 20
.

themylogin:
@undsoft in that case can you fix this properly please?

PR #6025, opened 2021-10-17
NAS-112769 / 22.02-RC.2 / Tests for LocalizationFormComponent

RehanY147:
NAS-112769


src/app/pages/system/general-settings/localization-form/localization-form.component.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+  const createComponent = createComponentFactory({
+    component: LocalizationFormComponent,
+    imports: [
+      IxFormsModule,
+      ReactiveFormsModule,
+    ],
+    providers: [
+      mockWebsocket([
+        mockCall('system.general.update'),
+        mockCall('system.general.language_choices', {
```
undsoft, line 28: You are not calling `system.general.language_choices` and others directly in `LocalizationFormComponent`. I think it would be better to mock `SystemGeneralService` directly since that's what being used.
There is no reason for tests for `LocalizationFormComponent` to be aware what's going on in other places, such as `SystemGeneralService`.

PR #6035, opened 2021-10-19
Update zh-hans.json

dpzhang314:
maybe use "状态" , instead of "州" to show the status of the virtual machines. Thanks.


undsoft (approved):
Thanks for the contribution.

PR #6039, opened 2021-10-20
NAS-111409 / 12.0 / Only the first seven custom groups appear in ACL manager

AlexKarpov98:


dszidi (approved):
Works as per ticket comments.

PR #6046, opened 2021-10-22
NAS-112556 / 22.02-RC.2 / Added `parseAndFormatInput` method to IxInput

RehanY147:
To test, check out the new added `Memory Size` form control in GUI Form (it will be removed before merging). Check both value formatting for inputs like `2 mb` and check submitted value as well. Formatted text seen on the form should be `2 MiB` and submitted value should be `2097152`.
NAS-112556


RehanY147:
Creating a new branch to fix the tests failure issue

PR #6079, opened 2021-11-01
NAS-113000 / 12.0 / Fixed idmap form

RehanY147:
To test, on TrueNAS Core, go to `Directory Services` -> `Active Directory` -> `Advanced Options` -> `Edit Idmap` -> `Add`
Next, on the field named `Idmap Backend` select the option `RCF2307`. Check that the relevant fields have correct types. Also check form submission.


dszidi (approved):
Code looks good. Altered fields show up as expected

PR #6081, opened 2021-11-02
NAS-112556 / 22.02-RC.2 / Setting up formatter and parser

RehanY147:
NAS-112556

### DIFF
--- a/src/app/pages/common/ix-forms/components/ix-input/ix-input.component.ts
+++ b/src/app/pages/common/ix-forms/components/ix-input/ix-input.component.ts
@@ -91,9 +91,9 @@ export class IxInputComponent implements ControlValueAccessor {
 
   blur(): void {
     this.onTouch();
-    if (this.value) {
+    if (this.formatted) {
       if (this.parse) {
-        this.value = this.parse(this.value);
+        this.value = this.parse(this.formatted);
       }
       if (this.format) {
         this.formatted = this.format(this.value);
--- a/src/app/pages/common/ix-forms/components/ix-input/ix-input.component.ts
+++ b/src/app/pages/common/ix-forms/components/ix-input/ix-input.component.ts
@@ -29,7 +29,6 @@ export class IxInputComponent implements ControlValueAccessor {
 
   value: string | number = '';
   formatted: string | number = '';
-  shouldUpdateValueAsap = false;
   isDisabled = false;
 
   onChange: (value: string | number) => void = (): void => {};
@@ -43,31 +42,19 @@ export class IxInputComponent implements ControlValueAccessor {
   }
 
   writeValue(value: string | number): void {
-    let parsed = value;
     let formatted = value;
     if (value) {
-      if (this.parse) {
-        parsed = this.parse(value);
-        /** This flag exists because when default value is used, onChange is not registered yet
-         * So parsed value isn't immediately updated. Furthermore, if the user submits the form with
-         * default value, the submitted value isn't parsed in that case either.
-         */
-        this.shouldUpdateValueAsap = true;
-      }
       if (this.format) {
         formatted = this.format(value);
       }
     }
-    this.value = parsed;
     this.formatted = formatted;
+    this.value = value;
     this.cdr.markForCheck();
   }
 
   registerOnChange(onChange: (value: string | number) => void): void {
     this.onChange = onChange;
-    if (this.shouldUpdateValueAsap) {
-      this.onChange(this.value);
-    }
   }
 
   registerOnTouched(onTouched: () => void): void {
--- a/src/app/pages/common/ix-forms/components/ix-input/ix-input.component.ts
+++ b/src/app/pages/common/ix-forms/components/ix-input/ix-input.component.ts
@@ -1,7 +1,7 @@
 import {
   ChangeDetectionStrategy, ChangeDetectorRef, Component, Input,
 } from '@angular/core';
-import { ControlValueAccessor, NgControl } from '@angular/forms';
+import { ControlValueAccessor, FormControl, NgControl } from '@angular/forms';
 import { UntilDestroy } from '@ngneat/until-destroy';
 
 @UntilDestroy()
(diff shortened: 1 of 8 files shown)


src/app/pages/common/ix-forms/components/ix-input/ix-input.component.ts
```diff
@@ -33,13 +42,32 @@ export class IxInputComponent implements ControlValueAccessor {
...
+    let parsed = value;
+    let formatted = value;
+    if (value) {
+      if (this.parse) {
+        parsed = this.parse(value);
+        /** This flag exists because when default value is used, onChange is not registered yet
+         * So parsed value isn't immediately updated. Furthermore, if the user submits the form with
+         * default value, the submitted value isn't parsed in that case either.
+         */
+        this.shouldUpdateValueAsap = true;
```
undsoft, line 55: I still don't think we need this. 
It's easier to just require everyone to use proper numeric values in default values or use a service, i.e.:
```
memory_size: [this.ixFormatter.memorySizeParsing('2 mib')],
```
And I would imagine that data is coming in numbers from middleware anyway.

src/app/pages/common/ix-forms/services/ix-formatter.service.ts
```diff
@@ -0,0 +1,34 @@
+import { Injectable } from '@angular/core';
+import { StorageService } from 'app/services';
+
+@Injectable()
+export class IxFormatterService {
+  constructor(
+    private storageService: StorageService,
+  ) {}
+
+  memorySizeFormatting: (val: string | number) => string = (value: string | number) => {
```
undsoft, line 10: Not sure if `memory` is the best word for it as this method will be used for disk space and file sizes.
@dszidi Requesting help from a person with better English.

undsoft (changes requested):
Looks better. 
One thing to also consider is that in absence of units (i.e. user writing `25`) we can default to MiBs not bytes.

src/app/pages/common/ix-forms/services/ix-formatter.service.ts
RehanY147: Well, both in case of disk space and file sizes, it is still memory so I thought it was appropriate. Still, suggestions welcome.

dszidi: I agree with @undsoft that calling it "memory" could be misleading. Since you're converting bytes and bytes are actually a way to measure size of data, perhaps dataSizeFormatting or byteSizeFormatting might be more appropriate.

dszidi (commented):
Tooltip says that there is a minimum of 256MiB yet the form lets me save with memory_size set to 4B

image

dszidi (commented):
No matter what value I save for memory_size it always comes back as 4 B when I go back to the form.

RehanY147:
> 
> 
> No matter what value I save for memory_size it always comes back as 4 B when I go back to the form.

The `memory_size` value isn't part of this form specifically. It's just a test value. It will be removed before the PR is merged.

RehanY147:
> 
> 
> Tooltip says that there is a minimum of 256MiB yet the form lets me save with memory_size set to 4B
> 
> image

The validation is also not part of the scope of this PR. This field is for testing purposes only and will be removed before the PR is merged.

src/app/pages/common/ix-forms/services/ix-formatter.service.ts
```diff
@@ -0,0 +1,183 @@
+export class IxFormatter {
+  static humanReadable: string;
+  static readonly IECUnits = ['B', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB'];
```
undsoft, line 3: IecUnits

src/app/pages/system/general-settings/gui-form/gui-form.component.ts
```diff
@@ -86,6 +88,9 @@ export class GuiFormComponent {
     });
   }
 
+  format = IxFormatter.memorySizeFormatting;
+  parse = IxFormatter.memorySizeParsing;
```
undsoft, line 92: I think it's better not to do static functions. They definitely look neater, but they would be much harder to deal with in tests if we ever need to mock this service.

src/app/pages/common/ix-forms/services/ix-formatter.service.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  static readonly memorySizeFormatting: (val: string | number) => string = (value: string | number) => {
+    value = value.toString();
+    if (!value) {
+      return '';
+    }
+    let formatted = '';
+    const memoryInNumbers = IxFormatter.convertHumanStringToNum(value);
```
undsoft, line 11: Why do we need to convert humanStringToNum here when we don't expect value to be a "human" string?

src/app/pages/common/ix-forms/services/ix-formatter.service.ts
```diff
@@ -0,0 +1,183 @@
...
+  // '12k4'      ''               NaN
+  // '12.4k'     ''               NaN
+  // ' 10G'      '10 GiB'         10*1024**3 (10,737,418,240)
+
+  // hstr = the human string from the form;
+  // dec = allow decimals;
+  // allowedUnits (optional) should include any or all of 'kmgtp', the first letters of KiB, Mib, etc. The first letter
+  // is used as the default, so for 'gtp', an entered value of 256 becomes 256 GiB. If you don't pass in allowedUnits,
+  // all of the above are accepted AND no unit is attached to an unlabeled number, so 256 is considered 256 bytes.
+  static readonly convertHumanStringToNum = (hstr: string, dec = false, allowedUnits?: string): number => {
```
undsoft, line 90: I don't know if these functions that you copied are meant to be ready for review or not.

If they are ready for review, I'd make the following changes:
1. Make them class methods as opposed to static.
2. Make them pure and not store anything in `humanReadable` property.
2. Return `null` if parsing failed.
3. Add a couple of unit tests for them.
4. Checked typing for `allowedUnits`.
5. Reformatted JSDoc

undsoft (changes requested):
Functionality wise works well.

PR #6090, opened 2021-11-04
NAS-113115 / 22.02-RC.2 / Layout improvements for enclosure UI

dszidi:
To test:

- Go to the System-Settings > Enclosure on a supported system. 
- Make sure that populated disk trays are selectable. 

NOTE: min-width is 640px for MINIs and 700 for rackmount systems so there will be a point where horizontal scrolling kicks in.


undsoft (approved):
Tested on several different systems. Looks better.
The only minor issue is that on Chrome on Ubuntu, I sometimes get an unnecessary vertical scrollbar.
Screenshot from 2021-11-05 12-25-47
:

PR #6099, opened 2021-11-08
NAS-113065: New Tables Draft

denysbutenko:


undsoft (commented):
So, as we've discussed, maybe let's try small, for example just by implementing a table using mat table and then going from there.
It may be then possible to extract api querying logic into Cdk DataSource and connect it to paginator and other components.

denysbutenko:
The newest PR is https://github.com/truenas/webui/pull/6126

PR #6100, opened 2021-11-09
NAS-111694 / 22.02-RC.2 / NAS-111694: Route added to unlock datasets

RehanY147:
Now, if the imported pool has encrypted datasets, the user will be prompted to go to a different page and unlock them if they want after import is successful.


src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
```diff
@@ -348,6 +350,32 @@ export class VolumeImportWizardComponent implements WizardConfiguration {
       dialogRef.componentInstance.submit();
       dialogRef.componentInstance.success.pipe(untilDestroyed(this)).subscribe(() => {
         dialogRef.close(false);
+
+        if (this.pool) {
+          this.modalService.closeSlideIn();
+          this.modalService.refreshTable();
```
undsoft, line 356: Duplicates lines below.

src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
```diff
@@ -348,6 +350,32 @@ export class VolumeImportWizardComponent implements WizardConfiguration {
...
+          this.ws.call('pool.dataset.query', [[['pool', '=', this.pool]]]).pipe(untilDestroyed(this)).subscribe((datasets) => {
+            for (const dataset of datasets) {
```
undsoft, line 358: So if we only care to see if there are any encrypted datasets, this needs to be rewritten to something like:
```
const hasLockedDataset = datasets.some((dataset) => dataset.encrypted && dataset.locked);
```

src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
```diff
@@ -348,6 +350,32 @@ export class VolumeImportWizardComponent implements WizardConfiguration {
...
+              if (dataset.encrypted && dataset.locked) {
+                this.dialogService.confirm({
+                  title: helptext.unlock_dataset_dialog_title,
+                  message: helptext.unlock_dataset_dialog_message,
+                  hideCheckBox: true,
+                  buttonMsg: helptext.unlock_dataset_dialog_button,
+                }).pipe(untilDestroyed(this)).subscribe((unlock) => {
+                  if (unlock) {
+                    const route_unlock = ['storage'].concat(['id', this.pool, 'dataset', 'unlock', this.pool]);
```
undsoft, line 367: route_unlock => routeUnlock

src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
```diff
@@ -348,6 +350,32 @@ export class VolumeImportWizardComponent implements WizardConfiguration {
...
+                    this.router.navigate(new Array('/').concat(route_unlock));
```
undsoft, line 368: new Array('/') is unnecessary

src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
```diff
@@ -348,6 +350,32 @@ export class VolumeImportWizardComponent implements WizardConfiguration {
...
+          this.ws.call('pool.dataset.query', [[['pool', '=', this.pool]]]).pipe(untilDestroyed(this)).subscribe((datasets) => {
+            for (const dataset of datasets) {
+              if (dataset.encrypted && dataset.locked) {
+                this.dialogService.confirm({
+                  title: helptext.unlock_dataset_dialog_title,
+                  message: helptext.unlock_dataset_dialog_message,
+                  hideCheckBox: true,
+                  buttonMsg: helptext.unlock_dataset_dialog_button,
+                }).pipe(untilDestroyed(this)).subscribe((unlock) => {
+                  if (unlock) {
```
undsoft, line 366: Would be good to rewrite this new code in a way that would prefer returning early instead of going deeper when possible.

src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
```diff
@@ -348,6 +350,32 @@ export class VolumeImportWizardComponent implements WizardConfiguration {
       dialogRef.componentInstance.submit();
```
undsoft, line 350: There is another if above:
```
if (value.encrypted) {
```
Do you know what it's for?

undsoft (changes requested):

RehanY147:
These changes were accidentally removed in one of the older PRs and I've brought them back. Will look at the suggested code refactoring.

src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
RehanY147: I think it's to check if the pool being imported is encrypted itself and needs a key or passphrase along with form submission.

RehanY147: It's for when there are encrypted disks involved. According to @sonicaj `12.0-stable` is a good example to follow and it does the same thing.

src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
```diff
@@ -348,9 +351,25 @@ export class VolumeImportWizardComponent implements WizardConfiguration {
...
+          this.ws.call('pool.dataset.query', [[['pool', '=', this.pool]]]).pipe(untilDestroyed(this)).subscribe((datasets) => {
+            const hasLockedDataset = datasets.some((dataset) => dataset.encrypted && dataset.locked);
+            if (hasLockedDataset) {
+              this.dialogService.confirm({
+                title: helptext.unlock_dataset_dialog_title,
+                message: helptext.unlock_dataset_dialog_message,
+                hideCheckBox: true,
+                buttonMsg: helptext.unlock_dataset_dialog_button,
+              }).pipe(filter(Boolean), untilDestroyed(this)).subscribe(() => {
+                this.router.navigate(['storage', 'id', this.pool, 'dataset', 'unlock', this.pool]);
```
undsoft, line 367: `['/storage']` no?

src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
undsoft: One of the two possibilities seem to exist. Either this is an old code and is unused (as I couldn't seem to hit that code even when using encrypted pools) or the code is used and it may need to be rewritten to add support for unlocking locked datasets.

RehanY147: Yes. Upon further investigation, it looks like Scale is missing this step from Core that works on decrypting disks
image

RehanY147: So, it can be removed. Just need @sonicaj to confirm.

RehanY147: This page has legacy code all over that changes the form based on whether we're working with Scale or Core. So, upon confirmation, all of that can be removed since this is a Scale branch now and checking for those is not needed.

sonicaj: @RehanY147 decrypting disks code can be removed from SCALE as that is specific to CORE. We would only in this case want to decrypt datasets after importing a pool.

undsoft (approved):
Looks good.

Nits:
- Let's show a loading indicator while we wait for `pool.dataset.query`.
- Let's try to restructure code in `success` clause to use return early.

PR #6105, opened 2021-11-10
NAS-112556 / 22.02-RC.2 / NAS-112556: Formatter and Parser

RehanY147:
Formatter and Parse

### DIFF
--- a/src/app/pages/common/ix-forms/services/ix-formatter.service.spec.ts
+++ b/src/app/pages/common/ix-forms/services/ix-formatter.service.spec.ts
@@ -12,10 +12,6 @@ describe('IxFormatterService', () => {
     spectator = createService();
   });
 
-  it('should instantiate', () => {
-    expect(spectator).toBeTruthy();
-  });
-
   describe('memorySizeFormatting', () => {
     it('should return empty string when invalid size string is passed', () => {
       const formatted = spectator.service.memorySizeFormatting('2u');
@@ -34,9 +30,9 @@ describe('IxFormatterService', () => {
   });
 
   describe('memorySizeParsing', () => {
-    it('should return NaN when invalid size string is passed', () => {
+    it('should return null when invalid size string is passed', () => {
       const parsed = spectator.service.memorySizeParsing('2u');
-      expect(parsed).toBe(NaN);
+      expect(parsed).toBe(null);
     });
 
     it('should return value converted in bytes when valid size string is passed', () => {
--- a/src/app/pages/common/ix-forms/services/ix-formatter.service.spec.ts
+++ b/src/app/pages/common/ix-forms/services/ix-formatter.service.spec.ts
@@ -18,8 +18,8 @@ describe('IxFormatterService', () => {
       expect(formatted).toBe('');
     });
 
-    it('should return normalized string when valid size string is passed', () => {
-      const formatted = spectator.service.memorySizeFormatting('2gb');
+    it('should return formatted size string when valid size in bytes is passed', () => {
+      const formatted = spectator.service.memorySizeFormatting('2147483648');
       expect(formatted).toBe('2 GiB');
     });
 
--- a/src/app/pages/common/ix-forms/services/ix-formatter.service.spec.ts
+++ b/src/app/pages/common/ix-forms/services/ix-formatter.service.spec.ts
@@ -1,0 +1,53 @@
+import { createServiceFactory, SpectatorService } from '@ngneat/spectator';
+import { IxFormatterService } from 'app/pages/common/ix-forms/services/ix-formatter.service';
+
+describe('IxFormatterService', () => {
+  let spectator: SpectatorService<IxFormatterService>;
+
+  const createService = createServiceFactory({
+    service: IxFormatterService,
+  });
+
+  beforeEach(() => {
+    spectator = createService();
+  });
+
+  it('should instantiate', () => {
+    expect(spectator).toBeTruthy();
+  });
+
+  describe('memorySizeFormatting', () => {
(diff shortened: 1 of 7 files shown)


src/app/pages/common/ix-forms/services/ix-formatter.service.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    service: IxFormatterService,
+  });
+
+  beforeEach(() => {
+    spectator = createService();
+  });
+
+  it('should instantiate', () => {
+    expect(spectator).toBeTruthy();
+  });
```
undsoft, line 17: This test case is not really needed.

src/app/pages/common/ix-forms/services/ix-formatter.service.ts
```diff
@@ -0,0 +1,227 @@
...
+  memorySizeFormatting: (val: string | number) => string = (value: string | number) => {
+    value = value.toString();
+    if (!value) {
+      return '';
+    }
+    let formatted = '';
+    if (value.replace(/\s/g, '').match(/[^0-9]/g) === null) {
+      formatted = this.convertBytestoHumanReadable(value.replace(/\s/g, ''), 0);
+    } else {
+      formatted = this.reformatHumanString(value);
```
undsoft, line 21: Why do this tho, if we always assume incoming value is a correct numeric representation?

src/app/pages/common/ix-forms/services/ix-formatter.service.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  describe('memorySizeFormatting', () => {
+    it('should return empty string when invalid size string is passed', () => {
+      const formatted = spectator.service.memorySizeFormatting('2u');
+      expect(formatted).toBe('');
+    });
+
+    it('should return normalized string when valid size string is passed', () => {
```
undsoft, line 25: It shouldn't. It's a job for parser.

src/app/pages/common/ix-forms/services/ix-formatter.service.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    it('should return the value with B when no unit is provided', () => {
+      const formatted = spectator.service.memorySizeFormatting('2');
+      expect(formatted).toBe('2 B');
+    });
+  });
+
+  describe('memorySizeParsing', () => {
+    it('should return NaN when invalid size string is passed', () => {
+      const parsed = spectator.service.memorySizeParsing('2u');
+      expect(parsed).toBe(NaN);
```
undsoft, line 39: I don't think we should use NaN. No one will remember to check for NaN and `typeof NaN` is still `number`. Let's use `null` or `undefined`.

src/app/pages/common/ix-forms/services/ix-formatter.service.ts
```diff
@@ -0,0 +1,227 @@
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class IxFormatterService {
+  readonly IecUnits = ['B', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB'];
```
undsoft, line 5: iecUnits

undsoft (changes requested):

src/app/pages/common/ix-forms/services/ix-formatter.service.spec.ts
RehanY147: By 'normalized' I mean with proper spacing and capitalization. It is only formatting `2gb` to `2 GiB`.

src/app/pages/common/ix-forms/services/ix-formatter.service.ts
RehanY147: The first check is if the passed in string is a huge number of raw bytes and has no unit alongside it. For example, `2147483648` which should formatted to `2 GiB`. The second check is for when the value passed has deformed units or wrong units. For example, `2048 MiB` should be `2 GiB` or `2mb` should be `2 MiB`

src/app/pages/common/ix-forms/services/ix-formatter.service.spec.ts
undsoft: Why would we expect '2gb' in this function? Haven't we agreed that this function will only deal with proper numeric values and it would be parse function that would deal with loose strings?

src/app/pages/common/ix-forms/services/ix-formatter.service.ts
undsoft: Same as above. This function should not receive `2048 MiB` or `2Mb` or anything else than `2147483648`

undsoft (changes requested):

src/app/pages/common/ix-forms/services/ix-formatter.service.ts
```diff
@@ -0,0 +1,189 @@
...
+export class IxFormatterService {
+  readonly iecUnits = ['B', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB'];
+
+  /**
+   * Formats any memory size in bytes to human readable string, e.g., '2147483648' to '2 GiB'
+   * @param value The string to be formatted
+   * @returns Formatted string
+   */
+  memorySizeFormatting: (val: string | number) => string = (value: string | number) => {
+    value = value.toString();
```
undsoft, line 13: Cannot read `toString` of null when user types in 'pizza' in the input.

undsoft (approved):
Missing accomplished :)

PR #6110, opened 2021-11-10
ICON for 'pool status' stays a 'green check-mark' even when unhealthy

AlexKarpov98:


dszidi (commented):
This looks fine in terms of not having any errors. However I'm wondering if we should just remove all the logic that determines health, since poolState provides a "healthy" property. 

This would allow us to do something like keep the status text as is but have poolState.healthy determine whether there is a green checkmark or a red circle_close icon.

@undsoft  What do you think of this?

undsoft:
@dszidi What kind of logic to determine health do we have on frontend now?

dszidi:
> @dszidi What kind of logic to determine health do we have on frontend now?

Currently the widget component's getStatusItemInfo() handles a case where the pool is ONLINE but poolState.healthy is false in which case we show a warning icon. So effectively the health is baked into the status field via the icon. Was wondering if we shouldn't simply show the health as another field and keep status and health separate.

undsoft:
In this case I'm okay with having a combined indicator as it was implemented.

dszidi (approved):
LGTM

PR #6111, opened 2021-11-10
NAS-113125 / 22.02-RC.2 / Should we display WAITING jobs in Task manager

AlexKarpov98:


denysbutenko (changes requested):
Waiting for a green checkmark on tests

denysbutenko (approved):
I wondering if we need to replace the `Pending` state with `Waiting` in other places. Seems like there is no `Pending` state on the middleware side.
Otherwise looks good 👍🏻

AlexKarpov98:
> I wondering if we need to replace the `Pending` state with `Waiting` in other places. Seems like there is no `Pending` state on the middleware side. Otherwise looks good 👍🏻

So this change was only happend on jobs and we don't have any other places using this.

PR #6112, opened 2021-11-10
NAS-112571 / 22.02-RC.2 / NAS-112571: Added tooltips

RehanY147:


src/app/pages/common/entity/entity-tree-table/entity-tree-table.component.html
```diff
@@ -27,12 +27,25 @@
...
+                  role="img" 
+                  fontSet="mdi-set" 
+                  [matTooltip]="'Encrypted dataset is locked' | translate"
+                  fontIcon="mdi-lock" 
+                  *ngIf="element.data.is_encrypted_root && element.data.locked"
+                ></mat-icon>
+                <mat-icon 
+                  svgIcon="anti-lock" 
+                  *ngIf="element.data.non_encrypted_on_encrypted || (!element.data.encrypted && element.data.has_encrypted_children)"
+                  [matTooltip]="'Pool has encrypted children' | translate"
```
sonicaj, line 47: Pool has datasets. Better terminology might be that root dataset has encrypted children. We should also reword this to show that the root dataset is not encrypted itself but has children which are encrypted.

src/app/pages/common/entity/entity-tree-table/entity-tree-table.component.html
```diff
@@ -27,12 +27,25 @@
...
+                  role="img"
+                  fontSet="mdi-set"
+                  fontIcon="mdi-lock-open-variant" 
+                  *ngIf="element.data.is_encrypted_root && !element.data.locked"
+                  [matTooltip]="'Encrypted dataset is unlocked' | translate"
+                ></mat-icon>
+                <mat-icon 
+                  role="img" 
+                  fontSet="mdi-set" 
+                  [matTooltip]="'Encrypted dataset is locked' | translate"
```
sonicaj, line 40: I think we should make it clear that these datasets are actually encryption roots as it gives a false impression that other datasets which might be children to this are not encrypted whereas they are also encrypted.

To summarise, we should reword this to show that the dataset in question is an encryption root and is locked/unlocked.

To clarify, now if children are created under this dataset, they would be encrypted as well as they would inherit encryption properties but we won't be showing an icon next to them unless they are an encryption root themselves.

sonicaj (changes requested):

sonicaj (approved):

PR #6117, opened 2021-11-12
NAS-113289 / 22.02-RC.2 / webui to preserve order for bond members

AlexKarpov98:


src/app/pages/network/forms/interfaces-form.component.ts
```diff
@@ -563,6 +564,15 @@ export class InterfacesFormComponent extends ViewControllerComponent implements
         this.entityForm.setDisabled('xmit_hash_policy', true, true);
       }
     });
+
+    this.entityForm.formGroup.get('lag_ports').valueChanges.pipe(untilDestroyed(this)).subscribe((value: string[]) => {
+      this.lag_ports.options.sort((a: Option, b: Option) => {
```
RehanY147, line 569: We shouldn't re-order options upon every selection/un-select. It can be confusing to the user.

src/app/pages/network/forms/interfaces-form.component.ts
```diff
@@ -631,10 +641,17 @@ export class InterfacesFormComponent extends ViewControllerComponent implements
     const id = data.id;
     this.setType(type);
     if (type === NetworkInterfaceType.LinkAggregation) {
+      const lag_ports = (data as LinkAggregationNetworkInterface).lag_ports;
       this.networkService.getLaggPortsChoices(id).pipe(untilDestroyed(this)).subscribe((choices) => {
         for (const key in choices) {
           this.lag_ports.options.push({ label: choices[key], value: key });
         }
+        this.lag_ports.options.sort((a: Option, b: Option) => {
```
RehanY147, line 649: Same as above

RehanY147 (changes requested):
It would be better to try to use the `form-chip` form control instead of `select` as that'll make it easier to have the values in order.
Should also have some indication in hint/tooltip that the order of selection is important here.

RehanY147 (changes requested):
A few notes
1. When the drop down is showing, it stays detached from the input and the form can be scrolled around while the dropdown stays in place.
2. There's an issue with the back ground of the last selected option
3. We should remove an option that's been selected from the dropdown while it's part of the selection. Because clicking on it again does nothing. When it's removed from the current selection, it should show up in the drop down again.
4. Once you select a value, the drop down goes away. I have to click out of the input and then click back inside to get it to show up again.
See video for examples
chip isssuse

src/app/pages/common/entity/entity-form/components/form-chip/form-chip.component.ts
```diff
@@ -67,27 +69,47 @@ export class FormChipComponent implements Field, OnInit {
...
         }
       });
       this.config.searchOptions = searchOptions;
     }
   }
+
+  openPanel(): void {
+    if (this.config.selectOnly) {
+      setTimeout(() => {
+        this.trigger._onChange('');
```
RehanY147, line 110: This `_onChange` doesn't seem to have any effect. Furthermore, it's a private method. We shouldn't be using it. Maybe it can be replaced with `this.trigger.writeValue('')`. I'm not sure. But as is, it doesn't seem to be doing anything. Please elaborate on its purpose here.

src/app/pages/common/entity/entity-form/components/form-chip/form-chip.component.ts
```diff
@@ -40,14 +41,15 @@ export class FormChipComponent implements Field, OnInit {
       if (this.chipLists !== this.group.controls[this.config.name].value && typeof this.group.controls[this.config.name].value === 'object') {
         this.chipLists = this.group.controls[this.config.name].value;
       }
+      this.updateSearchOptions('');
     });
   }
 
   add(event: MatChipInputEvent): void {
     const input = event.input;
```
RehanY147, line 49: This seems depricated and should be replaced with `const input = event.chipInput.inputElement;`

src/app/pages/network/forms/interfaces-form.component.ts
```diff
@@ -561,9 +571,22 @@ export class InterfacesFormComponent extends ViewControllerComponent implements
...
+        this.lagPortsField.tooltip = helptext.lagg_interfaces_tooltip + ' ' + helptext.lagg_interfaces_failover_tooltip;
+      } else {
+        this.lagPortsField.tooltip = helptext.lagg_interfaces_tooltip;
+      }
     });
   }
 
+  updateSearchOptions(value = '', values?: string[]): void {
+    this.lagPortsField.searchOptions = this.lagPortsOption.filter((option) => (
+      (!value
```
RehanY147, line 584: Break down these checks as separate variables so it's more readable what's being filtered?

src/app/pages/network/forms/interfaces-form.component.ts
```diff
@@ -561,9 +571,22 @@ export class InterfacesFormComponent extends ViewControllerComponent implements
         this.entityForm.setDisabled('lacpdu_rate', true, true);
         this.entityForm.setDisabled('xmit_hash_policy', true, true);
       }
+      if (value === LinkAggregationProtocol.Failover) {
+        this.lagPortsField.tooltip = helptext.lagg_interfaces_tooltip + ' ' + helptext.lagg_interfaces_failover_tooltip;
```
RehanY147, line 575: The tooltip should be one string which properly describes the special case for `Failover` link aggregation protocol at the end. So, I think it's better to make an addition to `helptext.lagg_interfaces_tooltip` instead of having another string that's added conditionally. This looks like someone moving code around could easily forget to add the extra bit and we'd lose important information that's very relevant to this input.

RehanY147 (changes requested):
Works well over all. Minor picks about the code that need to be addressed and we should be good.

src/app/pages/common/entity/entity-form/components/form-chip/form-chip.component.ts
AlexKarpov98: Right. Shouldn't change it manually.

RehanY147 (approved):
Looks good. Thanks!

PR #6118, opened 2021-11-12
NAS-113322 / 22.02-RC.2 / NAS-113322: Close modals on Escape

denysbutenko:
Add support to close slide-in panels using the Esc key.

Please let me know if there are other places that need to be supported


undsoft (approved):
Works well, two minor things:
1. ['$event'] is unused and can be removed.
2. If I first open a new ix-modal and hit escape, I'll get `this.modal is undefined` in console. The slide-in still closes.

PR #6122, opened 2021-11-14
NAS-113319 / None / NAS-113319: Make template checks even more strict

undsoft:
For testing maybe check some of the changes that look riskier.


denysbutenko (approved):
No regressions are found. Looks good 👍🏻

denysbutenko:
Noticed that drag-n-drop stopped working in `dual-listbox` component. It can be verified on `Crendentials -> Local Groups -> Pick any group -> Members`

PR #6123, opened 2021-11-15
NAS-112767 / 22.02-RC.2 / NAS-112767: Custom validators

RehanY147:
The new IxForm validators


undsoft (changes requested):
1.I think it would be good to have default translated error messages, so that developers don't have to specify them in every case.
2. It's possible for developer to forget and use an old Validators class from Angular. We need to make sure we are still showing errors in this case.

src/app/pages/common/ix-forms/services/ix-validators.service.ts
```diff
@@ -0,0 +1,204 @@
...
+   */
+  compose(validators: ValidatorFn[]): ValidatorFn | null {
+    return Validators.compose(validators);
+  }
+
+  /**
+   * For inclusion or Validators.composeAsync
+   * @returns the result of Validators.composeAsync called on the passed validator functions array
+   */
+  composeAsync(validators: AsyncValidatorFn[]): AsyncValidatorFn | null {
```
undsoft, line 173: `composeAsync` and `compose` are not really needed, could be used directly from Angular's `Validators`.

src/app/pages/common/ix-forms/services/ix-validators.service.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * This method applies the Validators.pattern validator to the input
+   * but applies errMessage as the custom error message.
+   * @param pattern The pattern to check the control value against
+   * @param errMessage The error message if this validator fails
+   * @returns A validator function that applies Validators.pattern and if failed,
+   * applies errMessage as the custom error message.
+   */
+  pattern(pattern: string | RegExp, errMessage?: string): ValidatorFn {
```
undsoft, line 143: I liked your approach of having default messages in the component. 
Could we have a more generic wrapper instead of wrapping each individual original validator?
Something like:
```
cidr_port: ['', this.ixValidators.withMessage(
  Validators.required, 
  this.translate('Yes, it is really required'),
)],
```

undsoft (changes requested):

undsoft (approved):
Nice, thanks!

RehanY147:
The original ticket is https://jira.ixsystems.com/browse/NAS-112901

PR #6125, opened 2021-11-15
NAS-112744 / 12.0 / NAS-112744: Bringing api.service closer to its Scale conterpart

undsoft:
This likely won't change much if anything at all, but just to be safer I've ported some changes that we made on `master` branch.

For testing you could try to reproduce the issue in the ticket, or do quick regression testing.


dszidi (approved):
Looks fine. The welcome dialog isn't showing up after logging back in. I tested the enclosure page, the UI preferences form and a few random pages that use tables and all seems to be functioning properly. LGTM

PR #6126, opened 2021-11-15
NAS-113294 / 22.02-RC.2 / Build an example of basic ix-table

denysbutenko:


src/app/pages/common/ix-tables/components/ix-table-column/ix-table-column.component.ts
```diff
@@ -0,0 +1,22 @@
...
+  styleUrls: ['./ix-table-column.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class IxTableColumnComponent<T> extends MatTextColumn<T> {
+  /**
+   * Define custom width of the column
+   */
+  @Input() width: string;
+  @ContentChild(TemplateRef) templateRef: TemplateRef<T>;
+}
```
undsoft, line 22: Does we even need this component?

src/app/pages/services/services.component.ts
```diff
@@ -1,132 +1,89 @@
...
-
-        setTimeout(() => {
-          service.state = incomingService.state;
-          service.enable = incomingService.enable;
-          service.onChanging = false;
-        }, 300);
-      });
+  getData(): void {
+    this.services.getAll().pipe(
+      take(1),
```
undsoft, line 62: Does `take(1)` actually do something here?

src/app/pages/services/services.component.ts
```diff
@@ -204,48 +170,81 @@ export class ServicesComponent implements EntityTableConfig, OnInit {
...
+        default:
+          this.router.navigate(['/services', row.service]);
+          break;
+      }
     }
   }
 
-  getServiceName(service: Service): string {
-    return serviceNames.get(service.service) || service.service;
+  setupToolbar(): void {
```
undsoft, line 219: I'm assuming this will change in the future PRs.

src/app/pages/services/services.service.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  startStopAction(rpc: 'service.start' | 'service.stop', service: ServiceName): Observable<boolean> {
+    return this.ws.call(rpc, [service]);
+  }
+
+  enableDisableAction(id: number, enable: boolean): Observable<number> {
+    return this.ws.call('service.update', [id, { enable }]);
+  }
+
+  transform(service: Service): ServiceRow {
```
undsoft, line 46: This method is very component specific and does not belong to a system-wide service.

src/app/pages/services/services.module.ts
```diff
@@ -61,7 +64,7 @@ import { routing } from './services.routing';
     ServiceWebdavComponent,
     ServiceS3Component,
   ],
-  providers: [UserService],
+  providers: [UserService, ServicesService],
```
undsoft, line 67: You should not need to add service to providers when `{ providedIn: 'root' }` is used.

src/app/pages/services/services.component.ts
```diff
@@ -177,13 +140,17 @@ export class ServicesComponent implements EntityTableConfig, OnInit {
 
   updateService(rpc: 'service.start' | 'service.stop', service: ServiceRow): void {
     let waiting = true;
+    this.cdr.markForCheck();
     // Delay spinner for fast API responses
     setTimeout(() => {
       if (waiting) service.onChanging = true;
+      this.cdr.markForCheck();
```
undsoft, line 147: I think in the future we may want to consider not mutating / extending original service, but instead have a separate `Map` of service name that is being loaded.

src/app/pages/services/services.component.html
```diff
@@ -0,0 +1,63 @@
+<mat-card [style.overflow]="'hidden'">
```
undsoft, line 1: Style can be moved to a scss file.

undsoft (changes requested):
The component itself will probably require more love, but the table template itself looks good.

src/app/pages/services/services.component.ts
denysbutenko: Yeah, probably it involves some refactoring for entity-toolbar or a new component for the page title that allows extending via templates

undsoft (approved):
Looks good.

PR #6130, opened 2021-11-16
NAS-113363 / None / NAS-113363: Enforce camelCase on function params 2

undsoft:
No testing is necessary.


src/app/core/services/interaction-manager.service.ts
```diff
@@ -90,7 +88,6 @@ export class InteractionManagerService {
         // this.displayObjectWithFocus.hasFocus = false;
```
dszidi, line 88: These should also be removed

src/app/core/services/interaction-manager.service.ts
```diff
@@ -135,7 +132,6 @@ export class InteractionManagerService {
 
   // Find the related Layout object for the displayObject if one exists
   getLayoutParent(displayObject: DisplayObject): LayoutObject {
-    // let index = this.displayList.indexOf(displayObject);
```
dszidi, line 138: Also should be removed

dszidi (approved):
Other than the two comments above the rest looks good!

PR #6132, opened 2021-11-17
NAS-112798 / 22.02-RC.2 / NAS-112798: Failover button requires disabled Input()

dszidi:
To test
- Use an HA system with SCALE installed
- On dashboard press the "Initiate Failover" button on the card for the standby node.
- After it reboots (you may need to refresh the page but that's a separate issue), Make sure the "Initiate Failover" button is usable again.


undsoft (approved):
Looks good.

PR #6135, opened 2021-11-18
NAS-113377 / 22.02-RC.2 /Refactor Rsync form to new ix forms

AlexKarpov98:


src/app/pages/services/components/service-rsync/rsync-configuration/configure-rsync/configure-rsync.component.ts
```diff
@@ -1,45 +1,83 @@
...
-      value: helptext.rsyncd_port_value,
-    },
-    {
-      type: 'textarea',
-      name: 'auxiliary',
-      placeholder: helptext.rsyncd_auxiliary_placeholder,
-      tooltip: helptext.rsyncd_auxiliary_tooltip,
-    },
-  ];
+export class CconfigureRYSNCComponent implements OnInit {
```
denysbutenko, line 22: Funny name :) Seems like it needs renaming to follow the naming convention. Also, we have `RYSNCConfigurationFormComponent` that makes me confused what is the difference between these components. It would be nice if we can distinguish them based on names.

src/app/pages/services/components/service-rsync/rsync-configuration/configure-rsync/configure-rsync.component.ts
```diff
@@ -1,45 +1,83 @@
...
+  form = this.fb.group({
+    port: [Number(helptext.rsyncd_port_value), Validators.pattern('^[0-9]*$')],
```
denysbutenko, line 24: I think we can remove `helptext.rsyncd_port_value` and use `873` as a number for the default value.

denysbutenko (approved):
Looks good, thanks 👍🏻

src/app/pages/services/components/service-rsync/rsync-configure/rsync-configure.component.ts
```diff
@@ -0,0 +1,83 @@
...
+@UntilDestroy()
+@Component({
+  selector: 'app-rsync-configure',
+  templateUrl: './rsync-configure.component.html',
+  styleUrls: ['./rsync-configure.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class RsyncConfigureComponent implements OnInit {
+  form = this.fb.group({
+    port: [873, Validators.pattern('^[0-9]*$')],
```
denysbutenko, line 24: the last thing, it may be worth considering adding min and max validators to define the allowed port range. For example
https://github.com/truenas/webui/blob/master/src/app/pages/system/general-settings/gui-form/gui-form.component.ts#L39-L40

PR #6143, opened 2021-11-20
NAS-113421 / None / NAS-113421: Enforce naming convention on class members 2

undsoft:
No testing is necessary.


denysbutenko (approved):
👍🏻

PR #6145, opened 2021-11-20
NAS-110081 / 22.02-RC.2 / NAS-110081: Setting up sentry

RehanY147:


undsoft (changes requested):
As mentioned in previous PR, we need to honour `Crash Reporting` setting in GUI Settings.
Also, we discussed enabling reporting for nightly builds only for now. What's the status on that one?

src/app/pages/system/general-settings/gui-form/gui-form.component.ts
```diff
@@ -121,6 +123,23 @@ export class GuiFormComponent {
       this.cdr.markForCheck();
       this.slideInService.close();
       this.handleServiceRestart(body);
+      if (body.crash_reporting) {
+        combineLatest([
```
undsoft, line 127: Don't like code duplication, especially keys in two places. Let's try to find a single place for it.
Let's move code from here and from successLogin to a common place (AdminLayoutComponent will do for now). Gui form is calling `this.sysGeneralService.refreshSysGeneral` when saving. This may be enough to enable/disable Sentry on settings change.
Key can be put to environment.ts

undsoft (changes requested):

undsoft (approved):
Looks good, thanks.

PR #6150, opened 2021-11-22
NAS-113298 / None / NAS-113298: Refactor NTP Server table to new ix table and add unit tests

denysbutenko:


src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.html
```diff
@@ -0,0 +1,61 @@
...
+      </ng-container>
+      <ng-container matColumnDef="prefer" width="40px">
+        <th mat-header-cell *matHeaderCellDef>{{ 'Prefer' | translate }}</th>
+        <td mat-cell *matCellDef="let server; dataSource: dataSource">{{ server.prefer }}</td>
+      </ng-container>
+      <ng-container matColumnDef="minpoll" width="60px">
+        <th mat-header-cell *matHeaderCellDef>{{ 'Min Poll' | translate }}</th>
+        <td mat-cell *matCellDef="let server; dataSource: dataSource">{{ server.minpoll }}</td>
+      </ng-container>
+      <ng-container matColumnDef="maxpoll" width="60px">
```
undsoft, line 32: I think we can remove hardcoded width, seems to be doing without it just fine.

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('should show table headers', async () => {
+    const table: IxTableHarness = await loader.getHarness(IxTableHarness);
+    const headerRow = await table.getHeaderRow();
+
+    expect(Object.keys(headerRow).length).toBe(7);
```
undsoft, line 83: You should also be able to do:
```
expect(headerRow).toMatchObject({
  address: Address,
 ...
});
```

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+      {
+        actions: 'delete',
+        address: '0.debian.pool.ntp.org',
+        burst: 'false',
+        iburst: 'true',
+        prefer: 'false',
+        minpoll: '6',
+        maxpoll: '10',
+      },
+    ];
```
undsoft, line 128: I guess my question is whether we care about column names in this case, since they are not shown to the user.
How about an even simplier method:
```
const cells = await table.getCells();
expect(cells).toMatchObject([
 ['Address', 'Burst', 'IBurst', ...],
 ['2.debian.pool.ntp.org', 'false', 'true'...],
);
```

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    expect(ws.call).toHaveBeenCalledWith('system.ntpserver.query');
+    expect(cells).toEqual(expectedRows);
+  });
+
+  it('should show empty table', async () => {
```
undsoft, line 134: should show empty table when ...

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+  });
+
+  it('should open add ntp server form', async () => {
+    const addButton = await loader.getHarness(MatButtonHarness.with({ text: 'Add' }));
+    await addButton.click();
+
+    expect(spectator.fixture.componentInstance.doAdd).toHaveBeenCalledTimes(1);
+  });
+
+  it('should display confirm dialog of deleting ntp server', async () => {
```
undsoft, line 161: Since deletion of data is also something that happens in this component, we also need to make sure that proper ws call is made.

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.ts
```diff
@@ -0,0 +1,120 @@
...
+  };
+  emptyConf: EmptyConfig = {
+    type: EmptyType.NoPageData,
+    large: true,
+    title: this.translate.instant('No servers have been added yet'),
+  };
+  errorConf: EmptyConfig = {
+    type: EmptyType.Errors,
+    large: true,
+    title: this.translate.instant('Can not retrieve response'),
```
undsoft, line 46: We can probably have a more generic function to handle errors (and show exactly what the error was). I'll create a ticket for that.

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.ts
```diff
@@ -0,0 +1,120 @@
...
+    private translate: TranslateService,
+  ) { }
+
+  ngOnInit(): void {
+    this.getData();
+  }
+
+  createDataSource(servers: NtpServer[]): void {
+    this.dataSource = new MatTableDataSource(servers);
+    this.cdr.markForCheck();
```
undsoft, line 74: markForCheck should not be necessary here.

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  getData(): void {
+    this.ws.call('system.ntpserver.query').pipe(
+      untilDestroyed(this),
+    ).subscribe((servers) => {
+      this.loading = false;
```
undsoft, line 81: Don't see you setting `loading` to `true` anywhere.

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.ts
```diff
@@ -0,0 +1,120 @@
...
+  constructor(
+    private ws: WebSocketService,
+    private cdr: ChangeDetectorRef,
+    private slideInService: IxSlideInService,
+    private dialog: DialogService,
+    private translate: TranslateService,
+  ) { }
+
+  ngOnInit(): void {
+    this.getData();
```
undsoft, line 69: You also need to listen for when slide in closes, to reload table when new row is added.

undsoft (changes requested):
Looks pretty good.

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.ts
denysbutenko: Some tests are failing without it

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.ts
denysbutenko: it is true by default for first data loading and then the table refreshes data without a spinner

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.ts
denysbutenko: Yeah, good idea 👍🏻

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.ts
undsoft: Would it work for subsequent calls, like when row is removed?

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.ts
undsoft: Can you add `spectator.markForCheck()` in tests then?

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.ts
denysbutenko: yeah, `getData` will update the table without showing a spinner

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.spec.ts
denysbutenko: Added the check but it does not work properly, so it is commented out until figure out a cause. Received `system.ntpserver.query` instead of expected `system.ntpserver.delete`. Even if `doDelete` does not use `dialog.confirm` and `getData()` and doing only `ws.call('system.ntpserver.delete')`

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.ts
undsoft: And why is this a good thing? :)

src/app/core/testing/utils/run-on-push.utils.ts
```diff
@@ -0,0 +1,12 @@
+import { ChangeDetectorRef } from '@angular/core';
+import { ComponentFixture } from '@angular/core/testing';
+
+/**
+ * Changes in components using OnPush strategy are only applied once when calling .detectChanges(),
+ * This function solves this issue.
+ */
+export async function runOnPushChangeDetection<T>(fixture: ComponentFixture<T>): Promise<void> {
+  const changeDetectorRef = fixture.debugElement.injector.get<ChangeDetectorRef>(ChangeDetectorRef);
+  changeDetectorRef.detectChanges();
```
undsoft, line 10: I'm not exactly sure what's the difference between `spectator.detectChanges()` and `spectator.detectComponentChanges()` is, but if you replace `runOnPushChangeDetection` with `spectator.detectComponentChanges()` it works.

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.spec.ts
undsoft: It may have something to do with:
```
    jest.spyOn(spectator.fixture.componentInstance, 'doDelete').mockImplementation();
```
😛

undsoft: On this note, instead of testing that `doAdd` was called, let's mock `slideInService` and check that `open` was called with correct class.

src/app/pages/system/general-settings/ntp-server/ntp-server-list/ntp-server-list.component.ts
denysbutenko: Initially, I thought showing the loading spinner on each call of getData could be annoying, e.g., `slideIn.onClose$`. Updated `getData` method to set loading to true before each call. It seems `*matNoDataRow` is visible only when a data source is empty, so the loading spinner won't appear on adding/editing/deleting actions when rows exist.

undsoft (approved):
Nice work.

PR #6158, opened 2021-11-25
NAS-112060 / 22.02-RC.2 / Rollback should use dropdown instead of textfield

AlexKarpov98:


Ornias1993 (approved):
Nice work, looks cleaner than I anticipated :)

undsoft (approved):
Looks good 👍🏻

PR #6159, opened 2021-11-25
NAS-113370 / 22.02-RC.2 / Add return button on the "Unlock Datasets" popup

AlexKarpov98:


denysbutenko (approved):
image
image
`Cancel` sounds weird. I think `Close` is more suitable in this case but that's up to you.

Looks good 👍🏻

PR #6162, opened 2021-11-26
NAS-113053 / 22.02-RC.2 / Add Cron Job, but schedule is lost upon save

AlexKarpov98:


dszidi (commented):
After editing the custom cron schedule the popup disappears and the schedule is not shown in the field in the parent form.

dszidi (approved):
Looks good!

image

PR #6165, opened 2021-11-30
NAS-113474 / 22.02-RC.2 / First time update UI doesn't allow apply updates and reboot checkbox

AlexKarpov98:
- [ ] Applied checked value to dialog ref for the checkbox `apply updates and reboot after downloading`.
- [ ] Disabled the save configuration dialog once saved it.


undsoft (approved):
Seems to be working well.

PR #6173, opened 2021-12-03
NAS-110535 / 22.02 / NAS-110535: Custom subscription to console messages

RehanY147:
To test,
1. Go to `System Settings` -> `General` -> `GUI Settings` and make sure `Show Console Messages` checkbox is checked before saving.
2. Go to `Shares` -> `Block (iSCSI) Shares Targets` Configure and on the next page, `Total Global Configuration` tab selected, just tap on `Save` button. The loader should open. Here I've added a delay of 10 seconds so testing this PR is easier. (Will remove the delay before merging). The loader should be showing the `Console` button underneath the spinner. Click on that button.
3. Previously, opening the console dialog like this would cause the UI to completely crash and reload. It shouldn't happen now.
Let me know if there are any questions.

### DIFF
--- a/src/app/services/ws.service.ts
+++ b/src/app/services/ws.service.ts
@@ -78,7 +78,7 @@ export class WebSocketService {
 
   onopen(): void {
     this.onOpenSubject$.next(true);
-    this.send({ msg: 'connect', version: '1', support: ['1'] });
+    this.send({ msg: ApiEventMessage.Connect, version: '1', support: ['1'] });
   }
 
   onconnect(): void {
@@ -100,7 +100,7 @@ export class WebSocketService {
 
   ping(): void {
     if (this.connected) {
-      this.socket.send(JSON.stringify({ msg: 'ping', id: UUID.UUID() }));
+      this.socket.send(JSON.stringify({ msg: ApiEventMessage.Ping, id: UUID.UUID() }));
       setTimeout(() => this.ping(), 20000);
     }
   }
@@ -194,7 +194,7 @@ export class WebSocketService {
   call<K extends ApiMethod>(method: K, params?: ApiDirectory[K]['params']): Observable<ApiDirectory[K]['response']> {
     const uuid = UUID.UUID();
     const payload = {
-      id: uuid, msg: 'method', method, params,
+      id: uuid, msg: ApiEventMessage.Method, method, params,
     };
 
     // Create the observable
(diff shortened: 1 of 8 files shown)


denysbutenko (commented):
The console dialog closing after the delay is it by design?

denysbutenko:
https://user-images.githubusercontent.com/351613/144594884-91a700c7-d381-41aa-8d36-cc1d4b568ad0.mp4
 
no errors in console, just empty dialog

denysbutenko (approved):
Works as described 👍🏻  UI did not crash in this case anymore

RehanY147:
> 

Once saved, the console middleware api doesn't seem to send any console messages.

RehanY147:
Need to confirm with UI code and middleware why there's no response from middleware end point if the dialog is opened the second time.

src/app/services/ws.service.ts
```diff
@@ -62,16 +60,6 @@ export class WebSocketService {
     return this.authStatus$.asObservable();
   }
 
-  get consoleMessages(): Observable<string> {
```
RehanY147, line 65: This getter was causing the original issue in the ticket. It was basically subscribing to the same end point without changing the request id. Which caused an error and caused the ui to crash.

src/app/services/ws.service.ts
```diff
@@ -221,18 +209,25 @@ export class WebSocketService {
...
-    const nom = name.replace('.', '_'); // Avoid weird behavior
+  /**
+   * This method subscribes to the provided api end point for real time updates
+   * @param api The api end point to subscribe to
+   * @param subscriptionId The unique id that will be used as request id and can be
+   * used to unsubscribe from the websocket subscription with the `unsub(api, subscriptionId)`
+   * method
+   * @returns
+   */
+  sub<T = any>(api: string, subscriptionId?: string): Observable<T> {
```
RehanY147, line 220: The purpose of the `subscriptonId` here is to provide a way for the calling component/service to unsubscribe from these subscriptions. This is a necessary step because if you don't unsubscribe from the subscription and leave it forever open, you can't subscribe to it again later on anywhere else if you have. The middleware won't respond to the new subscription request.

RehanY147 (commented):
A lot of integral changes have been made so here's some explanation

src/app/services/ws.service.ts
RehanY147: Here, by "unsubscribe" I don't mean unsubscribing from an observable. I mean calling `WebSocketService.send` method with `subscriptionId` and `unsub` message like so `WebSocketService.send({ id: subscriptionId, msg: 'unsub'})`.

undsoft:
This is a weird feature, I couldn't get it to work to test. 
The modal either closes too fast before I can see anything or if I enable Slow 3G in Chrome, it sometimes stays open saying `Loading...`.

RehanY147:
> 

You should be able to test this now with the provided description

undsoft (approved):
Okay, good I can see the logs, thanks.

PR #6176, opened 2021-12-03
NAS-113604 / 22.02-RC.2 / Keep sidenav state in user preferences

denysbutenko:
Changes:

- Save the sidenav state when clicking on the hamburger icon to user preferences and restore it on each page reload
- Fix 6px pixels on left when hovering on menu items when `collapsed: false`
- Fix 6px pixels on right  when `collapsed: true`
- Fix tooltip position for some items when `collapsed: true`

Known issues:

- Sometimes, you may see slide-out animation on page reload when the sidenav collapsed

Testing:

- Make sure these changes do not introduce new bugs
- Try to switch between Mobile / Desktop and check how it is going

### DIFF
--- a/src/app/components/common/layouts/admin-layout/admin-layout.component.ts
+++ b/src/app/components/common/layouts/admin-layout/admin-layout.component.ts
@@ -170,9 +170,11 @@ export class AdminLayoutComponent implements OnInit, AfterViewChecked {
       setTimeout(() => {
         this.sideNav.open();
       });
+      this.layoutService.isMenuCollapsed = this.prefService.preferences.sidenavStatus.isCollapsed;
+      this.isSidenavCollapsed = this.prefService.preferences.sidenavStatus.isCollapsed;
+    } else {
+      this.layoutService.isMenuCollapsed = false;
     }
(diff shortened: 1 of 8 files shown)


undsoft (approved):
The code for sidebar is painful, good luck :)

There is now a regression:
1. Desktop screen size.
2. Collapse menu.
3. Resize window to xs.
Menu is now open.

I guess it's not too important as it will only affect a narrow percentage of users in a specific scenario.

denysbutenko:
> The code for sidebar is painful, good luck :)
> 
> There is now a regression:
> 
> 1. Desktop screen size.
> 2. Collapse menu.
> 3. Resize window to xs.
>    Menu is now open.
> 
> I guess it's not too important as it will only affect a narrow percentage of users in a specific scenario.

I've tested it from a user perspective:
- Open on desktop -> collapse menu -> then
- Open on mobile -> open menu (menu shouldn't be collapsed) -> then
- Switch to desktop -> reload page -> check menu state (menu still collapsed from the 1st step)

Since resizing is a specific scenario let's keep it for another iteration in this area :)

PR #6177, opened 2021-12-05
NAS-113470 / 22.02 / when saving a new license and reloading the UI does not reload

AlexKarpov98:


src/app/pages/system/general-settings/support/license/license.component.ts
```diff
@@ -49,18 +49,20 @@ export class LicenseComponent {
...
-        hideCheckBox: true,
-        buttonMsg: helptext.update_license.reload_dialog_action,
-        hideCancel: true,
-        disableClose: true,
-      }).pipe(filter(Boolean), untilDestroyed(this)).subscribe(() => {
-        document.location.reload();
-      });
       this.slideInService.close();
       this.cdr.markForCheck();
+      setTimeout(() => {
```
denysbutenko, line 54: Is there another possible approach to solve this without setTimeout?

denysbutenko: For example, wrap `document.location.reload();` into `setTimeout()` or even add pipe `delay(200)`?

AlexKarpov98: It does not work for me. While working the `slideInService` and `dialogService` together, the subscription was destroyed.
I think should run sequentially.
How do you think to close slide after confirmation?

denysbutenko (approved):
Works as described

william-gr:
backport

PR #6193, opened 2021-12-13
NAS-113724 / 22.02-RC.2 / NAS-113724: Cannot Restore Config File

undsoft:
Make sure to actually press Upload after selecting a file.


denysbutenko (approved):
Works well!

PR #6195, opened 2021-12-14
NAS-113759 / 22.02 / NAS-113759: New FTP option: timeout_notransfer

undsoft:
Find a freshly updated system and see that there is a new settings in Services -> FTP.


dszidi (approved):
Code looks good. Behaves as expected

image

PR #6198, opened 2021-12-14
NAS-113738 / 22.02-RC.2 / NAS-113738: Hack-fixes alerts

undsoft:
Testing:
- Find a system with alerts. There are alerts that get dismissed and can be reopened and alerts that get removed. You need to test both.
- Dismiss/undismiss alerts.
- Open two tabs on the same page. Dismiss/undismiss alerts. Observe that changes propagate to another tab.
- Create alerts manually via shell: `midclt call alert.oneshot_create UnscheduledReboot "Something bad happened"`. Clear the alert. For additional alerts use different error message.

### DIFF
--- a/src/app/components/common/notifications/notifications.component.html
+++ b/src/app/components/common/notifications/notifications.component.html
@@ -62,17 +62,21 @@
   </mat-list-item>
 
   <mat-list-item
-    *ngFor="let n of notifications; let last = last"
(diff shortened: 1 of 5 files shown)


AlexKarpov98 (approved):
image
LGTM!

undsoft:
backport

PR #6202, opened 2021-12-14
NAS-113760 / 22.02 / Export keys from dataset does not render json output

AlexKarpov98:


undsoft (approved):
Looks good.

Снимок экрана 2021-12-15 в 12 07 52

PR #6203, opened 2021-12-14
NAS-111989 / 22.02 / Loading the dashboard calls to our update servers

AlexKarpov98:


dszidi (commented):
Please check the ticket comments. The ticket reporter has a valid concern but I don't think the optimal solution is to stop showing update availability on the dashboard altogether.

dszidi (commented):
Almost there. If the goal is to limit the update check/network calls to just initial page load , then you also need to handle the `UpdateCheck` event on line 115 of the widget-sys-info component file. That event triggers an update.check_available API call every time the user navigates to the dashboard too.

image

dszidi (commented):
This latest change doesn't really give us the desired result as now the widget doesn't check for updates at all. 

Let's try this instead. Add a new item to sessionStorage called "updateLastChecked" and log the date when we check for an update. The widget can check that value and only make the API call if updateLastChecked is older than a day. 

This will limit the check to once a day at the most giving the users the information they want and not make unnecessary network calls..

dszidi (approved):
LGTM 

image

PR #6205, opened 2021-12-15
NAS-113739 / 22.02 / Fix enclosure selector for R Series

dszidi:
There was a typo in the code due to two different variables with similar names. Correct variable is now being referenced with a more distinct and obvious name to avoid future issues. Also some error handling was added to deal with scenarios where middleware doesn't return data with enclosure.query

To test:
- try tn30 and make sure that the visualization comes up in both the main view and the enclosure selector on the right
- try tn40. Assuming the middleware issue is still present on that machine, it should show an error message about enclosure data missing.


AlexKarpov98 (approved):
I was able to check this working on tn30 and tn40, nice work!

PR #6207, opened 2021-12-15
NAS-110348 / 13.0 / Fixed property name returned from middleware on error to remove dots

RehanY147:
To test,
- Go to the jails page and click on the `Add` button to create a new jail
- Click on the `Advanced Jail Creation` button
- Select an `IPv4 Interface` and type in an IP that's already in use by the system. (It can be the ip that your NAS system points to. E.g, 192.168.18.21).
- Add a name for the jail and select any release version
- Click on save and you should see the validation error returned by middleware in red under the relevant form field.
- Do the same for the Jail creation wizard by clicking `Next` on step 2
 
Previously the error failed to appear under the form field. The flashing dialog is intentional. This is what was meant to happen.


undsoft (approved):
Works well.

Снимок экрана 2021-12-16 в 11 09 49

PR #6210, opened 2021-12-17
NAS-113575 / 22.02 / NAS-113575 - Refactoring NFS service form

undsoft:
Testing: Settings -> Services -> Edit NFS.
If you want to get something to show up in bind IPs dropdown, you need to go to Network, open your interface and add a static IP address to it. Static IP has to be added on the interface, not not in Static Routes.


RehanY147 (changes requested):
The validation for the port fields isn't applied. Entering alphabets gives no error, forms submits successfully. Invalid values not saved (correctly) when you open the form again.

Furthermore, I feel like there should be some indication of an empty value in the drop down if no value is available. Otherwise, it looks like it's stuck and not working properly. An empty state or a null value maybe?

RehanY147:
Here's a video of the issue
nfs form refactor

undsoft:
I've partially fixed it. Chrome will now show a range error in this case. Unfortunately I cannot fix the issue in Firefox easily, it has to do with how Firefox works with numeric inputs. I've created a separate ticket to address this: https://jira.ixsystems.com/browse/NAS-113903

PR #6212, opened 2021-12-17
NAS-113867 / 22.02 / NAS-113867: Unit tests should not be making real network requests

bvasilenko:
Several tests were crashing when `environment.remote` was reachable via network
For example at localization-form.component.spec.ts

### DIFF
--- a/src/app/core/testing/classes/mock-websocket.service.ts
+++ b/src/app/core/testing/classes/mock-websocket.service.ts
@@ -30,6 +30,8 @@ export class MockWebsocketService extends WebSocketService {
     this.call = jest.fn();
     this.job = jest.fn();
     this.subscribe = jest.fn(() => of());
+    this.socket.send = jest.fn();
+    this.socket.close = jest.fn();
     when(this.call).mockImplementation((method: ApiMethod, args: unknown[]) => {
       throw Error(`Unmocked websocket call ${method} with ${JSON.stringify(args)}`);
     });


undsoft (changes requested):
1. It works, but I would prefer to take a different approach - not using a base class. I feel like it makes system's code more complicated only to accommodate tests.
I would rather either take an approach that was used before in MockWebsocketService:
```
this.call = jest.fn();
this.job = jest.fn();
```
or just redefine `send` in MockWebsocketService.

2. I'm a bit confused as to why it helped. I don't see any new error message in console. Does it mean that there is now some code in localization form that still tries to make network request, gets an Error and catches it?

bvasilenko:
@undsoft 

> 1. I would prefer to take a different approach - not using a base class. I feel like it makes system's code more complicated only to accommodate tests.

Agree. Let me do it the other way, we'll see which one is better.

> 2. I'm a bit confused as to why it helped. I don't see any new error message in console. Does it mean that there is now some code in localization form that still tries to make network request, gets an Error and catches it?

Steps which cause the issue:
   1.  `new MockWebsocketService` (called by dependency injector)
   2.  `MockWebsocketService.constructor` -> `WebSocketService.constructor` ->  `this.connect();` -> `rxjs` emits, etc (side effects) -> broken test `localization-form.component.spec.ts` 


```
constructor(protected router: Router) {
    this.authStatus$ = new Subject<boolean>();
    this.onOpenSubject$ = new Subject();
    this.onCloseSubject$ = new Subject();
    this.pendingCalls = new Map();
    this.protocol = window.location.protocol;
    this.remote = environment.remote;
    this.connect();     // <== here's the beginning of the story
  }
```

undsoft (approved):
All tests pass.

PR #6216, opened 2021-12-20
NAS-113790 / 22.02 / NAS-113790: "Mixing disks of different sizes in a vdev is not allowed."

bvasilenko:
Testing:
You should have at least 2 available disks with different sizes
Go to Storage -> Create Pool -> Add these 2 disks to the right pane


src/app/pages/storage/volumes/manager/vdev/vdev.component.html
```diff
@@ -93,6 +93,7 @@
       </button>
     </div>
   </div>
-  <div *ngIf="error" [style.color]="'red'" [innerHTML]="error" class="datainfo"></div>
+  <div *ngIf="error" [innerHTML]="error" class="sizeMessage red"></div>
+  <div *ngIf="showDiskSizeError" [innerHTML]="diskSizeErrorMsg" class="sizeMessage orange"></div>
```
undsoft, line 97: There is no html in `diskSizeErrorMsg`, you can just do `{{ diskSizeErrorMsg }}`

src/app/pages/storage/volumes/manager/manager.component.scss
```diff
@@ -64,7 +64,7 @@ h4 {
   display: auto;
 }
 
-.sizeMessage {
+::ng-deep .sizeMessage {
```
undsoft, line 67: Looks like a good opportunity to change `.sizeMessage` to `.size-message`

undsoft (changes requested):
Still shows `Mixing disks of different sizes in a vdev is not allowed`, not what ticket suggests.

undsoft (approved):
Thanks.
Снимок экрана 2021-12-20 в 18 08 31

PR #6217, opened 2021-12-20
NAS-113728 / 13.0 / Make single SLOG device warning less scary for commercial users

AlexKarpov98:


undsoft (changes requested):
1. See image. Two errors in the background. One error in dialog.
2. Not a new issue, but it's also possible to get a dialog which would show no errors, but ask to overwrite them. To do this, I just immediately added one log vdev upon navigating to this page and then pressed Force.

undsoft (changes requested):
Eh, I don't really want a permanent "This layout is not recommended due to errors". What I want is the list of errors/warnings that were shown on the page to be shown in the dialog.
For example in this case:

Снимок экрана 2021-12-22 в 19 03 12

src/app/helptext/storage/volumes/manager/manager.ts
```diff
@@ -19,10 +19,10 @@ export default {
   manager_disknumExtendConfirmMessage: T('It is not recommended to extend a pool with one or\
  more vdevs containing different numbers of disks. Continue?'),
 
-  manager_vdevtypeErrorMessage: T('Adding data vdevs of different types is not supported.'),
+  manager_vdevtypeErrorMessage: T('Caution: Adding data vdevs of different types is not supported.'),
 
   manager_stripeVdevTypeErrorMessage: T('vdev is highly discouraged and will result in data loss if it fails'),
-  manager_logVdevWarningMessage: T('A stripe log vdev may result in data loss if it fails combined with a power outage.'),
+  manager_logVdevWarningMessage: T('Caution: a stripe log vdev layout can lose data in the unlikely event of simultaneous vdev failure and total power failure.'),
```
dszidi, line 25: The letter A should be capitalized eg. "Caution: A stripe log..."

dszidi (approved):
LGTM

image

undsoft (changes requested):
Issues found:
1. "Caution: A stripe data vdev is highly discouraged and will result in data loss if it fails" is no longer red for some reason.
2. Errors seem to be sticky? I.e. if I set layout to give me a "stripe log vdev" error, but then undo changes, the error still stays. See video:

https://user-images.githubusercontent.com/1481270/147354883-30e35f29-dc7f-410a-8013-cee835552c4f.mov

undsoft (approved):
Looks good now, thanks.

PR #6218, opened 2021-12-21
NAS-113453 / 13.0 / NAS-113453: Prevent recursive shutdowns and reboots

undsoft:
Basically in some scenarios the url for reboot/shutdown and some other pages got saved, which led to infinite reboot cycle.
More detailed steps are in ticket description.


denysbutenko (approved):
Tested with reboot looks good. Got redirected to the sign-in page

PR #6222, opened 2021-12-21
NAS-112919 / Bluefin / Refactor Jira Ticket Form

denysbutenko:
For testing, go to General Settings and click on the "File Ticket" button
- Try to submit a ticket
- Check if debug is attached to the ticket
- Check if screenshots are attached successfully to the ticket (50 MiB max size per file)


AlexKarpov98:
LOOK GOOD!

I attached test screenshot.
- [ ] Attach screenshots does not work
- [ ] Ticket detail is a few difference

image

denysbutenko:
> LOOK GOOD!
> 
> I attached test screenshot.
> 
> * [ ]  Attach screenshots does not work
> * [ ]  Ticket detail is a few difference
> 
> image

@AlexKarpov98 thanks, that's a good point. Screenshots attaching handled via a job, so it may take a while until it gets uploaded. I'll double-check it again.
About `Suggestion` not sure why it happens. The category is a mandatory field. My [bug ticket](https://jira.ixsystems.com/browse/NAS-113910) has all these params. Probably the bugclerk uses different logic for the `Bug` and `Feature` types. Needs further investigation.

@undsoft does it make sense to change the label from `Feature` to `Suggestion`?

undsoft:
If the change is only on the frontend, then yes, Suggestion sounds good.

src/app/interfaces/api-event-directory.interface.ts
```diff
@@ -17,4 +17,5 @@ export type ApiEventDirectory = {
   'truecommand.config': { response: TrueCommandConfig };
   'vm.query': { response: VirtualMachine };
   'zfs.pool.scan': { response: ResilverData };
+  'support.attach_ticket': { response: Job };
```
undsoft, line 20: This file is only used with `subscribe` method of `WebsocketService`. Doesn't seem to be needed.

src/app/pages/system/general-settings/support/file-ticket-form/file-ticket-form.component.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    ws = spectator.inject(WebSocketService);
+    matDialog = spectator.inject(MatDialog);
+    jest.spyOn(matDialog, 'open').mockImplementation();
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('shows current values when form is being opened', async () => {
```
undsoft, line 74: More like `it('loads ticket categories using ... token when form is opened')`

src/app/pages/system/general-settings/support/file-ticket-form/file-ticket-form.component.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+        Category: '',
+        Subject: '',
+        Body: '',
+        Type: 'Bug',
+        'Attach Debug': false,
+      },
+    );
+    expect(ws.call).toHaveBeenCalledWith('support.fetch_categories', ['token.is.mocked']);
+  });
+});
```
undsoft, line 90: Also need to test form submission.

src/app/pages/system/general-settings/support/file-ticket-form/file-ticket-form.component.ts
```diff
@@ -0,0 +1,224 @@
...
+      filter((token) => !!token),
+      untilDestroyed(this),
+    ).subscribe((token) => {
+      this.sysGeneralService.setTokenForJira(token);
+    });
+
+    this.form.get('screenshot').valueChanges.pipe(
+      untilDestroyed(this),
+    ).subscribe((fileList: FileList) => {
+      this.fileList = fileList;
```
undsoft, line 89: Let's add a // TODO: to move this to a separate file.

src/app/pages/system/general-settings/support/file-ticket-form/file-ticket-form.component.ts
```diff
@@ -0,0 +1,224 @@
...
+  getCategories(): Observable<Option[]> {
+    return this.form.get('token').value$.pipe(
+      filter((token) => !!token),
+      debounceTime(300),
+      switchMap((token) => this.ws.call('support.fetch_categories', [token])),
+      map((choices) => {
+        let options: Option[] = [];
+        for (const property in choices) {
+          if (choices.hasOwnProperty(property)) {
+            options.push({ label: property, value: choices[property] });
```
undsoft, line 116: choicesToOptions()?

src/app/pages/system/general-settings/support/file-ticket-form/file-ticket-form.component.html
```diff
@@ -0,0 +1,63 @@
...
+          [tooltip]="tooltips.body | translate"
+          [rows]="8"
+          [required]="true"
+        ></ix-textarea>
+        <ix-file-input
+          formControlName="screenshot"
+          [label]="'Attach screenshots' | translate"
+          [tooltip]="tooltips.screenshot | translate"
+          [multiple]="true"
+          [acceptedFiles]="'*.png,*.jpeg,*.jpg,*.gif'"
```
undsoft, line 49: Let's move to a constant somewhere.

src/app/pages/system/general-settings/support/file-ticket-form/file-ticket-form.component.ts
```diff
@@ -0,0 +1,224 @@
...
+          }
+        }
+        options = _.sortBy(options, ['label']);
+        return options;
+      }),
+    );
```
undsoft, line 122: Errors are unhandled.

src/app/pages/system/general-settings/support/file-ticket-form/file-ticket-form.component.ts
```diff
@@ -0,0 +1,224 @@
...
+        this.openSuccessDialog(ticket);
+      } else {
+        dialogRef.close();
+        this.slideIn.close();
+        this.openSuccessDialog(ticket);
+      }
+    });
+    dialogRef.componentInstance.failure.pipe(untilDestroyed(this)).subscribe((res) => {
+      this.isFormLoading = false;
+      dialogRef.componentInstance.setDescription(res.error);
```
undsoft, line 199: I think you may need to do `cdr.markForCheck` here to stop the progress bar.

src/app/pages/system/general-settings/support/file-ticket-form/file-ticket-form.component.ts
```diff
@@ -0,0 +1,224 @@
...
+      },
+    });
+    dialogRef.componentInstance.setCall('support.new_ticket', [payload]);
+    dialogRef.componentInstance.submit();
+    dialogRef.componentInstance.success.pipe(
+      untilDestroyed(this),
+    ).subscribe((res: Job<NewTicketResponse>) => {
+      let ticket: NewTicketResponse;
+      if (res.result) {
+        ticket = res.result;
```
undsoft, line 166: Are we expecting no result in res? What happens in this case?

src/app/pages/system/general-settings/support/file-ticket-form/file-ticket-form.component.ts
```diff
@@ -0,0 +1,224 @@
...
+            method: 'support.attach_ticket',
+            params: [{
+              ticket: res.result.ticket,
+              filename: file.name,
+              token: payload.token,
+            }],
+          }));
+          formData.append('file', file);
+          dialogRef.componentInstance.wspost(this.apiEndPoint, formData);
+          dialogRef.componentInstance.success.pipe(untilDestroyed(this)).subscribe(() => {
```
undsoft, line 181: 1. I've attached two files to my ticket and none of them appeared in Jira. I've waited for 7 minutes.
2. We should not be uploading images silently. I may be uploading a large image, see the success dialog, think everything is done and close the browser. See if you can rewrite the code to show user a dialog. Also, would be good to use switchMaps, not go deeper into callbacks.

src/app/pages/common/ix-forms/components/ix-jira-oauth/ix-jira-oauth.component.html
```diff
@@ -0,0 +1,41 @@
...
+  <div *ngIf="label" class="label-container">
+    <label class="label">
+      {{label}}
+      <span *ngIf="required" class="required">*</span>
+    </label>
+    <tooltip *ngIf="tooltip" [header]="label" class="tooltip" [message]="tooltip"></tooltip>
+  </div>
+
+  <div class="outlined input-container" [class.has-reset-button]="shouldShowResetInput()">
+    <input
```
undsoft, line 12: Help text on Login to Jira is good, but we don't really need to show token to the user - there is nothing he can do with it.

undsoft (changes requested):

AlexKarpov98:
Waiting for tests success

undsoft:
Doesn't build. Likely because of me moving files.

denysbutenko:
Fixed build

src/app/modules/ix-forms/components/ix-jira-oauth/ix-jira-oauth.component.html
```diff
@@ -0,0 +1,40 @@
+<button
+  mat-button
+  type="button"
+  (click)="initSession()"
+  class="action-button"
+  [disabled]="hasValue()"
+>{{ (hasValue() ? 'Logged In Jira' : 'Login to Jira') | translate }}</button>
```
undsoft, line 7: Logged In To Jira

src/app/pages/system/general-settings/support/file-ticket-form/file-ticket-form.component.html
```diff
@@ -0,0 +1,67 @@
+<ix-modal-header [title]="'File Ticket' | translate" [loading]="isFormLoading"></ix-modal-header>
+<mat-card>
+  <mat-card-content>
+    <form [formGroup]="form" class="ix-form-container" (submit)="onSubmit()">
+      <ix-fieldset [title]="'OAuth' | translate">
```
undsoft, line 5: Can we have no title?

src/app/services/ix-file-upload.service.ts
```diff
@@ -61,4 +63,58 @@ export class IxFileUploadService {
   get onUploaded$(): Subject<HttpResponse<unknown>> {
     return this.fileUploadSuccess$;
   }
+
+  validateFiles(fileList: FileList): Observable<ValidatedFile[]> {
```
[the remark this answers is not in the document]
undsoft: I'm also okay with not validating file types if it helps.

undsoft (changes requested):
1. Multiple file selector is neat, though my natural assumption was that if I already have something selected and I press "Choose File", it would add to the list, not replace.
2. For loading indicator, how about showing it instead of Save button or in the header instead of the progress bar.
@dszidi What do you think? Here's how it looks now https://drive.google.com/file/d/1-9BbykfMPnyy1yqkrCDhe4YIuzyIeS-g/view?usp=sharing
3. I've added two files, but they don't appear on my ticket. Weirdly `support.attach_ticket` reports SUCCESS.
I've recently created https://jira.ixsystems.com/browse/NAS-113995, may be related.
4. Since images are being added after the form is closed, we should inform user that it may take some time before images appear.

src/app/helpers/filelist.helper.ts
```diff
@@ -0,0 +1,9 @@
+/**
+  * @param files Array of files to add to the FileList
+  * @returns FileList
+  */
+export function fileListItems(files: File[]): FileList {
```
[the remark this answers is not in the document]
denysbutenko: This hack aims to allow FileList editing to provide the `remove screenshot` button. Unfortunately, there is no other way to modify FileList. Input with `type=file` has attribute `files` with `FileList` type, so it would require transformation at some point anyway in case of using `File[]`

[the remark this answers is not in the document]
undsoft: I understand, I'm saying that it's better to refactor ix-file-input to transform to `File[]` and work with that instead of trying to piece together new `FileList`.

denysbutenko:
> 1. Multiple file selector is neat, though my natural assumption was that if I already have something selected and I press "Choose File", it would add to the list, not replace.

Seems like replacing is common behavior, probably because of FileList immutability.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#using_file_inputs

> 2. For loading indicator, how about showing it instead of Save button or in the header instead of the progress bar.

Yep, good point. I did a try to put in `ix-modal-header` but have some issues with change detection and `app-job-item` sometimes does not appear. Also, Job Progress != Progress Bar in `ix-modal-header`, Progress Bar waits until all actions (job and screenshots uploading) are done.

> 3. I've added two files, but they don't appear on my ticket. Weirdly support.attach_ticket reports SUCCESS.
I've recently created https://jira.ixsystems.com/browse/NAS-113995, may be related.

The workflow is following, when a user clicks Save:

- Create a job `support.new_ticket`
- Wait until it is done (it may take a couple of minutes up to 5-10 when attaching debug)
- Loop through attached screenshots to make `http.post(support.attach_ticket)` query
- All screenshots uploaded to middleware
- Close the form and show success dialog
- Meanwhile, middleware handles uploading screenshots to Jira Ticket 

> 4. Since images are being added after the form is closed, we should inform user that it may take some time before images appear.

Yeah, it makes sense. Also, we could show jobs for uploading screenshots (mw->jira) (`job_id` could be extracted from response of http.post()`

denysbutenko:
addressed feedback & rebase on bluefin

src/app/modules/ix-forms/components/ix-slide-in/components/ix-modal-header/ix-modal-header.component.html
```diff
@@ -3,3 +3,4 @@ <h3 class="ix-formtitle">{{ title | translate }}</h3>
   <mat-icon id="ix-close-icon" (click)="slideInService.close()">cancel</mat-icon>
 </div>
 <mat-progress-bar *ngIf="loading" mode="indeterminate"></mat-progress-bar>
+<ng-content></ng-content>
```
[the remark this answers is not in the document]
denysbutenko: yep, moved `app-job-item` from the bottom of form to `ix-modal-header`

denysbutenko:
Note:
If you select `attach_debug` and the file is reached the size limit then a ticket will be created but neither debug nor screenshots won't be attached (`support.new_ticket` job will have `FAILED` status)

src/app/helpers/options.helper.ts
```diff
@@ -7,8 +7,17 @@ export function mapToOptions(map: Map<string, string>): Option[] {
...
 }
 
-export function choicesToOptions(): OperatorFunction<Choices, Option[]> {
+/**
+ * Convert choices to options
+ * @param [swap] use when value of object matched to option value
+ * @default key of object matched to option value
+ * @returns Option[]
+ */
+export function choicesToOptions(swap = false): OperatorFunction<Choices, Option[]> {
```
undsoft, line 16: I'd prefer not to add this option, and to instead just do map in that one place where it's needed. This really feels like an exception.

src/app/modules/ix-forms/components/ix-slide-in/components/ix-modal-header/ix-modal-header.component.html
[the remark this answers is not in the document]
undsoft: Can we just put `app-job-item` underneath `ix-modal-header` and not inside?

undsoft (approved):
Nice work 👍🏻

PR #6223, opened 2021-12-21
NAS-113851 / 22.02 / NAS-113851: Changes default train when adding a catalog.

undsoft:


dszidi (approved):
Tested and field defaults to stable. LGTM

PR #6224, opened 2021-12-21
NAS-113860 / 22.02 / Remove Authority Key Identifier option from CSR form

AlexKarpov98:
The error was from `AuthorityKeyIdentifier.enabled`. 
Kept to use `AuthorityKeyIdentifier.authority_cert_issuer` and `AuthorityKeyIdentifier.extension_critical`.


undsoft (changes requested):
1. Authority Key Identifier needs to be removed from CSR form only. Changes to this component will also affect Certificates form.
2. I think we need to remove the whole option. It doesn't make sense to me to be able to set some of the fields, if I cannot enable it.

undsoft (changes requested):
If I set Authority Key Identifier in a Add Certificate form it still won't be saved correctly.

undsoft (approved):
Mission accomplished.

PR #6225, opened 2021-12-21
NAS-113869 / 22.02 / NAS-113869: Users Page Add Button ix-auto value regression

denysbutenko:
For testing, go to the Users page, and right-click on the "Add" button choose "Inspect" then make sure that 
`ix-auto="button__Users_ADD"`


RehanY147 (approved):
Looks good

PR #6226, opened 2021-12-21
NAS-113814 / 13.0 / Making `getServerTime` async

RehanY147:
Not sure how to test. This was an issue on Chrome only that wasn't reproducible on localhost anyway, so don't know how you would test this. This needs to be tested directly on a VM.


undsoft (approved):
Works well, tested by replacing files within a vm.

PR #6233, opened 2021-12-22
NAS-113903 / 22.02 / Input type="number" validation in Firefox

AlexKarpov98:


dszidi (commented):
I configured the group add form's GID field with input="number". When I type something into the GID field it simply doesn't respond at all. No characters get added to the input field and no validation error message appears below. Also, if there is already a validation error present, the previous validation error persists.

image

If something is being entered that is not passing the field's validation, we should keep the behavior consistent and show a validation error message.

For a working example, we can look at how the "name" field located directly below the GID field behaves...

image

AlexKarpov98:
> I configured the group add form's GID field with input="number". When I type something into the GID field it simply doesn't respond at all. No characters get added to the input field and no validation error message appears below. Also, if there is already a validation error present, the previous validation error persists.
> 
> image
> 
> If something is being entered that is not passing the field's validation, we should keep the behavior consistent and show a validation error message.
> 
> For a working example, we can look at how the "name" field located directly below the GID field behaves...
> 
> image

@dszidi Thanks!
Input type="number" does not allow any character is not number.
I imitated it in Firefox. which prevents the default event, so would not obstruct to input validation.

You can reproduce this bug in `Web Interface HTTP Port` field of System/General/GUI Setting through two browsers.

image

dszidi:
@AlexKarpov98 Thanks for providing field that I can reference in the UI.

Apologies, as I rushed to write my previous comment I left out a key part.

>  "_... we should keep the behavior consistent and show a validation error message._" 

 should have been 
 > " _...we should keep the behavior consistent and show a validation error message **as well as the input.**_"

When the user types characters into the field, they should still see what they typed even if it breaks validation. Blocking the keypress could give the user the impression that something about their browser/computer is broken and is generally bad UX.

AlexKarpov98:
@undsoft  Input type="number" on Chrome does not allow to type any character that is not number (except "e").
To allow to see what they typed even if it breaks validation, it firstly should update on Chrome.
Could you update the description?

AlexKarpov98:
Inputs type:

- password
Use component `ix-input-password`

- number
Use component `ix-input`
        Notes: 
        Should use numberValidator() instead of type="number"
        Should use rangeValidator(1, 10000) instead of Validators.min() or Validators.max() if needed

src/app/pages/common/entity/entity-form/validators/number-validation.ts
```diff
@@ -0,0 +1,26 @@
+import { ValidatorFn, FormControl } from '@angular/forms';
+
+export function numberValidator(): ValidatorFn {
+  let thisControl: FormControl;
+
+  return function numberValidate(control: FormControl) {
```
undsoft, line 6: Validator can probably be simplified.
I don't understand why we need `thisControl`.

src/app/pages/common/ix-forms/components/ix-input/ix-input.component.ts
```diff
@@ -75,15 +73,7 @@ export class IxInputComponent implements ControlValueAccessor {
   }
 
   shouldShowResetInput(): boolean {
```
undsoft, line 75: Method is probably not needed.

src/app/pages/common/ix-forms/components/ix-input/ix-input.component.ts
```diff
@@ -18,24 +18,22 @@ export class IxInputComponent implements ControlValueAccessor {
...
-  @Input() type: string;
   @Input() autocomplete = 'off';
   @Input() autocapitalize = 'off';
 
   /** If formatted value returned by parseAndFormatInput has non-numeric letters
    * and input 'type' is a number, the input will stay empty on the form */
-  @Input() format: (value: string | number) => string;
-  @Input() parse: (value: string | number) => string | number;
+  @Input() format: (value: string) => string;
+  @Input() parse: (value: string) => string;
```
undsoft, line 27: return value can be a number

undsoft (changes requested):
It looks good to me, though the changes are somewhat scary.
Let's retarget this to `bluefin`.

@RehanY147 Any opinion on this PR?

src/app/pages/common/ix-forms/components/ix-input-password/ix-input-password.component.ts
```diff
@@ -0,0 +1,73 @@
...
+@Component({
+  selector: 'ix-input-password',
+  templateUrl: './ix-input-password.component.html',
+  styleUrls: ['./ix-input-password.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class IxInputPasswordComponent implements ControlValueAccessor {
+  @Input() label: string;
+  @Input() placeholder: string;
+  @Input() prefixIcon: string;
```
RehanY147, line 17: With such a specific input, we can remove such fields which we know we won't be using.

RehanY147 (changes requested):
The IxValidation stuff looks OK. I do have a couple of notes about the overall approach.
- We should see if it's possible to do all of this behind the scenes. It's very intuitive to just add `type="number"` and easy to forget that you're supposed to do the other stuff if you just want a numeric field. Maybe we can allow devs to just add the type and do the validation etc behind the scenes.
- A lot of the times, the API end point is expecting a numeric value and will throw out validation errors otherwise. As is, we will have to remember to convert strings to numbers and do trial and error every time we forget to do so. We should have a behind the scenes system which makes sure that a field that is meant to be numeric throws out a numeric value.

src/app/pages/common/ix-forms/components/ix-input/ix-input.component.ts
AlexKarpov98: image
 I think it is needed for reset input.

src/app/pages/common/ix-forms/components/ix-input-password/ix-input-password.component.ts
denysbutenko: What if we wish to refactor the sign-in form and want to use this component?

AlexKarpov98:
Retargeted to `bluefin` [ https://github.com/truenas/webui/pull/6348 ](https://github.com/truenas/webui/pull/6348)

PR #6234, opened 2021-12-22
NAS-113646 / 13.0 / Added fix for length check

RehanY147:
To reproduce this bug, create two smb shares at least and name the first one any letter out of `KMGTB` like `B` or `G`. If that doesn't work rename the other one to this.


AlexKarpov98:
I created two smb shares and edit/rename the first one `B`, then I can no longer see any smb shares, only a loading view.
Tried to reload the page, but still the same.

image

RehanY147:
@AlexKarpov98 Please share the names of your smb shares

undsoft (approved):
Worked well for me. Looks old code for @AlexKarpov98.

PR #6240, opened 2021-12-24
NAS-113953 / 22.02 / NAS-113953: When locale is not "English (en)", reports other than CPU cannot be accessed, as they redirect to the dashboard

bvasilenko:
Testing: switch between different reports and keep an eye on URL - the URL should not contain any invalid routes, such as "процессор" or "Datenträger"

### DIFF
--- a/src/app/pages/reports-dashboard/reports-dashboard.component.ts
+++ b/src/app/pages/reports-dashboard/reports-dashboard.component.ts
@@ -9,6 +9,7 @@ import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { TranslateService } from '@ngx-translate/core';
 import { Subject, BehaviorSubject } from 'rxjs';
 import { CoreService } from 'app/core/services/core-service/core.service';
+import { ReportTab } from 'app/enums/report-tab.enum';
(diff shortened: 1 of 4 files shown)


undsoft (changes requested):
There seems to be more code that relies on `label`, which may get broken.
See `activateTab()` and usages for `activeTab`.
Also looks like now is the time to introduce the enum for tab values.

src/app/pages/reports-dashboard/reports-dashboard.component.ts
```diff
@@ -54,7 +55,7 @@ export class ReportsDashboardComponent implements OnInit, OnDestroy, /* HandleCh
   otherReports: Report[];
   activeReports: Report[] = [];
 
-  activeTab = 'CPU'; // Tabs (lower case only): CPU, Disk, Memory, Network, NFS, Partition?, System, Target, UPS, ZFS
+  activeTab = { label: this.translate.instant('CPU'), value: ReportTab.Cpu } as Tab;
```
undsoft, line 58: You can also type `value` in `Tab` to `ReportTab`.

undsoft (approved):
Looks good.

PR #6242, opened 2021-12-27
NAS-111851 / 22.02 / NAS-111851: Updating manpage links

undsoft:
Mostly replaces with links to http://manpages.org/ with some specific exceptions.


RehanY147 (commented):
Instructions on testing

Mrt134 (approved):
Nice improvement!

undsoft:
Looks like replacing links broke message extraction for these messages. Will require more work.

undsoft:
I had to redo the work to keep translation extractor happy. 
Links were replaced again, this time mostly with https://man7.org/linux/man-pages/man8/statd.8.html, some with https://linux.die.net/man/8/smartd and a few with direct links to places like www.samba.org

src/app/helptext/system/tunable.ts
```diff
@@ -28,7 +28,7 @@ export const helptextSystemTunable = {
       'Enter a value to use for the <a\
  href="https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/boot-introduction.html#boot-loader-commands"\
```
undsoft, line 29: @Mrt134 Need advice on what to replace these links with.

Mrt134: I'm looking at the related area in the SCALE UI (22.02-RC.2) and noticing the section has been retitled "Sysctl" and no longer has a drop-down for switching between loader, sysctl, or rc.conf options. 

image

If I'm reading this correctly, then the entire tooltip can be removed except for the loader link: "Enter a value to use for the <a href="https://man7.org/linux/man-pages/man8/sysctl.8.html" target="_blank">sysctl</a>."

undsoft: Makes sense, fixed.

undsoft: @Mrt134 Bump.

PR #6246, opened 2021-12-28
NAS-113406 / 22.02 / NAS-113406: Change FreeBSD wording

undsoft:


src/app/helptext/network/interfaces/interfaces-form.ts
```diff
@@ -24,7 +24,7 @@ export default {
   ],
 
   int_interface_placeholder: T('NIC'),
-  int_interface_tooltip: T('Enter the FreeBSD device name of the\
+  int_interface_tooltip: T('Enter the device name of the\
  interface. This cannot be changed after the interface is created.'),
   int_interface_validation: [Validators.required],
```
RehanY147, line 29: What about place where links from www.freebsd.org are used? Are those alright?

RehanY147 (approved):
Looks good to me otherwise.

src/app/helptext/network/interfaces/interfaces-form.ts
undsoft: Addressed in another PR.

src/assets/i18n/af.json
```diff
@@ -506,7 +506,7 @@
   "Browse to the keytab file to upload.": "",
   "Browse to the locally saved configuration file.": "",
   "Browse to the operating system installer image file.": "",
-  "Browse to the path to be copied. FreeBSD  file path limits apply. Other operating systems can have different  limits which might affect how they can be used as sources or  destinations.": "",
+  "Browse to the path to be copied. Linux  file path limits apply. Other operating systems can have different  limits which might affect how they can be used as sources or  destinations.": "",
```
Mrt134, line 509: Found these redundant empty spaces. I see the same in the other i18n files - is this an error from translation file generation?
```suggestion
  "Browse to the path to be copied. Linux file path limits apply. Other operating systems can have different limits which might affect how they can be used as sources or destinations.": "",
```

Mrt134 (commented):
Found one small thing.

src/assets/i18n/af.json
undsoft: It's an artifact of how the code was written.

PR #6247, opened 2021-12-28
NAS-113756 / 22.02 / NAS-113756: form-chip property of undefined error

undsoft:
Go to Sharing -> iSCSI Configure. Observe no error in console.


dszidi (approved):
LGTM

PR #6249, opened 2021-12-29
NAS-113839 / None / NAS-113839: Making linter rule more strict

undsoft:
No testing is necessary, code review only.


RehanY147 (approved):
Code looks good. Just variable/class renaming. Nothing too complicated.

PR #6250, opened 2021-12-29
NAS-112767 / 22.02 / Creating a ix-user-combobox behavior with service methods

RehanY147:
This is a draft to look at the possible code towards implementing an IxUserCombobox using only method refs without creating a specific component outright.

### DIFF
--- a/src/app/pages/common/ix-forms/services/ix-users.service.ts
+++ b/src/app/pages/common/ix-forms/services/ix-users.service.ts
@@ -1,0 +1,27 @@
+import { Injectable } from '@angular/core';
+import { TranslateService } from '@ngx-translate/core';
+import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
+import { Option } from 'app/interfaces/option.interface';
+import { User } from 'app/interfaces/user.interface';
+import { UserService } from 'app/services';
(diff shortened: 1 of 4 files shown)


src/app/pages/common/ix-forms/services/ix-users.service.ts
```diff
@@ -0,0 +1,27 @@
+import { Injectable } from '@angular/core';
+import { TranslateService } from '@ngx-translate/core';
+import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
+import { Option } from 'app/interfaces/option.interface';
+import { User } from 'app/interfaces/user.interface';
+import { UserService } from 'app/services';
+
+@Injectable()
+export default class IxUsersService {
```
undsoft, line 10: UsersService would be okay.

undsoft (commented):
That's a lot of code that needs to be there in every component that uses a user combobox. Any way of extracting this to something reusable?
I.e. we could consider having a concept (interface) of DataProvider that then gets instantiated by ArrayDataProvider for simple cases (essentially replacing) or a more custom UserDataProvider that supports all of this stuff.
Basically something similar to a DataSource in Material.

PR #6253, opened 2022-01-04
NAS-114028 / 22.02 / Added the new field

RehanY147:
To test, go to the page Services -> S3 (Config/Edit icon click)
There, if you change the `Certificate` field to something other than empty/null, the new field should show up. It should be required only in this case and not required otherwise.

### DIFF
--- a/src/app/pages/services/components/service-s3/service-s3.component.html
+++ b/src/app/pages/services/components/service-s3/service-s3.component.html
@@ -49,6 +49,7 @@
           *ngIf="form.controls['certificate'].value !== null"
           formControlName="tls_server_uri"
           [label]="'TLS Server Uri' | translate"
+          [required]="true"
         >
         </ix-input>
       </ix-fieldset>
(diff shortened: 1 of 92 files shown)


src/app/pages/services/components/service-s3/service-s3.component.html
```diff
@@ -45,6 +45,13 @@
           [tooltip]="tooltips.certificate | translate"
           [options]="certificateOptions$"
         ></ix-select>
+        <ix-input
+          formControlName="tls_server_uri"
+          [label]="'TLS Server Uri' | translate"
+          [required]="true"
+          *ngIf="form.controls['certificate'].value !== null"
```
undsoft, line 52: Let's move *ngIf as a first line within ix-input.

undsoft (changes requested):
Looks like I no longer can submit form with empty Certificate.

RehanY147:
Should we working now 
certificate issue

src/app/pages/services/components/service-s3/service-s3.component.html
```diff
@@ -45,6 +45,12 @@
           [tooltip]="tooltips.certificate | translate"
           [options]="certificateOptions$"
         ></ix-select>
+        <ix-input
+          *ngIf="form.controls['certificate'].value !== null"
+          formControlName="tls_server_uri"
+          [label]="'TLS Server Uri' | translate"
```
undsoft, line 51: Let's also add [required]="true"

undsoft (approved):
Works well now.

PR #6260, opened 2022-01-06
NAS-105865 / 22.02 / Display parent dataset name in GUI in the "Add Dataset" form

AlexKarpov98:


src/app/helptext/storage/volumes/datasets/dataset-form.ts
```diff
@@ -3,8 +3,11 @@ import { marker as T } from '@biesbjerg/ngx-translate-extract-marker';
 import { matchOtherValidator } from 'app/pages/common/entity/entity-form/validators/password-validation/password-validation';
 
 export default {
+  dataset_parent_name_placeholder: T('Parent name'),
```
RehanY147, line 6: It's not parent name as much as it is path of the dataset being created. Terminology here should be changed.
For example, creating a dataset under `pool` -> `ds1` gives the impression that "parent name" would be `ds` since that's the parent dataset of the dataset being created. However, you see `pool/ds1` in the parent name field.

RehanY147 (changes requested):

RehanY147 (approved):
Looks good to me

PR #6262, opened 2022-01-06
NAS-114150 / 13.0 / Pressing reload for the license it reload to the Dashboard instead of Support page

AlexKarpov98:
Solved the problem that redirected to dashboard page when reloading the page.
Similar to Scale version [NAS-110533](https://jira.ixsystems.com/browse/NAS-110533)


denysbutenko (approved):
Seems like a backport. LGTM

PR #6266, opened 2022-01-07
NAS-112371 / 13.0 / Fixed wrong text

RehanY147:


RehanY147:
This should be back ported to 12.0 if possible as per @themylogin

undsoft (approved):
Messaging makes sense.

PR #6267, opened 2022-01-09
NAS-114137 / 22.02 / NAS-114137: Added the new field

RehanY147:
To test, go to `S3` -> `Edit Icon` and check that the new field works and has a default value `9001`


denysbutenko (changes requested):
Tests are failed

undsoft (approved):
The new field is there.

PR #6269, opened 2022-01-09
NAS-114137 / 13.0 / Added new field

RehanY147:


AlexKarpov98 (changes requested):
Got an error with a new field `console_bindport`
image

undsoft (approved):
Was able to save new field succesfully.

PR #6270, opened 2022-01-10
NAS-113560 / 22.02 / NAS-113560: Added check for a not auth message

RehanY147:
This is a preventative measure for getting `Not Authenticated` errors. Not sure how to reproduce the scenario where this code would be reached but as described by the ticket description, this case is reached somehow. Another ticket will be created to refactor the authentication process. That's expected to improve the process overall.


dszidi (approved):
Code looks good and those error messages are gone

PR #6275, opened 2022-01-10
NAS-114222 / 22.02 / default to sh for container shell

Ornias1993:
This moves the default container shell from /bin/bash to /bin/sh.
Generally speaking all containers have sh, but lightweight containers do not always contain bash.

This has lead to multiple instances of people getting confused and filing bugreports on either the forum, discord or TrueCharts-support


denysbutenko (approved):
Sounds reasonable. Thanks for the contribution 👍🏻

PR #6276, opened 2022-01-11
NAS-114088 / 22.02 / NAS-114088: Cloud Sync Tasks - prevent page from crashing (fix job schedule parsing)

bvasilenko:
Made a lot of attempts to replicate issue which user was facing, and finally discovered the following bug:
Parse logic was based on a hard-coded string "Disabled" and thus was tightly coupled with English localization

Testing:
1. Switch language to non English (German, Russian)
2. Credentials -> Cloud Storage -> Dropbox (fill in your dropbox creds, or any other cloud creds you have)
3. Go to Data Protection -> Add Cloud Sync task -> Create 2 tasks
4. Edit both tasks marking each as "Disabled"
5. Go to Data Protection -> click "Flyout" icon next to Cloud Sync Tasks 
Expected result:
image

### DIFF
--- a/src/app/pages/common/entity/entity-table/entity-table-row-details/entity-table-row-details.component.html
+++ b/src/app/pages/common/entity/entity-table/entity-table-row-details/entity-table-row-details.component.html
@@ -28,7 +28,7 @@
             ></app-task-schedule-list>
           </ng-container>
         </mat-menu>
-        <p>{{ getPropValue(column.prop, true) }}</p>
+        <p>{{ config.enabled ? getPropValue(column.prop, true) : getPropValue(column.prop) }}</p>
       </ng-container>
       <ng-template #notScheduleList>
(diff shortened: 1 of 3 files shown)


undsoft (changes requested):
Well, it works, but I think the solution can be improved.
Current solution is essentially "if anything bad happened, it's disabled" which is not necessarily true. It may be a good idea to distinguish between disabled and an error when parsing schedule.
It also produces a lot of errors in console when you move cursor over buttons in the table. These will get reported to Sentry.

I think you can check for `this.config.enabled` in `buildFutureRuns` instead. Also see `resourceTransformIncomingRestData` in `CloudsyncListComponent`.

undsoft (approved):
False alarm, the error was due to incorrect packages on my side.

Works well.
Снимок экрана 2022-01-12 в 16 23 35

PR #6277, opened 2022-01-11
NAS-113706 / 22.02 / Cloud Sync Tasks: null credentials is not allowed

AlexKarpov98:
To test:
Data Protection -> Cloud Sync Tasks -> Click Add
Select `+ Add a backup credential` option of Credential field
Navigated to `backup-credentials` slider to add a backup credential


RehanY147 (changes requested):
As is, the UI doesn't tell a straight forward story. I came to add a cloud sync task but now I'm on the credentials page and I have to find my way back. So, if we're taking the User to a completely different form and page, we should also bring them back and also keep in mind the values they had filled into the form before to provide a natural progression.

AlexKarpov98:
Allowed to add a credential in Dialog instead of redirection.

src/assets/styles/other/_tn-styles.scss
```diff
@@ -570,6 +570,11 @@ $primary-dark: darken(map-get($md-primary, 500), 8%);
   }
   /////////////////// END SpecSelectors
 
+  .cdk-overlay-pane {
```
RehanY147, line 573: This is a very generic selector. To avoid issues with styles throughout the system unintended, we should use a more specific css selector which only applies to where we want.

src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
```diff
@@ -719,6 +721,38 @@ export class CloudsyncFormComponent implements FormConfiguration {
     this.folderDestinationField = this.fieldSets.config('folder_destination') as FormExplorerConfig;
     this.folderSourceField = this.fieldSets.config('folder_source') as FormExplorerConfig;
     this.formGroup.controls['credentials'].valueChanges.pipe(untilDestroyed(this)).subscribe((res: number | typeof NULL_VALUE) => {
+      if (!res) {
+        const dialogRef = this.matDialog.open(CloudCredentialsFormComponent, {
+          width: '600px',
+          data: { title: this.translate.instant('Cloud Credentials') },
```
RehanY147, line 727: Title isn't showing up on the dialog form.

src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
```diff
@@ -719,6 +721,38 @@ export class CloudsyncFormComponent implements FormConfiguration {
     this.folderDestinationField = this.fieldSets.config('folder_destination') as FormExplorerConfig;
     this.folderSourceField = this.fieldSets.config('folder_source') as FormExplorerConfig;
     this.formGroup.controls['credentials'].valueChanges.pipe(untilDestroyed(this)).subscribe((res: number | typeof NULL_VALUE) => {
+      if (!res) {
```
RehanY147, line 724: If I click on the "Add new credential" option the first time, the dialog opens. But if I close the dialog without doing anything, the drop down has "Add new credential" selected and clicking on it again does nothing until you change the value to something else. We should handle that case too where dialog is just closed without success or failure

src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
```diff
@@ -719,6 +721,38 @@ export class CloudsyncFormComponent implements FormConfiguration {
...
+          width: '600px',
+          data: { title: this.translate.instant('Cloud Credentials') },
+        });
+        dialogRef.componentInstance.finishSubmit = (value) => {
+          dialogRef.componentInstance.prepareAttributes(value);
+          dialogRef.componentInstance.entityForm.submitFunction(value).pipe(untilDestroyed(this)).subscribe(
+            () => {
+              dialogRef.close();
+              this.cloudcredentialService.getCloudsyncCredentials().then((credentials) => {
+                credentials.forEach((item) => {
```
RehanY147, line 735: Could be simplified by just doing `this.credentialsField.options = credentials.map(c => { return { label: item.name + ' (' + item.provider + ')', value: item.id }; });` and `this.credentials = ...credentials;` or something similar.

RehanY147 (changes requested):

src/assets/styles/other/_tn-styles.scss
```diff
@@ -1051,7 +1047,7 @@ $primary-dark: darken(map-get($md-primary, 500), 8%);
   // Bring tooltips to the front in dialogs
   .mat-dialog-container,
   .mat-dialog-content {
-    overflow: visible !important;
```
RehanY147, line 1049: This is again change at a larger scale and might result in some dialog being too short. There's a diff between `visible` and `auto`. `auto` is closer to `scroll`. This will affect all the dialog styles which is out of the scope of this ticket. We should have a specific selector.
image

src/assets/styles/other/_tn-styles.scss
```diff
@@ -570,12 +570,8 @@ $primary-dark: darken(map-get($md-primary, 500), 8%);
   }
   /////////////////// END SpecSelectors
 
-  .cdk-overlay-pane {
-    overflow-x: hidden;
-    overflow-y: auto;
-  }
-
   .mat-dialog-container {
+    max-height: calc(100vh - 100px);
```
RehanY147, line 574: Same as before

RehanY147 (changes requested):

RehanY147 (approved):
Looks good to me. Nice!

PR #6280, opened 2022-01-11
NAS-113818 / None / NAS-113818: Refactoring alerts to ngrx

undsoft:
Alerts were refactored to ngrx. Everything related to alerts requires testing: indicator, dismissing, adding alerts, reopening, doing the same in different tab.

Some alerts are removed when dismissed, others will stay.

To create an alert you can run `midclt call alert.oneshot_create UnscheduledReboot "Unique name"` in CLI.

Changes from previous implementation:
- Alerts are now sorted chronologically with new alerts on top.
- There may be minor design differences.


denysbutenko (approved):
image

Looks good 👍🏻

PR #6282, opened 2022-01-12
NAS-114045 / 22.02 / NAS-114045: Fix Truenas Scale PCIe Passthru issue - Raid card not visible

bvasilenko:
After check the video attachment, I made a suggestion that `vm.device.passthrough_device_choices` has slow response. Assuming it may take ~15 seconds to retrieve such a long device list, it's possible to see the behavior depicted on the video.

The spinner I've added should help the user to know that he needs to wait until PCI devices will be loaded.

Testing: create a VM, and go to Devices -> Add. As a shortcut, I've adjusted response of `vm.device.passthrough_device_choices` query by hard-coding the JSON from the debugging logs which the user has uploaded.

### DIFF
--- a/src/app/pages/vm/devices/device-add/device-add.component.ts
+++ b/src/app/pages/vm/devices/device-add/device-add.component.ts
@@ -333,6 +333,8 @@ export class DeviceAddComponent implements OnInit, OnDestroy {
 
   protected ipAddress: FormSelectConfig;
 
+  isLoadingPci = false;
+
   constructor(
     protected router: Router,
     protected core: CoreService,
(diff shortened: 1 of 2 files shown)


src/app/pages/vm/devices/device-add/device-add.component.ts
```diff
@@ -402,11 +404,16 @@ export class DeviceAddComponent implements OnInit, OnDestroy {
...
     this.ws.call('vm.device.passthrough_device_choices').pipe(untilDestroyed(this)).subscribe((res) => {
+      this.isLoadingPci = true;
       this.pptdev = _.find(this.pciFieldConfig, { name: 'pptdev' }) as FormSelectConfig;
       this.pptdev.options = Object.keys(res || {}).map((pptdevId) => ({
         label: pptdevId,
         value: pptdevId,
       }));
+      this.isLoadingPci = false;
+    },
+    () => {
```
RehanY147, line 415: We should handle the error case here as well since we have added a specific call back, using `EntityUtils`.

src/app/pages/vm/devices/device-add/device-add.component.ts
```diff
@@ -402,11 +404,16 @@ export class DeviceAddComponent implements OnInit, OnDestroy {
 
     // pci
     this.ws.call('vm.device.passthrough_device_choices').pipe(untilDestroyed(this)).subscribe((res) => {
+      this.isLoadingPci = true;
```
RehanY147, line 407: Why set `isLoadingPci = true` after the call has already been subscribed to? At that point, you've already got the response. The loader won't show up while the call is loading but will show up for half a millisecond once we get the response. This line should be moved outside the subscribe on line 405 or something.

RehanY147 (changes requested):

PR #6285, opened 2022-01-12
NAS-114052 / 13.0 / Fixed checkbox clicks by adding unique ids

RehanY147:
Ticket description has details on the issue. One important detail is that you should have the same name for each table item on the alert services list page.


denysbutenko (approved):
Works well, good catch!

PR #6309, opened 2022-01-18
NAS-114214 / 22.02 / NAS-114214: Deleting a dataset causes the deletion of a replication task

bvasilenko:
Testing:
1. Storage -> Create 2 Datasets
2. Create ZVol inside each dataset
3. Data Protection -> Replication task ZVol -> ZVol2. You may also try ZVol -> Dataset, etc.
4. Storage -> Delete ZVol2
Expected result: confirmation dialog will have a list of related replication tasks, which will be deleted automatically


src/app/pages/storage/volumes/volumes-list/volumes-list-table-config.ts
```diff
@@ -1105,6 +997,186 @@ export class VolumesListTableConfig implements EntityTableConfig {
...
+
+    return res;
+  }
+
+  private getDatasetRelatedSummary(
+    datasetName: string,
+    attachments: PoolAttachment[],
+    processes: PoolProcess[],
+  ): string {
+    let res = '';
```
RehanY147, line 1053: Here too

src/app/pages/storage/volumes/volumes-list/volumes-list-table-config.ts
```diff
@@ -1105,6 +997,186 @@ export class VolumesListTableConfig implements EntityTableConfig {
     return actions as EntityTableAction[];
   }
 
+  private getStorageFlattenedAttachments(attachments: PoolAttachment[]): string {
+    const flattenAttachment = (obj: PoolAttachment): string => obj.attachments.map((a) => a.split(',').map((i) => `<br> - ${i}`).join()).join();
```
RehanY147, line 1001: We should break down this code into multiple properly named variable assignments on multiple lines so its more readable and understandable what's happening.

RehanY147 (changes requested):
The warning shows up if I try to delete the source dataset of the replication task. But deleting the destination dataset/zvol of the replication task shows no warning about replication task. Not sure if that's intended.

bvasilenko:
>Not sure if that's intended. 
To me that seems to be by design. Because if the destination Dataset/ZVol does not exists, it will be created automatically.

RehanY147 (approved):
Looks good to me

PR #6314, opened 2022-01-18
NAS-100796 / None / Accounts: Groups: Add column "membership"

AlexKarpov98:
image

To test: Credentials -> Local Group


RehanY147 (approved):
Works ok. The members column shows up with all member names. When deleting the members to be deleted are listed down Ok.

PR #6318, opened 2022-01-19
NAS-114347 / 22.02 / NAS-114347: Virtual Machines list don't update when starting or stopping a VM

bvasilenko:
Testing: start/stop existing VM and watch for issues as per description


src/app/pages/common/entity/entity-table/entity-table.component.ts
```diff
@@ -720,6 +720,19 @@ export class EntityTableComponent<Row = any> implements OnInit, AfterViewInit, A
     return res;
   }
 
+  patchCurrentRows(
+    rowSelector: (row: any) => boolean,
```
undsoft, line 724: Let's do `unknown` instead of anys.

undsoft (approved):
Works okay.

PR #6319, opened 2022-01-19
NAS-110571 / None / Confusing order of group fields when creating a user

AlexKarpov98:
To test:
Go to Credentials -> Users
Hidden `Primary Group` select when `New Primary Group` checked.

Screenshot_13


bvasilenko:
image

Looks and works fine. UI causes no confusion

bvasilenko (approved):

PR #6325, opened 2022-01-21
NAS-113675 / 22.02 / NAS-113675: Allow dashboard widgets to be reordered

bvasilenko:
image

**Testing:**
1. Desktop scenarios: 
   - Reorder, add/remove, cancel 
   - Reorder, add/remove, confirm (sequence of steps may vary)
2. Mobile - only regression testing

**Unit tests**

image

### DIFF
--- a/src/app/pages/common/entity/entity-toolbar/entity-toolbar.component.html
+++ b/src/app/pages/common/entity/entity-toolbar/entity-toolbar.component.html
@@ -1,5 +1,5 @@
 <div *ngIf="config" class="control-group">
-  <div *ngFor="let control of config.controls">
+  <ng-container *ngFor="let control of config.controls">
     <toolbar-button *ngIf="control.type == 'button'" [config]="control" [controller]="controller$" ></toolbar-button>
     <toolbar-slider *ngIf="control.type == 'slider'" [config]="control" [controller]="controller$ | cast"></toolbar-slider>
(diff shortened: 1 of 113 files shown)


dszidi (commented):
Behavior is off. When I drag the System Information card to position 3, it places it in position 2.  It seems to consistently put it 1 position behind where it actually is. Maybe it just needs a temporary card to fill the original position so when the user releases the card it stays where the user intended

image

dszidi:
Let me try to clarify. My screenshot shows the Sysinfo widget hovering over the 3rd position. Ideally that third position should show as empty and the Memory widget should be in position 2. This would indicate to the user that when they release the card it will stay at the position their cursor was at.

RehanY147 (approved):
Nice work! LGTM.

RehanY147 (changes requested):
The reorder button behaves weirdly if you switch between mobile and desktop screens. E.g., if you're going from mobile -> desktop, the reorder button won't show up until refresh. If you go from desktop -> mobile, the reorder button won't go away always.

RehanY147:
What I was talking about earlier.
reorder button issue

bvasilenko:
Additions: 
+ grid layout for dragging (both vertical and horizontal directions)
+ animations
+ reusable "ix-drop-grid" directive
+ unit tests

RehanY147 (approved):
Functionally works well. The code looks OK too.

PR #6327, opened 2022-01-21
NAS-114329: Removing usages of ApiService

undsoft:
Things to test:
- Storage pool widget on main dashboard.
- Enclosure UI


dszidi (approved):
Code looks ok. Tested this on tn11 and it works well. Storage related dashboard widgets and enclosure UI disks view all fetch and display data just fine

PR #6328, opened 2022-01-21
NAS-112767 / 22.02 / Better implementation of IxUserCombobox

RehanY147:
Go to Credentials -> Groups -> Add and observe the bottom two fields. Which now both use the same component to implement the combobox. Also check out the code.


src/app/pages/common/ix-forms/classes/user-combobox-provider.ts
```diff
@@ -0,0 +1,58 @@
+import { forkJoin, Observable, Subject } from 'rxjs';
+import { map, tap } from 'rxjs/operators';
+import { Option } from 'app/interfaces/option.interface';
+import { IxCombobox2Provider } from 'app/pages/common/ix-forms/components/ix-combobox2/ix-combobox2-provider.interface';
+import IxUsersService from 'app/pages/common/ix-forms/services/ix-users.service';
+
+export class UserComboboxProvider implements IxCombobox2Provider {
+  private ProviderUpdater$ = new Subject();
```
undsoft, line 8: bad case

src/app/pages/common/ix-forms/classes/user-combobox-provider.ts
```diff
@@ -0,0 +1,58 @@
...
+  isLoading = false;
```
undsoft, line 9: should be private

src/app/pages/common/ix-forms/classes/user-combobox-provider.ts
```diff
@@ -0,0 +1,58 @@
...
+          tap((options) => {
+            this.isLoading = false;
+            this.pageOffset += (options).length;
+          }),
+        ),
+    ]).pipe(map(([arr1, arr2]) => [...(arr1), ...(arr2)]));
+    this.updateProvider();
+  };
+
+  get providerUpdater$(): Subject<any> {
```
undsoft, line 47: Here and in definition this is probably a Subject<void>

src/app/pages/account/groups/group-form/group-form.component.ts
```diff
@@ -54,7 +127,13 @@ export class GroupFormComponent {
     private cdr: ChangeDetectorRef,
     private errorHandler: FormErrorHandlerService,
     private translate: TranslateService,
-  ) {}
+    public ixUsersProvider: IxUserComboboxProviderService,
+  ) {
+    this.userProvider.providerUpdater$.pipe(untilDestroyed(this)).subscribe(() => {
```
undsoft, line 132: I'm struggling to understand the point of this. Is this to make change detection work? If so, then `ix-combobox2` should be responsible for rerunning change detection after new options have been received from the provider.

src/app/pages/account/groups/group-form/group-form.component.ts
```diff
@@ -37,8 +42,76 @@ export class GroupFormComponent {
...
+          { label: 'Rehan - ' + UUID.UUID().substring(0, 6), value: UUID.UUID() },
+          { label: 'Rehan - ' + UUID.UUID().substring(0, 6), value: UUID.UUID() },
+          { label: 'Rehan - ' + UUID.UUID().substring(0, 6), value: UUID.UUID() },
+          { label: 'Rehan - ' + UUID.UUID().substring(0, 6), value: UUID.UUID() },
+        ] as Option[]),
+      ]).pipe(delay(3000)).pipe(map(([arr1, arr2]) => {
+        this.normalLoading = false;
+        return [...arr1, ...arr2];
+      }));
+      this.normalProvider = _.cloneDeep(this.normalProvider);
```
undsoft, line 110: Why clone?

src/app/pages/account/groups/group-form/group-form.component.ts
```diff
@@ -37,8 +42,76 @@ export class GroupFormComponent {
...
+      this.cdr.markForCheck();
+    },
+  };
+
```
undsoft, line 114: It's nice to see an example of a more advanced provider with static values. I would also like to see an example of simplified version for developer use, something like:
```
normalProvider = new ArrayProvider([
       { label: UUID.UUID().substring(0, 6), value: UUID.UUID() },
       { label: UUID.UUID().substring(0, 6), value: UUID.UUID() },
       { label: UUID.UUID().substring(0, 6), value: UUID.UUID() },
]);

```

src/app/pages/common/ix-forms/components/ix-combobox2/ix-combobox2-provider.interface.ts
```diff
@@ -0,0 +1,9 @@
+import { Observable } from 'rxjs';
+import { Option } from '../../../../../interfaces/option.interface';
+
+export interface IxCombobox2Provider {
+  pageOffset: number;
```
undsoft, line 5: I am not sure that we need `pageOffset` in this interface. Seems like this should be something for provider to figure out internally.

undsoft (changes requested):
General architecture looks good, had some issues when using comboboxes:
1. First combobox has 2.5 second delay every time I type in a letter.
2. There seems to be a large delay or a change detection issue between typing something and seeing dropdown values.
3. Sometimes I see an infinite scrolling bar, for example when I open normal checkbox and scroll down.

src/app/pages/common/ix-forms/classes/user-combobox-provider.ts
```diff
@@ -0,0 +1,57 @@
+import { Subject } from 'rxjs';
+import { Option } from 'app/interfaces/option.interface';
+import { IxCombobox2Provider } from 'app/pages/common/ix-forms/components/ix-combobox2/ix-combobox2-provider.interface';
+import IxUsersService from 'app/pages/common/ix-forms/services/ix-users.service';
+
+export class UserComboboxProvider implements IxCombobox2Provider {
+  private ProviderUpdater$ = new Subject<void>();
```
undsoft, line 7: The case is still incorrect.

src/app/pages/account/groups/group-form/group-form.component.ts
undsoft: Still would like to see this example.

src/app/pages/common/ix-forms/components/ix-combobox2/ix-combobox2-provider.interface.ts
```diff
@@ -0,0 +1,10 @@
+import { Subject } from 'rxjs';
+import { Option } from '../../../../../interfaces/option.interface';
+
+export interface IxCombobox2Provider {
+  providerUpdater$: Subject<void>;
+  options: Option[];
```
undsoft, line 6: How about not having `providerUpdater$` and instead just have `options$` be a stream?

undsoft (changes requested):
Works much better.
I think we should also consider having some indication that search is happening or about to happen. Maybe loading indicator should instead appear within the input box (like a spinner on the right or a very thin line) and appear immediately as I type.
Right now because of debounce and probably because of network request there is a noticeable delay between me typing something and seeing results.
If I was a user who didn't know that this UI control performs a search, I may give up sooner than I see suggestions.

RehanY147:
Going a different route for now.

PR #6330, opened 2022-01-21
NAS-110078 / None / NAS-110078: Replaces system config with ngrx

undsoft:
Testing: general regression and special attention to places that rely on general or advanced config, such as Settings.


src/app/store/system-config/system-config.actions.ts
```diff
@@ -0,0 +1,11 @@
+import { createAction, props } from '@ngrx/store';
+import { AdvancedConfig } from 'app/interfaces/advanced-config.interface';
+import { SystemGeneralConfig } from 'app/interfaces/system-config.interface';
+
+export const systemConfigLoaded = createAction(
+  '[System Config API] Loaded',
+  props<{ generalConfig: SystemGeneralConfig; advancedConfig: AdvancedConfig }>(),
+);
+
+export const generalConfigUpdated = createAction('[System Config] General Config Updated');
```
RehanY147, line 10: These names could be improved I feel like. They should be actions not actions already performed. E.g., `reloadGeneralConfig` and `reloadAdvancedConfig`.

RehanY147 (changes requested):
The console messages panel at the bottom not immediately showing up after a page refresh and form submit.
console not showing

src/app/store/system-config/system-config.actions.ts
undsoft: Even though they are called actions, these are really events as mentioned here: https://ngrx.io/guide/store/actions#writing-actions

> Event-Driven - capture events not commands as you are separating the description of an event and the handling of that event.

undsoft: These action names are incorrect in a sense that they should be more specific, like `guiFormSubmitted`. Current ones are too general.
If you are interested in more about this: https://www.youtube.com/watch?v=JmnsEvoy-gY

undsoft:
Fixed.

RehanY147 (approved):
Looks good to me. No obvious regressions. The code looks good too.

PR #6335, opened 2022-01-22
NAS-114432 / None / NAS-114432: Refactor iSCSI global configuration form to ix forms.

undsoft:
Testing: Shares -> iSCSI Shares -> Configure.


denysbutenko (commented):
image

Can we make `mat-progress-bar` to take full width?

Nit: when `Save` is pressed `Settings saved.` message do not disappear

denysbutenko (approved):
Works well!

PR #6337, opened 2022-01-24
NAS-114410 / 22.02 / Custom cron job modal is cut off on smaller screens

AlexKarpov98:
Updated `form-scheduler` style


undsoft (approved):
Scrolling on smaller screens is there, but there is now a small scrolling gap when modal is opened by default.

Снимок экрана 2022-01-24 в 11 03 24

PR #6341, opened 2022-01-24
NAS-114428 / 22.02 / NAS-114428: Validation Error Traceback from unsetting Include Dataset Properties

bvasilenko:
**Testing:**

1. Data Protection -> Replication Tasks -> Add one (use Wizard for simplicity)
2. Click on your replication task to Edit
3. Uncheck **Include Dataset Properties** and click **Save**

**Summary:**

When saving changes with **Include Dataset Properties** unticked, payload for `replication.update` was:

BEFORE:
```
      "name":"My Pool/My Dataset - My Pool/My Dataset 2",
      "properties":false,      <--- FALSE reflects unchecked state of "Include Dataset Properties"
      "properties_override":null, <--- NULL causes "null not allowed"
      "properties_exclude":[],
```

AFTER:
```
      "name":"My Pool/My Dataset - My Pool/My Dataset 2",
      "properties":false,
      "properties_override":{}, <--- dummy object
      "properties_exclude":[],
```

### DIFF
--- a/src/app/pages/data-protection/replication/replication-form/replication-form.component.ts
+++ b/src/app/pages/data-protection/replication/replication-form/replication-form.component.ts
@@ -1427,15 +1427,14 @@ export class ReplicationFormComponent implements FormConfiguration {
       data['properties'] = true;
       data['exclude'] = [];
     }
+    const propertiesExcludeObj: any = {};
     if (data['properties_override']) {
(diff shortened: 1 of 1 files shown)


dszidi (approved):
Code looks good. Was able to edit and save task after unchecking "Include Dataset Properties"

PR #6344, opened 2022-01-25
NAS-114471 / None / Sync link styles across pages

AlexKarpov98:
Screenshot_5
Screenshot_6


denysbutenko:
Just a thought: would it be possible to use a CSS selector like `a[href*="//"]` (an example) to determine external links instead of adding an `external-link` class?

undsoft (changes requested):
Would be good preserve some of the existing translations, as in fr.json and maybe others.

undsoft:
> Would be good preserve some of the existing translations, as in fr.json and maybe others.

This point still stands.

undsoft (changes requested):
Need to recover translation files.

undsoft (approved):
Thanks

PR #6345, opened 2022-01-26
NAS-114467 / None / NAS-114467: Extracting Encryption Options to a separate component

undsoft:
Testing: create encrypted datasets and test context menu -> Encryption Options.
Additional logic:
- Dataset that is a child of passphrase encrypted dataset can only be passphrase encrypted.
- Dataset with a key encrypted dataset can only be key encrypted.

There are minor stylistic differences, which are expected.


src/app/pages/storage/volumes/encyption-options-dialog/encryption-options-dialog.component.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  private setControlDependencies(): void {
+    this.form.controls.algorithm.disable();
+
+    if (this.data.hasPassphraseParent || this.data.hasKeyChild) {
+      this.form.controls['encryption_type'].disable();
+    }
+
+    this.form.controls['key'].disabledWhile(combineLatestIsAny([
```
bvasilenko, line 177: I noticed that `control.disableWhile()` internally calls `.subscribe` , but never calls `.unsubsribe` .  
Additionally, I found out that `form.select()` call won't stop emitting after component is destroyed.
I also noticed that `.disableWhile` returns a `Subscription` object, which brings to us the responsibility to call `.unsubscribe`.

Summarizing these points, I suggest to call `.unsubscribe` at `ngOnDestroy()`, or alternatively, to apply `untilDestroyed` pipe after `form.select()`.

bvasilenko (approved):

src/app/pages/storage/volumes/encyption-options-dialog/encryption-options-dialog.component.ts
undsoft: You are right 😔 
I'll fix it as a separate PR, because this is not the only place with `disabledWhile` and I want to discuss this with library authors.

PR #6352, opened 2022-01-27
NAS-112767 / 22.02 / NAS-112767: Functional approach to IxCombobox

RehanY147:


undsoft (changes requested):
The approach looks good, but I'd to see a loading indicator when user starts typing in user combobox. Right now you have to know that network request is being sent and wait for results.

undsoft (approved):
Looks good to me. Spinner can be optimized further:
- not to show it before user touches control (i.e. on initial page load).
- not to show it when user selects an option from the list?

@dszidi 
You okay with this indication?

https://user-images.githubusercontent.com/1481270/151559123-6b31ae46-a875-45c7-b9e2-18ca3194f76a.mov

AlexKarpov98 (approved):
LGTM

RehanY147:
Have to change loader placement because it will be hidden in this case
image

RehanY147:
New loader placement
new loader

undsoft (approved):
Yes, looks better.

PR #6354, opened 2022-01-27
NAS-114516 / 22.02 / NAS-114516: Removed checkbox

RehanY147:
Go to System -> Services -> SNMP (Configure) check that the checkbox `iftop` is no longer there.


denysbutenko (approved):
image

Looks good

PR #6355, opened 2022-01-27
NAS-114394 / 22.02 / NAS-114394: Making sure the service restarts

RehanY147:
Go to the GUI settings under System -> General. Change the certificate and save. You should be prompted to restart the service. If you restart the service, the websocket will disconnect and connect again. Make sure that login on reconnect is working and you are still able to use the UI or redirected to the signin page if sign in is not possible.

Save anything other setting other than certificate (like one of the ports) and check that service restart happens (`service.restart` http is called) without calling `system.general.ui_restart` being called. And again that upon reconnection, login works and UI is usable.

You can test this on the Scale VM available to the team.

### DIFF
--- a/src/app/pages/system/general-settings/gui-form/gui-form.component.ts
+++ b/src/app/pages/system/general-settings/gui-form/gui-form.component.ts
@@ -8,7 +8,8 @@ import {
 import { FormBuilder } from '@ngneat/reactive-forms';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { TranslateService } from '@ngx-translate/core';
-import { filter } from 'rxjs/operators';
+import { combineLatest } from 'rxjs';
+import { filter, tap } from 'rxjs/operators';
 import { ServiceName } from 'app/enums/service-name.enum';
 import { choicesToOptions } from 'app/helpers/options.helper';
 import { helptextSystemGeneral as helptext } from 'app/helptext/system/general';
@@ -119,7 +120,6 @@ export class GuiFormComponent {
     ).subscribe(() => {
       this.isFormLoading = false;
       this.cdr.markForCheck();
-      this.slideInService.close();
       this.handleServiceRestart(body);
     }, (error) => {
       this.isFormLoading = false;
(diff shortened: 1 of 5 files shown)


src/app/pages/system/general-settings/gui-form/gui-form.component.ts
```diff
@@ -179,7 +181,11 @@ export class GuiFormComponent {
 
         this.loader.open();
         this.ws.shuttingdown = true; // not really shutting down, just stop websocket detection temporarily
-        this.ws.call('service.restart', [ServiceName.Http]).pipe(
+        const obs = [this.ws.call('service.restart', [ServiceName.Http])];
+        if (uiCertificateChanged) {
+          obs.push(this.ws.call('system.general.ui_restart'));
```
undsoft, line 186: I feel like these two are either the same or `system.general.ui_restart` should replace `service.restart`. Please confirm with middleware.

src/app/pages/system/general-settings/gui-form/gui-form.component.ts
```diff
@@ -179,7 +181,11 @@ export class GuiFormComponent {
...
+        }
+        combineLatest(obs).pipe(
```
undsoft, line 188: What is `obs`?

src/app/pages/system/general-settings/gui-form/gui-form.component.ts
```diff
@@ -8,7 +8,8 @@ import {
 import { FormBuilder } from '@ngneat/reactive-forms';
```
undsoft, line 8: Tests need to be updated.

undsoft (changes requested):

src/app/pages/system/general-settings/gui-form/gui-form.component.ts
RehanY147: 'service.restart` is called with `HTTP` param so I think it's meant to restart the `HTTP` service. `system.general_ui_restart` restarts the UI service at least from the names.

src/app/pages/system/general-settings/gui-form/gui-form.component.ts
RehanY147: `observables` shorthand

src/app/pages/system/general-settings/gui-form/gui-form.component.ts
undsoft: I don't think there is such a thing as "UI service"

undsoft (approved):
Can you also fix the following error, that I get after service is restarted:

> Cannot read properties of undefined (reading 'classList')
>   at modal.component.ts:98:18

https://drive.google.com/file/d/1PcULC8vC9KZLASrPOgyKC_Uz0DUDi5TC/view?usp=sharing

PR #6356, opened 2022-01-28
NAS-112795 / None / NAS-112795: fixed form-input value overlap icon

Raz-Dva:
Check the password fields that have a button to toggle the display of the password


undsoft (changes requested):
It's better to add styles to form-input styles and not to global styles.

undsoft (changes requested):
Doesn't look to good here.
Снимок экрана 2022-01-30 в 12 42 58

undsoft (changes requested):
As discussed on Slack
> width: 372px !important;
needs to be removed completely.

RehanY147 (commented):
Waiting for changes on previous comments by @undsoft

undsoft (approved):
Looks good.

PR #6362, opened 2022-01-31
NAS-114549 / None / NAS-114549: Removing unused styles and other code

undsoft:
General regression testing is needed.


RehanY147 (approved):
Looks good.

PR #6364, opened 2022-01-31
NAS-113939 / None / Nas 113939

Raz-Dva:
Check window with the cron time setting (e.g. backup with rclone or rsync).
Months and days of the week must not go beyond the window.
Make a check in German and Spanish


bvasilenko (approved):
Translations look good to me.

PR #6370, opened 2022-02-01
NAS-112343 / None / Websocket observables are not properly multicasted

AlexKarpov98:
Testing:
Confirm multicasting in any component:
```
const request$ = this.ws.call('system.general.language_choices');     
request$.subscribe((config) => console.log('subscription 1', config));
request$.subscribe((config) => console.log('subscription 2', config));
request$.subscribe((config) => console.log('subscription 3', config));
```


undsoft (changes requested):
Looks like `ConnectableObservable` is deprecated, I think there should be an easier way with new pipeable operators.

undsoft (approved):
Looks good to me.

PR #6371, opened 2022-02-02
NAS-114530 / 22.02 / NAS-114530: Added filter for ix-chart

RehanY147:
Go to the apps page and go to the `Available Applications` tab. You shouldn't see the `ix-chart` app as mentioned in the ticket.


undsoft (commented):
1. There is a `ixChartApp` constant that you can use.
2. Waqar is saying that this looks like a regression. Can you check if this is so?

AlexKarpov98 (approved):
LGTM!

PR #6380, opened 2022-02-04
Expose `servers` key for NFS service in SCALE

AlexKarpov98:
Testing

System Settings -> Services -> NFS service
Exposed servers key as the `Number of threads` .


RehanY147 (approved):
Looks good to me

PR #6383, opened 2022-02-04
NAS-114607 / None / Refactor ServiceUpsComponent to ix forms

AlexKarpov98:
Testing

System Settings -> Services -> UPS


dszidi (commented):
Trying to test this on a VM. I've filled out every field but save button is still disabled

image

PR #6386, opened 2022-02-05
NAS-114620 / 22.02.1 / NAS-114620: Fixed field hide order

RehanY147:
Go to the certificates page and create a CA. In the form, on the first step, change the type to `Intermediate CA`. Now on the third step `Subject`, all of the fields should be showing up. Previously, this step would be just empty.


undsoft (approved):
The fields are there.

PR #6387, opened 2022-02-07
NAS-114686 / None / NAS-114686: Enabling additional linter rules

undsoft:
Careful code review should be enough.


RehanY147 (approved):
Looks good to me.

PR #6388, opened 2022-02-07
NAS-114653 / 22.02.1 / NAS-114653: Altered warning when has_debug is false

RehanY147:
Reproduction steps are mentioned in the ticket


bvasilenko (approved):
Looks good to me

PR #6391, opened 2022-02-08
NAS-114688 / None / NAS-114688: Refactoring preferences to ngrx

undsoft:
Testing:
- UI should remember if sidebar was collapsed
- Form in topbar -> Preferences should work correctly.
- Some tables allow to select which columns to show and should remember this setting.
- Dashboard should remember its settings
- Date and time formats should be correctly saved in Settings -> Localization form
- User and Group lists should remember setting for showing built-in users.

Also removed: VM wizard used to store values selected in some dropdowns.


RehanY147 (changes requested):
The preferences form isn't working. There's no indication that the form submitted or is in progress etc. And the form doesn't go away either. It just stays there. And I get too many calls errors in the console if I click it a couple of times to check.

undsoft:
Added back indication that form is submitted and fix console errors. The form will not redirect anywhere, this is by design.
The indication is stupid (shown before request completes), but this is fine, as I'm refactoring this form in next PR anyway.

RehanY147 (changes requested):
Also no indication for when Toggle Built-in Users button is clicked on the Users page. On a slow internet, the prompt goes away and I sat there confused whether the UI malfunctioned and nothing changed. Until, it suddenly clicked in and I realized that it did work.

RehanY147 (approved):
The rest looks good functionality wise. The code looks fine too.

PR #6392, opened 2022-02-08
NAS-114597 / None / webUI network interface icons not working

AlexKarpov98:
Removed states icons completely.


undsoft (changes requested):
I am not convinced we should remove this completely.
1. We still have something very similar on the dashboard.
2. As a user there is value in seeing what network interfaces are connected.

How about either making it more reliable and fixing the underlying undefined issue or finding if we can change this indicator to only show whether interface is connected, but leave out bytes sent/received.

AlexKarpov98:
Updated:
- Hide tooltip when `sent` or `received` is undefined.
- Update link_status of the interface from interfaces value of the subscription `reporting.realtime`

src/app/pages/dashboard/components/widget-network/widget-network.component.ts
```diff
@@ -425,7 +425,10 @@ export class WidgetNetworkComponent extends WidgetComponent implements AfterView
     const lastSent = this.storage.convertBytestoHumanReadable(this.nicInfoMap[nic.state.name].lastSent);
     const lastReceived = this.storage.convertBytestoHumanReadable(this.nicInfoMap[nic.state.name].lastReceived);
 
-    return `${this.translate.instant('Sent')}: ${lastSent} ${this.translate.instant('Received')}: ${lastReceived}`;
+    if (lastSent !== undefined && lastReceived !== undefined) {
+      return `${this.translate.instant('Sent')}: ${lastSent} ${this.translate.instant('Received')}: ${lastReceived}`;
+    }
+    return null;
```
undsoft, line 431: Usually successful scenario is the last one.

undsoft (changes requested):
1. It looks like at least for machines from here https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20(Full%20-%20Nightly%20ISO)/ the indication is wrong on Network page. NIC is shown as disconnected, which is not true.
2. When I then switch from Network page to Dashboard, arrows won't show anything. If I reload the page, they will work.

https://user-images.githubusercontent.com/1481270/154961070-7a5cbef6-554a-4021-a547-0d9d7e53bc84.mov

undsoft (approved):
Tested basic configurations on 4 machines. They seem to work fine.

PR #6393, opened 2022-02-09
NAS-114633 / None / NAS-114633: Fixed ix-input value overlap icon

Raz-Dva:
Check inputs that have a reset button


denysbutenko (approved):
image

Looks good!

PR #6399, opened 2022-02-10
Add some Simplified Chinese Translation

hjkl950217:


hjkl950217:
I want to ask how to ensure that the localization file contains these multilingual keys when new text is added in subsequent versions? I checked I18N files in several languages and found that the order of keys is different

RehanY147 (approved):
Looks good. Thank you!

RehanY147:
@hjkl950217 We need help from contributors like yourself to translate new text as its added. Messages and text strings that is added to the code gets extracted automatically and put into `.json` files (such as `zh-hans.json`). Contributors/translators can help us by adding translations there when new strings are added without translation.

PR #6402, opened 2022-02-10
NAS-113654: Fixed styles for password autocomplete that overlap the b…

Raz-Dva:
Check fields with autocompletion (passwords, etc.)


bvasilenko (approved):
Looks good

PR #6404, opened 2022-02-10
NAS-114608 / None / Refactor TwoFactorComponent to ix-forms

AlexKarpov98:
Testing:

Credentials -> 2FA


src/app/pages/system/two-factor/two-factor.component.html
```diff
@@ -0,0 +1,94 @@
+<mat-card class="form-card">
+  <mat-progress-bar *ngIf="isFormLoading" mode="indeterminate"></mat-progress-bar>
+  <mat-card-content>
+    <form [formGroup]="form" class="ix-form-container" (submit)="onSubmit()">
+      <div class="fieldset-content">
+        <ix-fieldset [title]="helptext.two_factor.title">
+          <p class="help-text">
+            {{helptext.two_factor.message}}
```
undsoft, line 8: Need `| translate` here and in tooltips and labels below.

undsoft (changes requested):
1. I'm thinking we should rearrange the form a little bit. For most users the workflow is to click Enable and scan the code. The rest of the form is advanced settings really.
Let's move status text ("Two-factor authentication is NOT enabled."), Enable Two-Factor Authentication button and Show QR to be above the form. 
We should also remove the primary color from Save (since it's not a primary action) and give it to Enable/Disable button.
2. I think we should always show the QR code after Enable button is pressed.

AlexKarpov98:
Rearranged the form:
Screenshot_18

Read-only input behaviors:
- Shallow background style
- Select all when focus
- Copy cursor
Screenshot_19

Added tooltip to Title:
Screenshot_20

AlexKarpov98:
@undsoft I need help with:
Current there are 3 skipped failed tests in [two-factor.component.spec.ts](https://github.com/truenas/webui/blob/0257ee95d482b1cb9f558926ad44690485e8bc37/src/app/pages/system/two-factor/two-factor.component.spec.ts)
image
I have a problem expecting calling after dialog confirmation in the test.
Let me know how to get it sorted. Thank you!

undsoft (approved):
Nice work! Functionality works well, will message you about tests in Slack.

undsoft:
Basically, the problem (at least in some tests) is that `this.dialogService.confirm` is not set to return anything in `mockProvider`.
It should be something like this:
```
mockProvider(DialogService, {
  confirm: jest.fn(() => of()),
}),
```

As for `disable 2FA when click `Disable Two-Factor Authentication` button` case, you need to set up necessary pre-conditions for `Disable Two-Factor Authentication` button to appear.
I think it would be okay to do something like:
```
spectator.inject(MockWebsocketService).mockCallOnce('auth.twofactor.config', ...config to make button appear);
spectator.component.ngOnInit();
```

PR #6405, opened 2022-02-11
NAS-113044 / None / NAS-113044: Handle errors in options in ix-select and ix-combobox

bvasilenko:
Testing:
1. find usage of a ix-select
2. simulate error by replacing API calls with incorrect strings
3. repeat for ix-combobox


RehanY147 (approved):
Looks good to me

PR #6406, opened 2022-02-11
NAS-114682 / 13.0 / Unable to select 0 in special vdev for metadata size

AlexKarpov98:
Testing:

Storage -> Create dataset / Edit dataset option
Try to input 0 in the `Metadata (Special) Small Block Size` field.


denysbutenko (approved):
image

Works as expected.

PR #6408, opened 2022-02-11
NAS-114731 / None / NAS-114731: fixed width for .mat-card-table

Raz-Dva:
Check borders for empty tables 
border


bvasilenko (approved):
Looks good

PR #6409, opened 2022-02-14
NAS-114840 / None / NAS-114840: Renaming single letter variables

undsoft:
Code review should be enough.


RehanY147 (approved):
Looks good to me.

PR #6410, opened 2022-02-14
NAS-114141 / None / NAS-114141: fixed scrollbar display in pool widget

Raz-Dva:
Check widgets. The scrollbar should only appear if the content block is overflow.
pastedImage


undsoft (approved):
The scrollbars are no longer there.

PR #6384, opened 2022-02-04
NAS-113713 / Bluefin / NAS-113713: Refactoring Snapshot List

denysbutenko:
PR is up for early feedback. It is greatly appreciated.

Go to `/storage/snapshots` to check it.

What's new?

Updated components:
- [x] refactor `snapshot-list` to use ix-tables
- [x] add unit tests for `snapshot-list`
- [x] add `snapshot-rollback-dialog` component
- [x] add unit tests for `snapshot-rollback-dialog`
- [x] add `snapshot-batch-delete-dialog` component
- [x] add unit tests for `snapshot-batch-delete-dialog`
- [x] refactor `snapshot-add` component to use ix-forms
- [x] add unit tests for `snapshot-add-form`

Changes:

- [x] add `toolbar-slide-toggle` component  it's for `Entity Toolbar` (use it on Users page)
- [x] improve `slide-toggle` behavior when user clicks `Cancel`
- [x] lazy loading of `SnapshotsModule`
- [x] use store for snapshots
- [x] use store for `showExtraColumns`, previously `snapshotXtraCols` in local storage
- [x] add loading indicator on toggle column (Users page should adapt these improvements)
- [x] remove `SnapshotListRow` interface, using `ZfsSnapshot` instead
- [x] Address feedback items
- [x] fix unit tests

Legacy:
- [x] ~refactor `snapshot-details` to use `ix-expand-toggle-column`~ (removed)
- [x] ~add pipe `convertBytestoHumanReadable`~ (replaced with filesize)
- [x] ~add unit tests for `convertBytestoHumanReadable`~ (replaced with filesize)

PR #6411, opened 2022-02-14
NAS-109881 / None / Added buttons around icons to make them interactable and other minor fixes for snapshots page

RehanY147:
Added transparent `<button>` around mat-icons as per the recommendations here https://material.angular.io/components/icon/overview. The snapshots page has special cases where columns are changed on the go and some additions were also made to fix issues with that.

To test
1. Check the snapshots page and use it with a screen reader (NVDA) or just use the system with a keyboard. The icons should be intractable. (Previously, those icons would be completely skipped and not intractable without a mouse click). The icons should have proper descriptions when in focus for the screen reader via `aria-label`
2. Change the columns from the cog icon on the top right. Make sure that the columns change and the action icon on the right changes as well based on whichever configuration is selected currently. (Previously, in some cases after a reload, switching back to all columns would not show the actions menu icon on the row right)
3. Also check that the setting is persistent and UI remembers the last configuration on reload. (Previously, you would always see fewer columns selected no matter what after a reload)
4. The columns drop down trigger button at the top right previously didn't behave correctly. It wouldn't hide or show if columns config changed. That should be fixed now as well.


denysbutenko:
From first view it looks good. These changes affected each `entity-table` not only `snapshot-list`. Even if refactored `snapshot-list` don't use `entity-table-actions` component it has value for other pages. I'll test it in details a bit later.
Thanks for pointing out snapshots-related code in `entity-table`. I'll remove it in PR #6384.

`ix-expand-column` [component](https://github.com/truenas/webui/blob/master/src/app/pages/common/ix-tables/components/ix-expand-toggle-column/ix-expand-toggle-column.component.html) is not affected

cc @undsoft

denysbutenko (changes requested):
It works using a keyboard. Something weird happens when I click on the three dots button. It doesn't open the menu, but sometimes it appears if I click multiple times.

https://user-images.githubusercontent.com/351613/154030561-c6607ae2-0f2d-48fa-bcbc-aba7181e315f.mp4

For column toggling, it works for a single column only. `Reset to defaults` and `Unselect All` and `Select All` don't anything

https://user-images.githubusercontent.com/351613/154030842-f2f6ea75-c951-4c06-9250-11ef96e05d39.mp4

RehanY147:
Nice catch @denysbutenko! Fixes added.

denysbutenko:
@RehanY147  the build is failed and new image is not available for testing

RehanY147:
@denysbutenko fixed now.

src/app/modules/entity/entity-table/entity-table-actions/entity-table-actions.component.html
```diff
@@ -17,31 +17,36 @@
 
   <ng-container *ngIf="!isSingleAction && !inlineActions; else inlineActionsTemplate">
     <!-- Material Design Icon Set -->
-    <mat-icon
+    <button
+      class="transparent-btn"
+      [attr.aria-label]="translateLabels('Actionable Options')"
```
denysbutenko, line 22: why use the `translateLabels` method instead of translate pipe?

denysbutenko (approved):
Looks good!

PR #6412, opened 2022-02-14
NAS-114750 / None / NAS-114750: changed strings from "TLS Server URI" to "TLS Server Host…

Raz-Dva:
The field name "TLS Server URI" has been changed to "TLS Server Hostname" for S3 Service. Check this label and field.
Services -> S3 ->Certificate (choose freenas_default)


dszidi (approved):
LGTM

image

PR #6413, opened 2022-02-14
changed string from 'TLS Server URI' to 'TLS Server Hostname'

Raz-Dva:
The field name "TLS Server URI" has been changed to "TLS Server Hostname" for S3 Service. Check this label and field.
Services -> S3 ->Certificate (choose freenas_default)


bvasilenko (approved):
LGTM!

PR #6414, opened 2022-02-14
NAS-114639 / None / NAS-114639: CSR cannot be signed immediately after it has been added

bvasilenko:
On /credentials/certificates UI route, we had a bug - dropdown menu in Authorities block does not refresh it's list of options, when a new option gets added.

Testing: see https://jira.ixsystems.com/browse/NAS-114639

### DIFF
--- a/src/app/pages/credentials/certificates-dash/certificates-dash.component.ts
+++ b/src/app/pages/credentials/certificates-dash/certificates-dash.component.ts
@@ -65,7 +65,7 @@ export class CertificatesDashComponent implements OnInit {
     });
     this.systemGeneralService.getUnsignedCertificates().pipe(untilDestroyed(this)).subscribe((res) => {
       res.forEach((item) => {
-        this.signCsrFieldConfSelect.options.push(
+        this.unsignedCsrSelectField.options.push(
           { label: item.name, value: item.id },
         );
       });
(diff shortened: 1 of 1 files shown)


src/app/pages/credentials/certificates-dash/certificates-dash.component.ts
```diff
@@ -390,15 +396,17 @@ export class CertificatesDashComponent implements OnInit {
     }, 200);
   }
 
+  private signCsrFieldConfSelect: FormSelectConfig = {
```
AlexKarpov98, line 399: Please name as `unsignedCertificatesField` or `unsignedCsrSelectField` instead of `signCsrFieldConfSelect`

src/app/pages/credentials/certificates-dash/certificates-dash.component.ts
```diff
@@ -67,7 +65,7 @@ export class CertificatesDashComponent implements OnInit {
     });
     this.systemGeneralService.getUnsignedCertificates().pipe(untilDestroyed(this)).subscribe((res) => {
       res.forEach((item) => {
-        this.unsignedCertificateAuthorities.push(
+        this.signCsrFieldConfSelect.options.push(
           { label: item.name, value: item.id },
         );
       });
```
AlexKarpov98, line 71: I think we don't need call this subscribe on init because it would called whenever clicking checkmark icon.

AlexKarpov98 (approved):
Look good! 👍

PR #6416, opened 2022-02-15
NAS-113471 / 13.0 / NAS-113471: Dashboard and Storage -> Pools degraded symbology do not match

bvasilenko:
image

image

image

Testing: 

mock properties of the pool object
- `healthy` to `false` 
- `status` to `DEGRADED` (also try `LOCKED` and `FAULTED`)

### DIFF
--- a/src/app/core/components/widgets/widgetpool/widgetpool.component.css
+++ b/src/app/core/components/widgets/widgetpool/widgetpool.component.css
@@ -323,6 +323,22 @@ span.capitalize{
   text-transform:capitalize;
 }
 
+.orange-icon {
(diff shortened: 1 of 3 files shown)


undsoft (changes requested):
Pending changes for [NAS-114412](https://jira.ixsystems.com/browse/NAS-114412)

bvasilenko:
Now we're displaying 

image

when status is Online/Healthy, but `healthy` property is `false`. This logic matches similar logic in SCALE.

undsoft (approved):
Thanks, confirmed via mocking.

PR #6420, opened 2022-02-16
NAS-113464 / None / Added ix-table-extended

RehanY147:
The `ix-table-extended` approach needs to be reviewed. Plus typical code review. The NTP Servers List under General Settings is used as an example.

### DIFF
--- a/src/app/modules/ix-tables/components/ix-table-extended/ix-table-extended.component.ts
+++ b/src/app/modules/ix-tables/components/ix-table-extended/ix-table-extended.component.ts
@@ -10,6 +10,7 @@ import { IxTableStatus } from 'app/modules/ix-tables/enums/ix-table-status.enum'
 @Component({
   selector: 'ix-table-extended',
   templateUrl: 'ix-table-extended.component.html',
+  styleUrls: ['ix-table-extended.component.scss'],
 })
 export class IxTableExtendedComponent implements AfterContentInit {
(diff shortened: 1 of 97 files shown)


denysbutenko (approved):
Looks good to me. Perhaps it can replace `app-table` on the Sharing and Data Protection pages

src/app/modules/ix-tables/components/ix-table-extended/ix-table-extended.component.ts
```diff
@@ -0,0 +1,64 @@
...
+    switch (this.status) {
+      case IxTableStatus.Loading:
+        return this.loadingConf;
+        break;
+      case IxTableStatus.Error:
+        return this.errorConf;
+        break;
+      default:
+        return this.emptyConf;
+        break;
```
AlexKarpov98, line 59: May remove break lines.

AlexKarpov98 (approved):
LGTM! 👍
Screenshot_15
Screenshot_16

undsoft (changes requested):
I cannot say that I like the approach. 
One of the major design goals with entity-table was to have a set of components that can be connected together to build the table. As in not do what entity-table was doing by wrapping logic inside of component. This goes back on this goal.
For example, I don't see why title or buttons should be part of the `table` component.
The table tag is now hidden from a developer.

What I really wanted by this ticket is to reduce boilerplate by taking all of these emptyConf, loadingConf, errorConf and extracting that part in a way that can be connected to ix-table.

PR #6422, opened 2022-02-16
NAS-114810 / 22.02 / NAS-114810: Create Pool UI Page - No page/scroll option shown for available disks (SCALE)

bvasilenko:
Testing:
- Storage->Create Pool-> Available Disks
- You will need 10+ disks. One way to do it is mock data


undsoft (approved):
The pagination is back. Not really sure how it disappeared.

PR #6423, opened 2022-02-16
NAS-114746 / None / Nas 114746 Replace statements for Webdav, SMB, NFS 

Raz-Dva:
Statements "No NFS", "Add NFS", "any NFS" don't make sense and shouldn't be exposed to end users. Replace with something like "No NFS shares have been configured yet", "Add NFS share", "any NFS shares yet".
This applies to NFS, WebDAV, SMB.
Open Sharing-> NFS, WebDAV, SMB check text and messages while interacting with these protocols


denysbutenko (commented):
image

maybe we could use some string with placeholders like in the last one?

denysbutenko (commented):
We also need to check Data Protection dashboard to make sure new messages in sync

denysbutenko (approved):
Otherwise looks good :)

PR #6432, opened 2022-02-21
NAS-114872 / None / NAS-114872: Hide expandable rows from DOM to improve performance

denysbutenko:
Changes:

- Refactor `user-details-row` component
- Add `ix-detail-row` directive to reduce boilerplate code
- One row can be expanded at the same time
- One row is present in DOM when it's expanded
- Added unit tests

Testing:
Open Users page and try to expand rows (click or expand button)


src/app/modules/ix-tables/directives/ix-detail-row.directive.ts
```diff
@@ -0,0 +1,67 @@
...
+      this.row = value;
+    }
+  }
+
+  get ixDetailRow(): T {
+    return this.row;
+  }
+
+  // eslint-disable-next-line @angular-eslint/no-input-rename
+  @Input('ixDetailRowTpl')
```
undsoft, line 32: Nothing wrong with `ixDetailRowTemplate`

undsoft (changes requested):
I liked the idea of having a separate component for user details. `UserListComponent` is already kinda large with a lot of concerns, so anything we can do to make it smaller is good.
Even with the current approach you could still have a details component and use it within:
```
<ng-template #expandedRowTemplate>
  <user-details-row>
```

Current approach is not type-safe.
I wonder if we can do a structural directive instead, something like a `*matNoDataRow`, but keep it type-safe like `*matCellDef`.

undsoft:
Also, is there any way `when` property in `MatRowDef` can help us?

bvasilenko (approved):
Works and looks good to me.

undsoft (approved):
I'm assuming `when` property wasn't too useful?

In any case, looks good.

PR #6433, opened 2022-02-21
NAS-112662 / None / NAS-112662: Added keyboard state for buttons and menu items.

Raz-Dva:
Check keyboard state for buttons, links and menu items.
Check if any other controls in the form need focus states.
This should work for different color themes.


RehanY147:
tab issue

RehanY147:
@undsoft ping

undsoft:
@Raz-Dva 
Let's try to fix Rehan's suggestion. If it's too hard, please create another ticket for yourself.

Raz-Dva:
Ticket "Fix tab order for sub menu" created 
`https://jira.ixsystems.com/browse/NAS-115062`

src/app/components/common/navigation/navigation.component.scss
```diff
@@ -23,4 +23,9 @@
   .nav-section-divider {
     margin: 0 -24px;
   }
+
+  .sidenav-link:focus-visible {
+    box-shadow: 0 0 0 1px var(--alt-fg2) inset;
```
RehanY147, line 28: Should be a bit more subtle. Try `0.5px` here and all other places like this.

src/assets/styles/other/_tn-styles.scss
```diff
@@ -1558,4 +1559,9 @@ $primary-dark: darken(map-get($md-primary, 500), 8%);
       visibility: hidden;
     }
   }
+
+  // focus state styles for mat-form-field
+  .mat-form-field.mat-focused:focus-within {
+    outline: 1px solid var(--alt-fg2);
```
RehanY147, line 1565: This should look the same as other field's focus states
image
image

RehanY147 (changes requested):

src/app/components/common/navigation/navigation.component.scss
Raz-Dva: Unfortunately 0.5px does not work correctly, because A pixel is the smallest unit value to render something.

RehanY147: Please review this
https://stackoverflow.com/a/45522942/6664821

RehanY147:
This doesn't work for white heavy themes. The focus color looks ugly.
image

RehanY147 (changes requested):
The focus color doesn't work for white heavy themes.

src/app/modules/common/layouts/admin-layout/admin-layout.component.scss
```diff
@@ -58,9 +59,17 @@
   top: 13px;
 }
 
+.logo:focus-visible,
+.logo-text:focus-visible {
+  background: rgba(120, 120, 120, 0.2);
+  box-shadow: 0 0 0 1px rgb(255, 255, 255, 0.3) inset;
```
RehanY147, line 65: We should store these values in variables so they're are easily changeable. If I want to change the focus color of a theme, I should just have to change the value of that css variable. See how `primary` `accent` and other colors are declared project wide for different themes.

src/app/modules/common/layouts/admin-layout/admin-layout.component.scss
```diff
@@ -58,6 +59,13 @@
   top: 13px;
 }
 
+.logo:focus-visible,
+.logo-text:focus-visible {
+  background: rgba(120, 120, 120, 0.2) !important;
```
RehanY147: Avoid using `!important` so much if possible. It can be used when nothing else works. Try to find out what's stopping the style from getting applied in the browser inspector maybe.

RehanY147 (changes requested):

RehanY147 (approved):
Looks good. Nice job!

PR #6434, opened 2022-02-21
NAS-114718 / 13.0 / NAS-114718: Fix for Pool Status -> Edit

bvasilenko:
## Testing

To see **Pool Status** page,

1. Storage -> Pools

2. Click image next to the pool's name, and choose **Status** from the menu

image

3. When **(Context menu next to one of the disks) -> Edit** is clicked,

**Expected result**: editing form is opened.

**Actual result**: error in console. form won't open,

```
[Error] ERROR
TypeError: undefined is not an object (evaluating 'e[0].identifier')
```

## Summary of changes

Error was caused by incorrect query sent over WS, which resulted in an empty result:

```
"disk.query"
      [
        "devname",
        "=",
        "sdc2"    <--- incorrect. should be "sdc"
      ]
```

After looking into the code, I've found out that in 13 data bindings for device names were incorrectly calculated on UI side.
I've confirmed this by comparing with 22, where **Edit** action worked correctly. Example: 
   - in 13 name of one of my disks was "sdc2" , which is incorrect device name
      image

   - in 22 the same disk had a correct device name "sdc"
     image

So, in this PR I've synced data displaying logic from 22 to 13.

### DIFF
--- a/src/app/pages/storage/volumes/volume-status/volume-status.component.ts
+++ b/src/app/pages/storage/volumes/volume-status/volume-status.component.ts
@@ -524,7 +524,7 @@ export class VolumeStatusComponent implements OnInit {
       stats = data.stats;
     }
     if (data.type && data.type != 'DISK') {
-      data.name = data.type;
+      data.disk = data.type;
     }
     // use path as the device name if the device name is null
     if (!data.device || data.device == null) {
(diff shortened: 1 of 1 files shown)


denysbutenko (approved):
Before             |  After
:-------------------------:|:-------------------------:
image  |  image

Looks good and works well.

PR #6436, opened 2022-02-22
NAS-114857: widget buttons alignment

Raz-Dva:
Check icon buttons in widgets


dszidi (commented):
This looks better now that the icon is centered! However there isn't supposed to be a background color to these mat icon buttons. It's supposed to be transparent.

`.mat-icon-button` needs a rule `background: transparent` but another generic button rule elsewhere is overriding it with an !important. It would be great if you could get that collision resolved and the background transparent again please.


image


image

dszidi (approved):
LGTM

image

PR #6438, opened 2022-02-22
NAS-113464 / None / NAS-113464: Added service for default empty configs

RehanY147:
Less boiler code for future use


src/app/modules/ix-tables/services/ix-empty.service.ts
```diff
@@ -0,0 +1,38 @@
+import { Injectable } from '@angular/core';
+import { TranslateService } from '@ngx-translate/core';
+import { EmptyConfig, EmptyType } from 'app/modules/entity/entity-empty/entity-empty.component';
+
+@Injectable()
+export class IxEmptyService {
```
undsoft, line 6: No need for `Ix` prefix.

PR #6446, opened 2022-02-25
NAS-114915 / 22.12 / Option to global set SED password missing from Advanced

AlexKarpov98:
Testing:

System -> Advanced
Screenshot_22


src/app/pages/system/advanced/advanced-settings.component.ts
```diff
@@ -490,6 +506,9 @@ export class AdvancedSettingsComponent implements OnInit {
     if ([AdvancedCardId.Kernel].includes(name)) {
```
undsoft, line 506: Or just add an array element here.

undsoft (changes requested):
1. Ticket's fix version is set to SCALE-22.02.1. Please retarget PR to stable/angelfish.
2. Shouldn't there be some indication when password is set?

AlexKarpov98:
Targeted to stable/angelfish https://github.com/truenas/webui/pull/6454

AlexKarpov98:
> 2. Shouldn't there be some indication when password is set?

How can confirm if a password was set?
Advanced Config does not let to know it.

PR #6442, opened 2022-02-24
NAS-114950 / 22.12 / Rename "Disable FreeNAS cache" in AD form to "Disable AD User / Group Cache"

AlexKarpov98:
Testing:

Credentials -> Directory Service -> Advanced Settings -> Idmap -> Click Add -> Active Directory -> Advanced Options 

Renamed `Disable FreeNAS cache` to `Disable AD User / Group Cache`

PR #6450, opened 2022-02-27
NAS-114950 / 22.02 / Rename "Disable FreeNAS cache" in AD form to "Disable AD User / Group Cache"

AlexKarpov98:
Re-targeted from https://github.com/truenas/webui/pull/6442


undsoft (approved):
Approving previously approved PR.

PR #6441, opened 2022-02-24
NAS-114881 / 22.12 / Remove Disable Hardware Offload checkbox from webUI

AlexKarpov98:
Testing:

Network -> Interfaces -> Add/Edit interface

Removed `Disable Hardware Offload` Checkbox

PR #6452, opened 2022-02-27
NAS-114881 / 22.02 / Remove Disable Hardware Offload checkbox from webUI

AlexKarpov98:
Re-targeted from https://github.com/truenas/webui/pull/6441


undsoft (approved):
Approving because previous PR was approved.

PR #6443, opened 2022-02-24
NAS-114769 / 22.12 / Changing Kerberos Principal from nothing to "---" breaks AD form

AlexKarpov98:
Testing:

Credentials -> Directory Service -> Advanced Settings -> Idmap -> Click Add -> Active Directory -> Advanced Options -> `Kerberos Principal` select field

Ignored `null_value`.

PR #6453, opened 2022-02-27
NAS-114769 / 22.02 / Changing Kerberos Principal from nothing to "---" breaks AD form

AlexKarpov98:
Re-targeted from https://github.com/truenas/webui/pull/6443


undsoft (approved):
Approved, because previous PR was approved.

PR #6454, opened 2022-02-27
NAS-114915 / 22.02 / Option to global set SED password missing from Advanced

AlexKarpov98:
Re-targeted from https://github.com/truenas/webui/pull/6446


undsoft (approved):
Form looks good. I will add a ticket to have something on middleware to indicate that sed password has been set.

PR #6460, opened 2022-03-02
NAS-114911 / 22.12 / Add a new option in network bridge sidebar

AlexKarpov98:
Testing:

Network -> Interfaces ->  Add Bridge Interface

Added STP option as checkbox.

image


bvasilenko (commented):
I noticed that validation message is not clear enough. Can we override the text to make it more clear to the user?

image

bvasilenko:
I've also noticed that it's impossible to save changes while the STP checkbox unchecked.

bvasilenko:
And last thing I discovered, is that the parameters  of `interface.create` query is being sent without `stp` field being populated at all (no `true` nor `false` )

AlexKarpov98:
> I noticed that validation message is not clear enough. Can we override the text to make it more clear to the user?
> 
> image

Please confirm testing is on the version 22.12 with update https://jira.ixsystems.com/browse/NAS-113826 (added a new option STP). - there isn't validation message such.

AlexKarpov98:
> And last thing I discovered, is that the parameters of `interface.create` query is being sent without `stp` field being populated at all (no `true` nor `false` ) . Not sure if that's by design or not.

`stp: false` would be added to payload in the only case Interface type is Bridge and unchecked STP.
Screenshot_27

AlexKarpov98:
> I've also noticed that it's impossible to save changes while the STP checkbox unchecked. Not sure if that's by design or not.

Waiting for reply from backend developer
image

undsoft:
Let's try to do something with this PR. If backend issue still persists, please create a backend ticket and link to our ticket.

AlexKarpov98:
Tested with the latest vm (TrueNAS-SCALE-22.12-MASTER-20220330-015452.iso)
It works to update `stp` option.

bvasilenko (approved):
Looks good and works for me.
Tested that UI sends proper value of `stp` to the back-end.

PR #6461, opened 2022-03-02
NAS-115089 / 22.12 / NAS-115089: Removing toggle built-in users and groups + cleanup

undsoft:
Testing: on Users and Groups pages there is no longer a warning about built-in users. Instead there is just a button saying Show/Hide built-in users.

Known issue: there is no loading indication in some cases when button is pressed.


denysbutenko (approved):
Looks good and works well except unit tests

denysbutenko (changes requested):
While testing the latest changes noticed a difference order of items on toolbar:
image
image

denysbutenko (approved):
Looks good!

PR #6467, opened 2022-03-03
NAS-114730 / 22.12 / Incorrect interface behavior during tab switching

AlexKarpov98:
Testing:

Apps -> Switching between Available Applications Tab and others
Prevented repeated filtering


bvasilenko (approved):
Tested and works well.

PR #6477, opened 2022-03-04
NAS-115136 / 13.0 / Replace deprecated dependency.

dszidi:
To test simply make sure UI compiles and no new errors are present.


undsoft (approved):
Don't see any obvious regressions.

PR #6481, opened 2022-03-05
NAS-115134 / None / NAS-115124: Making it more clear how cron rules work

undsoft:
Cron has a special quirk in that when both day of month and day of week have something besides *, it will use these two conditions as OR.
I.e. `0 0 1 * MON` is `every first day of the month OR Monday` and not `every first day of the month that is Monday`.

This PR makes it more clear in scheduler UI.
You can test it in Advanced settings -> CRON Add -> Schedule custom.
Scheduler in general requires regression testing.


denysbutenko (approved):
image

Looks good

PR #6485, opened 2022-03-07
NAS-115162 / None / NAS-115162: Replacing old form-scheduler with new implementation

undsoft:
Testing: test any of the forms that rely on old scheduler (search for type: 'scheduler') in code.
Also pay attention to a special case. In some forms, such as Data Protection -> Periodic Snapshot Task, when you select hourly, there are custom Begin and End fields. These fields will affect what's going to be shown in examples in scheduler modal when you select Custom.


dszidi (commented):
Tested all the slide out forms on the Data Protection dashboard. The schedule preview works as expected and I did not notice any regressions. There was one regression on the SMART Tests form. The save button on the form itself looks like it doesn't have style rules applied

image


Other than that, this looks good to me

undsoft:
Fixed.

dszidi (approved):
Looks good

image

PR #6486, opened 2022-03-07
NAS-114597 / 22.12 / webUI network interface icons not working

AlexKarpov98:
Implemented realtime Network interface link state by updated API `reporting.realtime`


bvasilenko (approved):
Tested with 1 network interface so far. Looks good to me

PR #6491, opened 2022-03-08
NAS-115181 / 22.12 / NAS-115181: Added missing module

RehanY147:
Go to System -> Services and previously the column "Start Automatically" would have issues showing checkboxes. You would get the "No value accessor for control" error. This PR should fix that.


undsoft (approved):
The checkboxes are there. 

Снимок экрана 2022-03-08 в 10 47 08

PR #6493, opened 2022-03-08
NAS-115186 / 22.12 / NAS-115186: Added missing module

RehanY147:
The close icon on tooltips wasn't showing due to the missing module. That should be fixed now.


denysbutenko (approved):
Looks good

PR #6494, opened 2022-03-08
NAS-115187: Fixed custom configs for different cases

RehanY147:
The ticket has a description of the issue. The fix aims to enable setting custom config from the view component instead of relying on EntityTableComponent.


AlexKarpov98 (approved):
LGTM! 👍

PR #6497, opened 2022-03-09
NAS-115191 / 22.02.1 / "Network / Global Configuration / Host Name Database" is inconvenient

AlexKarpov98:
Testing:

Nextwork -> Gloabal Configuration -> Host Name Database
Try to paste the following text

192.168.0.220 tn1
192.168.0.228 tn2
192.168.0.225 tn3

Split it to set the chip list value if the paste value includes line break.


dszidi (approved):
LGTM

image

PR #6499, opened 2022-03-09
NAS-114752 / 22.12 / Improve loading experience/performance in ix-tables

denysbutenko:
Changes:

- Use ngrx store for users
- Update unit tests
- Sync changes with snapshots store

TODO:

WebSocket Events doesn't dispatch changes for `user.query`. I need to sync users state with new users and updated
- Dispatch `userChanged()` and `userAdded()` actions from User Form. It requires refactoring User Form to use ix-forms.

### DIFF
--- a/src/app/pages/account/users/store/user.effects.ts
+++ b/src/app/pages/account/users/store/user.effects.ts
@@ -1,0 +1,78 @@
+import { Injectable } from '@angular/core';
+import { Actions, createEffect, ofType } from '@ngrx/effects';
+import { Store } from '@ngrx/store';
+import { TranslateService } from '@ngx-translate/core';
+import { of } from 'rxjs';
+import {
+  catchError, filter, map, switchMap,
+} from 'rxjs/operators';
+import { ApiEventMessage } from 'app/enums/api-event-message.enum';
+import { QueryParams } from 'app/interfaces/query-api.interface';
+import { User } from 'app/interfaces/user.interface';
+import {
+  userPageEntered, usersLoaded, usersNotLoaded, userAdded, userChanged, userRemoved,
+} from 'app/pages/account/users/store/user.actions';
+import { WebSocketService } from 'app/services';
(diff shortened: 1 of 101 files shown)


src/app/pages/account/users/store/user.effects.ts
```diff
@@ -0,0 +1,78 @@
...
+@Injectable()
+export class UserEffects {
+  loadUsers$ = createEffect(() => this.actions$.pipe(
+    ofType(userPageEntered, builtinUsersToggled),
+    switchMap(() => this.store$.pipe(waitForPreferences)),
+    switchMap((preferences) => {
+      let params: QueryParams<User> = [];
+      if (preferences.hideBuiltinUsers) {
+        // TODO: Fix any
+        params = [[['OR', [['builtin', '=', false], ['username', '=', 'root']]]]] as any;
```
denysbutenko, line 29: I'm sorry about that.
If you know how to fix type checking to make this expression valid please let me know.
image

undsoft: From standpoint of typings, there are too many braces.
```
[['OR', [['builtin', '=', false], ['username', '=', 'root']]]];
```

denysbutenko: > From standpoint of typings, there are too many braces.
> 
> ```
> [['OR', [['builtin', '=', false], ['username', '=', 'root']]]];
> ```

Yeah, it passes typescript validation but middleware throwing invalid syntax error in this case

undsoft: Then there is an issue in typings, either in QueryParams or ApiDirectory. You can try to fix it or forcecast with `as unknown as QueryParams<User>` leaving TODO comment.

undsoft (approved):
Code looks okay.
From the point of functionality, editing a user does not update the table, but it works in next PR, so I'm approving based on that.

PR #6501, opened 2022-03-10
NAS-113755 / 22.12 / Fix handleWsError

AlexKarpov98:
If `res.extra` is not array, this error would be handled without any error report dialog.
image
(tested with mock data)

Added the case for the exception, opened a report dialog.


undsoft (approved):
Should be okay.

PR #6502, opened 2022-03-10
NAS-115224 / 22.12 / NAS-115224: Fixed drag

RehanY147:
The help widget is draggable all the time. Even when the dashboard is not being re-ordered. You can click on it and drag it around. This PR fixes that.


undsoft (approved):
Let's retarget this to stable/angelfish.

PR #6503, opened 2022-03-10
NAS-115224 / 22.02.1 / NAS-115224: Fixed drag

RehanY147:


dszidi (approved):
The help widget is no longer draggable when not in reorder mode. Nice catch

PR #6508, opened 2022-03-12
NAS-114089 / 22.12 / NAS-114089: Fixed overlap and scroll issue

RehanY147:
The virtualization table had issues of overlap columns by first and last columns due to them being `sticky`. Check responsive of the table with different column configurations and responsiveness on different screen sizes. Also, the border on the left column for scroll shouldn't touch the next column (usually the toggle) as complained about in the ticket comments.


bvasilenko:
Tested, and looks good now

image

bvasilenko (approved):

PR #6511, opened 2022-03-13
NAS-115146: Refactor VolumeImportWizardComponent to ix-forms

bvasilenko:
Testing:

1. Storage -> Export/Disconnect your pool. Make sure you don't delete the data. Just disconnect.
2. Storage -> Import -> Select your disconnected pool


src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
```diff
@@ -1,115 +1,72 @@
...
-    dialogRef.componentInstance.setCall('pool.import_find');
-    dialogRef.componentInstance.submit();
-    dialogRef.componentInstance.success.pipe(
-      untilDestroyed(this),
-    ).subscribe((poolFindResult: Job<PoolFindResult[]>) => {
-      dialogRef.close(false);
-      if (!poolFindResult?.result) {
-        return;
+  ngOnInit(): void {
+    this.ws.job('pool.import_find').pipe(untilDestroyed(this)).subscribe((res) => {
```
RehanY147, line 56: We want to avoid naming variable `res`

src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
```diff
@@ -118,48 +75,17 @@ export class VolumeImportWizardComponent implements WizardConfiguration {
...
-            this.router.navigate(['/storage', 'id', this.pool, 'dataset', 'unlock', this.pool]);
-          });
-        }, (err) => {
-          this.modalService.closeSlideIn();
-          this.modalService.refreshTable();
-          new EntityUtils().handleWsError(this, err, this.dialogService);
-        });
+      this.isFormLoading = false;
+      this.slideInService.close();
+      this.modalService.refreshTable();
```
RehanY147, line 84: We don't need to use this. The `IxSlideInService` has an `onClose$` subject that you can subscribe to in the table component and get notified when the slide in panel closes.

src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+  beforeEach(() => {
+    spectator = createComponent();
+    loader = TestbedHarnessEnvironment.loader(spectator.fixture);
+    ws = spectator.inject(WebSocketService);
+  });
+
+  it('loads and shows the current list of pools to import when form is opened', (done) => {
+    spectator.component.pool.options.subscribe((options: Option[]) => {
+      expect(ws.job).toHaveBeenCalledWith('pool.import_find');
+      expect(options).toEqual([
```
RehanY147, line 78: This is basically checking for a static variable twice since our mockup is returning this result. Instead maybe we should test the template element and see if that got the options.

RehanY147 (changes requested):

src/app/pages/storage/volumes/volume-import-wizard/volume-import-wizard.component.ts
bvasilenko: I tried the approach you've proposed, but it turns out that the data in the table is updated every time after the sidebar is closed. I believe that this is not entirely correct, because there might not be a new pool import.

So I'd like to keep it as it is.

RehanY147 (approved):
Looks good

PR #6512, opened 2022-03-14
NAS-115269: Cleaning up translation strings

undsoft:
Testing: Maybe test some of the updated strings in UI.


src/app/helptext/directory-service/active-directory.ts
```diff
@@ -93,7 +93,7 @@ export default {
   activedirectory_kerberos_principal_tooltip: T('Select the location of the principal in the \
  keytab created in <b>Directory Services > Kerberos Keytabs</b>.'),
 
-  computer_account_OU_name: T('createcomputer'),
+  computer_account_OU_name: 'createcomputer',
```
dszidi, line 96: Shouldn't this also be wrapped in `T(...)` ?

dszidi: It seems this is not actually used in any help text. LGTM

dszidi (approved):

PR #6515, opened 2022-03-15
NAS-114605 / 22.12 / Added IxPermissionsComponent

RehanY147:
Go to System Settings -> General -> GUI Settings

Test the IxPermissions input in the form. Since the actual form submission probably won't succeed. In order to see changed values show on form start, you will have to update default value in code. Also test validation on the new component.
Also added testing.

Changes in this form must be removed before merging!!


src/app/modules/ix-forms/components/ix-permissions/ix-permissions.component.scss
```diff
@@ -0,0 +1,51 @@
...
+  }
+}
+
+.label-container {
+  margin-bottom: 1px;
+  margin-left: 1px;
+}
+
+.label {
+  font-size: x-small;
```
denysbutenko, line 44: it makes all labels smaller than others in a form:
image
image

RehanY147: It looks consistent with the other inputs on the new ix-forms.
image

AlexKarpov98: Screenshot_40

Looks label is small . I think it's better to make same font-size with above checkbox label `Show console Messages` considering new type of input.

src/app/pages/system/general-settings/gui-form/gui-form.component.ts
```diff
@@ -52,6 +52,7 @@ export class GuiFormComponent {
     crash_reporting: [false, [Validators.required]],
     usage_collection: [false, [Validators.required]],
     ui_consolemsg: [false, [Validators.required]],
+    ui_perms: ['235', Validators.required],
```
AlexKarpov98, line 55: Seems `Validators.required` does not work.

RehanY147:
@AlexKarpov98 I'm not sure what `required` would mean for this input. At the very least, the input will give the value `'000'`. So, the input always has a value.

denysbutenko (approved):
Tested in User Form. Works well 👍🏻

AlexKarpov98 (approved):
LGTM! 👍

denysbutenko:
https://github.com/truenas/webui/pull/6526 - you can test ix-permissions in this PR if you want

PR #6519, opened 2022-03-16
NAS-115309 / 22.02.1 / No error is shown if update servers cannot be reached

AlexKarpov98:
Opened dialog if TrueNAS was unable to reach update servers

image


RehanY147 (approved):
LGTM

PR #6520, opened 2022-03-16
NAS-115077 / 22.02.1 / NAS-115077: Show close button after abort

RehanY147:
Go to the Add new cloud sync task form and after entering the values (Google Photos preferably), do a dry run. You should see the abort button and after abort, the dialog should be closable.


bvasilenko (approved):
Looks good.
The dialog is closable now:

   image

PR #6521, opened 2022-03-17
NAS-115256 / 22.12 / Dashboard widget is broken after pool is removed

AlexKarpov98:
Testing:

1. Create a pool.
2. Go to dashboard, click Configure and select widget for this pool.
3. Delete pool.
4. Go to dashboard.

Fixed broken pool widget


undsoft (approved):
It works, but it should also/instead be targeted for Angelfish.

PR #6526, opened 2022-03-18
NAS-115183 / 22.12 / Refactoring User Form

denysbutenko:
Changes:

- Refactor User Form to ix forms
- Add unit tests
- Add two-column form styles for the wide layout
- Add method to check the disabled state of inputs for `IxFormHarness`

Testing:
- Try to add a new user
- Try to edit a user
- Try to delete a user

Known issue: Some tests are failing, keep trying to fix them.


src/app/pages/account/users/user-form/user-form.component.ts
```diff
@@ -1,545 +1,352 @@
...
+    password: ['', [Validators.required, Validators.pattern(UserService.passwordPattern)]],
+    password_conf: ['', [
+      Validators.required,
+      this.validatorsService.withMessage(
+        matchOtherValidator('password'),
+        this.translate.instant('Password and confirmation should match.'),
+      ),
+    ]],
+    password_edit: ['', [Validators.pattern(UserService.passwordPattern)]],
+    password_conf_edit: ['', [
```
undsoft, line 66: Is there a reason for another set of fields, as opposed to a conditional required validator?

src/app/pages/account/users/store/user.effects.ts
```diff
@@ -41,23 +41,7 @@ export class UserEffects {
...
-            case ApiEventMessage.Added:
-              return userAdded({ user: event.fields });
-            case ApiEventMessage.Changed:
-              return userChanged({ user: event.fields });
-          }
-        }),
-      );
-    }),
-  ));
+  // userAdded() and userChanged() are dispatched from the User Form
```
undsoft, line 44: I think a slight benefit to the old system was that if for whatever reason new user is added in new tab / by another user (if even possible), this would have updated it.

src/app/pages/account/users/user-form/user-form.component.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+      expect(disabled).toEqual(expect.objectContaining({
+        'Home Directory Permissions': true,
+        'Home Directory': true,
+        'Primary Group': true,
+        Username: true,
+      }));
+    });
+  });
+
+  it('checks download ssh public key button', async () => {
```
undsoft, line 262: Here and in other places I think it would better to describe test cases from the standpoint of how system works, i.e. `it('only shows download ssh key button when ssh key is filled in` or something similar.

undsoft (changes requested):
1. When creating a user, I get `null is not allowed` on `auxillary groups` for some reason.
2. I also get null is not allowed on Primary Group when `Create New Primary Group` is on.
3. Not too critical, but feels a bit weird that  `Download key` appears as I'm adding key myself.
4. Old form doesn't allow to create a new primary group on edit.
5. Probably middleware related: if I have a user without email, enable Microsoft Account ,save and try to edit again, I will get validation error that email is required. This will happen on second save only.

src/app/pages/account/users/store/user.effects.ts
denysbutenko: I completely agree. The reason is `ws.subscribe()/ws.sub()` for `user.create|user.update|user.query` endpoints does not produce websocket messages on adding/editing user.
Websocket output right after form submitted:
image
Based on that, I getting `id` in response of `user.query`, then making a call `user.query` and dispatch `userAdded()` or `userChanged()`.

src/app/pages/account/users/user-form/user-form.component.ts
```diff
@@ -1,545 +1,352 @@
...
+      this.validatorsService.withMessage(
+        matchOtherValidator('password_edit'),
+        this.translate.instant('New password and confirmation should match.'),
+      ),
+    ]],
+    uid: [null as number, [Validators.required, numberValidator()]],
+    group: [null as number],
+    group_create: [true],
+    groups: [[] as number[]],
+    home: ['/nonexistent', []],
```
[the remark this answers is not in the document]
denysbutenko: Yes, the old form uses this as the default value.

image

https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/plugins/account.py#L108-L110

src/app/pages/account/users/user-form/user-form.component.ts
```diff
@@ -1,545 +1,352 @@
...
-  private group: FormSelectConfig;
-  private groups: FormSelectConfig;
-  private password_disabled: AbstractControl;
+  readonly groupOptions$ = this.ws.call('group.query').pipe(
+    map((groups) => groups.map((group) => ({ label: group.group, value: group.id }))),
+  );
+  readonly shellOptions$ = this.ws.call('user.shell_choices').pipe(choicesToOptions());
+  readonly treeNodeProvider = this.filesystemService.getFilesystemNodeProvider();
+  readonly shellProvider = new SimpleAsyncComboboxProvider(this.shellOptions$);
+  readonly groupProvider = new SimpleAsyncComboboxProvider(this.groupOptions$);
```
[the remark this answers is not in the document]
denysbutenko: I'm not sure. The `GroupProvider` uses `group.group` as option value but in UserForm it requires `group.id`. And we still need for `groupOptions$` for `Auxiliary Groups`

src/app/pages/account/users/user-form/user-form.component.ts
denysbutenko: Fixed 👍🏻

denysbutenko:
> Not too critical, but feels a bit weird that Download key appears as I'm adding key myself.
We can hide the button if `isNew=true`, would it help?

undsoft (approved):
Looks good.
With regards to download key, I guess adding the `isNew` condition would be okay.

PR #6529, opened 2022-03-18
NAS-115330 / 22.12 / NAS-115330: Refactoring POSIX and NFS permissions forms

undsoft:
Testing:
1. Create two datasets, one with Acl Type = Posix, another with Acl Type = NFS.
2. In Storage page, click View Permissions -> Edit -> Set ACL.
3. Test the form and acl editor.
4. Do the same for another dataset.

Known issues: labels on some of the components are of different sizes. I will fix this in future PRs.


RehanY147 (changes requested):
Why not remove the `dynamicFields` from the parent container component? It's only 4 more fields looks like. Otherwise, the PR looks good. The changes are persistent and show up on edits. the form works ok.

RehanY147 (approved):
Looks good

PR #6530, opened 2022-03-18
NAS-114658 / 22.12 / NAS-114658: Improving coverage reporting

undsoft:
This is mostly an experiment to check if coverage reporting works correctly.


dszidi (approved):
LGTM

PR #6539, opened 2022-03-21
NAS-115393 / 22.12 / NAS-115393: Fixes console errors in tests

undsoft:
Some errors in tests are reported in console, but do not prevent tests from passing. This fixes those errors.


denysbutenko (approved):
Looks good

PR #6542, opened 2022-03-21
NAS-115137 / 22.12 / New method "webdav.cert_choices". Get valid certs for webdav service

Raz-Dva:
Replaced 'certificate.query' API endpoint with "webdav.cert_choices" in Services -> WebDAV form
Getting, updating and saving of certificates in the form.


bvasilenko (approved):
LGTM

PR #6544, opened 2022-03-22
NAS-114069 / 22.12 / NAS-114069: "hidden" class removed

Raz-Dva:
Removed class 'hidden' which had no effect.
Check rendering and efficiency "pool operations" button

image


AlexKarpov98 (approved):
LGTM! 👍

PR #6545, opened 2022-03-22
NAS-115358 / 22.02.1 / Updating french translation

Dremor:


Dremor:
Closed to avoid unnecessary CI builds. Is there a way to disable them while the PR is in draft state?

PR #6547, opened 2022-03-22
Refactor iSCSI portal form to ix forms

AlexKarpov98:
Testing: 

Shares -> iSCSI -> Configure -> Portals -> Add/Edit

Created two component `ix-list` & `ix-list-item`
Refactored `PortalFormComponent` with ix-form


bvasilenko (approved):
Works and looks good to me

PR #6549, opened 2022-03-23
NAS-115373 / 22.12 / NAS-115373: Added checkbox and calls to show data

RehanY147:
To test, try to delete an app after installation.
1. The checkbox to delete unused images should be there
2. If marked checked, you should see another dialog showing the images that can't be deleted because they are still under use.
3. Clicking OK on this dialog should delete the app with the images
4. If not checked, the app should be deleted but the flag `delete_unused_images` should be set to false in the middleware call


src/app/services/dialog.service.ts
```diff
@@ -77,13 +77,16 @@ export class DialogService {
       dialogRef.componentInstance.method = options.method;
       dialogRef.componentInstance.switchSelectionEmitter.pipe(untilDestroyed(this)).subscribe((selection: boolean) => {
         const data = options.data;
-        if (selection) {
+        if (selection && data[0]) {
```
AlexKarpov98, line 80: Looks weird. It's good to revert it.

AlexKarpov98:
Waiting for success testing

src/app/services/dialog.service.ts
RehanY147: Without the check `data[0]`, I have to pass an object in the dialog config and introduce new ifs into the service with special code just for this PR. This check helps me avoid that scenario. I can just subscribe to the selection change emitter and do my own thing.

src/app/services/dialog.service.ts
```diff
@@ -77,13 +77,16 @@ export class DialogService {
...
           if (data[0] && data[0].hasOwnProperty('reboot')) {
             data[0].reboot = !data[0].reboot;
           }
           if (data[0] && data[0].hasOwnProperty('overcommit')) {
             data[0].overcommit = !data[0].overcommit;
           }
+          if (data[0] && data[0].hasOwnProperty('delete_unused_images')) {
```
AlexKarpov98, line 87: If so, remove data[0] from the next line.

AlexKarpov98 (approved):

RehanY147:
Needs more work. Can't merge yet.

PR #6552, opened 2022-03-23
NAS-115403 / 22.12 / On a encrypted pool with passphrase, giving different password does not give an error

AlexKarpov98:
Fixed validator
Screenshot_17


bvasilenko (approved):
LGTM 👍

PR #6558, opened 2022-03-24
NAS-115375 / 22.12 / Some translation string do not translate despite having a translation string in the translation file

AlexKarpov98:


bvasilenko (approved):
Code looks good to me

PR #6559, opened 2022-03-25
NAS-115339 / 22.12 / NAS-115339: Refactor iSCSI target form to ix-forms

bvasilenko:
Summary: 
- converted old iSCSI target form into ix-forms
- created new component for iSCSI groups

Testing:
 Shares -> iSCSI -> Configure -> Targets -> Add/Edit.


src/app/pages/sharing/iscsi/target/target-form/target-form.component.ts
```diff
@@ -1,255 +1,182 @@
...
-        this.loader.close();
-        new EntityUtils().handleWsError(this.entityForm, res);
-      },
-    );
+  addGroup(): void {
+    const newIndex = this.groups.length;
+    const newListItem: any = {};
+    this.groupsFromControls.forEach((fc) => {
+      newListItem[fc.name] = fc.default;
+      this.form.addControl(`${fc.name}${this.listPrefix}${newIndex}`, new FormControl(fc.default, fc?.validator));
```
RehanY147, line 169: The `?` in `fc?.validator` doesn't really do anything since we're already accessing `fc` without checking in `fc.default`. Better check should be used if the purpose is to check if `fc.validator` exists. Same in another place above.

RehanY147 (changes requested):
The `ix-list` have no margins and clear separations. If this issue is out of scope for this ticket, let me know.
image

src/app/pages/sharing/iscsi/target/target-form/target-form.component.ts
bvasilenko: > If this issue is out of scope for this ticket
Yes, I suppose this is out of scope. I guess it's also needs fixing in other places where `ix-list` is used. Let's create a separate ticket

bvasilenko: >fc?.validator doesn't really do anything

It wasn't intentional, just a typo. Fixed now!

RehanY147 (approved):
Looks good

PR #6560, opened 2022-03-25
Handle empty state in ix-select

AlexKarpov98:
Screenshot_18

### DIFF
--- a/src/app/helptext/services/components/service-snmp.ts
+++ b/src/app/helptext/services/components/service-snmp.ts
@@ -18,14 +18,13 @@ export default {
   v3_username_tooltip: T('Enter a username to register with this service.'),
   v3_authtype_tooltip: T('Choose an authentication method.'),
   v3_authtype_options: [
-    { label: '---', value: '' }, { label: 'MD5', value: 'MD5' },
+    { label: 'MD5', value: 'MD5' },
     { label: 'SHA', value: 'SHA' },
   ],
 
   v3_password_tooltip: T('Enter a password of at least eight characters.'),
   v3_privproto_tooltip: T('Choose a privacy protocol.'),
   v3_privproto_options: [
-    { label: '---', value: null },
     { label: 'AES', value: 'AES' },
     { label: 'DES', value: 'DES' },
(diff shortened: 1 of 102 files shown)


undsoft (changes requested):
Sorry, the ticket I created was misleading. What I wanted is to allow user to select no option at all or to clear selection in some cases. For example in ServiceFtpComponent there is a certificate field and it's optional - user doesn't have to select any certificates. 
The problem we currently have is that once user selects a certificate, there is no way to unselect it.

What you've already done is good to keep too.

src/app/helptext/services/components/service-snmp.ts
```diff
@@ -18,14 +18,13 @@ export default {
   v3_username_tooltip: T('Enter a username to register with this service.'),
   v3_authtype_tooltip: T('Choose an authentication method.'),
   v3_authtype_options: [
-    { label: '---', value: '' }, { label: 'MD5', value: 'MD5' },
```
undsoft, line 21: I get null is not allowed here.

src/app/helptext/services/components/service-ssh.ts
```diff
@@ -34,7 +34,6 @@ export default {
  href="https://man7.org/linux/man-pages/man3/syslog.3.html"\
  target="_blank">syslog(3)</a> level of the SFTP server.'),
   ssh_sftp_log_level_options: [
-    { label: '', value: '' },
```
undsoft, line 37: Also get null is not allowed in this file.

src/app/pages/storage/snapshots/snapshot-add-form/snapshot-add-form.component.ts
```diff
@@ -125,7 +125,6 @@ export class SnapshotAddFormComponent implements OnInit {
   private getNamingSchemaOptions(): Observable<Option[]> {
```
undsoft, line 125: Please also add [required]=true to Dataset field.

undsoft (approved):
Looks good with some fixes required.

PR #6563, opened 2022-03-28
NAS-115470 / 22.12 / NAS-115470: Disable ms account field for builtin users

denysbutenko:
Related middleware PR - https://github.com/truenas/middleware/pull/8639


undsoft (approved):
The checkbox is disabled for built-in users.

PR #6571, opened 2022-03-30
NAS-115551 / 22.12 / NAS-115551: Extract sign CSR into a separate component

undsoft:
Go to Certificates and create a CSR.
Then create a Certificate Authority.
Then use Certificate Authority to sign a CSR. This can be done in the table and by clicking on a CA.

### DIFF
--- a/src/app/helptext/system/ca.ts
+++ b/src/app/helptext/system/ca.ts
@@ -372,7 +372,6 @@ certificate still approved.'),
   },
 
   sign: {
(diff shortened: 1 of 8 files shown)


dszidi (commented):
The dialog works great! Code lgtm as well. 

I am wondering though, do we want the SignCSR button in the slide out form? The form is for editing the CA. Signing a CSR is an action the CA can take. It is not a property of a CA

image

undsoft:
Fair enough, removed.

PR #6573, opened 2022-03-31
Fix the tooltip text for path field configuring a ddns service

GermanDZ:
Just fix the text used for tooltips, now is showing the same text as in `server`


undsoft (approved):
Thanks for the fix 👍🏻

PR #6574, opened 2022-03-31
NAS-115110 / 13.0 / Added option to remove invalid user quotas

RehanY147:
To test,
1. Create a two new users `user1` and `user2`
2. Set `User Quotas` for a dataset using the `Bulk` option
3. Remove the users from the system
4. Go back to the `User Quotas` page and you should see errors for quotas
5. Check the `Actions` button and click on the `Remove quotas for invalid users` button and follow through. It should remove the erroneous quota entries.
Test the same thing with `Group Quotas`


AlexKarpov98 (approved):
Nice work! 👍
Screenshot_17
Screenshot_18

PR #6575, opened 2022-04-01
Correct wrong translations

qwec01:
Correct zh-hans translation of "take snapshot" to "创建快照" from wrong translation "拍摄快照".


denysbutenko (approved):
Thanks for the contribution!

PR #6577, opened 2022-04-01
NAS-115466 / 22.12 / "Select All" checkbox stays checked after completing operation

AlexKarpov98:
There was already the feature of clearing selection after closed dialog.
Disabled close dialog by clicking backdrop.


RehanY147 (approved):
Looks good

PR #6580, opened 2022-04-04
NAS-115357 / 22.12 / Added logs for replication tasks and cloud sync tasks

RehanY147:
To test, go to the Data Protection dashboard and run cloud sync or replication tasks from there. After starting the task, click on the state button and the dialog should come up with the loading bar. The dialog should also have logs of the task as it progresses. Real time logs. Then, click on the title and go to the cloud sync and replication tasks dedicated pages. Test the same thing there too.


AlexKarpov98 (approved):
LGTM! 👍
Screenshot_12

Please update darker background style of `logs-container` like
Screenshot_14

PR #6588, opened 2022-04-06
NAS-115291: Adding force flag to unlock operation

undsoft:
Testing: 
1. Create a dataset.
2. Edit dataset encryption options and change to password.
3. Lock dataset.
4. Unlock and observe Force flag.


RehanY147 (approved):
See no difference in the actual websocket request whether the force is checked or not.

RehanY147 (changes requested):
Mentioned above

undsoft:
Fixed, please take a look again.

PR #6591, opened 2022-04-07
NAS-115677 / 22.12 / failover.disabled.reasons requires webUI changes

AlexKarpov98:
Replaced API `failover.disabled_reasons` to `failover.disabled.reasons`

Replaced key `DISAGREE_CARP` to `DISAGREE_VIP`

Replaced return string `Nodes CARP states do not agree.` to `Nodes Virtual IP states do not agree.`

Related middleware PR [8659](https://github.com/truenas/middleware/pull/8659)


src/app/services/system-profile.service.ts
```diff
@@ -48,8 +48,8 @@ export class SystemProfileService extends BaseService {
     });
 
     // HA Status change events
-    this.websocket.subscribe('failover.disabled_reasons').subscribe((res) => {
-      this.updateHa(res.fields.disabled_reasons);
+    this.websocket.call('failover.disabled.reasons').subscribe((res) => {
```
undsoft, line 51: I don't think replacing `subscribe` with `call` is correct here.

undsoft (approved):
Code looks okay, one comment.
Remember to schedule merge with Caleb.

PR #6598, opened 2022-04-07
Network graph: incorrect legend units

AlexKarpov98:
Testing: Reporting -> Network

Formatted Interface legend data and series


undsoft (approved):
Seems to be working okay, though the old version also worked okay for me.

PR #6602, opened 2022-04-09
Nas 115241 Add unit tests to one of the ix-forms components

Raz-Dva:
Added unit test for ix-select component.


src/app/modules/ix-forms/components/ix-select/ix-select.component.ts
```diff
@@ -37,21 +42,18 @@ export class IxSelectComponent implements ControlValueAccessor, OnChanges {
   }
 
   ngOnChanges(): void {
-    if (this.options) {
+    if (!this.options) {
+      this.hasErrorInOptions = true;
+    } else {
       this.opts$ = this.options.pipe(
```
RehanY147, line 48: Set `hasErrorInOptions = false;` here so the options load when correct value is assigned to `this.options`.

RehanY147 (changes requested):
This is the wrong methodology to test these `ix-*` components. We're testing basic simple code instead of checking how these controls behave in the dom when they are setup with certain properties. A good example of what I'm talking about can be seen in this question on StackOverflow https://stackoverflow.com/q/57732352/6664821
The issue the post author was having here is irrelevant. The approach is what I want us to look at.

@undsoft thoughts?

src/app/modules/ix-forms/components/ix-select/ix-select.component.ts
RehanY147: @Raz-Dva Please fix the requested issue

RehanY147 (commented):
Waiting for the fix

src/app/modules/ix-forms/components/ix-select/ix-select.component.ts
Raz-Dva: @[RehanY147](https://github.com/RehanY147) Ok

RehanY147 (approved):
Looks good.

PR #6605, opened 2022-04-11
NAS-115487 / 22.12 / SMB form refactor

RehanY147:
Test the smb form. All the correct warnings should show up. Also test that under the correct conditions, correct service restarts are prompted and executed.


RehanY147:
Working on `ChangeDetectionStrategy` changes

undsoft (changes requested):
Please self-review yourself first.

Try to clean it up to:
- have smaller methods with semantic names
- have methods do one thing at a time and not have unexpected side effects
fix naming of some variables
- restructure methods to return early
- streamline observables via pipes and switches instead of nested subscribes

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -1,652 +1,399 @@
...
-    entityForm.formGroup.controls['path'].valueChanges.pipe(untilDestroyed(this)).subscribe((path) => {
-      const nameControl = entityForm.formGroup.controls['name'];
+    pathFormControl.valueChanges.pipe(debounceTime(50), untilDestroyed(this)).subscribe((path) => {
+      const nameControl = this.form.get('name');
       if (path && !nameControl.value) {
         const name = path.split('/').pop();
         nameControl.setValue(name);
       }
 
       if (!this.stripACLWarningSent) {
```
undsoft, line 166: 1. Bad property name. => `wasStripAclWarningShown` or something.
2. Can return early.
3. Same for res here and in other places.
4. This block and block below look similar. Any way to combine (e.g. with `combineLatest`).

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -1,652 +1,399 @@
...
+    }
+    this.presetFields = [];
+  }
+
+  setSmbShareForEdit(smbShare: SmbShare): void {
+    this.existingSmbShare = smbShare;
+    this.hostsAllowOnLoad = smbShare.hostsallow ? [...smbShare.hostsallow] : [];
+    this.mangle = smbShare.aapl_name_mangling;
+    this.hostsDenyOnLoad = smbShare.hostsdeny ? [...smbShare.hostsdeny] : [];
+    this.isTimeMachineOn = smbShare.timemachine;
```
undsoft, line 211: You save a bunch of fields for values before edit, but you also have `existingSmbShare` that you can use.

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -1,652 +1,399 @@
...
+    this.title = helptextSharingSmb.formTitleEdit;
+    const index = this.namesInUse.findIndex((name) => name === smbShare.name);
+    if (index >= 0) {
+      this.namesInUse.splice(index, 1);
+    }
+    this.form.get('aapl_name_mangling').valueChanges.pipe(untilDestroyed(this)).subscribe((value) => {
+      if (value !== this.mangle && !this.mangleWarningSent) {
```
undsoft, line 218: 1. Looks like this listener belongs somewhere else, like an `ngOnInit`.
2. Probably can be refactored to something like:
```
this.form.get(..).valueChanges.pipe(
  filter(...),
  take(1), // instead of mangleWarningSent
  switchMap(() => this.dialog.confirm(..)),
  untilDestroyed(),
).subscribe();
```

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -1,652 +1,399 @@
...
       }
     });
+    this.form.patchValue(smbShare);
   }
 
-  afpConfirm(): void {
-    const afpControl = this.entityForm.formGroup.controls['afp'];
+  /* If user blurs name field with empty value, try to auto-populate based on path */
+  setEmptyNameFromPath(): void {
+    const nameControl = this.form.get('name');
```
undsoft, line 234: Listener similar to what you have in `ngOnInit`? Can be combined?

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -1,652 +1,399 @@
...
+          return of();
+        }),
+        filter(() => {
+          const sharePath: string = this.form.get('path').value;
+          const homeShare = this.form.get('home').value;
+
+          if (homeShare && this.isNew) {
+            const datasetId = sharePath.replace('/mnt/', '');
+            const poolName = datasetId.split('/')[0];
+            this.router.navigate(['/'].concat(['storage', 'id', poolName, 'dataset', 'acl', datasetId]), { queryParams: { homeShare: true } });
```
undsoft, line 294: filter has a side-effect.

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -1,652 +1,399 @@
...
+      request$ = this.ws.call('sharing.smb.create', [smbShare]);
+    } else {
+      request$ = this.ws.call('sharing.smb.update', [this.existingSmbShare.id, smbShare]);
+    }
+
+    let cifsServiceInstance: Service;
+    request$.pipe(
+      untilDestroyed(this),
+    ).subscribe(() => {
+      this.shouldServiceRestart().pipe(
```
undsoft, line 280: Large body - hard to follow. Maybe extract some of this code into separate methods that would return observables.

src/app/pages/sharing/smb/smb-form/smb-form.component.spec.ts
```diff
@@ -0,0 +1,449 @@
...
+        aapl_name_mangling: false,
+        acl: true,
+        durablehandle: true,
+        shadowcopy: true,
+        streams: true,
+        fsrvp: false,
+        auxsmbconf: '',
+      },
+    },
+    ENHANCED_TIMEMACHINE: {
```
undsoft, line 108: Since this is coming from backend as far as I understand, one should be enough to test our code.

src/app/pages/sharing/smb/smb-form/smb-form.component.spec.ts
```diff
@@ -0,0 +1,449 @@
...
+    ]);
+  });
+
+  it('should have error for duplicate share name', async () => {
+    const nameControl = await loader.getHarness(IxInputHarness.with({ label: 'Name' }));
+    await nameControl.setValue('ds222');
+    expect(await nameControl.getErrorText()).toEqual('The name "ds222" is already in use.');
+  });
+
+  it('when a preset is selected, the relevent fields should be impacted', async () => {
```
undsoft, line 218: `releavant`

src/app/pages/sharing/smb/smb-form/smb-form.component.spec.ts
```diff
@@ -0,0 +1,449 @@
...
+
+    const advancedButton = await loader.getHarness(MatButtonHarness.with({ text: 'Advanced Options' }));
+    await advancedButton.click();
+
+    const aaplNameManglingCheckbox = await loader.getHarness(
+      IxCheckboxHarness.with({ label: formLabels.aapl_name_mangling }),
+    );
+
+    if (existingShare.aapl_name_mangling) {
+      await (await aaplNameManglingCheckbox.getMatCheckboxHarness()).uncheck();
```
undsoft, line 290: `aaplNameManglingCheckbox.setValue(false)` should also work

src/app/pages/sharing/smb/smb-form/smb-form.component.spec.ts
```diff
@@ -0,0 +1,449 @@
...
+  it('shows values of existing share when editing', async () => {
+    spectator.component.setSmbShareForEdit(existingShare);
+
+    const advancedButton = await loader.getHarness(MatButtonHarness.with({ text: 'Advanced Options' }));
+    await advancedButton.click();
+
+    const values = await form.getValues();
+
+    const existingShareWithLabels: { [key: string]: any } = {};
+    for (const key in existingShare) {
```
undsoft, line 268: This is my personal opinion, so no change necessary for this PR – I find it easier to read when tests are less programmatic and more close to what user would see, even if it's more lines, i.e. just showing labels and their values.

undsoft (changes requested):

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
RehanY147: Since, we're changing the value of the `name` control, setting up a subscription to valuechanges triggers recursion issues here and inside `ix-input`

undsoft: When you patchValue in form there is an optional argument that you can set no to trigger updates. emitEvent: false

undsoft (changes requested):

src/app/pages/sharing/components/shares-dashboard/shares-dashboard.component.ts
```diff
@@ -167,10 +163,29 @@ export class SharesDashboardComponent implements AfterViewInit {
           ],
           detailsHref: '/sharing/nfs',
           add() {
-            this.parent.add(this.tableComponent, ShareType.Nfs);
+            this.parent.slideInService.open(NfsFormComponent);
+            this.parent.slideInService.onClose$
```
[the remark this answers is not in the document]
RehanY147: The code inside the `onClose$` callback looks similar but it isn't really. `this` inside the `add` method doesn't point to `SharesDashboardComponent`. It points to the `TableComponent` inside the `ExpandableTableComponent` that we have 4 of in the `SharesDashboardComponent`. Meaning in each of the `add` methods, the access we have to the `this.getData` method, we won't have outside the `add` method. We can subscribe to it once in someplace else but the `SlideInService` doesn't really has nice way of communicating what page it's closing. It's configured to close with a response or error object. But that response is meant to be a response from the form submission. Not the form identifying itself. It can be done but it's unclean. Which is why I was avoiding that particular route.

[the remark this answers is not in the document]
RehanY147: I've extended the `IxSlideinService` to return the modal type when it closes so we have a way of knowing what form just closed.

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -1,652 +1,411 @@
...
+        switchMap((restart) => {
+          if (restart) {
+            return this.restartServices();
+          }
+          return of(true);
+        }),
+        switchMap(this.shouldRedirectToAclEdit),
+        filter(Boolean),
+        switchMap(this.shouldEnableServiceAutomaticRestart),
+        switchMap(this.shouldStartAndUpdateCifsService),
```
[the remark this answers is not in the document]
RehanY147: I already messaged someone from middleware for that reason. Should be able to get a response tomorrow.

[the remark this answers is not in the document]
RehanY147: Couldn't get a response from middleware. I've made changes to address the issues. Not sure how exactly locked dataset comes into play when enabling or disabling the service. Gonna wait for a response from middleware.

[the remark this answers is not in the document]
RehanY147: Quoting from @anodos325 

> Okay, the reasoning behind a service restart after applying change is whether the change should impact existing users. In an enterprise / business environment, making a service restart can be very undesirable and so we want to avoid it whenever possible.

> The reasons why we want to restart is when something has change requiring us to force everyone to restart:

- new time machine share - we need to change how the SMB server advertises itself (and clients need to see this change)

- hosts allow / hosts deny - security settings of share have changed and we need restart in order to enforce them

- home share - this changes details of how paths / shares are evaluated. Clients need to refresh this

- any path change - clients won't see this parameter change until they're forced to negotiate a new session

> The user should be prompted for a service restart so that they have a choice about whether to do it.... think of environment with 1000 users actively on server. We don't want to boot them unless administrator says "go for it".

undsoft (changes requested):
Ok, started testing the form:
1. There is a change detection issue when purposes are loaded. If you don't do anything once the form is opened, purposes won't show.
2. Default share parameters preset disables a bunch of checkboxes, but not when the form is opened.
3. Presets select needs required=true to hide -- option.
4. Multiprotocol preset and private SMB preset has a different set of checkboxes pre-set in old and new forms.
5. mangle warning used to appear only on edits
6. If you select a preset that clears Path Suffix and then select Multi-protocol, the path suffix will stay empty.
7. Can we rename Cancel in RestartSmbDialog to "Don't restart" not to confuse user that this cancels saving?

RehanY147:
@undsoft Can't reproduce issue regarding different values being disabled for the two preset options

For the prefix issue, same happens for old smb form
prefix isssue2

For the preset fields issue, same fields are disabled on both old and new forms. Can you provide screenshots maybe?

undsoft (approved):
Couldn't reproduce those two points mentioned before. Maybe they got resolved somehow.
In either case, nice work, mission accomplished. :)

PR #6607, opened 2022-04-12
"Import pool" dialog says "No options"

AlexKarpov98:
Removed first option when changed options.


src/app/modules/ix-forms/components/ix-select/ix-select.component.ts
```diff
@@ -38,6 +40,7 @@ export class IxSelectComponent implements ControlValueAccessor, OnChanges {
 
   ngOnChanges(): void {
     if (this.options) {
+      this.matSelect?.panel?.nativeElement.remove(0);
```
bvasilenko, line 43: I noticed that here we manipulate DOM tree to remove an element. Could you please clarify why it wasn't possible to achieve by modifying the model? Is there a reference to a known issue with `matSelect` UI control which we're trying to solve here?

bvasilenko (changes requested):

src/app/modules/ix-forms/components/ix-select/ix-select.component.ts
AlexKarpov98: Model not binding to select option that has blank value and selected default. I reverted and add `null` value to `No options`.

bvasilenko (approved):
Ok, looks good to me!

Works good.

image

image

PR #6611, opened 2022-04-15
NAS-115376 / 22.12 / NAS-115376: Added portals

RehanY147:
Test the page header for UserListComponent in comparison to the typical headers

### DIFF
--- a/src/app/pages/account/users/user-list/user-list.component.html
+++ b/src/app/pages/account/users/user-list/user-list.component.html
@@ -1,3 +1,46 @@
+<ng-template #pageHeader>
+  <div class="header-container">
+    <div class="title-container">
+      <a [routerLink]="['/', 'network']"><strong>NETWORK</strong></a>
+      <h1>Users</h1>
+    </div>
+    <div class="header-actions">
+      <button
+        color="primary" 
+        class="btn" 
+        (click)="doAdd()" 
+        mat-button
+      >Add</button>
+      <mat-slide-toggle 
+        [checked]="!hideBuiltinUsers" 
+        (change)="toggleBuiltins()" 
+        class="slider"
+      >Show Built-in Users</mat-slide-toggle>
+      <div class="input-container">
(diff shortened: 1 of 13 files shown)


src/app/pages/account/users/user-list/user-list.component.html
```diff
@@ -1,3 +1,46 @@
+<ng-template #pageHeader>
+  <div class="header-container">
+    <div class="title-container">
+      <a [routerLink]="['/', 'network']"><strong>NETWORK</strong></a>
+      <h1>Users</h1>
+    </div>
```
undsoft, line 6: Let's also simplify life for other developers in the future, so that they don't have to copy css.
For example, you can allow `<pagetitle>` to be put here. It can also be made a bit simpler (for example by defaulting `breadcrumbs` to true and nuking `productType`.

src/app/services/layout.service.ts
```diff
@@ -7,6 +8,12 @@ import * as domHelper from 'app/helpers/dom.helper';
 export class LayoutService {
   private readonly collapsedMenuClass = 'collapsed-menu';
 
+  private headerUpdater$: ReplaySubject<TemplateRef<any>> = new ReplaySubject();
+
+  get pageHeaderUpdater$(): ReplaySubject<TemplateRef<any>> {
```
undsoft, line 13: Let's update `anys` to `unknowns`.

src/app/services/page-title.service.ts
```diff
@@ -31,6 +31,10 @@ export class PageTitleService {
   }
 
   private observeRouteChanges(): void {
+    const routeParts = this.routePartsService.generateRouteParts(this.route.snapshot);
+    if (routeParts.length > 0) {
+      this.title$.next(routeParts[0]?.title || '');
+    }
```
undsoft, line 37: Duplicated code should be extracted into a separate method.

undsoft (approved):
Nice work!

PR #6613, opened 2022-04-18
NAS-111786 / 22.12 / NAS-111786: Added silent flag for all calls and handled errors

RehanY147:
Check `service.start` and `service.stop` calls for failure details and stack trace


bvasilenko:
@RehanY147 
What version of the back-end can I test with? 
I've tried latest stable night build, and it throws error stating "silent: Field was not expected"

bvasilenko (approved):
Looks good!

PR #6625, opened 2022-04-22
Further improvements to zh_hans translation

WilliamLi0623:


bvasilenko (approved):
Ok, I don't read chineese :) so it looks good to me

bvasilenko:
I've performed automated translation and it looks correct to me.

PR #6629, opened 2022-04-22
NAS-115862 / 22.12 / NAS-115862: Refactoring API Keys table

denysbutenko:
Go to `Topbar -> Account -> API Keys` try to add/edit/delete api keys


src/app/pages/api-keys/store/api-key.effects.ts
```diff
@@ -0,0 +1,47 @@
...
+import {
+  catchError, filter, map, switchMap,
+} from 'rxjs/operators';
+import { ApiEventMessage } from 'app/enums/api-event-message.enum';
+import { WebSocketService } from 'app/services';
+import {
+  apiKeyRemoved, apiKeysLoaded, apiKeysNotLoaded, apiKeysPageEntered,
+} from './api-key.actions';
+
+@Injectable()
```
undsoft, line 14: I think having fully fledged ngrx store for api keys is too much.
ngrx store is meant more for global things, like users and system configuration.
In here, just having a component store should be enough.

src/app/modules/common/dialog/general-dialog/general-dialog.component.html
```diff
@@ -11,8 +11,8 @@ <h1 mat-dialog-title>
   </div>
 </div>
 <div mat-dialog-actions>
-  <mat-checkbox color="primary" *ngIf="conf.confirmCheckbox" (change)="confirmed = !confirmed">{{conf.confirmCheckboxMsg || 'CONFIRM'}}</mat-checkbox>
+  <mat-checkbox color="primary" *ngIf="conf.confirmCheckbox" (change)="confirmed = !confirmed">{{conf.confirmCheckboxMsg || 'Confirm' }}</mat-checkbox>
```
undsoft, line 14: Also needs translation.

src/app/pages/api-keys/components/api-key-list/api-key-list.component.ts
```diff
@@ -0,0 +1,210 @@
...
+export class ApiKeyListComponent implements OnInit, AfterViewInit {
+  @ViewChild(MatSort, { static: false }) sort: MatSort;
+  @ViewChild('pageHeader') pageHeader: TemplateRef<unknown>;
+  toolbarActionsConfig: GlobalActionConfig = null;
+
+  displayedColumns: string[] = ['name', 'created_time', 'actions'];
+  settingsEvent$: Subject<CoreEvent> = new Subject();
+  filterString = '';
+  dataSource: MatTableDataSource<ApiKey> = new MatTableDataSource([]);
+  defaultSort: Sort = { active: 'name', direction: 'asc' };
```
undsoft, line 46: Clicking on Created Date header doesn't do anything.

undsoft (changes requested):
Works okay, but I think ngrx store is too much and we should have something simplier: load data in component, have a component store or have a generic component store that can be used in many places.

src/app/pages/api-keys/store/api-key.effects.ts
denysbutenko: Yeah, good point 👍🏻

src/app/pages/api-keys/store/api-key.store.ts
```diff
@@ -0,0 +1,90 @@
...
+          isLoading: false,
+          error: 'API Keys could not be loaded',
+        });
+
+        return EMPTY;
+      }),
+    );
+  });
+
+  addApiKey = this.updater((state, key: ApiKey) => ({
```
undsoft, line 61: I think one of two things should happen. Either we should include websocket calls for add/edit/delete in store (probably better?) or we should rename these to `apiKeyAdded`, `apiKeyEdited`, etc.

undsoft (approved):
👍🏻 
I wonder if in the future, we can have a generic component store (or generator of stores) so that you only need to provide API endpoints or something for it to work.

src/app/pages/api-keys/components/api-key-form-dialog/api-key-form-dialog.component.ts
```diff
@@ -33,10 +27,7 @@ export class ApiKeyFormDialogComponent implements OnInit {
   constructor(
     private fb: FormBuilder,
     private dialogRef: MatDialogRef<ApiKeyFormDialogComponent>,
-    private matDialog: MatDialog,
-    private ws: WebSocketService,
-    private loader: AppLoaderService,
-    private errorHandler: FormErrorHandlerService,
+    private store$: ApiKeyComponentStore,
```
undsoft, line 30: Original `store$` had dollar sign because it was an observable. It's not needed here.

src/app/pages/api-keys/components/api-key-form-dialog/api-key-form-dialog.component.ts
```diff
@@ -51,26 +42,13 @@ export class ApiKeyFormDialogComponent implements OnInit {
...
-          });
-        }
-      }, (error) => {
-        this.loader.close();
-        this.errorHandler.handleWsFormError(error, this.form);
-      });
+    if (this.isNew) {
+      this.store$.addApiKey({ name: values.name });
+    } else {
+      this.store$.editApiKey([this.editingRow.id, values]);
```
undsoft, line 50: We are missing loading indicators, dialogRef now closes even if operation was unsuccessful and previous error reporting was better in a way that it would show errors in the form.
I believe `addApiKey` and `editApiKey` return observables. You should be able to pipe them in the same way we used to do before and move presentation logic and error handling back here.

undsoft (changes requested):

src/app/pages/api-keys/components/api-key-form-dialog/api-key-form-dialog.component.ts
denysbutenko: Fixed

src/app/pages/api-keys/components/api-key-form-dialog/api-key-form-dialog.component.ts
denysbutenko: Yeah, good point. Store `effect` returns `Subscription` type

src/app/pages/api-keys/store/api-key.store.ts
denysbutenko: > I think one of two things should happen. Either we should include websocket calls for add/edit/delete in store (probably better?) or we should rename these to `apiKeyAdded`, `apiKeyEdited`, etc.

Since the first one is not working as expected. Probably make sense to go the second way.

denysbutenko:
Reverted the latest changes to follow the second way – `apiKeyAdded`, `apiKeyEdited` and `apiKeyDeleted`. Also rebased on top of master to resolve merge conflicts.

undsoft (approved):
I guess it's also possible to just write ordinary functions that would return observables instead of effects.
In any case, this is good too.

denysbutenko:
> I guess it's also possible to just write ordinary functions that would return observables instead of effects. In any case, this is good too.

Yeah, let's keep consistency until we find a better approach

PR #6640, opened 2022-04-27
NAS-115902 / 22.02.2 / [SCALE] Quota critical/warning alert cannot be 0

AlexKarpov98:
Testing:

Storage -> Edit Dataset -> Advanced Options -> Quota critical/warning

### DIFF
--- a/src/app/pages/storage/volumes/datasets/dataset-form/dataset-form.component.ts
+++ b/src/app/pages/storage/volumes/datasets/dataset-form/dataset-form.component.ts
@@ -87,6 +87,7 @@ export class DatasetFormComponent implements FormConfiguration {
 
   protected warning = 80;
   protected critical = 95;
+  private dataset: Dataset;
 
   custActions = [
     {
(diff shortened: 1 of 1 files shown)


src/app/pages/storage/volumes/datasets/dataset-form/dataset-form.component.ts
```diff
@@ -1466,34 +1468,22 @@ export class DatasetFormComponent implements FormConfiguration {
       delete data.special_small_block_size;
     }
 
-    if (data.quota_warning_inherit) {
-      delete (data.quota_warning);
+    if (data.quota_warning_inherit && this.dataset?.quota_warning) {
```
RehanY147, line 1471: Are the same changes required in `addSubmit`?

RehanY147 (changes requested):

src/app/pages/storage/volumes/datasets/dataset-form/dataset-form.component.ts
AlexKarpov98: No need to change in `addSubmit`
For example, if `quota_warning_inherit` is true, just need only to delete `quota_warning` in `addSubmit`. but need to submit value `INHERIT` if exists origin value of `quota_warning` in `editSubmit`

RehanY147 (approved):
Looks good to me

PR #6628, opened 2022-04-22
NAS-115593 / 22.12 / WIP NAS-115593: Add support for dynamic forms in ix-forms

bvasilenko:
Summary:

   This task is not complete yet, but at the moment I need someone to review and review the changes made so far. After your approval, the changes won't be merged until the task is complete.

So far I did:
1) Create `ix-dynamic-form` shared component
2) Declare a strong typed schema (based on the one we had for `ChartFormComponent`)
3) Integrate new dynamic form component into `AcmednsFormComponent` 
3) Integrate new dynamic form component and a new schema into `ChartFormComponent`, and display all components using ix-components (everything except `ix-list`)

PR #6650, opened 2022-05-01
WIP NAS-115593: Add support for dynamic forms in ix-forms, part 2

bvasilenko:
Continued from https://github.com/truenas/webui/pull/6628

## Summary
Implemented form submission and solved UI rendering issues. Now the `ChartFormComponent` has been rewritten for the new logic.
The following steps remain:
1. Rewriting `ChartWizardComponent for` new logic
2. Tests


undsoft (commented):
Generally looks good, let's try to extract some stuff out to make it more reusable. Also let's try to tighten type-safety, `any` is last resort when everything else failed. If full interface is impossible, alternatives may be `index signatures`, `unknown`.

Testing:
ACME DNS form:
1. Extra paddings.
2. Every input has a placeholder. Is not really needed.
3. If it takes a bit of time for schemas to load, form will be in a weird state with no loading indication.

Charts:
1. Wizard page looks much cleaner without the "wizard", but it's unclear what the difference between chart-form and chart-wizard is now. Can we have one component now?
1. Selects now have -- selectable as an option. Likely need `[required]` on them.
2. Missing some section groups and checkbox titles. They may not be needed actually, but we shouldn't make this change now.
Снимок экрана 2022-05-03 в 11 58 39
3. Missing some paddings in lists.
4. Got an error editing installed Test app on http://10.234.30.8/.
5. Missing tooltips.

PR #6651, opened 2022-05-02
NAS-115481 / 22.12 / NAS-115481 - Added extra check for group by name and disabled save button when group form is invalid

RehanY147:


undsoft (changes requested):
If the goal was to prevent Save button from working when incorrect value is in Group field, then I can still press that button. Seems more related to how ix-combobox works and not groupQueryDsCache.

RehanY147:
@undsoft  There are two bugs to be fixed here
1. We need to disable the save button if the owner user and group form is invalid
2. We need to ask middleware if it can find a group with the exact name in the search query outside the cache. The original query to search groups with similar names doesn't do that. It only asks for similar groups in the cache. The exact name match query also searches for groups outside the cache which is the core issue addressed in this ticket.

RehanY147:
@undsoft example of disabled button
button disable

undsoft (changes requested):
https://user-images.githubusercontent.com/1481270/167138904-a7d58875-e390-4b12-a5f4-7b16ccba29d1.mov

RehanY147:
@undsoft This is because the previous selection doesn't change when you start typing in to search for something. That's an IxCombobox issue. Should that behave differently?

undsoft:
@RehanY147 Yes, in this case user will be under impression that form was saved. Combobox should set value to empty if an invalid option is entered. Also I should be able to remove the value and type-in correct value manually.

RehanY147:
> @RehanY147 Yes, in this case user will be under impression that form was saved. Combobox should set value to empty if an invalid option is entered. Also I should be able to remove the value and type-in correct value manually.

That only makes sense if the typed value matches one of the available options actual value. But without selecting, that might have unintentional consequences. Like the combobox is set to show the label of a option when a value is selected. If i type in 'root' and that has a label 'rehan' then suddenly, the input will have 'rehan' which wasn't intended.

RehanY147:
@undsoft The idea behind this behavior by `ix-combobox` was that it should behave as much as possible like a normal select with just one added feature of being able to search options. The more custom behavior it has, the more confusing it becomes.

undsoft:
@RehanY147 It's not really a select. Also in vanilla select, you can type to select a value.
Consider this: I get to the form, it has 'root' in combobox, I type to replace 'root' with 'sys' (one of the valid options). I save the form, but nothing gets saved, because I didn't click on 'sys'.

RehanY147:
@undsoft The questions that need answers if we are to implement this behavior
We want to show labels for the selected option instead of the value. What happens when the user is typing `root` and there are two options with the values `root` with label `Rehan` and `root1` with label `Rehan1`. If the user types `root` do we show the selected value label in the input field now i.e., `Rehan`. What if they wanted to type in `root1` and the input interrupts them? Or should we only try to select a value when the input loses focus? In that case, what happens if the user clicks on the save button directly. Whatever the value, they won't see the label for it. They could then assume that it is like a text input. Whatever they input, will be treated like a valid value.

undsoft:
@RehanY147
Users won't be typing in values, they will still be typing in labels. So, in this example typing in `root` or `root1` will set form control to empty, because there is no option to match. Typing `Rehan` and `Rehan1` will set form control value to `root` and `root1` as user types (or maybe on change event).

undsoft (changes requested):
I can still do the same thing as I've showed in video.
Also, can you explain what the difference between querying for groups by name vs by `group`?

RehanY147:
> Also, can you explain what the difference between querying for groups by name vs by `group`?

When we do a search with the `^` operator, middleware will look for groups with similar names in the cache. But that excludes a group that exists but wasn't fetched so it's not in the middleware cache. This is the issue the ticket reporter was having. With the `=` operator, middleware searches outside the cache too basically. To see if a group exists somewhere with the exact name match. Next time, that group will be included in the cache too.

undsoft (approved):
Thanks, looks good.

PR #6652, opened 2022-05-02
NAS-115852 / 22.12 / NAS-115852 - Better header with custom html elements

RehanY147:


src/app/pages/account/users/user-list/user-list.component.html
```diff
@@ -1,5 +1,31 @@
...
+        mat-button 
+        type="button" 
+        color="primary"
+        (click)="doAdd()"
+      >{{ 'Add' | translate }}</button>
+      <mat-slide-toggle 
+        (change)="toggleBuiltins()" 
+        [checked]="!this.hideBuiltinUsers"
+      >
+        <label>Show Built-in Users</label>
```
denysbutenko, line 13: Missing translation pipe

src/app/modules/common/page-title-header/page-title-header.component.scss
```diff
@@ -0,0 +1,21 @@
+.header-container {
+  color: white;
```
denysbutenko, line 2: Seems better to use some variable here

src/app/modules/common/page-header-actions-wrapper/header-input.directive.ts
```diff
@@ -0,0 +1,29 @@
...
+@Directive({
+  selector: '[headerInput]',
+})
+export class HeaderInputDirective implements OnInit {
+  compRef: ComponentRef<HeaderInputWrapperComponent>;
+  @Input() prefixIcon: string;
+  @Input() shouldShowReset: Observable<boolean>;
+  @Output() resetInput = new EventEmitter<void>();
+  constructor(private vcRef: ViewContainerRef,
+    private templateRef: TemplateRef<any>,
```
denysbutenko, line 18: any is not good

denysbutenko (changes requested):
Will provide more feedback after testing

denysbutenko:
https://user-images.githubusercontent.com/351613/166207691-00208382-31ef-4825-9377-5758a0bd6962.mp4

1. The search input does not look well. No focus when clicking on the icon. UI jumps on focus.
2. Hover effect on the primary button looks inconsistent to other pages

denysbutenko:
Groups old one:
image
Users new one:
image
Still looks and behaves inconsistently. 

- Content goes below on a few pixels when search input is focused.
- search icon requires `pointer-events: none` to pass clicks on the input
- The primary button does not have a hover effect on other pages but there become gray

denysbutenko (changes requested):

denysbutenko:
Focus state on buttons looks nice 👍🏻

denysbutenko (changes requested):
image
image
Looks better now. I like how the page looks now without the border and gap below

Still see minor differences:

- breadcrumb and title on the users page look collapsed in comparison to the groups page
- search icon size mismatch between groups and users page
- search input has different border width on focus state
- height of search input is a bit different

RehanY147:
@denysbutenko The height and width issue of the input might be related to different screen zoom while comparing. Otherwise, I don't see significant difference. Can you confirm please?

denysbutenko:
> @denysbutenko The height and width issue of the input might be related to different screen zoom while comparing. Otherwise, I don't see significant difference. Can you confirm please?

There are 36px and 40px same zoom ~ 10% diff

PR #6654, opened 2022-05-02
NAS-115709 / 22.12 / Nas 115709 Add linter rule to standartize import styles

Raz-Dva:
Added a new import rule. The "../../..." part of the path should be replaced with "app /..."
Check project for any errors after replacing paths.


undsoft (approved):
So, I'm assuming there was no way to have paths be fixed automatically?
This is okay too, though.

Raz-Dva:
> 



> So, I'm assuming there was no way to have paths be fixed automatically? This is okay too, though.


I couldn't find a decent plugin that implements auto-fix.

PR #6655, opened 2022-05-03
NAS-115713 / 22.12 / NAS-115713: New api endpoint to be used for recordsize choices for webUI

undsoft:
Testing: in dataset form, zvol form or zvol wizard (available when adding a Disk device in Virtual machines) there is record/block size field.
Field takes its values from API now. If the selected value is lower than recommended minimum, a warning is shown.


AlexKarpov98 (approved):
LGTM! 👍
Screenshot_41
Screenshot_42

undsoft:
backport

PR #6657, opened 2022-05-04
NAS-115713 / 13.0 / NAS-115713: Replacing static recordsizes with api call

undsoft:
Similar to previous PR: dataset form and zvol form need testing for record/block size field.


AlexKarpov98 (approved):
Nice Works like prev PR! 👍
Screenshot_45
Screenshot_46

PR #6665, opened 2022-05-07
NAS-115759 / 22.12 / Added readonly and disabled to immutable fields when editing

RehanY147:


undsoft (changes requested):
I got this error when editing Netdata on our cloud server. I don't have to change anything, I'm just saving form as is.
Снимок экрана 2022-05-09 в 12 35 26

PR #6672, opened 2022-05-08
NAS-116104 / 22.12 / ix-input shouldn't set value to '' when type is number

AlexKarpov98:
Testing:

System Settings -> Services -> NFS -> Ports fields


src/app/modules/ix-forms/components/ix-input/ix-input.component.ts
```diff
@@ -59,6 +59,13 @@ export class IxInputComponent implements ControlValueAccessor {
     this.value = value;
   }
 
+  formatValue(value: string | number): string | number {
```
RehanY147, line 62: Instead of this method, maybe we can have getter and setter for `value`. Let me know if you disagree.

RehanY147 (changes requested):

PR #6673, opened 2022-05-08
NAS-116096 / 22.02.2 / Migrating system dataset pool does not change displayed system dataset pool until page reload

AlexKarpov98:
Testing:

System Settings -> Advanced -> System Dataset Pool -> Configure


dszidi (approved):
Updates system dataset without needing to refresh the page.
LGTM

undsoft (changes requested):
I don't see what the difference between old code and new code is.
I still get old system dataset pool shown in the table.

Also, there is another issue if user selects Cancel in restart dialog. This stops saving, but does not clear loading indicators.

AlexKarpov98:
Solved issues already in [6678](https://github.com/truenas/webui/pull/6678) 

- if user selects Cancel in restart dialog. This stops saving, but does not clear loading indicators
- closes slide even if the state of the job is not successful
- not displayed any validation error in the interface

so close this PR

PR #6674, opened 2022-05-08
Migrating system dataset pool does not change displayed system dataset pool until page reload

AlexKarpov98:
Testing:

System Settings -> Advanced -> System Dataset Pool -> Configure


denysbutenko (approved):
Works as expected 👍🏻

PR #6675, opened 2022-05-09
NAS-115951 /  Add linter for component selectors

Raz-Dva:
Check functionality after changing selectors


AlexKarpov98 (commented):
Changes looks good, but seems not applied to all the components:
- [ ] Use ix- as prefix.
- [ ] Rename existing components

PR #6676, opened 2022-05-09
NAS-113293 / 22.12 / NAS-113293: Updating to Angular 13

undsoft:
General regression testing.


RehanY147 (approved):
Looks good. No obvious regressions. UI seems to be working OK.

PR #6678, opened 2022-05-10
NAS-116092 / 22.02.2 / System dataset update validation errors are not displayed

AlexKarpov98:
Fixed error that goes to next even if the state of the job is not successful when using `switchMap` with `ws.job` in `SyslogFormComponent`, `KubernetesSettingsComponent`, `SystemDatasetPoolComponent`.

### DIFF
--- a/src/app/pages/applications/kubernetes-settings/kubernetes-settings.component.ts
+++ b/src/app/pages/applications/kubernetes-settings/kubernetes-settings.component.ts
@@ -10,6 +10,7 @@ import {
 import {
   catchError, filter, map, switchMap, tap,
 } from 'rxjs/operators';
+import { JobState } from 'app/enums/job-state.enum';
 import { choicesToOptions } from 'app/helpers/options.helper';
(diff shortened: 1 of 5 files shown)


undsoft (changes requested):
Looking at `job` at `WebsocketService` observable will complete when job completes.
You can also either use `finalize()` in `pipe` or update `subscribe` to use `complete` callback.

If not, `tap` statements can be updated to return early.

I've tried to simulate error in syslog form by changing code to:
```
switchMap(() => this.ws.job('systemdataset.update', [{ syslog: 'pizza' as any }])
```
I did not see any error in the interface. Could be because `FormErrorHandlerService` is not used.

undsoft (changes requested):
I meant so that it's always:
```
this.errorHandler.handleWsFormError(error, this.form);
```
but inside it handles 'exc_info', so that developers don't have to do spreads in every error handler for jobs.

undsoft (approved):
Works well.

PR #6681, opened 2022-05-10
NAS-116088 / 22.12 / NAS-116088: Refactoring quota forms

undsoft:
Testing: 
Two forms.
Dataset context menu -> User Quotas and Group Quotas -> Add.

We used to have to fields to specify users/groups, they were deliberately merged into one.
IxChips gained support for autocomplete.


denysbutenko (commented):
Getting some errors when submitting user form
- when selected `root` user – `[EINVAL] quotas.2.id: user: Setting u quota on s [0] is not permitted.`
- when selecting the same user twice in the chips component - `'NoneType' object has no attribute 'decode'`

denysbutenko:
The action says `Add` but the form allows updating existing user quotas. Not sure it's worth to notify somehow

denysbutenko (approved):
Otherwise looks good 👍🏻

denysbutenko (commented):
image
image
Breadcrumbs does not work

undsoft:
I'll create additional tickets for some of the issues found: one is middleware and one happens on other pages too.

PR #6684, opened 2022-05-12
NAS-115593: Add support for dynamic forms in ix-forms 

bvasilenko:


bvasilenko:
this is PR with entire code of the feature
closing to create another partial PR as a draft
after the review, will re-open

PR #6685, opened 2022-05-12
NAS-115593 / None / WIP NAS-115593: Add support for dynamic forms in ix-forms, part 4

bvasilenko:


src/app/enums/chart-schema-type.enum.ts
```diff
@@ -1,9 +1,10 @@
-export enum DynamicSchemaType {
+export enum ChartSchemaType {
   Int = 'int',
   String = 'string',
   Boolean = 'boolean',
   Hostpath = 'hostpath',
   Path = 'path',
   List = 'list',
   Dict = 'dict',
+  Ipaddr = 'ipaddr',
```
undsoft, line 9: Unused key.

src/app/modules/entity/utils.ts
```diff
@@ -296,173 +285,6 @@ export class EntityUtils {
     return result;
```
undsoft, line 285: `createRelations` is also unused.

src/app/modules/ix-forms/components/ix-dynamic-form/ix-dynamic-form.component.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  const createComponent = createComponentFactory({
+    component: IxDynamicFormComponent,
+  });
+
+  beforeEach(() => {
+    spectator = createComponent();
+  });
+
+  describe('Control Emit', () => {
+    it('addControlNext()', () => {
```
undsoft, line 15: Same here, it's better to do `MockComponent(IxDynamicFormItem)` and then check what's being passed in.
```
const dynamicItem = spectator.query(IxDynamicFormItem);
expect(dynamicItem.schema).toEqual(...);

// In another case
dynamicItem.addListItem.emit(...)
// expect that parent listener has been called.
```

src/app/modules/ix-forms/components/ix-list/ix-list-item/ix-list-item.component.scss
```diff
@@ -9,6 +9,10 @@
   width: 100%;
 }
 
+.list-item-content {
```
undsoft, line 12: This is okay, but this will break other components that use ix-list and had margins added manually.

undsoft (commented):
Still noticed that on http://10.234.30.8/ when editing pihole I get a different set of attributes sent.
For example in `appVolumeMounts.config.datasetName`

bvasilenko:
@undsoft 

> Still noticed that on http://10.234.30.8/ when editing pihole I get a different set of attributes sent. For example in `appVolumeMounts.config.datasetName`

A bit more details would be useful. Could you please provide some screencapture or screenshot, depicting the comparison between what's expected and actual result?

bvasilenko:
Ok, now everything what's above should be resolved.

PR #6688, opened 2022-05-12
NAS-116082 / 22.02.2 / NAS-116082: Update Tunable/Sysctl Variable tooltips for SCALE

Mrt134:
Remove text about rc.conf and loader variables as these no longer apply in SCALE.


denysbutenko (approved):
Looks good!

PR #6693, opened 2022-05-14
NAS-112562 / 22.02.2 / NAS-112562: Showing error properly

RehanY147:
The method to recreate this issue is described in the ticket

1. Create a mirrored pool
2. Run the command mentioned in the ticket on your NAS
3. Attempt to add more disks to the pool
4. Previously, you would only see error "Failed to extend"
5. Now you should see error returned by middleware


dszidi (commented):
image


This looks to be consistent with what Vladimir was suggesting. I'm wondering if we should filter out the "[EFAULT]"? @undsoft what do you think?

undsoft:
@dszidi Maybe as another ticket later. This is not specific to extending disks.

dszidi (approved):
LGTM

PR #6661, opened 2022-05-05
NAS-116009 / 22.12 / Refactoring Dashboard

denysbutenko:
Changes:

- Sync styles for widget header and remove `&nbsp;`. 
- Sync widget gaps
- On the `System Information` widget fill more space on the `Version` row
- Better error handling
- More progressive loading but still have space for improvements
- Fixed a couple of bugs on mobile widgets
- PR includes a fix from #6680 
- add throttle for network updates
- use a new approach for page-header

For testing enable available widgets and check if everything looks good. Try it with a slow connection to feel a full experience. Any comments are appreciated.

PR #6694, opened 2022-05-15
NAS-116193 / 22.12 / Added `system.info` to NgRx Store

RehanY147:


src/app/pages/system/general-settings/general-settings.component.ts
```diff
@@ -267,52 +268,54 @@ export class GeneralSettingsComponent implements OnInit {
...
-          mimetype = 'application/x-tar';
-          fileName += '.tar';
-        } else {
-          mimetype = 'application/x-sqlite3';
-          fileName += '.db';
+    this.store$.pipe(waitForSystemInfo, untilDestroyed(this)).subscribe(
+      (systemInfo) => {
+        let fileName = '';
+        let mimetype: string;
+        if (systemInfo) {
```
denysbutenko, line 275: seems like this check is unnecessary

denysbutenko (changes requested):
Like this idea. The changes look good except for one issue:

https://user-images.githubusercontent.com/351613/168555070-009ba3b3-42f2-4a99-b019-81285440f764.mp4

For some reason, the dashboard page is still empty and no errors in logs were tested against a cloud VM.

undsoft:
Nice cleanup 👍🏻

src/app/pages/system/general-settings/general-settings.component.ts
RehanY147: > Like this idea. The changes look good except for one issue:
> Kapture.2022-05-16.at.14.19.09.mp4
> 
> For some reason, the dashboard page is still empty and no errors in logs were tested against a cloud VM.

The issue with the empty dashboard is only an issue of slow network/slow cloud vm responses. If you wait long enough, the cloud vm responds and the dashboard shows up. Maybe we should have a loading indication on the dashboard too.

denysbutenko: Yeah, it might be. It just happened in this branch. When opened cloud VM on the same network directly, it shows correctly. I'll check on a local VM and let you know 👍🏻 
The loading indicator will be addressed in #6661

RehanY147: Sounds good. Please do.

denysbutenko (approved):
Works well on local and cloud VMs. Nice job 👍🏻

PR #6695, opened 2022-05-16
NAS-115951 / None / Nas 115951- 4 Add linter for component selectors

Raz-Dva:
Check functionality after changing selectors

All component selector prefixes "app-" have been changed to "ix-".
All component selectors are prefixed with "ix-".
Changes have also been made in the "tests" folder.


bvasilenko (approved):
I've done smoke testing and it looks good to me

PR #6697, opened 2022-05-16
NAS-116237 / 22.12 / Optimize unit tests executing time in CI

denysbutenko:
Small improvement for jest tests execution. 
image


undsoft (approved):
Less is more :)

PR #6699, opened 2022-05-17
NAS-116149 / 22.12 / Useless / broken breadcrumbs

AlexKarpov98:
Re-configured router-part-service and updated `BreadcrumbComponent`.
Please review breadcrumbs for all pages.


dszidi (commented):
@undsoft  IMO if we're going to remove these breadcrumbs, then the titles themselves will probably need to be a little bit more descriptive. Couple of examples are below

image

This could be something like "Advanced System Settings"

image

This title could be something like "Memory Reports"

undsoft (changes requested):
Yes, agreed, we should either improve wording, i.e. Advanced -> Advanced Settings, CPU -> CPU Reports, etc. or maybe we could have non-clickable tooltips, but show them styled differently?
Also Local Users and other subpages in Credentials have broken breadcrumbs.

undsoft (approved):
Maybe let's use a slightly more muted color for disabled breadcrumbs (like --alt-fg2).
Otherwise looks good.

PR #6700, opened 2022-05-17
NAS-116223: Extracting restore dialogs into separate components

undsoft:
1. Create a cloudsync task.
2. Create a replication task.
3. Use Restore buttons on both of them.


denysbutenko (approved):
Looks good

PR #6701, opened 2022-05-17
Docs 3538 - Resort UI tests for new Jira Numbering

q5sys:


q5sys:
The manchinaris test failure is a false positive, it failed due to the assert timing out, but checking the system after the cycle was finished and Manchinaris was installed and running.  It had passed in the 4 tests prior, so a random false positive.

ericbsd (approved):

pkgdemon (approved):

PR #6702, opened 2022-05-18
NAS-116175 / 22.12 / Show most important alerts first

AlexKarpov98:
Updated `sortComparer` function by sorting by level, klass, datetime in `alert.reducer.ts`.


bvasilenko (approved):
Looks good to me!

PR #6703, opened 2022-05-18
NAS-116139 / 22.12 / Optimize theme loading and fallback styles

denysbutenko:
Changes:

- Fix saving theme when GUI form submitted
- Save the active theme to session storage to activate it before user preferences are loaded
- Getting user theme from the store
- Add fallback values for CSS variables
- Use new `page-title-header` inside old `pagetitle` to keep consistent look :)
- Some code cleanup


undsoft (changes requested):
Fonts on buttons and headers looks funny.
I also have a console error, not sure if it's related.
Снимок экрана 2022-05-18 в 10 35 59

denysbutenko:
> Fonts on buttons and headers looks funny. I also have a console error, not sure if it's related. Снимок экрана 2022-05-18 в 10 35 59

Yeah, there are two issues with font:

1. font-smoothing property is not applied in Safari. Does it make sense to enable it for all browsers or just remove it?
2. `IBM Plex Sans` font-face not properly defined, it will be fixed

denysbutenko:
`Enabled` means these styles applied:
```
body {
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
}
```

|Browser|Enabled|Browser Defaults|
|---|---|---|
|Chrome|image|image|
|Safari|image|image|

undsoft (changes requested):
I just want fonts to look like they used to look before the changes.
Снимок экрана 2022-05-18 в 18 35 59

denysbutenko:
> I just want fonts to look like they used to look before the changes. Снимок экрана 2022-05-18 в 18 35 59

Yeah, overlooked. Reverted to use `Epilogue` font for h1-h2. Thanks

Other possible mismatches. I'm trying to compare the master branch to keep the same appearance. `IBM Plex Sans` font uses for body text and it has many classes with different font style and font weight. And in the master branch, font-face is defined to use Regular type only. Perhaps each browser is trying to render thin/bold types on its own. . In that PR, font-face uses all font files according to type.

An example (same look before 700, now 500)
image

undsoft (approved):
- There used to be a thin line underneath top toolbar. It's gone now. I'm not opposed to change, but you need to confirm this with Damian.
- Buttons and headers look different because they use different font now. They used to use IBM Plex Sans, now they use IBM Plex Sand Medium.

Otherwise looks good.

PR #6705, opened 2022-05-19
NAS-116230 / 22.12 / NAS-116230: Remove Core mentions and convert Enterprise

denysbutenko:
Changes:

- `ProductType.Core` has been removed
- `ProductType.Enterprise` has been converted to `ProductType.ScaleEnterprise` and removed
- Some refactoring for getting productType

### DIFF
--- a/src/app/pages/system/update/update.component.html
+++ b/src/app/pages/system/update/update.component.html
@@ -124,26 +124,32 @@
   </mat-card-content>
 </mat-card>
 
-<mat-card *ngIf="updates_available && pre_release_train && product_type === ProductType.ScaleEnterprise" id="stable-warning-card">
-  <div class="stable-warning"><i class="material-icons" id="update-info">info</i>
-    <a href="https://www.truenas.com/docs/hub/intro/release-notes/" target="new">
+<mat-card *ngIf="updates_available && pre_release_train && product_type === ProductType.ScaleEnterprise" class="train-info-card">
+  <h4 class="stable-warning"><i class="material-icons">info</i>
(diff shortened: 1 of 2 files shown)


src/app/modules/common/dialog/about/about-dialog.component.html
```diff
@@ -2,8 +2,7 @@
   <div class="header">
     <div class="header-overlay">
       <div class="logo-wrapper">
-        <mat-icon *ngIf="systemType === ProductType.Enterprise" svgIcon="truenas_enterprise_logo_full" class="logo" alt="TrueNAS ENTERPRISE logo"></mat-icon>
```
undsoft, line 5: You may be able to remove the image too.

src/app/pages/system/update/update.component.html
```diff
@@ -124,34 +124,15 @@
   </mat-card-content>
 </mat-card>
 
-<mat-card *ngIf="updates_available && pre_release_train && product_type === ProductType.Core" id="stable-warning-card">
-  <div class="stable-warning"><i class="material-icons" id="update-info">info</i>
-    <strong>{{"This is not a production release, and should only be used for testing." | translate}}</strong>
```
undsoft, line 129: This looks useful.

denysbutenko: Is there some case when condition can be true? The same block with other сondition is present below

undsoft: I'm just saying that I think it makes sense to reimplement this for scale (assuming that pre_release_train works correctly).

PR #6711, opened 2022-05-20
NAS-115593 / None / NAS-115593: Add support for dynamic forms in ix-forms, improvements

bvasilenko:
Iteration after comparative testing of Application create/edit dynamic forms. I've done the necessary additions to the code (`dependsOn` property for the model). And also fixed several bugs.

### DIFF
--- a/src/app/enums/chart-schema-type.enum.ts
+++ b/src/app/enums/chart-schema-type.enum.ts
@@ -6,4 +6,5 @@ export enum ChartSchemaType {
   Path = 'path',
   List = 'list',
   Dict = 'dict',
+  Ipaddr = 'ipaddr',
 }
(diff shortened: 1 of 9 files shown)


undsoft (changes requested):
I'm a bit confused about the purpose of dependsOn. It seems that it only plays a role in change detection?

Testing from m40:
- When installing unpackerr, if I select `Show Expert Config` under `Networking and Services`, the form will not let me save with `dnsPolicy` = `Default`.
- Probably not an issue, but when installing kanboard, `(Advanced) Network Policy` - `Egress` - `Ports` are saved as strings, when previously numbers were sent.
- Same app, `(Advanced) Network Policy` - `Egress` - `To`, when cidr and except are filled in, except is supplied as an object (used to be an array of strings).

With regards to unit tests, they can still be improved. Let's talk about it after the team meeting.

PR #6717, opened 2022-05-24
NAS-116345 / 22.12 / "Maximum Upload Parts" is shown in "Authentication" section of S3 cloud sync credentials

AlexKarpov98:
Testing:
Credentials -> Backup Credentials -> Cloud Credentials -> Add (Provider: Amazon S3)
Removed `Maximum Upload Parts` field


undsoft (changes requested):
I think Vladimir meant that field doesn't belong to this section and thus should be in some other section, not that it should be removed completely.

themylogin (changes requested):
😱😱😱

undsoft (approved):
Looks good.

PR #6718, opened 2022-05-24
NAS-116346 / 22.12 / Not able to retrieve logs for a successful cloud sync task

AlexKarpov98:
Testing:
Data Protection -> Cloud Sync Tasks -> Run a task
After it's completed, click the `Success` or `Failed` button to view the logs.


undsoft (changes requested):
I don't like the logic here. `onComplete` is dependent on `getJobStatus` being called. This side-effect is implicit.
If you want a subject, you can add a subject to data-protection-dashboard itself.

PR #6720, opened 2022-05-25
NAS-115593: Add support for dynamic forms in ix-forms, improvements 2

bvasilenko:


src/app/interfaces/dynamic-form-schema.interface.ts
```diff
@@ -33,13 +33,15 @@ export interface DynamicFormSchemaInput extends DynamicFormSchemaBase {
   type: DynamicFormSchemaType.Input;
   tooltip?: string;
   private?: boolean;
+  number?: boolean;
```
undsoft, line 36: Maybe we ought to have something like `type` instead of `private` and `number` with enum of values like `password` and `number`. Would be more inline with how ix-input works.

src/app/modules/ix-forms/components/ix-dynamic-form/ix-dynamic-form-item/ix-dynamic-form-item.component.spec.ts
```diff
@@ -1,89 +1,254 @@
...
     spectator = createComponent();
   });
 
-  it('isHidden', () => {
-    spectator.component.dynamicForm = dynamicForm;
-    spectator.component.dynamicSchema = dynamicSchema;
-    expect(spectator.component.isHidden).toBeTruthy();
-  });
+  describe('UI tests', () => {
+    it('input', () => {
```
undsoft, line 125: tests also serve as documentation, so:
```
describe('Component rendering', () => {
  it('renders an ix-input when schema with "input" type is supplied'
```

src/app/modules/ix-forms/components/ix-dynamic-form/ix-dynamic-form-item/ix-dynamic-form-item.component.spec.ts
```diff
@@ -1,89 +1,254 @@
...
+      spectator.component.dynamicForm = dynamicForm;
+      spectator.component.dynamicSchema = inputSchema;
```
undsoft, line 127: Optionally, you can also do here:
```
spectator = createComponent({
  props: {
    dynamicForm,
    dynamicSchema: inputSchema,
  },
});

File: src/app/modules/ix-forms/components/ix-dynamic-form/ix-dynamic-form-item/ix-dynamic-form-item.component.spec.ts [hunk 4]
```diff
@@ -1,89 +1,254 @@
...
+      spectator.detectComponentChanges();
+      expect(spectator.query('ix-input')).toBeVisible();
+      expect(spectator.query('ix-input')).not.toBeDisabled();
```
undsoft, line 130: I don't think it's every going to be disabled, as this probably just checks for [disabled] attribute on `ix-input` tag.
You can test for inputs being passed correctly:
```
expect(spectator.query(IxInputComponent).required).toBe(true);
```
etc.

src/app/modules/ix-forms/components/ix-dynamic-form/ix-dynamic-form-item/ix-dynamic-form-item.component.spec.ts
```diff
@@ -1,89 +1,254 @@
...
+      spectator.component.dynamicForm = dynamicForm;
+      spectator.component.dynamicSchema = listSchema;
+      expect(spectator.component.isHidden).toBeTruthy();
+    });
+
+    it('getFormArray()', () => {
+      spectator.component.dynamicForm = dynamicForm;
+      spectator.component.dynamicSchema = listSchema;
+      expect(spectator.component.getFormArray).toEqual(dynamicForm.controls.list);
+    });
```
undsoft, line 238: I don't see value in testing this. `getFormArray` is public only because Angular doesn't allow private methods to be called in templates.
In spirit this method is private and is not meant to be interacted with by others, i.e. it's not really part of the public interface of this component.
How you do interact with this method is through a template. Ix-list renders a list of items and this is what needs to be checked in 'ix-list' case above.

src/app/modules/ix-forms/components/ix-dynamic-form/ix-dynamic-form-item/ix-dynamic-form-item.component.spec.ts
```diff
@@ -1,89 +1,254 @@
...
+
     it('addControl()', () => {
       spectator.component.dynamicForm = dynamicForm;
-      spectator.component.dynamicSchema = dynamicSchema;
+      spectator.component.dynamicSchema = listSchema;
 
       jest.spyOn(spectator.component.addListItem, 'emit').mockImplementation();
       spectator.component.addControl();
       expect(spectator.component.addListItem.emit).toHaveBeenCalledTimes(1);
     });
```
undsoft, line 247: Here too `addControl` is not mean to be called directly by anything but the template. It's only public de jure.
What really happens is: user presses add in ix-list and we want to emit `addListItem`.
We can simulate this:
```
    it('emits addListItem when (add) is emitted by a list control', () => {
      spectator = createComponent({
        props: {
          dynamicForm,
          dynamicSchema: listSchema,
        },
      });
      jest.spyOn(spectator.component.addListItem, 'emit').mockImplementation();

      spectator.query(IxListComponent).add.emit();

      expect(spectator.component.addListItem.emit).toHaveBeenCalledWith({
        array: dynamicForm.controls.list,
        schema: listSchema.itemsSchema, // itemsSchema needs to be added
      });
    });
```

src/app/services/app-schema.service.ts
```diff
@@ -381,4 +381,28 @@ export class AppSchemaService {
   deleteFormListItem(event: DeleteListItemEvent): void {
     event.array.removeAt(event.index);
   }
+
+  remapAppSubmitData(data: HierarchicalObjectMap<ChartFormValue>): HierarchicalObjectMap<ChartFormValue> {
```
undsoft, line 385: What does this do?
It doesn't seem that typing is correct because both `data` and return value can be simple values like strings.

bvasilenko: I've rewritten this function, so it would be obvious what it does from it's signature.

bvasilenko:
Ok, now we have more meaningful tests.
I've also found and fixed another minor bug

undsoft (commented):
When installing `appdaemon`, I've added `Additional app storage`, but kept `name` empty as suggested. It got sent as null (previously key would be absent). Middleware doesn't like null as value.
Same for when installing `website-shot` and selecting `Show Advanced settings` under `Networking and Services`.

Other than that, looks good.
Let's do something with rest of PRs, sync with master and I'll retest it again.

bvasilenko:
@undsoft 

I've synced with master, and fixed minor bugs. Would you please take a look?

Once I receive a confirmation from you, I'll prepare a full PR for this feature.

PR #6727, opened 2022-05-29
Update ko.json

hottodogu1919:
translated 1 to 32


undsoft (changes requested):
Sorry, you misunderstood. You need to keep the original string on the left, but provide translation on the right.
For example: 
```
"(This Controller)": "이 컨트롤러",
```

PR #6730, opened 2022-05-30
NAS-115812: Fixed tab loading an unsubbing

RehanY147:
Go to the Apps page, switch tabs a bunch of times. The original UI slows down. This PR should fix the lag.


bvasilenko (commented):
Code changes look good to me.

I have made before/after comparison , and observed an improvement of the performance after user rapidly switches between various tabs a lot of times.

While testing, I've noticed another problem, which is endless flickering of UI displayed in the video below. This problem was present before these code changes, so I guess we need to address it in another PR


https://user-images.githubusercontent.com/20611516/171187920-12a27591-d10e-48c1-98af-bfda96216041.mp4



This can be reproduced by rapidly switching between tabs.

bvasilenko (approved):

PR #6731, opened 2022-05-30
NAS-114484: Storage Redesign

RehanY147:
Todo before merging this PR
- Fix routing for `Device Management` and `DatasetManagement` pages from sidenav to buttons on the pools dashboard.
- Fix route structure on the sidenav for the new Storage

This branch sets up the base for the storage redesign as `Storage 2.0`. All changes meant for storage redesign should be committed to this branch.
Extend branches meant for storage redesign from this branch.


undsoft (commented):
Looks good, however let's wait for designs to be posted somewhere. I want to add a note in UI explaining what's going on.

undsoft:
It's okay, this is pre-release software that tinkerers will use. We can let them see how the sausage is made.

undsoft (changes requested):
Let's add a note on top of each page, explaining that this is a work a progress.
Something like:
> This is a work in progress for new Whatever page. You can see what we are building an provide suggestions in < a >Github discussions< /a >.

undsoft (approved):
Looks good.
Снимок экрана 2022-06-10 в 13 54 25

PR #6733, opened 2022-05-30
NAS-115711 / VM perf tuning

heppu:
UI to for https://github.com/truenas/middleware/pull/9045


heppu:
For some reason the input fields in edit form are not picking up the correct styles and I'm not quite sure what I'm doing wrong there.

PR #6737, opened 2022-06-01
NAS-116347 / 22.12 / "View Logs" button in Task Manager shows filesystem path to log file but does not offer an ability to download them (like e.g. cloud sync task logs view does)

AlexKarpov98:
Testing:
Task Manager -> Job History

Added `Download Logs` button if job has logs


dszidi (commented):
@AlexKarpov98 Asked you a question via slack DM about how to test.

dszidi (approved):
The code looks good and behaves as expected.

PR #6739, opened 2022-06-01
NAS-116244 / 22.02.2 / Added alerts on dashboard and reports

RehanY147:
Check the alerts on the dashboard system info widget and on the reports page.


denysbutenko (approved):
image
image

If I set browser time in the future, the warning did not appear, don't know if it is needed.
Other than that looks good.

PR #6743, opened 2022-06-02
NAS-116134: Adding loader

RehanY147:
Go to Shares Dashboard -> Iscsi Targets -> Extents tab -> Add Extent. The loading should show up while UI is loading the devices list.

### DIFF
--- a/src/app/pages/sharing/iscsi/extent/extent-form/extent-form.component.ts
+++ b/src/app/pages/sharing/iscsi/extent/extent-form/extent-form.component.ts
@@ -4,6 +4,7 @@ import {
 } from '@angular/forms';
 import { Router, ActivatedRoute } from '@angular/router';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
+import { TranslateService } from '@ngx-translate/core';
 import * as _ from 'lodash';
 import { ExplorerType } from 'app/enums/explorer-type.enum';
 import { IscsiExtentType } from 'app/enums/iscsi.enum';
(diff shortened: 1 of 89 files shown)


src/app/pages/sharing/iscsi/extent/extent-form/extent-form.component.ts
```diff
@@ -292,7 +294,9 @@ export class ExtentFormComponent implements FormConfiguration {
     this.fieldConfig = entityForm.fieldConfig;
     const extentDiskField = _.find(this.fieldConfig, { name: 'disk' }) as FormSelectConfig;
     // get device options
+    this.loader.open(this.translate.instant('Loading devices. Please wait.'));
     this.iscsiService.getExtentDevices().pipe(untilDestroyed(this)).subscribe((res) => {
+      this.loader.close();
```
bvasilenko, line 299: I noticed that loader is hidden if API call succeeds. However, it will stay open if API call fails. I suggest adding an failure handler as a 2nd argument for `.subscribe`, for example:

```
(res) => {
        this.loader.close();
        new EntityUtils().handleWsError(this.entityForm, res);
      },
```

RehanY147: Good catch

undsoft (changes requested):
There is also a device dropdown in iscsi wizard. The ticket also talks that the list is truncated and doesn't show all items. We probably need to to have something like a combobox similar to src/app/pages/sharing/nfs/nfs-form/nfs-form.component.ts:113

undsoft (approved):
Looks okay, will be improved for Bluefin.

PR #6746, opened 2022-06-02
NAS-115740 / 13.0 / NAS-115740: Azure Custom Endpoint

undsoft:
Check that Cloud Credentials form under System has Endpoint field for Microsoft Azure provider and it gets correctly saved.


RehanY147 (approved):
Looks good to me.

PR #6748, opened 2022-06-03
NAS-115740 / 22.02.2 / NAS-115740: Azure Custom Endpoint

undsoft:
Check that Cloud Credentials form under System has Endpoint field for Microsoft Azure provider and it gets correctly saved.


undsoft:
Same as https://github.com/truenas/webui/pull/6746, will mere.

PR #6750, opened 2022-06-04
NAS-116519 / 22.12 / Allow using keychain SSH connections for rsync tasks

AlexKarpov98:
Testing:
Data protection -> Rsync task -> Set `Remote Mode` as `SSH`

Added a new field `Connect using:`
Reordered the inputs


bvasilenko (approved):
Works and looks good to me

PR #6752, opened 2022-06-05
NAS-116514 / 22.02.2 / Added warning

RehanY147:
Go to Apps -> Manage Catalogs (tab) -> Add Catalog
You should see a warning prompt.


src/app/helptext/apps/apps.ts
```diff
@@ -555,6 +555,24 @@ export default {
...
+    title: T('Warning'),
+    message: T(`iXsystems does not audit or otherwise validate the contents of third-party applications catalogs. \
+    It is incumbent on the user to verify that the new catalog is from a trusted source and that the third-party \
+    properly audits its chart contents. Failure to exercise due diligence can expose the user and their data to \
+    the some or all of the following:<br/>\
+    <ul>
+      <li>Malicious software</li>
+      <li>Broken services on TrueNAS host</li>
+      <li>Service disruption on TrueNAS host</li>
+      <li>Broken filesystem permissions on Host or within application</li>
```
Ornias1993, line 570: It might be best to remove this one.
As it gives the impression that iX-Systems does guarantee this, which is factually not the case (or not completely at the very least).

(To give insight: Bitnami and TrueCharts, for example, at least ship tools to have us manage the permissions of user-added storage, where iX does not)

Ornias1993 (changes requested):
Generally good advice, baring one caveat...

denysbutenko (approved):
image

The warning appears when trying to add a new catalog 👍🏻

src/app/helptext/apps/apps.ts
undsoft: @Ornias1993 Not sure I understand. What would make user think that ixsystems guarantees broken  filesystem permissions on Host or within application?

src/app/helptext/apps/apps.ts
```diff
@@ -555,6 +555,24 @@ export default {
...
     pulling: T('Pulling...'),
   },
+  thirdPartyRepoWarning: {
+    btnMsg: T('Continue'),
+    cancelMsg: T('Cancel'),
+    title: T('Warning'),
+    message: T(`iXsystems does not audit or otherwise validate the contents of third-party applications catalogs. \
+    It is incumbent on the user to verify that the new catalog is from a trusted source and that the third-party \
+    properly audits its chart contents. Failure to exercise due diligence can expose the user and their data to \
+    the some or all of the following:<br/>\
```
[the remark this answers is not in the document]
undsoft: It says '_can_ expose the user', so not guaranteed. Saying 'can expose to none' doesn't sound right.
How about we change 'can' to 'may'?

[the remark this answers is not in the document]
stavros-k: Right, `may` also works yea

src/app/helptext/apps/apps.ts
```diff
@@ -561,7 +561,7 @@ export default {
     title: T('Warning'),
     message: T(`iXsystems does not audit or otherwise validate the contents of third-party applications catalogs. \
     It is incumbent on the user to verify that the new catalog is from a trusted source and that the third-party \
-    properly audits its chart contents. Failure to exercise due diligence can expose the user and their data to \
+    properly audits its chart contents. Failure to exercise due diligence may expose the user and their data to \
     the some or all of the following:<br/>\
```
stavros-k, line 565:
```suggestion
    some or all of the following:<br/>\
```
I'm not an native English speaker, but it doesn't sounds correct to me.
If that's correct, please ignore this.

src/app/helptext/apps/apps.ts
william-gr: This statement is staying. Apps can have access to the filesystem and do whatever they want with it.

Ornias1993: I should've indeed been more clear @william-gr sorry.

What i meant was the last part:
`on Host or within application`

Primarily the "within applications" part.

I would suggest the following:
```suggestion
      <li>Broken filesystem permissions</li>
```

PR #6754, opened 2022-06-05
NAS-116447 / 22.02.2 / NAS-116447: Added new checkbox

RehanY147:
To test, open the `Alerts` drop down and go to the `Alerts Settings` by clicking the cog icon. On the next page, on the top right from the select, click on `Storage`. Now on the list of new inputs, you should see `Proactive Support` as a checkbox for some inputs. (Examples are described in the ticket). Values should be persistent once saved. Default value should be `true` if you've never touched it.

Also, these checkboxes aren't supposed to show for non-enterprise products so find a way around that. Either disable the enterprise check in the new code for testing or edit the enterprise string in `localStorage` or buy a license.

### DIFF
--- a/src/app/interfaces/alert.interface.ts
+++ b/src/app/interfaces/alert.interface.ts
@@ -29,6 +29,7 @@ export interface AlertClass {
   id: string;
   level: AlertLevel;
   title: string;
+  proactive_support?: boolean;
 }
 
 export interface AlertClasses {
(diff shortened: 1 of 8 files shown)


undsoft (approved):
I had one of the proactive checkboxes come from middleware as false by default. I'm assuming this is okay.

Otherwise, checkboxes were there and saved okay.
Снимок экрана 2022-06-06 в 12 05 33

PR #6756, opened 2022-06-06
NAS-116609 / 22.02.2 / add proper strings so enclosure mapping works

yocalebo:
This causes webUI to display the following error on SCALE MINI XL+
image

Since we have rebranded to TrueNAS, most of the systems we ship out the door are flashed with the `TRUENAS-` prefix in SMBIOS. I know basically 0 about typescript but I think this will fix it?

### DIFF
--- a/src/app/pages/system/view-enclosure/enclosure-disks/enclosure-disks-mini.component.ts
+++ b/src/app/pages/system/view-enclosure/enclosure-disks/enclosure-disks-mini.component.ts
@@ -48,15 +48,22 @@ export class EnclosureDisksMiniComponent extends EnclosureDisksComponent {
 
   createEnclosure(enclosure: EnclosureMetadata = this.selectedEnclosure): void {
     switch (enclosure.model) {
-      case ('FREENAS-MINI-3.0-E' || 'TRUENAS-MINI-3.0-E'):
-      case ('FREENAS-MINI-3.0-E+' || 'TRUENAS-MINI-3.0-E+'):
+      case 'FREENAS-MINI-3.0-E':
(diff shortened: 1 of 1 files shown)


undsoft (changes requested):
This won't work correctly.
`'string1 || 'string2'` in brackets will evaluate to `'string1'`
Has to be:
```
case 'FREENAS-MINI-3.0-E+':
case 'TRUENAS-MINI-3.0-E+':
// etc.
```

yocalebo:
ping @dszidi @undsoft

undsoft (changes requested):
Did you mean to remove `FREENAS-MINI-3.0-E+` and `FREENAS-MINI-3.0-X+`?

Also, you can fall-through:
```
case 'FREENAS-MINI-3.0-E':
case 'FREENAS-MINI-3.0-E+':
case 'TRUENAS-MINI-3.0-E':
case 'TRUENAS-MINI-3.0-E+':
     this.chassis = new MINI();
     break;
```

yocalebo:
> Did you mean to remove `FREENAS-MINI-3.0-E+` and `FREENAS-MINI-3.0-X+`?
> 
> Also, you can fall-through:
> 
> ```
> case 'FREENAS-MINI-3.0-E':
> case 'FREENAS-MINI-3.0-E+':
> case 'TRUENAS-MINI-3.0-E':
> case 'TRUENAS-MINI-3.0-E+':
>      this.chassis = new MINI();
>      break;
> ```

Woops, should be fixed.

undsoft (approved):
Thanks 👍🏻

PR #6757, opened 2022-06-06
NAS-115675 / 22.02.2 / NAS-115675: Fixed race condition

RehanY147:
This addresses the Storage widget not showing up. But pool specific widgets simply don't show up because you have to enable them on the dashboard via the config button at the top right. But this fixes the issue with the storage widget. Normally, if you navigate away to a different page and then come back to the dashboard without refreshing the page, the storage page wouldn't show up 99% of the time. This fixes the issue with that.


denysbutenko (approved):
Looks good

PR #6758, opened 2022-06-07
Fail to create manual ssh connection

AlexKarpov98:
Testing: 
System -> SSH Connections -> Create a SSH Connection with an existing SSH keypair

Fixed error in https://github.com/truenas/webui/commit/628008d8a9086df12f406fa528167aea1be54ff6


RehanY147 (approved):
Form submits successfully for both Manual and Semi-Automatic options

PR #6759, opened 2022-06-07
NAS-116627 / 22.12 / NAS-116627: Removing GlobalActions

undsoft:
Changes:
- Removes GlobalActions and old page title component.
- Makes tabs in Apps lazy loaded.

Testing:
- Toolbar on one of the pages that uses entity-table
- Toolbar in Terminal (for example in Shell and in app shell).
- Toolbar in enclosure UI (elements menu)
- Toolbar on all tabs in Apps, including interaction between page and toolbar.


src/app/pages/applications/select-pool-dialog/select-pool-dialog.component.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  private showNoPoolsWarning(): void {
+    this.dialogService.confirm({
+      title: helptext.noPool.title,
+      message: helptext.noPool.message,
+      hideCheckBox: true,
+      buttonMsg: helptext.noPool.action,
+    }).pipe(untilDestroyed(this)).subscribe((confirmed) => {
+      if (!confirmed) {
+        this.dialogRef.close(false);
```
AlexKarpov98, line 120:
```
this.dialogRef.close(false);
if (!confirmed) {
```
chrome-capture-2022-6-8

AlexKarpov98 (approved):
All the page title header and toolbar work well! 👍

PR #6761, opened 2022-06-07
NAS-116609 / 22.12 / add proper strings so enclosure mapping works

yocalebo:
Cherry-picked from https://github.com/truenas/webui/pull/6756


undsoft (approved):
Backport.

PR #6764, opened 2022-06-08
NAS-116450 / 22.02.2 / NAS-116450: Changing the number of the first slot from 1 to 0

bvasilenko:


yocalebo:
@bvasilenko thanks for looking into this but it seems clear that the webUI has always started at 1 right? If so, then I think this is probably a regression in middlewared would you agree?

yocalebo (changes requested):
You can close this. I've fixed this on the backend side here: https://github.com/truenas/middleware/pull/9168

PR #6765, opened 2022-06-08
Scrolling is not available in shell

AlexKarpov98:
Testing:
System -> Shell

Fill the screen with the output and check the scroll up/down.


RehanY147 (changes requested):
The scroll works on the shell. However, the page goes out of bounds now. Which means there are two scrolls working. That wasn't happening before.

RehanY147 (approved):
LGTM

PR #6772, opened 2022-06-09
NAS-116422 / 22.02.3 / NAS-116422: re-establish CPU/memory data stream if it's aborted by the server

bvasilenko:
We had a bug reported by a user:

The CPU and RAM usage on dashboard circles endlessly. If I go to reports they are reported fine.

Testing:
1. Enter **Dashboard** page. 
2. Switch to **Storage** page and switch back to **Dashboard** page.
3. No matter how many times you repeat step "2", every time you should end up with CPU/Memory metrics updated in realtime in Dashboard widgets


undsoft (changes requested):
We need to get to the bottom of why this happens. I suggest examining first if there is something wrong going on with frontend subscriptions (race condition, frontend unsubscribing, etc) and if that doesn't help creating a ticket for middleware.

bvasilenko:
Will create a ticket for middleware

PR #6775, opened 2022-06-09
NAS-116408 / None / NAS-116408: Permissions editor for new storage section

undsoft:
Mostly copies existing dataset permissions over.
Testing: select a dataset (some rows may not work, because they shouldn't be in the table), test dataset card and edit functionality.
For edit, 3 screens exist: trivial permissions, ACL for datasets with POSIX ACL type and ACL for datasets with NFS ACL type.

Known issues: styles are not finished and some TODO comments are not addressed. Will fix separately.


denysbutenko (approved):
Looks good to me

RehanY147 (approved):
Looks good.

RehanY147:
Although, the font size for these new cards looks too large relatively.

undsoft:
I agree, but it's based on design.

PR #6776, opened 2022-06-09
NAS-116661 / 22.02.2 / NAS-116661: Replaced column

RehanY147:
Go to Apps -> Manage Catalogs. Click on one of the table rows and you should get a pop up that has replaced the `Version` column with `Status` column and shows app status now.


AlexKarpov98 (approved):
LGTM! 👍

Screenshot_38

PR #6778, opened 2022-06-10
NAS-116461 / 13.0 / Unsetting "UPS - No Communication Warning Time" gives error

AlexKarpov98:
Testing:
Services -> UPS -> Unset `No Communication Warning Time`

Changes:
Set null if the value of type `number` input is empty string


undsoft (changes requested):
The fix version on the ticket was Scale. I don't want to merge it in 13, as it contains a change in one of the core components. There is a high possibility of unwanted consequences.
Please check if the fix is needed for Scale. If not, you can close the ticket. 
The only thing we can do for 13 is to fix it in ups component specifically without changing core components.

undsoft (approved):
👍🏻

PR #6779, opened 2022-06-10
NAS-116585 / 13.0 / Incorrect Sort by Date

AlexKarpov98:
Testing:
Storage -> Snapshots -> Sort by `Date Created`

Changes:
Updated format as matched locale service.


undsoft (approved):
It works, but when I set date format to `11/06/2022` in settings, I have to click on the column header twice to get it to resort.
Let's fix it, if it's a simple fix, otherwise ok to merge as is.

AlexKarpov98:
Fixed

PR #6783, opened 2022-06-13
NAS-116405 / 22.12 / Added details card widget

RehanY147:
Check the details card widget under Datasets page.
Check the new loading card info component. it should show a blinking/fade-in-fade-out animation
Check that the details card should only show up for dataset rows and not for pool rows
Also test the `Edit` and `Delete` buttons

### DIFF
--- a/src/app/interfaces/dataset.interface.ts
+++ b/src/app/interfaces/dataset.interface.ts
@@ -32,6 +32,10 @@ export interface Dataset {
   pool: string;
   readonly: ZfsProperty<boolean>;
   used: ZfsProperty<number>;
+  usedbychildren: ZfsProperty<number>;
+  usedbydataset: ZfsProperty<number>;
+  usedbyrefreservation: ZfsProperty<number>;
+  usedbysnapshots: ZfsProperty<number>;
   type: DatasetType;
   managedby: ZfsProperty<string>;
   aclmode: ZfsProperty<string>;
--- a/src/app/pages/applications/common-apps-toolbar-buttons/common-apps-toolbar-buttons.component.ts
+++ b/src/app/pages/applications/common-apps-toolbar-buttons/common-apps-toolbar-buttons.component.ts
@@ -96,11 +96,12 @@ export class CommonAppsToolbarButtonsComponent implements OnInit {
 
         const catalogAppInfo = {
           ...catalogApp,
-          id: ixChartApp,
-          train: officialCatalog,
+          catalog: {
+            id: officialCatalog,
+            train: chartsTrain,
+          },
           schema: catalogApp.versions[catalogApp.latest_version].schema,
         } as CatalogApp;
-
         const chartWizard = this.slideInService.open(ChartFormComponent, { wide: true });
         chartWizard.setChartCreate(catalogAppInfo);
       });
--- a/src/app/pages/applications/forms/chart-form/chart-form.component.spec.ts
+++ b/src/app/pages/applications/forms/chart-form/chart-form.component.spec.ts
@@ -4,6 +4,7 @@ import { MatDialog, MatDialogRef } from '@angular/material/dialog';
 import { createComponentFactory, mockProvider, Spectator } from '@ngneat/spectator/jest';
 import { mockCall, mockWebsocket } from 'app/core/testing/utils/mock-websocket.utils';
 import { CatalogApp } from 'app/interfaces/catalog.interface';
+import { ChartSchemaNodeConf } from 'app/interfaces/chart-release.interface';
 import { EntityJobComponent } from 'app/modules/entity/entity-job/entity-job.component';
 import { IxFormsModule } from 'app/modules/ix-forms/ix-forms.module';
 import { ChartFormComponent } from 'app/pages/applications/forms/chart-form/chart-form.component';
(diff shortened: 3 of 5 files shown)


undsoft (approved):
Nice, but I think we can go even more granular in some cases.
For example here: https://jira.ixsystems.com/browse/NAS-116416 most of the content is static with only `Password is set` being a dynamic string. We could show loading indication for those strings specifically.
Also something to think about is error handling. If request for Global SED Password fails, it may be ? a good idea to show error inline with a retry button.

Approving because what you have here may be useful for some cards.

dszidi:
So far looks pretty good. 

 > Check that the details card should only show up for dataset rows and not for pool rows

 Why only show details for dataset rows? The first row is the root dataset and should also be showing details

src/app/pages/datasets/components/dataset-details-panel/dataset-details-panel.component.html
```diff
@@ -1,7 +1,8 @@
-<h3>Details for {{ dataset.name }}</h3>
+<h3 class="title">Details for {{ dataset.name }}</h3>
 
 <div class="cards">
   <ix-permissions-card [dataset]="dataset"></ix-permissions-card>
+  <ix-dataset-details-card *ngIf="dataset.id !== dataset.pool" [dataset]="dataset"></ix-dataset-details-card>
```
dszidi, line 5: This condition should be removed. The root dataset has the same name as the pool and has details we need to show

RehanY147:
Showing `1TiB of 3TiB` space data doesn't work for ZFS according to middleware. We can show used or available space but not total space. @dszidi @undsoft Should we remove that attr completely or change it to show one of the two options possible?

anodos325:
> Showing `1TiB of 3TiB` space data doesn't work for ZFS according to middleware. We can show used or available space but not total space. @dszidi @undsoft Should we remove that attr completely or change it to show one of the two options possible?

As mentioned in the ticket, I believe that the Dataset Details should include breakout of used stats. Example:
```
used: 776K (dataset: 144K, snapshots: 320K, children: 312K)
```

We'll update the `pool.dataset.query` return to include this info.

dszidi:
@RehanY147 @undsoft Unfortunately this branch is not compiling for me so I can't really comment on the animation issues.

Wrt the total storage stat. We can possibly break things down as @anodos325 suggested but put it in the capacity management card instead of the dataset details card. The layout is already busting at the seams, so the only way I can see that working is by ditching the little chart and going text only.

As for the overview section, perhaps we can just go with the used stat since that is what we highlight in the storage dashboard usage card. Users are probably most concerned with hitting that 80% usage mark. Perhaps it would be wise to prioritize that in the datasets overview as well. This would also cut down on screen real estate

@undsoft @RehanY147 @anodos325 what do you think? ^^

anodos325:
> @RehanY147 @undsoft Unfortunately this branch is not compiling for me so I can't really comment on the animation issues.
> 
> Wrt the total storage stat. We can possibly break things down as @anodos325 suggested but put it in the capacity management card instead of the dataset details card. The layout is already busting at the seams, so the only way I can see that working is by ditching the little chart and going text only.
> 
> As for the overview section, perhaps we can just go with the used stat since that is what we highlight in the storage dashboard usage card. Users are probably most concerned with hitting that 80% usage mark. Perhaps it would be wise to prioritize that in the datasets overview as well. This would also cut down on screen real estate
> 
> @undsoft @RehanY147 @anodos325 what do you think? ^^

I know from forums interactions there have been many cases where users are quite confused about why their USED is really high even after deleting a bunch of files. Showing how it's used may be incredibly helpful for them and also avoid some Tier 1 support calls.

RehanY147:
I will remove remnants of custom info loading components if we like the `ngx-skeleton-loader`.

undsoft (approved):
I think we should keep the 3rd party package and adjust animation.
For example, setting opacity: 0.25 makes it better for me on dark theme.
We should also consider extracting as much as possible to `forRoot`. `findTheme` can become `var(--alt-bg2)`

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
```diff
@@ -0,0 +1,66 @@
...
+          <span *ngIf="!loading || !!datasetCompression">{{ datasetCompression }}</span>
+          <ngx-skeleton-loader
+            *ngIf="loading && !datasetCompression"
+            class="loader-skel" 
+          ></ngx-skeleton-loader>
+        </div>
+        <div class="row">
+          <span class="header">Enable Atime: </span>
+          <span *ngIf="!loading || !!dataset.atime?.value">{{ dataset.atime?.value }}</span>
+          <ngx-skeleton-loader
```
denysbutenko, line 35: It makes sense to wrap `ngx-skeleton-loader` in a component with common styles and etc

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
```diff
@@ -0,0 +1,66 @@
+<div>
+    <mat-card class="container-card">
+      <mat-card-header>
+        <h3 mat-card-title>{{ 'Dataset Details' | translate }}</h3>
+        <button mat-button (click)="editDataset()">{{ 'Edit' | translate }}</button>
+      </mat-card-header>
+      <mat-card-content>
+        <div class="row">
+          <span class="header">Type: </span>
```
denysbutenko, line 9: Missing translate pipe on labels

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
undsoft: Let's fix this later. Extracting to `forRoot` should be good enough for now.

PR #6789, opened 2022-06-14
NAS-116727 / 22.12 / Remove unnecessary dependencies

sonicaj:
This commit adds changes to install yarn package as that will effectively bring in latest nodejs and install npm as well. No need to define it separately. Also we are moving away from debian yarnpkg as it brings legacy nodejs dependencies which we have no use of as we have moved on to newer nodejs version.


undsoft (approved):
👍🏻

PR #6793, opened 2022-06-15
NAS-116718 / 22.12 / Not all fields are rendered in Launch Docker Image

AlexKarpov98:
Testing:
1. Go to Apps in Angelfish and Bluefin.
2. Click Launch Docker Image on both.
3. Compare the two and Check if there is missing field.


RehanY147 (approved):
Looks good to me. All the same fields are loading on older and newer versions. Still not see the `Restart` field mentioned in the older or newer versions. Wonder what that's about.

undsoft:
Please also add a unit test case to cover this issue.

AlexKarpov98:
> Looks good to me. All the same fields are loading on older and newer versions. Still not see the `Restart` field mentioned in the older or newer versions. Wonder what that's about.

image
'Restart' field set to hidden `true`.

PR #6801, opened 2022-06-16
NAS-116699 / 22.12 / [SCALE] Downloading Debug not working & freezes the GUI

AlexKarpov98:
Testing:
System -> Advanced -> Click `Save Debug` button

Added failure pipe.


RehanY147 (changes requested):
We should enable the abort button on that job dialog. If it's taking too long it can seem like it's frozen. The user should be able to abort the process and get back to the UI. As is, the dialog forces the user to stay on this screen until it's done as there's no way to close it.

RehanY147:
`streamDownloadFile` is the method that seems to be crashing. Should have error checking there.

AlexKarpov98:
It is unable to abort `system.debug` job.
Added error report for HTTP requests failing.

RehanY147 (changes requested):
We should add the to the warning shown to the user. `This operation might take a long time. It cannot be aborted once started. It may seem like the UI is stuck at times. Proceed?` Or something to that effect.

AlexKarpov98:
Updated warning text to `This operation might take a long time. It cannot be aborted once started. Proceed?`

RehanY147 (commented):
I think the issue here is also that since we're showing a progress bar, the job goes to 80% and stays there for a while. That means neither the percentage is updating nor the progress bar is moving. And you can't close the dialog either. Which makes it seem like the UI is stuck. We should at least show some indication that we are still working on things. One way of doing that could be making the progress bar indetermined. Or show another warning if the job takes longer than a certain number of seconds for example that reads "We're still working on it...". Something that shows the user that the UI is not stuck. Because for all intents and purposes it does feel like the UI is stuck. @undsoft thouhgts?

AlexKarpov98:
We have a warning `This operation might take a long time` before proceeding.
The main problem with this ticket was that the UI didn't work at all when got an Http error:
Screenshot_75

Added error handlers for HTTP/job failure on PR.

RehanY147 (approved):
Looks good

PR #6803, opened 2022-06-17
NAS-114179 / 22.12 / NAS-114179: Adding checkbox "inherit domain from DHCP"

bvasilenko:
Testing:

1. On the **Network** page , open **Common settings** form
2. Locate **domain** and **inherit domain from DHCP lease** checkbox.
3. Adjust it's checked state and verify saving & loading


src/app/helptext/network/configuration/configuration.ts
```diff
@@ -13,6 +13,9 @@ export default {
   hostname_virtual_tooltip: T('When using a virtual host, this is also \
  used as the Kerberos principal name.'),
 
+  inherit_dhcp_placeholder: T('Inherit domain from DHCP'),
+  inherit_dhcp_tooltip: T('When this checkbox is checked, domain inherit from DHCP.'),
```
undsoft, line 17: `domain will be inherited` or `domain is inherited`.

undsoft (approved):
Looks good, probably a good idea to update `NetworkConfigurationCardComponent` to also handle the case when domain is inherited.

PR #6807, opened 2022-06-18
[SCALE] GUI displays wrong warning about not matching time

AlexKarpov98:
Testing:
Dashboard -> System Information
Try visit other pages, then back to Dashboard page, check if time is correct.

PS: Removed the Shell page scroll


src/app/pages/dashboard/components/widget-sys-info/widget-sys-info.component.ts
```diff
@@ -106,7 +106,7 @@ export class WidgetSysInfoComponent extends WidgetComponent implements OnInit, O
         this.ha_status = haStatus.status;
       });
     } else {
-      this.store$.pipe(waitForSystemInfo, untilDestroyed(this)).subscribe(
+      this.ws.call('system.info').pipe(untilDestroyed(this)).subscribe(
```
RehanY147, line 109: Instead of making a new call, I think it would be better if we store a timestamp of when the original call came back with a response. And have another selector that adds the difference to the nas time stored in state.
Something like
`this.store$.select(addedNasTime())` or something.

RehanY147 (changes requested):

denysbutenko (changes requested):
The warning appears if browser time and NAS time do not match after I synced browser time the warning is still visible even if reload the page

AlexKarpov98:
> The warning appears if browser time and NAS time do not match after I synced browser time the warning is still visible even if reload the page

Thanks for feedback. Could you please attach the screenshots?

AlexKarpov98:
@denysbutenko Changing the time, Chrome syncs new Date() on the computer after 30-60 seconds.
IE or Firefox works fine. What do you think of the delay?

denysbutenko:
image
image
I'm using docker image with cloud vm for testing.

denysbutenko (approved):
Works well 👍🏻 Good job

RehanY147 (approved):
Looks good

PR #6809, opened 2022-06-19
NAS-116785: Enforce code style on property members

undsoft:
No testing is necessary. Code review only.


RehanY147 (approved):
Looks good to me

PR #6810, opened 2022-06-20
NAS-116407 / None / NAS-116407: ZFS Encryption Card

undsoft:
Also adds support for routing and storing dataset path in URL.

Testing: Test ZFS Encryption card in a dataset.


denysbutenko (approved):
image
Looks nice

PR #6813, opened 2022-06-20
NAS-116795 / 22.12 / NAS-116795: translations

RehanY147:


src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
```diff
@@ -6,47 +6,47 @@ <h3 mat-card-title>{{ 'Dataset Details' | translate }}</h3>
       </mat-card-header>
       <mat-card-content>
         <div class="row">
-          <span class="header">Type: </span>
+          <span class="header">{{ 'Type: ' | translate }}</span>
```
denysbutenko, line 9: I believe semicolons and spaces are better to keep outside of the translatable string. More chances to be translated

denysbutenko (approved):
image

Generated translation files show that there exist similar strings with a small difference.
Since neither have translations yet I don't mind.

PR #6820, opened 2022-06-23
NAS-116828 / 22.12 / Fix styles for Hardware Disk Encryption card

AlexKarpov98:
image


RehanY147 (changes requested):
Let's also fix the loading indication. Right now, it just shows text `Loading...`

RehanY147 (approved):
Nice!

PR #6821, opened 2022-06-23
NAS-116398 / 22.12 / Unused Resources component for Pools Dashboard

AlexKarpov98:
image


src/app/pages/storage2/components/pools-dashboard/pools-dashboard.component.html
```diff
@@ -1,6 +1,29 @@
+<ng-template #pageHeader>
+  <ix-page-title-header>
+    <a
+      mat-button
+      color="primary"
+      [matTooltip]="'Create a new pool.' | translate"
+      ix-auto
+      ix-auto-type="button"
+      ix-auto-identifier="_POOL_CREATE"
```
undsoft, line 9: ix-auto identifiers and matTooltip can be removed.

src/app/pages/storage2/components/unassigned-disk/manage-unassigned-disk-dialog/manage-unassigned-disk-dialog.component.ts
```diff
@@ -0,0 +1,26 @@
+import { Component, Inject } from '@angular/core';
+import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
+import { Router } from '@angular/router';
+import { ManageUnassignedDiskDialogResource } from './manage-unassigned-disk-dialog.interface';
+
+@Component({
+  templateUrl: './manage-unassigned-disk-dialog.component.html',
+  styleUrls: ['./manage-unassigned-disk-dialog.component.scss'],
+})
```
undsoft, line 9: changeDetection.OnPush

src/app/pages/storage2/components/unassigned-disk/unassigned-disk.component.ts
```diff
@@ -0,0 +1,40 @@
...
+  ManageUnassignedDiskDialogComponent,
+} from 'app/pages/storage2/components/unassigned-disk/manage-unassigned-disk-dialog/manage-unassigned-disk-dialog.component';
+import { WebSocketService } from 'app/services';
+
+@UntilDestroy()
+@Component({
+  selector: 'ix-unassigned-disk',
+  templateUrl: './unassigned-disk.component.html',
+  styleUrls: ['./unassigned-disk.component.scss'],
+})
```
undsoft, line 16: onPush

RehanY147 (changes requested):
Instead of listing down the pools in a single column table, I think a select with the pool options followed by a button would be a better. Also, since this does not need much vertical space, giving it its own column in the model will waste space.
Furthermore, if you don't intent to add disks to pool, showing the pool selection is confusing and unnecessary. I think it should only be shown if the user clicks a button "Add to Pool" similar to "Create Pool". We can maybe show another confirm dialog that takes the pool selection before proceeding. Or conditionally show the pool select from a checkbox maybe.
"Add disks to existing pool" that would show the pool select and update the button to say "Add disks" instead of "Create a Pool"

AlexKarpov98:
Screenshot_100

RehanY147 (changes requested):
Unused heading shows when there's nothing to show
image

Furthermore, when you have many disks, the view doesn't work.
disks issue

We should limit disks list view to a portion of the screen and have a nested scroll. The bottom checkboxes should be always visible. And let's replace the radio button list of pools with a select drop down.

AlexKarpov98:
image

RehanY147 (approved):
Looks good. Nice work!

undsoft (approved):
I don't know if I fully like how Unused Disks modal looks, but it's probably more of a question to Damian.
Good to merge.

dszidi (commented):
Just from looking at the images posted so far, I would suggest that the unassigned disks should be shown as an inventory. The current approach would have us listing potentially hundreds of meaningless disk labels. They should instead be broken down by type and capacity. 
For example:

### Unused Disks

- 2 X 2 TB NVME
- 10 X 4 TB SSD
- 12 X 12 TB HDD

PR #6823, opened 2022-06-23
NAS-116871 / 22.12 / Toggle tree nodes recursively on alt+click

denysbutenko:
Like in Chrome Devtools.
For testing open the new Datasets page and try to toggle tree nodes with Alt(Option) key pressed


undsoft (approved):
A bit niche functionality, I wonder how many people will figure it out.
But we can keep. It would be good to have something like a test case for this or something to remind us in future why this code is here.

PR #6825, opened 2022-06-23
NAS-116854 / 22.12 / NAS-116854: Import pool and Create pool buttons

undsoft:
Buttons in header in new Storage Dashboard.
Most of it is just a copy from previous storage.


denysbutenko (approved):
It works well 👍🏻

PR #6832, opened 2022-06-28
NAS-116929 / 22.12 / NAS-116929: Fix the new refactored chart form

bvasilenko:


undsoft:
Can we have a unit test to cover this issue?

RehanY147 (approved):
Getting the same fields as old form. Nice

bvasilenko:
@undsoft I am going to write more tests in a separate PR

PR #6834, opened 2022-06-28
NAS-116397 / 22.12 / NAS-116397: Disks page

undsoft:
Mostly just ports existing disk pages. You can access them via Storage2 -> Manage Disks.


undsoft:
Couldn't reproduce. I've added something that may fix the issue.
In either case, this code is not new - this issue should exist in old smart test results page.

dszidi (commented):
> ... this code is not new - this issue should exist in old smart test results page.

You're right. It is also available in current master branch. I pulled in your latest changes and it seems there is an error with one of the sass files

image

undsoft:
Let's try again.

dszidi (approved):
LGTM

PR #6835, opened 2022-06-28
NAS-116334: Add search to nested tree datasource

denysbutenko:
Add support to filter data based on string query for IxNestedTreeDataSource

For testing go to `Dataset Management` or `Devices` page and try it out.

Previous iterations:

- [Search](https://github.com/truenas/webui/tree/NAS-116334-2)
- [Dynamic filters](https://github.com/truenas/webui/tree/NAS-116334-2)

### DIFF
--- a/src/app/modules/ix-tree/utils/find-in-tree.utils.ts
+++ b/src/app/modules/ix-tree/utils/find-in-tree.utils.ts
@@ -1,0 +1,19 @@
+export function findInTree<T extends { children?: T[] }>(
+  items: T[],
+  predicate: (item: T) => boolean,
+): T {
+  if (!items?.length) {
+    return;
+  }
+
+  for (const item of items) {
+    if (predicate(item)) {
+      return item;
+    }
+
+    const child = findInTree(item.children, predicate);
+    if (child) {
+      return child;
+    }
+  }
+}
--- a/src/app/modules/common/search-input/search-input.component.scss
+++ b/src/app/modules/common/search-input/search-input.component.scss
@@ -2,20 +2,23 @@
   align-items: center;
   background: rgba(0, 0, 0, 0.1);
   border: 1px solid var(--lines);
-  color: white;
+  color: var(--fg1);
   display: flex;
   flex-direction: row;
-  padding: 3px;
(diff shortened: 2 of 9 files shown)


src/app/modules/ix-tree/utils/find-in-tree.utils.ts
```diff
@@ -0,0 +1,19 @@
+export function findInTree<T extends { children?: T[] }>(
```
undsoft, line 1: `findInTree` in datasets needs to be removed now and replaced with this.

undsoft (approved):
It works well, but I do wonder if we should show all parent rows as well. Especially, when we will have separate rows for pools.
@dszidi What do you think?

https://user-images.githubusercontent.com/1481270/176255712-2ebe8573-0640-43b7-89f5-4a95c3331b8e.mov

denysbutenko:
> It works well, but I do wonder if we should show all parent rows as well. Especially, when we will have separate rows for pools.

That's a good point. Initially, I did something like this in the previous iterations. The approach was based on an additional wrapper for nodes to provide `children` and `parent` objects and some property for node visibility. While working on the current approach it's possible to show the root dataset that matches the filter criteria. The problem is how to notify the parent component where `treeControl` to handle which node needs to be expanded so it can be visible. If that's important I can do more trying to achieve it.


Previous one:

https://user-images.githubusercontent.com/351613/176367268-055931cd-19db-484b-bcd4-9221a763dbb2.mp4

Current one:

https://user-images.githubusercontent.com/351613/176367372-1bc948c6-88e1-461c-894f-7f282705963e.mp4

RehanY147 (approved):
I like the current approach. As the dataset path is available in the dataset details. So, if they really need to look at the path, they can look at that. Showing the hierarchy in the search kind of goes against the expected result by the user in my opinion.

denysbutenko:
> I like the current approach. As the dataset path is available in the dataset details. So, if they really need to look at the path, they can look at that. Showing the hierarchy in the search kind of goes against the expected result by the user in my opinion.

Okay, sounds good. We can extend search results later.

PR #6838, opened 2022-06-29
NAS-116799 / 22.12 / NAS-116799: Fixed release_name

RehanY147:
Test the PR on the team cloud scale vm. The `home` app on there has the issue mentioned on the ticket. Before the app name wasn't filled in when the app was being edited. That should be fixed.


bvasilenko (approved):
Works good. **Save** button is clickable now.

image

PR #6839, opened 2022-06-29
NAS-116973 / 22.12 / NAS-116973: Expand pool button

undsoft:
Testing: Expand button a pool in new storage dashboard.


AlexKarpov98 (approved):
LGTM! 👍

I think the header of the confirmation dialog should have the pool name like Export.
Screenshot_103

PR #6840, opened 2022-06-29
NAS-116247 / 22.12 / NAS-116247: added update-dialog component

Raz-Dva:
Test the new component UpdateDialogComponent.
During the update, a popup has been added instead of a modal window to display the status of the job.


dszidi (approved):
LGTM

image

Raz-Dva:
While the update is downloading, the "Update in Progress" button is visible on the topbar panel. By clicking on this button, we should see popover with progress instead of old modal. Could you check this popover too. 

image

dszidi (approved):
Looks good

image

PR #6842, opened 2022-06-30
NAS-116937 / 22.12 / Locking a dataset doesn't refresh a tree

AlexKarpov98:
Testing:
Datasets -> Select a Dataset (password protected) -> ZFS Encryption Card -> Click `Lock/Unlock` button
Check if refresh a tree by locking the dataset.


denysbutenko (changes requested):
It works when I click `Unlock` dataset is displayed as `Unlocked`. Then if I click `Lock` again it still displays `Unlocked` in the tree

denysbutenko (approved):
Sorry was in the wrong tab. It works as described 👍🏻

PR #6844, opened 2022-06-30
NAS-116393: new WidgetTopologyComponent

bvasilenko:


AlexKarpov98 (commented):
Screenshot_105
Screenshot_104

The layout looks broken.
I think it's better to use the standard styled Card template instead of the current extended `WidgetComponent`.
Please refer `ZfsHealthCardComponent`, and mixins in `cards.scss`.

undsoft:
Can we call it TopologyCard? We used -Card naming convention for other things.

RehanY147 (changes requested):
Let's also remove the empty card with the `Manage Devices` button.

AlexKarpov98 (changes requested):
image

A gap between an icon and a text isn't proper.

AlexKarpov98 (changes requested):
A prior issue(margin between icon and text) was fixed. LGTM!:thumbsup:

- The color and symbol are wrong in the state of a card icon.
A health icon is correct.
The rest icons was wrong like a screenshort.
Screenshot_42

- Unit tests are failed
This `mockCall('disk.query', [])` need to called in the mock websockets test, I think

bvasilenko:
@AlexKarpov98 

> A prior issue(margin between icon and text) was fixed. LGTM!👍
> 
> * The color and symbol are wrong in the state of a card icon.
>   A health icon is correct.
>   The rest icons was wrong like a screenshort.
>   Screenshot_42
> * Unit tests are failed
>   This `mockCall('disk.query', [])` need to called in the mock websockets test, I think

Fixed!
Would you please take a look and approve if it's ok

bvasilenko:
@RehanY147 

> Let's also remove the empty card with the `Manage Devices` button.

Done! Would you please take another look and approve if no other issues

RehanY147 (changes requested):
image
There's a slight misalignment there. The header is top aligned but the text on the right is center aligned.
Also, the msg `VDevs are missing` gives the wrong impression I think. It implies that the VDev was once available but isn't anymore. Which means there's an issue and we should show more of an indication of that. Like some red cross or a warning icon at least. But if this isn't an error, we should change the msg to say something like `VDevs not assigned` or something similar.

AlexKarpov98 (approved):
Issues were fixed.
Everything is looks good.

Screenshot_14

bvasilenko:
> image There's a slight misalignment there. The header is top aligned but the text on the right is center aligned. Also, the msg `VDevs are missing` gives the wrong impression I think. It implies that the VDev was once available but isn't anymore. Which means there's an issue and we should show more of an indication of that. Like some red cross or a warning icon at least. But if this isn't an error, we should change the msg to say something like `VDevs not assigned` or something similar.

@RehanY147 I've made the suggested improvements. Would you please take a look and approve if it's ok?

RehanY147 (approved):
The rest looks good. The messaging `VDevs are missing` should be changed. Maybe `NONE` or `Not Assigned`. @dszidi  thoughts?

PR #6845, opened 2022-06-30
NAS-116974 / 22.12 / NAS-116974: Add Dataset and Add Zvol buttons

undsoft:
Corresponding button appear on Datasets page. Both forms are just a copy of what we previously had.

Known issue: after dataset or zvol is added, it doesn't appear in tree table without reload. This will be handled in a separate ticket.


bvasilenko (approved):
Works good!

image

image

Tested with top-level datasets and nested datasets/Zvols

PR #6848, opened 2022-07-01
NAS-116937 / 22.12 / Locking a dataset doesn't refresh a tree

AlexKarpov98:


src/app/pages/datasets/components/dataset-management/dataset-management.component.ts
```diff
@@ -26,7 +26,6 @@ export class DatasetsManagementComponent implements OnInit {
   treeControl = new NestedTreeControl<Dataset, string>((dataset) => dataset.children, {
     trackBy: (dataset) => dataset.id,
   });
-  readonly trackByFn: TrackByFunction<Dataset> = (_, dataset) => dataset.id;
```
denysbutenko, line 29: ?

AlexKarpov98: The idea of 'trackBy' is to reinitialize the items need in the list and avoid reinitializing items don't need. Just because an item has a unique ID doesn't mean it's suitable for use in the `trackBy` function. it's just to tell the angle when an item needs to be re-rendered, i.e. its lifecycle hook needs to be re-executed. we need to be able to change the content even if Id  reamains the same, so I removed it.

undsoft (changes requested):
So, sounds like `reloadDatasetDetails` in `DatasetDetailsCardComponent` from previous PR isn't needed anymore and you can rely on `dataset` `@Input`.

undsoft (approved):
Looks good.

PR #6849, opened 2022-07-01
NAS-116414 / 22.12 / ZFS Info Card and Tree Node Groups

denysbutenko:
Work in progress. Appreciate feedback.

Preview:
ZFS Info Card:
image

 Tree Node Groups:
image

image


src/app/interfaces/storage.interface.ts
```diff
@@ -17,6 +17,7 @@ export interface VDev {
   stats: VDevStats;
   device?: string;
   disk?: string;
+  name?: string;
```
undsoft, line 20: I would prefer to keep backend interfaces clean and add extra fields to a new interface.

undsoft (commented):
Don't think we need to show empty groups like Log, Cache, etc.

src/app/interfaces/storage.interface.ts
denysbutenko: When VDev has type VDevType.Mirror then name property is present in the API response

denysbutenko:
New PR is #6855

PR #6850, opened 2022-07-01
NAS-116805 / 22.12 / NAS-116805: Remove code duplication for product images

Raz-Dva:
Created ProductImageService for select product image.
Check image render for different products in WidgetSysInfoComponent and SupportComponent

image
image


denysbutenko (approved):
Looks good

PR #6852, opened 2022-07-03
NAS-117036 / 22.12 / NAS-117036: Upgrade pool button

undsoft:
Testing:
`zpool create -o compatibility=legacy tank /dev/sdx`
where sdx is one of the unused disks.
`zpool export tank`
Import pool via UI and use Upgrade button on pools dashboard.

Pool will likely become offline after this one. This is likely middleware and unrelated to this PR.


bvasilenko (approved):
Works good!

image

image

image

PR #6853, opened 2022-07-04
NAS-117052 / 22.12 / NAS-117052: Refactor Create ACME Certificate form

undsoft:
Testing:
Go to Certificates, create an ACME DNS Authenticator, create a Certificate Signing Request, use a button next to a CSR to create an ACME certificate.
It's unlikely that middleware will let you through, but you can compare network requests.


dszidi (approved):
Code looks good. Form works and sends the correct payload. LGTM

image

PR #6854, opened 2022-07-04
NAS-114866 / 22.12 / NAS-114866: Fix text alignment

Raz-Dva:
The text is aligned.
Added value total raw data.

Go to Storage -> Create Pool
Check  text "Estimated total raw data capacity:"


undsoft (approved):
Looks good.

PR #6855, opened 2022-07-04
NAS-116414: ZFS Info Card

denysbutenko:
image

For testing go to Devices page and select an disk


src/app/pages/storage2/modules/devices/components/zfs-info-card/zfs-info-card.component.ts
```diff
@@ -0,0 +1,178 @@
...
+    private slideInService: IxSlideInService,
+    private dialogService: DialogService,
+    private matDialog: MatDialog,
+    private translate: TranslateService,
+    private devicesStore: DevicesStore,
+  ) {}
+
+  ngOnInit(): void {
+    this.poolId = +this.route.snapshot.paramMap.get('poolId');
+    this.slideInService.onClose$?.pipe(untilDestroyed(this)).subscribe(() => {
```
RehanY147, line 83: This will likely always reload disk even if the model is closed by clicking cancel/cross button. Let's make sure all instances of `sliceInService.close()` in the form return `true` for successful changes and then check here for that boolean. So we can only reload when required and not always.

src/app/pages/storage2/modules/devices/components/zfs-info-card/zfs-info-card.component.ts
```diff
@@ -0,0 +1,178 @@
...
+      title: this.translate.instant('Offline Disk'),
+      message: this.translate.instant('Offline disk {name}?', { name: this.disk.devname }),
+      buttonMsg: this.translate.instant('Offline'),
+    }).pipe(
+      filter(Boolean),
+      untilDestroyed(this),
+    ).subscribe(() => {
+      this.ws.call('pool.offline', [this.poolId, { label: this.topologyItem.guid }]).pipe(untilDestroyed(this)).subscribe(
+        () => {
+          this.devicesStore.reloadList();
```
RehanY147, line 104: These subscribe and error call backs are same in `onOnline`, `onOffline` and `onDetach`. Let's create named functions vars and use them as call backs to avoid redundancy.

RehanY147 (changes requested):
When I detach a disk from a mirror, the table doesn't reload. Also, the details pain continues to show old disk data that was removed. Also, when detach action is confirmed, there's no loading indication for a while. Nothing happens until the operation is complete. This might be a change detection issue.

src/app/pages/storage2/modules/devices/components/zfs-info-card/zfs-info-card.component.ts
denysbutenko: Sounds reasonable but it seems there are no places where it is implemented. I'll keep consistency with DiskInfoWidget and other forms for now.

src/app/pages/storage2/modules/devices/components/zfs-info-card/zfs-info-card.component.ts
denysbutenko: Could you show an example?

denysbutenko:
> When I detach a disk from a mirror, the table doesn't reload. 
Added a call to reload data on success

> Also, the details pain continues to show old disk data that was removed. 

Seems this issue not related to ZFS Info Card

> Also, when detach action is confirmed, there's no loading indication for a while. Nothing happens until the operation is complete. This might be a change detection issue.

Added loading indicator for operations

RehanY147:
> Seems this issue not related to ZFS Info Card
Since, the remove/detach actions were introduces by the ZFS Info card, I think it should be handled there. Let me know if you'd like to discuss this.

RehanY147 (changes requested):
You can add the following filter pipe to `this.slideInService.onClose$.pipe(...)` in both `DiskInfoComponent` and `ZfsInfoComponent` to avoid loading disks even when the slide in was closed without any changes.
`filter((value) => !!value.response && value.modalType === DiskFormComponent)`

denysbutenko:
> You can add the following filter pipe to `this.slideInService.onClose$.pipe(...)` in both `DiskInfoComponent` and `ZfsInfoComponent` to avoid loading disks even when the slide in was closed without any changes.
> `filter((value) => !!value.response && value.modalType === DiskFormComponent)`

Ok, added

denysbutenko:
> > Seems this issue not related to ZFS Info Card
> 
> Since, the remove/detach actions were introduces by the ZFS Info card, I think it should be handled there. Let me know if you'd like to discuss this.

Ok, fixed

src/app/pages/storage2/modules/devices/components/zfs-info-card/zfs-info-card.component.ts
denysbutenko: It wont work, keep as is

RehanY147 (approved):
Looks good. But let's also capitalize the `mirror-0` for `Parent` attribute. That's also capitalized on the pool status page on old storage page.

PR #6861, opened 2022-07-06
NAS-116916 / 22.12 / NAS-116916: 'status' column at pools table

bvasilenko:
Testing:

Navigate to new **Storage 2.0** page to see a "Status" column in the `ix-entity-tree-table` at the bottom of the page


RehanY147 (changes requested):
@bvasilenko This issue is meant for the Devices page. The table that needs changes is on the devices page that is accessed by clicking the "Manage Devices" button on the "Storage 2.0" page.

bvasilenko:
@RehanY147 I've re-applied my changes to the Devices page. Would you please make another iteration of testing?

src/app/pages/storage2/modules/devices/components/disk-node/disk-node.component.scss
```diff
@@ -0,0 +1,73 @@
...
+      padding: 5px;
+    }
+  }
+}
+
+// TODO: Fragile (at least with css selectors)
+:host-context(ix-tree-node:hover),
+:host-context(ix-tree-node.selected),
+:host-context(.disk-nested-tree-root-node:hover),
+:host-context(.disk-nested-tree-root-node.selected) {
```
undsoft, line 47: I think something got broken here. It used to be that when table got smaller and first item overlapped  other cells, it would have dark background with gradient.
You can compare how it is in datasets.
Снимок экрана 2022-07-11 в 23 25 52

src/app/pages/storage2/modules/devices/components/devices/devices.component.ts
```diff
@@ -104,8 +104,22 @@ export class DevicesComponent implements OnInit {
           tap((disks) => {
             this.diskDictionary = _.keyBy(disks, (disk) => disk.devname);
             this.topology = pools[0].topology;
-            this.treeControl.dataNodes = this.topology.data;
-            this.createDataSource(this.topology.data);
+            const data = { children: this.topology.data, type: 'Data VDEVs', guid: 'data' } as unknown as VDev;
```
undsoft, line 107: Needs proper typing. This is not really a VDev, but a separate thing in tree table, so we need to introduce another type and replace VDev in `NestedTreeControl`.

Also method starts to get fat, time to extract some logic to a separate method, e.g. `createDataSource`.

undsoft (changes requested):
1. Capacity and errors should be reported on Vdevs for their children. This only happens sometimes now.
2. Cache and spare groups have icons  - I don't think they should.
Снимок экрана 2022-07-11 в 23 23 28
3. Time to add unit tests to disk-node.

src/app/pages/storage2/modules/devices/components/devices/devices.component.ts
denysbutenko: Probably type should be an enum

src/app/pages/storage2/modules/devices/components/disk-node/disk-node.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  get diskErrors(): string {
+    if (this.disk.stats) {
+      const errors = this.disk.stats?.checksum_errors + this.disk.stats?.read_errors + this.disk.stats?.write_errors;
+      return errors ? `${errors} Errors` : this.translate.instant('No Errors');
+    }
+    return '';
+  }
+
+  get diskIcon(): string {
+    let icon = '';
```
denysbutenko, line 54: `return early` approach is preferred

src/app/pages/storage2/modules/devices/components/disk-node/disk-node.component.scss
denysbutenko: image

Group rows has different styles on mockup

bvasilenko:
@undsoft  I've introduced a new type `DeviceNestedDataNode`, just to separate it from `VDev`, however I realized that technically it still should have the save set of properties as `VDev`. So we have an alternative option, just to keep `VDev`

src/app/pages/storage2/modules/devices/components/devices/devices.component.ts
```diff
@@ -53,40 +54,61 @@ export class DevicesComponent implements OnInit {
...
           }
         });
       }).filter(Boolean);
     };
   }
 
+  private createDataNodes(topology: PoolTopology): DeviceNestedDataNode[] {
+    const dataNodes: DeviceNestedDataNode[] = [];
+    if (topology.data.length) {
+      dataNodes.push({ children: topology.data, disk: 'Data VDEVs', guid: 'data' } as DeviceNestedDataNode);
```
undsoft, line 76: Interface fields are "the same" only because they've been artificially made to be the same. Like here, a `disk` field doesn't really make sense for an arbitrary group of vdevs.
At the very list we should use `name` here, but ideally for `DeviceNestedDataNode` is to be typed as 
```
type DeviceNestedDataNode = VdevGroup | Vdev;
```
where VdevGroup only has fields used in a group.
Or something similar to help communicate to other developers that these are really different things.

Not a problem of this PR, but `Vdev` is a misleading name as well and we should fix it in the future. What we currently call a vdev is either a vdev or a disk in a vdev with different set of attributes.

src/app/pages/storage2/modules/devices/components/disk-node/disk-node.component.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+import { VDev } from 'app/interfaces/storage.interface';
+import { DiskNodeComponent } from 'app/pages/storage2/modules/devices/components/disk-node/disk-node.component';
+
+describe('DiskNodeComponent', () => {
+  let spectator: Spectator<DiskNodeComponent>;
+  const disk = {
+    type: 'DISK',
+    path: null,
+    guid: '123',
+    status: 'OFFLINE',
```
undsoft, line 13: Enums can be used here.

src/app/pages/storage2/modules/devices/components/disk-node/disk-node.component.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  get diskCapacity(): string {
+    return this.size !== null && this.size !== undefined ? this.utils.convert(this.size).value
+      + this.utils.convert(this.size).units : '';
+  }
+
+  get diskErrors(): string {
+    if (this.disk.stats) {
+      const errors = this.disk.stats?.checksum_errors + this.disk.stats?.read_errors + this.disk.stats?.write_errors;
+      return errors ? `${errors} Errors` : this.translate.instant('No Errors');
```
undsoft, line 48: You can leave it to translators to deal with how these strings should look.
```
this.translate.instant('{n, plural, =0 No Errors, one {# Error} other {# Errors}}', { n: errors });
```

undsoft (changes requested):
Looks better, but I believe Denyses comment about styling for the group is still not address.
You can do it as a separate ticket if you want.

src/app/pages/storage2/modules/devices/components/disk-node/disk-node.component.scss
bvasilenko: Now both when scrolled horizontally and when not, it looks exactly like in Dataset Management page:
image

denysbutenko: image

Sorry for the confusion. I mean `Data VDev` and other group rows not Mirror/Raidz etc

src/app/pages/storage2/modules/devices/components/devices/devices.component.html
```diff
@@ -9,45 +9,59 @@ <h4 mat-card-title>Note: This is a work in progress for new 'Device Management'
...
+                (click)="$event.preventDefault()"
+                [attr.aria-label]="'Toggle {row}' | translate: { row: disk.guid }"
+              >
+                <mat-icon>
+                  {{ treeControl.isExpanded(disk) ? 'expand_more' : 'chevron_right' }}
+                </mat-icon>
+              </button>
+              <ix-disk-node
+                [disk]="disk"
+                [size]="disk.type ? diskDictionary[disk.children[0].disk]?.size : null"
```
denysbutenko, line 55: Is it intentional to have two properties `size` and `type`? Perhaps it makes sense to pass a single object and move logic inside of disk-node

src/app/pages/storage2/modules/devices/components/disk-node/disk-node.component.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  get diskErrors(): string {
+    if (this.disk.stats) {
+      const errors = this.disk.stats?.checksum_errors + this.disk.stats?.read_errors + this.disk.stats?.write_errors;
+      return this.translate.instant('{n, plural, =0 {No Errors} other {# Errors}}', { n: errors });
+    }
+    return '';
+  }
+
+  get diskIcon(): string {
```
denysbutenko, line 53: We need to display a disk icon in the title on the Details Panel, it must be reuseable like `ix-dataset-icon`

src/app/pages/storage2/modules/devices/components/disk-node/disk-node.component.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  get diskCapacity(): string {
+    return this.size !== null && this.size !== undefined ? this.utils.convert(this.size).value
+      + this.utils.convert(this.size).units : '';
+  }
+
+  get diskErrors(): string {
+    if (this.disk.stats) {
+      const errors = this.disk.stats?.checksum_errors + this.disk.stats?.read_errors + this.disk.stats?.write_errors;
+      return this.translate.instant('{n, plural, =0 {No Errors} other {# Errors}}', { n: errors });
```
denysbutenko, line 48: The singular case is not covered, e.g. `1 Error`

src/app/pages/storage2/modules/devices/components/devices/devices.component.ts
```diff
@@ -53,40 +54,61 @@ export class DevicesComponent implements OnInit {
...
-              return vdev.name.toLowerCase().includes(query.toLowerCase());
+              return dataNode.name?.toLowerCase().includes(query.toLowerCase());
           }
         });
       }).filter(Boolean);
     };
   }
 
+  private createDataNodes(topology: PoolTopology): DeviceNestedDataNode[] {
+    const dataNodes: DeviceNestedDataNode[] = [];
```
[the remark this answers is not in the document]
bvasilenko: Here are 3 reasons why I prefer not to over use dynamic approaches to coding at this method:
- `PoolTopology` has each of it's properties strictly typed, it is not declared as an array or dictionary, so it is obviously not meant to be iterated through `.map` at first place
- `Object.entries` would be maintained in strict sync with the API changes. example: what if middleware would add a new key to the object, it would potentially lead to displaying unwanted root nodes in the tree
- creating a cycle around a `topology` object will lead to a creation of an unnecessary structures/condition checks just to inject side-values such as `'Data VDEVs', 'Cache'` and will lead to unnecessary over-complications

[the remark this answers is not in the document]
denysbutenko: The second point is debatable. I believe if new group will be added we should apply the same rules as for displaying it as for existing ones. But okay explicit is better than implicit. Thanks for clarifying 👍

RehanY147 (changes requested):
The filter isn't working properly on this PR anymore. It works OK on master. Here it only shows the first result I think.

Furthermore, there's a slight misalignment issue with the status column. The other two columns and row values are left aligned but for this column the header is left aligned but the values are centered which makes it look weird.
alignment issue

src/app/pages/storage2/modules/devices/components/devices/devices.component.ts
bvasilenko: I have addressed that in the [facbb39](https://github.com/truenas/webui/pull/6861/commits/facbb398c79b7af98e40a72e9b23df762cb54e86) commit

bvasilenko: @undsoft

src/app/pages/storage2/modules/devices/components/devices/devices.component.html
```diff
@@ -20,45 +20,51 @@ <h4 mat-card-title>Note: This is a work in progress for new 'Device Management'
...
+            <ix-disk-node [vdev]="vdev | cast" [disk]="diskDictionary[vdev.disk]"></ix-disk-node>
           </ix-tree-node>
 
           <ix-nested-tree-node
-            *ixTreeNodeDef="let disk; dataSource: dataSource, when: hasNestedChild"
+            *ixTreeNodeDef="let vdev; dataSource: dataSource, when: isVdevGroup"
+            [ixTreeNodeDefDataSource]="dataSource"
+          >
+            <div class="disk-nested-tree-root-node">
+              <ix-vdev-group-node [vdev]="vdev | cast"></ix-vdev-group-node>
```
denysbutenko, line 39: Looks very specific, wouldn't we plan to reuse groups on the dataset tree? I think it may be a part of the ix-tree module. Up to you

undsoft (approved):
Code looks good, header is supposed to allow section to be collapsed, but this can be added later.

bvasilenko:
> The filter isn't working properly on this PR anymore. It works OK on master. Here it only shows the first result I think.



> Furthermore, there's a slight misalignment issue with the status column. The other two columns and row values are left aligned but for this column the header is left aligned but the values are centered which makes it look weird. alignment issue [ alignment issue ](https://user-images.githubusercontent.com/24204846/179130332-cc2c0e72-478c-463a-8f90-9bbc0ca41253.gif) [ ](https://user-images.githubusercontent.com/24204846/179130332-cc2c0e72-478c-463a-8f90-9bbc0ca41253.gif)

Both issues should be fixed now.

bvasilenko:
> Looks better now. Noticed tree nodes are misaligned.
> 
> image
> 
> image

@denysbutenko 

Thanks for noticing this. Now alignment is much closer to design:

image

bvasilenko:
> header is supposed to allow section to be collapsed, but this can be added later.

@undsoft Thanks for a reminder. Header is now collapsible.
Jul-18-2022 16-37-23

RehanY147 (approved):
Looks good. Just one minor nit, we show `cursor: pointer` for the whole header row but the click only works on the down arrow button on the right. Let's make the whole row clickable.

denysbutenko (approved):
Looks good 👍🏻

PR #6862, opened 2022-07-06
fix state translation in certificate creation ui

pgollor:


pgollor:
sorry for that. same problem which was discussed in https://github.com/truenas/webui/pull/5313

PR #6863, opened 2022-07-07
NAS-117024 / 22.12 / NAS-117024: Fix the column layout of details panel

bvasilenko:
Check correctness of the layout formed by widgets at the right panel 
- at the **Dataset Management** (/datasets)
- at the **Storage 2.0**  > **Manage Devices** button  > **Devices** page


undsoft (changes requested):
Please note that the card layout also needs to be fixed.
Note how there are no vertical gaps between cards in the design.
Снимок экрана 2022-07-07 в 11 32 31

RehanY147 (changes requested):
Let's also fix the viewport height taken by the page. Even removing the note doesn't fix this. The page is always scrolling a little bit which doesn't happen on other pages like the `Network` page.
Furthermore, as Evgeny said, let's aim to fix all design issues with both the Datasets and Devices pages.
The table height has no end if you keep scrolling further and further down with the details panel. If the table needs to scroll down (because of more items in the tree table), let's keep the filter and column headers sticky.

AlexKarpov98:
image

There is still scrolling a little bit on Datasets Management page.

AlexKarpov98 (changes requested):
https://github.com/truenas/webui/pull/6863#issuecomment-1179432168

bvasilenko:
I've addressed all issues mentioned above

undsoft (changes requested):
I think when window is smaller, tree shouldn't just scroll inside, but should scroll with the page.
https://user-images.githubusercontent.com/1481270/178500232-50e17147-d265-4fd1-b3ad-2a9e2bd6983d.mov
In this case when I scroll up, I expect header to go away, but tree table header to stay on top. Same with details pane, it should be scrollable inside if there is not enough space, but stick to the top when page is scrolled.

Card layout is good, but could use a little bit more vertical margins.

AlexKarpov98 (changes requested):
- If some case that a tree table has a self vertical stretch space there is a bit scrolling still on the page
Screenshot_1


- I think that the padding/margin between all the cards should be keep the same size in a detail section like this screenshort.
Screenshot_3


- As Undsoft said the search and header's stick issue of a tree table should be fixed.
header-issue

undsoft:
Actually, here's a decent example of how this should work - Github itself (except that the columns are swapped).
https://github.com/truenas/webui/pull/6872/files

bvasilenko:
> https://github.com/truenas/webui/pull/6872/files

This is not exactly clear. A screenshot with marks would help. 

I am struggling with combining of "whole page" scroll with the "inner scroll" of the details panel. As you can see from the screenshot, it will lead to 2 scrollbars, which do not look good to me. I think we need some better example here

image

AlexKarpov98:
"_all of your screenshots display outdated version of the code. Most of these issues already fixed at https://github.com/truenas/webui/commit/e4a7b6aea388b6394163d4ce2c0e3a7c52d92076 commit.
Please make sure you have pulled the latest commit https://github.com/truenas/webui/commit/984419537b4e8e51b4a8b1f088fe89ad24580071_"

Regarding to your last comments in https://github.com/truenas/webui/pull/6863#issuecomment-1184210810...
I left that comment yesterday and you pushed this commit https://github.com/truenas/webui/commit/984419537b4e8e51b4a8b1f088fe89ad24580071 today.

Now the rest issues except an header issue of a tree table were fixed by this commit https://github.com/truenas/webui/commit/984419537b4e8e51b4a8b1f088fe89ad24580071

AlexKarpov98 (approved):
The rest of the issues except an header issue of a tree table were fixed.
I approve this PR because you suggested creating new tickets for the header issue of a tree table

Screenshot_13
.

RehanY147 (changes requested):
This is what the dataset page looks like at 1440p with 100% normal zoom on the browser. We should fix the alignment to have only two columns and center the page if there's extra space left.
image

Same goes for the Devices page
image

RehanY147 (dismissed):
In the video you shared earlier, both table and the details pane have independent scrolling, and if we don't need more vertical space, then the details pane shouldn't be scrolling to match the table height. It should stay within the viewport unless it has to go outside it.

src/app/core/components/directives/details-height/details-height.directive.ts
```diff
@@ -0,0 +1,41 @@
+import {
+  Directive, ElementRef, Inject, Input, OnDestroy, OnInit,
+} from '@angular/core';
+import { WINDOW } from 'app/helpers/window.helper';
+
+@Directive({
+  selector: '[ixDetailsHeight]',
+})
+export class IxDetailsHeightDirective implements OnInit, OnDestroy {
+  @Input() ixDetailsHeightParantClass: string;
```
undsoft, line 10: `Parent`

src/app/core/components/directives/details-height/details-height.directive.ts
```diff
@@ -0,0 +1,41 @@
...
+    this.el.nativeElement.style.height = this.heightDetails;
+    this.window.addEventListener('scroll', this.onScrollHandler, true);
+  }
+
+  ngOnDestroy(): void {
+    this.window.removeEventListener('scroll', this.onScrollHandler, true);
+  }
+
+  onScroll(event: Event): void {
+    if ((event.target as HTMLElement).className === this.ixDetailsHeightParantClass) {
```
undsoft, line 32: How about returning early to avoid nesting?

src/app/core/components/directives/details-height/details-height.directive.ts
```diff
@@ -0,0 +1,41 @@
...
+  heightDetails = `calc(100vh - ${this.hiddenHeight}px)`;
+
+  constructor(
+    @Inject(WINDOW) private window: Window,
+    private el: ElementRef,
+  ) {}
+
+  ngOnInit(): void {
+    this.el.nativeElement.style.height = this.heightDetails;
+    this.window.addEventListener('scroll', this.onScrollHandler, true);
```
undsoft, line 24: MDN recommends throttling scroll listeners. Should we?

src/app/core/components/directives/details-height/details-height.directive.ts
```diff
@@ -0,0 +1,41 @@
+import {
+  Directive, ElementRef, Inject, Input, OnDestroy, OnInit,
+} from '@angular/core';
+import { WINDOW } from 'app/helpers/window.helper';
+
+@Directive({
```
undsoft, line 6: Let's add a comment explaining briefly explaining why this is needed for future generations.

src/app/core/components/directives/details-height/details-height.directive.ts
```diff
@@ -0,0 +1,41 @@
...
+  selector: '[ixDetailsHeight]',
+})
+export class IxDetailsHeightDirective implements OnInit, OnDestroy {
+  @Input() ixDetailsHeightParantClass: string;
+  @Input() hiddenHeight = 230;
+  @Input() maxShowHeight = 150;
```
undsoft, line 12: Not immediately obvious what these are. Can we add more description or rename?
Also I'm assuming one of these is responsible for getting height of the header. If so, can we query for that height (maybe in parent component)? Trying to make sure this doesn't break when header height changes.

undsoft (changes requested):
Scroll behaviour is much better, nice work.
Things to improve:
1. I don't think that scrolling in details pane should push main scrolling after it reaches the end of the container. See [overscroll-behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior).

https://user-images.githubusercontent.com/1481270/179684998-c9bbe505-c37c-406e-8b08-bd6d7476d99a.mov

2. When scrolled up details pane has a gap in the bottom which seems larger than it should be. I.e. bottom gap in the beginning of the video is too large, in the end - good.
3. I'd rather see table header being sticky instead of search. Search is probably not that useful. If hard to implement, you can create a separate ticket for this.

src/app/core/components/directives/details-height/details-height.directive.ts
bvasilenko: Agreed. Now these 2 parameters have better names, and adjusted default values

bvasilenko:
> Scroll behaviour is much better, nice work. Things to improve:
> 
> 1. I don't think that scrolling in details pane should push main scrolling after it reaches the end of the container. See [overscroll-behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior).
> 
>  2022-07-19.08.50.39.mov 
> 2. When scrolled up details pane has a gap in the bottom which seems larger than it should be. I.e. bottom gap in the beginning of the video is too large, in the end - good.
> 3. I'd rather see table header being sticky instead of search. Search is probably not that useful. If hard to implement, you can create a separate ticket for this.

@undsoft 

Points 1 and 2 are addressed with my latest commits.

Point 3 is hard to implement, so I'm going to create a separate ticket for it

bvasilenko:
> In the video you shared earlier, both table and the details pane have independent scrolling, and if we don't need more vertical space, then the details pane shouldn't be scrolling to match the table height. It should stay within the viewport unless it has to go outside it.

@RehanY147  Details block already stays within the viewport unless it needs more vertical space. In other words, this is already implemented

src/app/core/components/directives/details-height/details-height.directive.ts
```diff
@@ -0,0 +1,52 @@
...
+import { WINDOW } from 'app/helpers/window.helper';
+
+const headingHeightInitialDefault = 213;
+const headingHeightDefault = 137;
+
+///
+/// This directive is used to dynamically adjust height of the "details" block in a "master-details" layout
+/// to fill the bottom space, which becomes available when user scrolls the page down,
+/// so the page's heading is shifted off the screen
+///
```
undsoft, line 13: Prefer jsdoc style
```
/**
 *
 */
```

src/app/core/components/directives/details-height/details-height.directive.ts
undsoft: Okay, but can we avoid magic numbers and query for height in the parent?

src/app/core/components/directives/details-height/details-height.directive.ts
bvasilenko: This is addressed in one of the latest commits

src/app/core/components/directives/details-height/details-height.directive.ts
bvasilenko: I suppose throttling of this scroll listener would be an overkill, and would potentially degrade scrolling animation, by making it sluggish

undsoft (changes requested):
As discussed on Slack, best to replace magic numbers in `[ixDetailsHeight]` with some code to get height of the header / offset of the details pane.

bvasilenko:
@undsoft 

- no magic numbers in `details-height.directive.ts`
- now calculation is partially dynamic (for example, initial "y" coordinate is taken from DOM)
- values provided to constants are well-known (for example, height of the header)
- safe to add/remove any content (for example, safe to remove "Work in progress notice")

undsoft (approved):
Two notes:
1. If you go to datasets, make window shorter and scroll details, they will not be fully scrolled - you won't be able to reach bottom of details panel.
2. 
```
const headerHeight = 48;
const footerHeight = 45;
```
can be extracted somewhere else to avoid values in two places.

PR #6867, opened 2022-07-08
Fix test mail subject

themylogin:
Middleware adds TrueNAS and hostname too so the resulting subject is
`TrueNAS truenas.local: TrueNAS Test Message hostname: truenas` which
looks strange.


themylogin:
time 0:15

themylogin:
backport

PR #6874, opened 2022-07-13
NAS-116964 / 22.12 / Nas 116964-1 Add URL support to Devices tree view

Raz-Dva:
Storage 2.0  -> Topology -> Manager Devices
Added a new router in the devices.routing
Check URL for disks in Devices. Check rendering Details cards. Reload the page, Details cards should be loaded with correct row selected based on URL. 
devices


src/app/pages/storage2/modules/devices/components/devices/devices.component.ts
```diff
@@ -109,6 +112,15 @@ export class DevicesComponent implements OnInit {
             this.selectFirstNode();
             this.loader.close();
             this.cdr.markForCheck();
+
+            this.route.params.pipe(
+              pluck('guid'),
+              untilDestroyed(this),
+            ).subscribe((guid) => {
+              this.selectedItem = this.treeControl.dataNodes[0].children.find((child) => {
```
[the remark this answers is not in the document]
undsoft: There is a similar function called `getDatasetAndParentsById`, maybe you could rename and repurpose it.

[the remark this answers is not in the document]
Raz-Dva: > There is a similar function called `getDatasetAndParentsById`, maybe you could rename and repurpose it.

At first I wanted to apply this function, but then I discovered that the IDs of the parents are written in the IDs of the their children "datasets".
 For example  pool -> parent -> child 2, the id looks like "poolio/parent/child2". The getDatasetAndParentsById function uses the url to search for daset and its parent.
Currently for devices, the url looks like this ['/storage2', pool id, 'devices', disk.guid]
Perhaps we should change the url for devices in the same way with the datasets?

src/app/pages/storage2/modules/devices/components/devices/devices.component.ts
```diff
@@ -83,14 +90,24 @@ export class DevicesComponent implements OnInit {
...
-    event.stopPropagation();
-    this.selectedItem = vdev;
-    this.selectedParentItem = [...Object.values(this.topology.data)].find((group: VDev) => {
-      return group?.children.find((child) => {
-        return child.guid === vdev.guid;
-      });
-    });
+  private listenForRouteChanges(id: string): void {
+    if (id && this.dataSource?.data) {
+      const traverseTree = (children: VDev[], parent?: VDev): { item: VDev; parent: VDev } => {
```
bvasilenko, line 95: I noticed that you've inlined the `traverseTree` function into the devices.component. 
We now have another tree traversing function in `app/modules/ix-tree/utils/flatten-tree-with-filter`, which I've used to recursively find multiple items in the tree. It's up to you to merge these 2 methods into a single method

bvasilenko (commented):
Looks good so far, 
However it's still necessary to fetch updated "devices' page from the latest `master`

src/app/pages/storage2/modules/devices/components/devices/devices.component.ts
```diff
@@ -104,12 +114,30 @@ export class DevicesComponent implements OnInit {
...
-    event.stopPropagation();
-    this.selectedItem = dataNodeSelected;
-    this.selectedParentItem = findInTree(this.treeControl.dataNodes, (dataNode: DeviceNestedDataNode) => {
-      return dataNode.guid === dataNodeSelected.guid;
-    });
+  private listenForRouteChanges(id: string): void {
+    this.selectedItem = undefined;
+    this.selectedParentItem = undefined;
+
+    if (id && this.dataSource?.data) {
```
bvasilenko, line 121: Suggestion for a minor improvement - can we apply "return early" principle here? Just to improve readability.

src/app/pages/storage2/modules/devices/components/devices/devices.component.ts
```diff
@@ -104,12 +114,30 @@ export class DevicesComponent implements OnInit {
...
+        }
+
+        if (dataNode.guid === id) {
+          this.selectedItem = dataNode;
+          this.selectedParentItem = undefined;
+          return true;
+        }
+        return false;
+      });
+    }
```
bvasilenko, line 140: When we're assigning a value to a `selectedItem` we're missing a method call which will expand all the parent nodes to let the user see the node, which we're  set as selected. This video depicts the bug:

   https://user-images.githubusercontent.com/20611516/180776241-03e42cfe-7c47-4de5-97f0-300d82c92f4b.mp4

  `this.treeControl.expand(node)` looks like suitable method to fix that.

bvasilenko (changes requested):
Good overall! I've just noticed a minor bug

src/app/pages/storage2/modules/devices/components/devices/devices.component.ts
Raz-Dva: Fixed.

bvasilenko (approved):
Works good!

PR #6875, opened 2022-07-15
NAS-117082 / 22.12 / NAS-117082: Improving loading indication on Datasets and Devices

undsoft:
Testing: go to Datasets and Devices.
If you go to Datasets, somewhere else and back, the tree will still be populated with old data. Same happens when you perform an action on a tree (like adding/removing a dataset).
This is a deliberate choice.

Also renames AppLoaderModule to LoaderModule.


src/app/pages/datasets/store/dataset-store.service.ts
```diff
@@ -25,6 +27,7 @@ const initialState: DatasetTreeState = {
 @Injectable()
 export class DatasetTreeStore extends ComponentStore<DatasetTreeState> {
   readonly isLoading$ = this.select((state) => state.isLoading);
+  // TODO
```
RehanY147, line 30: What's TODO?

RehanY147 (changes requested):
If you move around the dashboard a couple of times, the pools dashboard is loading but shows no indication. There's no errors in the console. But no loading indication either.
loading issue pools

src/app/pages/datasets/store/dataset-store.service.ts
undsoft: This is a TODO for next line to show error somewhere. Not included in this PR.

undsoft:
@RehanY147 Didn't work on loading indication for Storage Dashboard in this PR. Just Datasets and Devices (as in list of Devices).

dszidi (commented):
When I go between Storage 2.0 link and the devices page (via topology card) I get a console error.

image

EDIT: Nvm, this also seems to be there in master branch.

PR #6876, opened 2022-07-15
NAS-117017 / 22.12 / NAS-117017: Investigate the conditions for Encryption card on Dataset…

AlexKarpov98:
In a result of investigated if encryption can be changed post dataset creation, that edit button is to update the encrypt properties of a dataset encrypted already.
If to use the edit button post dataset creation with non-encryption, additional process to encrypte that dataset is needed before that at least, I think.
Because there is no feature like that currently, so it can't active an edit button I removed a ZFS encryption card in a case of non-encryption.


RehanY147 (approved):
LGTM

PR #6879, opened 2022-07-17
NAS-117216 / 22.12 / Pools dashboard cards scalling and ordering improved

AlexKarpov98:


src/app/pages/datasets/components/dataset-management/dataset-management.component.ts
```diff
@@ -86,11 +87,22 @@ export class DatasetsManagementComponent implements OnInit {
...
+    this.dataSource = new IxNestedTreeDataSource<DatasetInTree>(filteredDatasets);
+    this.dataSource.filterPredicate = (filteredDatasets, query = '') => {
+      return filteredDatasets.map((datasetRoot) => {
         return findInTree([datasetRoot], (dataset) => dataset.id.toLowerCase().includes(query.toLowerCase()));
       }).filter(Boolean);
     };
   }
+
+  private initFilterDataSource(datasets: DatasetInTree[]): DatasetInTree[] {
+    const filtered = (datasets: DatasetInTree[]): DatasetInTree[] => {
```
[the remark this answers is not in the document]
AlexKarpov98: That was code of another ticket.
Removed that code this branch.

bvasilenko (approved):
Works good to me!

Jul-20-2022 12-19-16

RehanY147 (changes requested):
I think a better approach would be to spread the content out dynamically. Extra height at the bottom of cards that isn't used looks odd. Also, no margin/padding on the right edge.
image

RehanY147:
We can figure out the tallest card. Fix that height and then make all the other cards match that height by spreading the content out with margins.

AlexKarpov98:
> I think a better approach would be to spread the content out dynamically. Extra height at the bottom of cards that isn't used looks odd. Also, no margin/padding on the right edge. image


The  "no margin/padding on the right edge" issue was fixed.
Please check for the last update agian.

https://github.com/truenas/webui/pull/6879/commits/e84ecc6125f4ec4c7d5aa49cc0a143317d037ce8#diff-8a22545de0ceef327f73ff7d5c5c20e4b846a1b5316318bcc3164097da03b892

AlexKarpov98:
> We can figure out the tallest card. Fix that height and then make all the other cards match that height by spreading the content out with margins. All of these cards are going to be static in vertical height. So, we can fix their margins without worrying about stuff changing.

If all the cards are arranged with a row, that's possible definitely but, I am not sure if the cards are arranged with 2 rows the height of all the cards will be fixed as the same size by the css.

RehanY147:
@AlexKarpov98 I didn't mean do it dynamically. I meant that let's figure out the height of the tallest card once. Let's say it's 500px. Let's set the height of all cards to 500px. Then spread the rows out so there isn't much space at the bottom left out. Let me know if you want to discuss this further.

RehanY147:
If there's too little to show, like in the ZFS health card in your screenshot, there's only three rows, we can spread them out at least a little bit and then leave some space at the bottom. Just make it less obvious that we're forcing cards to have the same height. Doesn't have to be perfect. Just a little bit better.

RehanY147 (commented):
^

AlexKarpov98:
Removed fixed height of the cards.
All the cards on the same row takes the same height with an height of the highest card of them.

RehanY147 (commented):
Discussed on Slack

AlexKarpov98:
> Rows need more vertical margins in the ZFS card.

Displayed as spread row.

image

RehanY147 (changes requested):
As discussed on slack, let's standardize the margins between rows on all cards similar to how they were previously on the topology card.

AlexKarpov98:
> As discussed on slack, let's standardize the margins between rows on all cards similar to how they were previously on the topology card.

Standardized a vertical margin between row as 24px on all the cards.
image

src/app/pages/storage2/components/zfs-health-card/zfs-health-card.component.scss
```diff
@@ -14,7 +14,7 @@
   }
 
   .details-item {
-    margin: 0;
+    margin: 12px 0;
```
RehanY147, line 17: I think `8px` would be a better fit

src/app/pages/storage2/components/pools-dashboard/widget-topology/widget-topology.component.html
```diff
@@ -22,7 +22,7 @@ <h3 mat-card-title>
   </mat-card-header>
 
   <mat-card-content fxFlex="100">
-    <div class="content-container" fxLayout="column" fxLayoutAlign="space-between">
+    <div class="content-container" fxLayout="column">
```
RehanY147, line 25: Let's switch the `<div>` underneath with `*ngIf="!loading"` with `ng-container` like so
`<ng-container *ngIf="!loading" fxLayout="row" ...>`
Also bring back the `fxLayoutAlign="space-between"` here to fix the weird alignment of the `Data VDEVs` row.

src/app/pages/storage2/components/disk-health-card/disk-health-card.component.scss
```diff
@@ -14,15 +14,8 @@
...
-  .mat-card-content {
-    display: flex;
-    flex: 1;
-    flex-direction: column;
-    justify-content: space-between;
-  }
-
   .details-item {
-    margin: 0;
+    margin: 12px 0;
```
RehanY147, line 18: `8px` here too

RehanY147 (changes requested):

src/app/pages/storage2/components/zfs-health-card/zfs-health-card.component.scss
AlexKarpov98: Done

src/app/pages/storage2/components/disk-health-card/disk-health-card.component.scss
AlexKarpov98: Done

src/app/pages/storage2/components/pools-dashboard/widget-topology/widget-topology.component.html
AlexKarpov98: Added `<ng-container *ngIf="!loading">...</ng-container>`.
image

A `fxLayoutAlign="space-between"` attribute was removed already.

RehanY147 (changes requested):
Data VDEVs still is in the middle of the card unlike the other rows. Was this an intentional change? Because I think it's coming from master. That's why I was asking fxLayoutAlign="space-between" to be brought back.
image

AlexKarpov98:
> Data VDEVs still is in the middle of the card unlike the other rows. Was this an intentional change? Because I think it's coming from master. That's why I was asking fxLayoutAlign="space-between" to be brought back. image



> Data VDEVs still is in the middle of the card unlike the other rows. Was this an intentional change? Because I think it's coming from master. That's why I was asking fxLayoutAlign="space-between" to be brought back. image

That wasn't an  intentional change.
Maybe that has been removed as a mistake before previous PR, I think.
But that attribute should be removed in a `content-container` div.
Instead, the lines should be arranged by `fxLayoutAlign="space-between"` attribute in a `vdev-line` div.
That was fixed by this PR https://github.com/truenas/webui/pull/6917.

This is the code fixed.
image

RehanY147 (commented):
The build is failing. Also, I'm seeing 0 files changed. What's that about? Do you know?

RehanY147:
Nothing left to merge.

PR #6881, opened 2022-07-19
NAS-117203: Separate capacity management form

undsoft:
Testing: 
In new storage pages:
- Dataset creation form - quota and capacity management fields are still shown.
- Dataset edit from - quota and capacity management fields are hidden.
- New edit form in Capacity management.


AlexKarpov98 (approved):
LGTM
Hidden them in Dataset edit from.
New edit form in Capacity management does work well.
Screenshot_16

PR #6883, opened 2022-07-20
NAS-116915 / 22.12 / NAS-116915: Add the pool header row on Datasets IxTreeTable

bvasilenko:
Testing:

Navigate to **Datasets (WIP)** page and check that collapsible header of the Pool matches the designer's mockup.

Designer's Mockup:
image

Pool header implemented:
image


src/app/pages/datasets/components/dataset-management/dataset-management.component.html
```diff
@@ -29,9 +29,35 @@ <h4 mat-card-title>Note: This is a work in progress for the new 'Dataset Managem
...
 
+          <ix-nested-tree-node
+            *ixTreeNodeDef="let dataset; dataSource: dataSource, when: isDatasetRoot"
+            [ixTreeNodeDefDataSource]="dataSource"
+          >
+            <div
+              class="dataset-nested-tree-root-node"
+              (click)="onRowGroupSelected(dataset, $event)"
+            >
+              <ix-dataset-group-node [dataset]="dataset | cast">
```
denysbutenko, line 43: `ix-dataset-group-node` and `ix-vdev-group-node` looks the same. I'd suggest making it reusable to avoid creating `ix-something-group-node` in the future. The common group node component should be part of the IxTree module.

How it might look like:
image

undsoft: I'm okay with having separate components. They are similar, but have slightly different responsibilities.

bvasilenko: Each of `ix-dataset-group-node` and `ix-vdev-group-node` are presenting the data from their underlying data node, so it's good in terms of separation of responsibilities.

undsoft:
Mmm, now that it has been implemented, I think I like what we had before more.
@dszidi What do you think?

https://user-images.githubusercontent.com/1481270/181739426-e8ebf721-77e2-446d-bbab-80be14607693.mov

dszidi (commented):
Let's try it with the following changes.

- Remove the collapse trigger from the group header.
- Expand the root dataset rows by default. This is what we've done up until now and it seems to work well.

It should look like this...


image

image

If the root dataset is always going to be visible due to toggle being removed, I wonder if we should remove the pool name and openzfs icon from the group header as well.

@undsoft What do you think?

undsoft:
@dszidi My suggestion would be not to have these groupings completely. Our original intent was to communicate to the user that root dataset != pool. While this is still something we need to communicate to users, I just think it makes UI looks worse and is not worth it.

denysbutenko:
Maybe we could move a pool name somewhere. For example after root dataset names as a badge or as a chip? It will be visible and don't take an extra row.

denysbutenko:
One more thought: We're using the `sand clock` icon for pools. Does it make sense to use it in the sidebar as well?

undsoft (changes requested):
Let's keep current branch and PR open and try a different option with just having some visual separation between pools:

image (3)

undsoft:
We decided against making changes.

PR #6884, opened 2022-07-20
NAS-116894 / 22.12 / NAS-116894: Application Events is not sorted

Raz-Dva:
Apps -> Installed Applications 
Check sorting of Application Events
_metadata.creation_timestamp?.$date_ was used instead of _first_timestamp?.$date_

image


denysbutenko (approved):
image

Nice work 👍🏻

PR #6885, opened 2022-07-20
NAS-117269 / 22.02.3 / add NO_FENCED to disabled reason enum

yocalebo:
Changes were made to fix a problem with HA systems here: https://github.com/truenas/middleware/pull/9439. I added a new `NO_FENCED` value to be returned so I'm adding it here on front-end.


undsoft (approved):
👍🏻

PR #6886, opened 2022-07-20
NAS-117198 / 22.12 / Upgrade linter packages and some fixes

denysbutenko:
- Enabled `no-trailing-spaces` but it seems already was in other set rules
- Upgraded packages to make it work again
- Fixed reported issues

For testing, make sure `yarn lint` is working


bvasilenko (approved):
linter works for me!

PR #6888, opened 2022-07-20
NAS-116483 / 22.12 / NAS-116483: Using new `vm.device.disk_choices` endpoint in VMs.

undsoft:
Testing:
1. Create two zvols with spaces in the name.
2. Create a VM using first zvol.
3. Use Devices -> Add to add second zvol.

Also fixes:
- VM button that was missing from the toolbar in VMs.
- setup_env script will now respect port

### DIFF
--- a/setup_env.js
+++ b/setup_env.js
@@ -8,9 +8,9 @@ program
 
 var proxy_config_json = './proxy.config.json';
 var environment_ts = './src/environments/environment.ts';
-var hostname = (program.ip || '').match(/^(?:https?:\/\/)?(?:[^@\n]+@)?([^:\/\n?]+)/);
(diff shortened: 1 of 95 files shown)


dszidi (commented):
Overall the process works. There is one thing about adding a disk via the devices page I noticed. The select list gives me the option to choose the zvol that is already used by the VM. If I select that disk it will fail validation of course, but IMO that option should never be shown to the user to begin with. 

I'm not sure if this is something that the UI needs to address or if the disk_choices data needs to be corrected.

undsoft:
@dszidi I'll vote for this being a middleware issue.

undsoft:
@dszidi But actually the error is unobtrusive, we can keep as is for now.

dszidi (approved):
LGTM

PR #6893, opened 2022-07-22
NAS-117233 / 22.12 / NAS-117233: New Google Cloud Storage task field: `bucket_policy_only`

bvasilenko:
Testing:

At the **Data Protection > Cloud Sync Tasks > Add** (And also at **... > Update**)

There's a new checkbox field:

- **Title:** Bucket Policy Only

- **Description:** Access checks should use bucket-level IAM policies.

It should include a new item `bucket_policy_only` Into the `attributes` object at these server queries.

Back-end PR: https://github.com/truenas/middleware/pull/9421


undsoft (changes requested):
I think this is only supposed to show for providers of type Google Cloud Storage.

src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
```diff
@@ -865,6 +877,12 @@ export class CloudsyncFormComponent implements FormConfiguration {
               this.setDisabled('bucket_input', true, true);
             }
 
+            if (targetProvider && targetProvider.name === 'GOOGLE_CLOUD_STORAGE') {
```
AlexKarpov98, line 880: It'll better if the Provider name is used as an enum type, I think.

AlexKarpov98 (changes requested):

src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
bvasilenko: Ok, there's no suitable enum, so I'll ping you once I'll replace all the providers from this dropdown with the enum

image

AlexKarpov98 (approved):
The code change looks good!

PR #6894, opened 2022-07-22
NAS-117301 / 22.12 / NAS-117301: Remove close icon

dszidi:
We don't need this since there is a breadcrumb leading back to the VM page.
To test just go to VMs page and choose "Devices" from the row actions. On the Devices page there should not be a close icon (x) next to the Add button in the page title area


undsoft (approved):
The close icon is gone :)

PR #6896, opened 2022-07-24
NAS-116397 / 22.12 / NAS-116397: Disk Health card

AlexKarpov98:
I tried with `alert.list`, `alert.list_categories` api but I couldn't get the disk temperature related alerts.
A `Disks temperature related alerts` item displays a default 0 value currently.
This item should be updated by a correct API.


denysbutenko (approved):
image

Looks good

undsoft:
Let's hide `Disks temperature related alerts` before merging, so that we don't forget about it.

AlexKarpov98:
> Let's hide `Disks temperature related alerts` before merging, so that we don't forget about it.

Yes, I will do that.

PR #6898, opened 2022-07-25
NAS-117333 / 22.12 / Use fill icons for datasets and outline icons for locked/unlocked by ancestor

dszidi:
To test, go to new datasets page and make sure the lock/unlock icons are outline for the "by ancestor" rows and filled icons for the rest


src/app/pages/datasets/components/dataset-node/dataset-encryption-cell/dataset-encryption-cell.component.html
```diff
@@ -1,15 +1,26 @@
 <div class="encryption-description" *ngIf="!dataset.encrypted; else encryptedCell">{{ 'Unencrypted' | translate }}</div>
 <ng-template #encryptedCell>
+
   <ng-container *ngIf="dataset.locked; else unlockedCell">
-    <mat-icon>{{ isEncryptionRoot ? 'lock' : 'lock_outline' }}</mat-icon>
+
```
undsoft, line 5: What was wrong with old code here?

undsoft (changes requested):
Unlock icon is okay, lock state just shows unlocked icons now.

src/app/pages/datasets/components/dataset-node/dataset-encryption-cell/dataset-encryption-cell.component.html
dszidi: The lock icons we need are not in the standard mat-icon set. I had to get them from the mdi icon set.

dszidi:
I switched to using the unlock variant icon that we were using before. It makes the difference a little more obvious 

image

undsoft (approved):
Unlock icon looks good, still don't see differences between new and old locked icons:
Снимок экрана 2022-07-26 в 16 20 46

dszidi:
> Unlock icon looks good, still don't see differences between new and old locked icons...

The icons in master branch, it's a bit hard to distinguish between locked and unlocked icons.

image


The mdi icons include a variant of unlocked that makes the difference more obvious

image

undsoft:
No questions about unlocked icon. Unlocked icon is good. I'm saying that I don't see any difference between previously lock icon and new lock icon.

dszidi:
The lock icon was switched to mdi set to make markup consistent. Using mdi icon set requires additional attributes

PR #6899, opened 2022-07-26
NAS-117285 / 22.12 / NAS-117285: Removing some FormsModule imports

undsoft:
Testing: in new Datasets check Creating a pool and adding unused disk to a pool for regressions.


bvasilenko (approved):
Works good!

PR #6902, opened 2022-07-26
NAS-117366 / 22.12 / NAS-117366: Make `patch` status check optional on codecov

undsoft:
This _should_ disable the status check that complains that not enough % was covered in the diff.


RehanY147 (approved):
Looks good

PR #6903, opened 2022-07-27
NAS-117348 / 22.12 / Replaced `pool.dataset.query` with `pool.dataset.details`

RehanY147:


src/app/interfaces/dataset.interface.ts
```diff
@@ -158,3 +159,17 @@ export interface DatasetUpdate {
   create_ancestors?: boolean;
   user_properties_update?: { key: string; value: string; remove?: boolean }[];
 }
+
+export interface DatasetDetails extends Omit<Dataset, 'children'> {
```
undsoft, line 163: Are other properties from Dataset really available here. I think it may be better to create a completely new interface and remove `DatasetInTree`

denysbutenko (approved):
image

Perhaps the cards require updates to reduce the amount of `pool.dataset.query` calls. Maybe in a separate ticket.

Looks good 👍🏻

PR #6905, opened 2022-07-27
NAS-117153 / 22.12 / NAS-117153: Created checkbox create_empty_src_dirs in Cloud Sync Task

Raz-Dva:
Data Protection -> Cloud Sync Tasks -> button "Add"
For the "Cloud Sync Task" form, the "Create empty ..." checkbox has been added.
Check the new checkbox. TrueNas Scale needs to be updated.


AlexKarpov98 (changes requested):
- A `create_empty_src_dirs` attribute isn't registered into `CloudSyncTask` type.
- A **tooltip** needs to be defined, I think.

Raz-Dva:
- _create_empty_src_dirs_ added to CloudSyncTask interface
- I checked with Eugene. There will be no tooltip for this field

AlexKarpov98 (approved):
Code change looks good.

PR #6906, opened 2022-07-27
NAS-117255 / 22.12 / NAS-117255: Add units tests for Dataset Capacity Management Card

bvasilenko:


src/app/pages/datasets/components/dataset-capacity-management-card/dataset-capacity-management-card.component.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+    beforeEach(() => {
+      spectator = createComponent({
+        props: {
+          dataset: datasetFilesystem,
+        },
+      });
+    });
+
+    it('shows header', () => {
+      expect(spectator.query('mat-card-header')).toHaveText('Dataset Space ManagementEdit');
```
undsoft, line 83: Maybe split it into two check to make it prettier, i.e.
```
expect(spectator.query('mat-card-header')).toHaveText('Dataset Space Management');
expect(spectator.query('mat-card-header')).toHaveText('Edit');
```

undsoft (approved):
Looks good, nice work.
You can also add a test verifying that Edit button opens correct thing when pressed.

PR #6908, opened 2022-07-27
NAS-117221 / 22.12 / Add IxIcon module

denysbutenko:
The idea behind this PR is to have single way of using icons from various sources (material font set, mdi font set, own SVG icons) across the app. The problem I want to solve is that we sometimes want to simultaneously use icons from all sources. The `MatIcon` provides a couple of parameters depending on what type of icon you want to use `fontIcon`, `svgIcon`, `icon ligature name inside a tag`, etc.

Existing `MatIcon` examples of usages (still supported and will be migrated in the future):
`<mat-icon>icon_ligature</mat-icon>` - it was preferred way of using icon before `14.1.0`
`<mat-icon fontIcon="icon_ligature"></mat-icon>` - is added in `14.1.0` version and the [preferred way](https://github.com/angular/components/commit/b37c96dd7f95e9a72cbc9ed0f556f8b240e3a10f) in the Material library
`<mat-icon svgIcon="ix:truenas_scale_logomark"></mat-icon>`
`<mat-icon fontSet="mdi-set" fontIcon="mdi-account"></mat-icon>`

`IxIcon` examples of usage:
`<ix-icon name="ix:truenas_scale_logomark"></ix-icon>` – custom SVG icon from `assets/customicons`
`<ix-icon name="mdi-account"></ix-icon>` – an icon from `mdi-set` font set
`<ix-icon name="material_icon_ligature"></ix-icon>` – an icon from `material` font set
`<ix-icon [name]="name"></ix-icon>` – is preferred way of using it

Changes:

- `IxIconComponent` extends `MatIcon` so it has same methods
- `IxIconRegistry` extends `MatIconRegistry` and now loading custom SVG icons
- Upgrade `@angular/material` and its dependencies to `14.1.0`
- Optimize custom SVG icons with SVGO
- Move custom icons to the `ix` namespace, e.g. `ix:dataset`, `ix:truenas_scale_logomark`
- Update sidebar navigation to consume `ix-icon` and update icon for `Datasets`
- Update `disk-icon` and `dataset-icon` to consume `ix-icon`
- `IxIconModule` still provides `<mat-icon>` for existing uses
- Update eslint rules to prefer `IxIconModule` over `MatIconModule`

For testing, go across pages to see if everything looks good (as before).

PS. The tests are broken in master.


undsoft (commented):
This is an improvement over what we currently have and it's nice to be able to use icons with better syntax.

In the future we should however discuss how to improve this further. Right now we have:
- classic material icons (one request)
- mdi font, which is supposed to be a superset of classic icons? (another request)
- custom icons with each a separate network request for each icon.

Ideas to improve this in the future may be something that packs our own font/svg sprite with only the icons we use.
There is also something that's called [Material Symbols](https://fonts.google.com/icons), which looks better than mdi font. We could potentially replace first two fonts with this and add a sprite builder for custom icons.

denysbutenko:
> This is an improvement over what we currently have and it's nice to be able to use icons with better syntax.
> 
> In the future we should however discuss how to improve this further. Right now we have:
> 
> * classic material icons (one request)
> * mdi font, which is supposed to be a superset of classic icons? (another request)
> * custom icons with each a separate network request for each icon.
> 
> Ideas to improve this in the future may be something that packs our own font/svg sprite with only the icons we use. There is also something that's called [Material Symbols](https://fonts.google.com/icons), which looks better than mdi font. We could potentially replace first two fonts with this and add a sprite builder for custom icons.

Good ideas. Material Symbols look fresh and flexible but it involves much more resources to replace existing icons. I've seen [@mdi/font-build](https://github.com/Templarian/MaterialDesign-Font-Build) to build a custom icon font using our SVG icons. As far as I remember, there could be issues with non-square SVG (e.g. logo) when using it from icon font. I'll take a look at SVG sprite tools 👍🏻 

Funny observations:
-  `@mdi/font` has 6,096 unique icons
-  `material-design-icons/font` has 1124 unique icons x 5 styles ~ 5620 variations
- `material-symbols` has 3552 unique icons x 3 styles x 7 font weight ~ 74592 variations

denysbutenko:
Random fact to bump PR: `mdi.svg` sprite weights seven times more over `mdi.woff2` font – `2.74 MB` vs `376 KB`

dszidi:
I'm getting missing icons in the sidenav area

image

denysbutenko:
> I'm getting missing icons in the sidenav area
> 
> image

If you testing from the branch, please make sure to run `yarn` to update dependencies. Thanks for the feedback.

dszidi (commented):
Ah ok that did it. Everything is better now

image

bvasilenko (approved):
Thanks for improvement! `ix-icon` component looks more convenient than `mat-icon`

Icons are displayed correctly:
image

image

PR #6912, opened 2022-07-29
NAS-117438 / 22.12 / Remove `Microsoft Account` option from users form

denysbutenko:


bvasilenko (approved):
Looks good to me

image
image

PR #6917, opened 2022-07-30
NAS-117439 / 22.12 / Misaligned text on Topology Card

AlexKarpov98:
image


undsoft (approved):
It's not introduced in this PR, but something that is worth fixing here.
The margins on top of the card are too large (should be lower than bottom of the card).
In fact, we could use a bit of empty space in the bottom of the card.
Снимок экрана 2022-07-30 в 21 57 03

AlexKarpov98:
> It's not introduced in this PR, but something that is worth fixing here. The margins on top of the card are too large (should be lower than bottom of the card). In fact, we could use a bit of empty space in the bottom of the card. Снимок экрана 2022-07-30 в 21 57 03

Fixed the top&bottom margins inside the card.
image

PR #6920, opened 2022-07-31
NAS-116138 / 22.12 / NAS-116138: Adding client-side validation for name in interface form

undsoft:
Testing: name field in Network interface form now has client side validation.


dszidi:
Network interface form's name field works fine. For that particular  field it might be useful to also add a rule that the Type field must have a value? I noticed that the validation in the name field doesn't kick in if I fill the name field first.

image



Otherwise it works great. LGTM

image

dszidi (approved):
LGTM

undsoft:
@dszidi If you enter name first and select Type second, validation for name should kick in the moment you select type.

PR #6923, opened 2022-08-02
NAS-102765 / 22.12 / NAS-102765: Ask for EC2 Instance ID when setting initial root password

bvasilenko:


undsoft (approved):
Original ticket mentions `account.set_root_password`, but that looks like an error.
Tested via code, queries look good.

PR #6927, opened 2022-08-02
NAS-116715 / 22.12 / NAS-116715

dszidi:
To test view the new datasets page at the following resolutions.

- 720p (1280x720)
- 1080p (1920x1080)
- 1440p (2560x1440)
- 4k (3840x2160)

Expected Behavior:
The details cards should switch between one two and three columns. On largest screen (4k+), details cards will try to fill up the horizontal space as the height decreases.

On narrower screens, when there isn't enough screen width left for a details card column, the details section is hidden completely. This is by design. How we expose the cards to the user on smaller screens will be dealt with in another ticket/PR


undsoft:
I get:
```
SassError: 0.1879699248%/px isn't a valid CSS value.
   ╷
16 │   width: calc(100% / ($card-width + $gap));
   │          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
  src/assets/styles/mixins/cards.scss 16:10  root stylesheet
```
locally because of this line in `cards.scss`:
```
width: calc(100% / ($card-width + $gap));
```

when I try to compile it locally.

dszidi:
That line should not be in there. Are you sure you have latest changes?

undsoft (changes requested):
No, my mistake.

This is how it looks on my screen:
Снимок экрана 2022-08-03 в 13 37 06

This is how is used to look before:
Снимок экрана 2022-08-03 в 13 38 52

Old table didn't quite fit, but it almost did if we made Used column smaller (e.g. only show numbers) or changed Encryption column to only use icons, or even allowed cards to be narrower.

Other issues:
* Expanding children shift other columns
Снимок экрана 2022-08-03 в 13 41 50

* There is now horizontal scroll in details pane.
* Cards have very little right margin in the header.

dszidi:
This is what the latest designs have for the used column content. I'll update the PR to match

image

bvasilenko (approved):
Looks good to me!

One remark which could be outside of the scope of the ticket.

I've noticed quadruple word wrapping of the header, which is caused for lack of width after layout has been changed to 1-column when window width is 1420px.
image

I have an idea how we could improve that in future:
image

undsoft (changes requested):
1. I still have horizontal scrolling in details.
2. Can we try to fit 2 columns on my screen? :) My width is 1792. Worth checking resolution on common laptops as well. Okay with hiding text in Encryption column on these resolutions.

dszidi:
>  I still have horizontal scrolling in details.

At what resolution?

undsoft:
At any resolution when details pane is visible. But only when my mouse is connected and vertical scrollbar is fully visible. There is a setting in Macos general settings to make scrollbars always visible.

undsoft (approved):
Seems to be working correctly. No obvious regressions found.

PR #6928, opened 2022-08-03
NAS-117476 / 22.12 / NAS-117476: Remove 'bucket_policy_only' attribute from 'cloudsync.list_directory' call

bvasilenko:


dszidi (approved):
Was able to create cloud sync task without errors

image

PR #6929, opened 2022-08-03
NAS-112616 / 22.12 / NAS-112616: Adding return early linter rule and others

undsoft:
- "Fixes" indentation rule.
- Enums are now required to be single case.
- Prefer return early.

Testing: code review only


RehanY147 (approved):
Looks good to me.

PR #6935, opened 2022-08-04
NAS-117319 / 22.12 / NAS-117319: Improving ZFS info card

undsoft:
Additional actions and checks were added to ZFS info card.
Logic for how some of the checks work was copied from `volume-status.component.ts 358-375`.

Testing: check actions when selecting a disk and a mirror in Devices. Compare with previous logic.

Note: Sometimes network requests fail and I am not always super sure about condition checks. In this case I think it's better to be too permissive and allow backend to validate if the action is possible.


RehanY147 (approved):
Looks good

PR #6941, opened 2022-08-05
NAS-116915 / 22.12 / NAS-116915: Add the pool header row on Datasets IxTreeTable (v2)

bvasilenko:
V2 of #6883 

image

**Update**: the purpose of this PR

> Let's keep current branch and PR open and try a different option with just having some visual separation between pools:
> 
> image (3)

### DIFF
--- a/src/app/pages/datasets/components/dataset-management/dataset-management.component.scss
+++ b/src/app/pages/datasets/components/dataset-management/dataset-management.component.scss
@@ -69,6 +69,10 @@ ix-dataset-details-panel {
 
 .dataset-tree {
   border-top: 2px solid var(--lines);
+
+  ix-nested-tree-node:not(:last-child) {
+    border-bottom: 10px solid var(--bg1);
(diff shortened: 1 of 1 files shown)


undsoft (changes requested):
Doesn't look right
Снимок экрана 2022-08-05 в 11 45 43

AlexKarpov98 (changes requested):
The pool header row should be added like this.
image

bvasilenko:
@AlexKarpov98 

> The pool header row should be added like this.

The picture you've attached is depicting the previous version (v1), which is reviewed in [another PR](https://github.com/truenas/webui/pull/6883). At this PR we're implementing an alternative version. I've just updated ticket's description above, so it contains this remark.

bvasilenko:
> Doesn't look right Снимок экрана 2022-08-05 в 11 45 43

@undsoft This should be fixed by latest commit

PR #6942, opened 2022-08-05
NAS-117275 / 13.0 / NAS-117275: Logging into WebUI immediately after failover results in …

undsoft:
…both nodes rebooting

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -79,7 +79,9 @@ export class AppComponent {
     router.events.subscribe((s) => {
       // save currenturl
       if (s instanceof NavigationEnd) {
-        if (this.ws.loggedIn && s.url != '/sessions/signin') {
+        const navigation = this.router.getCurrentNavigation();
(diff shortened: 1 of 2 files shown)


src/app/app.component.ts
```diff
@@ -79,7 +79,9 @@ export class AppComponent {
     router.events.subscribe((s) => {
```
RehanY147, line 79: Should rename `s`. Ambiguous.

RehanY147 (commented):
Need testing instructions and VM to test on

src/app/app.component.ts
undsoft: Fair enough :)

undsoft:
Testing instructions are: connect to the machine in ticket, click Initiate Failover on main dashboard, login. The machine should not reboot a second time.

bvasilenko (approved):
Looks good to me. However it needs more testing with the HA enabled setups

PR #6946, opened 2022-08-05
NAS-117249 / 22.12 / NAS-117249: Fixing errors in replication tasks

undsoft:
Testing: go to Data Protection, try to create a replication task, but make request fail (for example by changing something in code).
Check that there are no unexpected errors in the console and that there are no new periodic snapshot tasks.


RehanY147 (changes requested):
I changed the payload to `{}` wherever a call was being made to `this.createCalls` and got this error on the console.
image

undsoft:
I've added a small check, but still this method of testing is bit contrived. I would suggest adding a new dummy field in the payload.

RehanY147 (approved):
Looks good

PR #6947, opened 2022-08-05
NAS-116755 / 22.12 / NAS-116755: Reword label for ACL types in SCALE permission editor

Raz-Dva:
Storage -> Edit Options
Dataset ACL Type NFSv4 label changed to SMB/NFSv4.


denysbutenko (approved):
Looks good

PR #6948, opened 2022-08-06
NAS-117320 / 22.12 / Making immutable fields uneditable when editing an app

RehanY147:
To test, you can go to the scale cloud vm and edit the asdas app there. On submission there should be no errors about uneditable fields.


undsoft (changes requested):
I've added rehan-test branch of truecharts and tried installing netdata.

https://user-images.githubusercontent.com/1481270/183381551-66074ac7-14d4-4a34-9238-94b5c7486b5d.mov

src/app/modules/ix-forms/components/ix-dynamic-form/ix-dynamic-form-item/ix-dynamic-form-item.component.html
```diff
@@ -36,53 +32,48 @@
...
-    *ngIf="dynamicSchema.type === DynamicFormSchemaType.Input"
+    *ngIf="dynamicSchema.type === DynamicFormSchemaType.Input && !isHidden"
     [formControlName]="dynamicSchema.controlName"
     [label]="dynamicSchema.title | translate"
     [placeholder]="dynamicSchema.placeholder | translate"
     [tooltip]="dynamicSchema.tooltip | translate"
     [required]="dynamicSchema.required"
-    [fxHide]="isHidden || dynamicSchema.hidden"
-    [class.disabled]="dynamicSchema.editable === false"
+    [disabled]="dynamicSchema.editable === false"
```
undsoft, line 41: This now adds a console warning:
```
 It looks like you're using the disabled attribute with a reactive form directive. If you set disabled to true
  when you set up this control in your component class, the disabled attribute will actually be set in the DOM for
  you. We recommend using this approach to avoid 'changed after checked' errors.
```

undsoft:
Tried to install netdata 3 times on two separate machines, it just gets stuck at 75%.
Created a middleware ticket for it: https://ixsystems.atlassian.net/browse/NAS-117599
If you have other instructions on how I can test this, lmk, otherwise I'll wait for response from middleware.

RehanY147:
I'm not sure what the issue is. I can install `netdata` from `rehan-catalog` on this branch using the Scale vm available to the team.

RehanY147:
install issue

undsoft (changes requested):
Installing/editing Netdata seems to be okay.
I've tried editing a different app that I had installed and I couldn't press Save. Also name was editable when previously it was disabled.
Ideally, we should use reactive forms syntax for disabling controls, not [disabled] attribute.

undsoft (changes requested):
If you go to our m-40, installed apps, Bora and click `Enable Pod resource limits` this will no longer show additional fields.

undsoft (changes requested):
I've installed ordinary Truecharts on our m-40. I get an error dialog when I try to install apps from it (e.g. when installing `ddns-go`), which didn't happen before.

undsoft (approved):
Seems like this time it worked, couldn't find any regressions.

PR #6949, opened 2022-08-07
NAS-117520 / 22.12 / Reimplement disk health card

AlexKarpov98:
Testing:
Storage => Disk Health Card


src/app/interfaces/storage.interface.ts
```diff
@@ -207,3 +207,13 @@ export type DiskWipeParams = [
...
 }
+
+export interface DiskTemperatureAgg {
+  [disk: string]: TemperatureAgg;
+}
+
+export interface TemperatureAgg {
+  min: number | 0;
+  max: number | 0;
+  avg: number | 0;
```
bvasilenko, line 218: I think `number | 0` is literally same as `number`. Is there a reason we can't keep it simply `number` here?

bvasilenko (approved):

src/app/interfaces/storage.interface.ts
AlexKarpov98: Thanks. That should be the only number. I'll fix it.

PR #6950, opened 2022-08-08
NAS-117509 / 22.12 / NAS-117509: Sync colors and fonts

bvasilenko:
**Testing**
Go to **Storage** page. On the **Usage** card gauge chart should have:
- same share of red color as in other places
- same font as the number of "Unassigned disks"


AlexKarpov98 (approved):
LGTM!
Screenshot_39

PR #6951, opened 2022-08-08
NAS-117550 / 13.0 / NAS-117550: Jira Oauth for support form

undsoft:
Testing: go to System -> Support and use the form to submit a bug.


undsoft:
Wrong ticket number.

PR #6953, opened 2022-08-08
NAS-117540 / 22.12 / NAS-117540: Select the correct pool on datasets page

bvasilenko:
**Testing**

On **Storage** (/storage2), click **Manage Datasets**. URL previously was changing to http://localhost:4200/datasets/ . 

I've changed the code, so that URL would contain the name of the pool:

http://localhost:4200/datasets/[pool name]


denysbutenko (approved):
👍🏻

PR #6955, opened 2022-08-08
NAS-117119 / 22.12 / NAS-117119: Change loading animation in permissions card

Raz-Dva:
Check the skeleton-loader in Permissions card (in storage2 module).


undsoft (changes requested):
You can always show `Owner`, `Group` and `Path` labels as they are static and show loading indicators for values only. Similar to how other cards do it.
I believe `Path` can always be shown as is not affected by loading.
For permission lines in the card you can also use:
```
<ngx-skeleton-loader [count]="3"></ngx-skeleton-loader>
```

undsoft (approved):
Looks good.

PR #6957, opened 2022-08-09
NAS-113681 / 22.12 / NAS-113681: Add support for prefer-as-const for class members

undsoft:
+ Removing unused code

Testing: general shallow regression testing with attention to widgets on main dashboard.


dszidi (commented):
Configuring the dashboard doesn't seem to work. I can fill out the form but it keeps reverting back to initial state

dszidi (approved):
You can disregard previous comment. Seems the problem is also present in master.

LGTM

PR #6961, opened 2022-08-10
NAS-117574 / 22.12 / Update Roles column on Datasets page

denysbutenko:
For testing, go to Datasets page and check Roles column


bvasilenko (approved):
Looks good to me!

PR #6963, opened 2022-08-10
NAS-116715 / None / NAS-116715 mobile support

dszidi:


src/assets/styles/mixins/breakpoints.scss
```diff
@@ -1,10 +1,19 @@
+// Global Breakpoints
+$breakpoint-laptop: 1024px;
```
undsoft, line 2: We already had breakpoints based on material in `scss-imports/cssvars`.
Need to consolidate with them in some way.

src/assets/styles/mixins/breakpoints.scss
```diff
@@ -1,10 +1,19 @@
...
+$breakpoint-tablet: 767px;
+$breakpoint-min-mobile: 320px;
```
undsoft, line 4: I think cool kids these days try to avoid `mobile` and `tablet` in names, though I don't have any strong preference about it.

src/app/pages/datasets/components/dataset-management/dataset-management.component.ts
```diff
@@ -52,6 +68,12 @@ export class DatasetsManagementComponent implements OnInit {
       });
   }
 
+  ngAfterViewInit(): void {
+    if (window.innerWidth < 1280) {
```
undsoft, line 72: Magic number needs to be extracted somewhere. If this breakpoint matches what Material uses, I think you can import and use `Breakpoints/BREAKPOINTS` constants.

src/app/pages/datasets/components/dataset-management/dataset-management.component.scss
```diff
@@ -73,6 +91,92 @@
   @media (max-width: $breakpoint-hidden) {
     display: none;
   }
+
+  // One Card Column
+  /*@media (min-width: $breakpoint-hidden) and (max-width: calc($breakpoint-singlecolumn - 1px)) {
```
undsoft, line 96: Comments need to go.

src/app/pages/datasets/components/dataset-management/dataset-management.component.html
```diff
@@ -67,7 +71,7 @@
     </div>
   </div>
 
-  <div class="details-container"
+  <div class="details-container" [ngClass]="{ 'details-container-mobile': showMobileDetails }"
```
undsoft, line 74: [class.details-container-mobile]="showMobileDetails"

src/app/pages/datasets/components/dataset-details-panel/dataset-details-panel.component.ts
```diff
@@ -22,6 +22,8 @@ export class DatasetDetailsPanelComponent implements OnInit {
   @Input() dataset: DatasetDetails;
   @Input() parentDataset: DatasetDetails | undefined;
 
+  @Output() CloseMobileDetails: EventEmitter<void> = new EventEmitter<void>();
```
undsoft, line 25:
```
@Output() closeMobileDetails = new EventEmitter<void>();
```

undsoft (changes requested):
Look much better than what we currently have.
Some issues that I've found:
1. On mobile when you press on Jobs icon, the jobs popover still appears on top.
2. On mobile on some forms, bottom panel overlaps content.
3. Columns are still uneven in children:
Снимок экрана 2022-08-11 в 12 38 20
4. Did we rearrange cards? This is how zvols look:
Снимок экрана 2022-08-11 в 12 41 31
5. I think we need to shrink dataset space management chart on some resolutions. Here's how it looks on my screen. I'd rather not have scrolling instead of the big blue donut.
Снимок экрана 2022-08-11 в 12 46 53
6. Devices page would need to get the same treatment.

src/assets/styles/mixins/breakpoints.scss
dszidi: I used those names because that is what the Chrome developer tools call them. I can change them if you like.

image

image

src/app/pages/datasets/components/dataset-management/dataset-management.component.scss
dszidi: Done

src/app/pages/datasets/components/dataset-management/dataset-management.component.html
dszidi: Done

src/app/pages/datasets/components/dataset-details-panel/dataset-details-panel.component.ts
dszidi: Done

src/app/pages/datasets/components/dataset-management/dataset-management.component.ts
dszidi: I found that Breakpoints also contains a BreakpointsObserver I can subscribe to. I switched to using that instead of hostlistener

dszidi:
> 1. On mobile when you press on Jobs icon, the jobs popover still appears on top.

Fixed

> 2. On mobile on some forms, bottom panel overlaps content.

Which forms and what resolution? From what I have seen most forms need to be resized but I didn't come across one that had the footer at a higher z-index.

> 3. Columns are still uneven in children:

Fixed

> 4.  Did we rearrange cards?...

Yes. The main motivator was that permissions had the most dynamic card height possibilities so it was put last. I chose the order of the rest of the cards based on both importance as well as card height predictability. This was required to make the columns wrap which relies on max height rules on the container element.

> 5.  I think we need to shrink dataset space management chart on some resolutions...

Done. I dialed down the doughnut on single column and dual column slim layouts.

> 6.  Devices page would need to get the same treatment

Will do. Just wanted to get datasets page dialed in first

undsoft (changes requested):
1. I've noticed that cards wrapping can be better. Like here, there is enough space in the second column to fit everything without scrolling, but it won't try to reflow.
Снимок экрана 2022-08-12 в 10 45 26
2. Donut chart can be 25% of its current size on smaller screens (we could use a media query for height).

denysbutenko:
It looks much better now.
 
https://user-images.githubusercontent.com/351613/184382005-9b26eaa2-0cfe-497a-adaa-3495e4deab01.mp4

Some observations:

- Horizontal scrolling appears when the dataset name does not fit on the card
- `Add dataset` button can disappear

Maybe it happens rarely and does not require action for now. Just FYI

dszidi:
> * Horizontal scrolling appears when the dataset name does not fit on the card
> * `Add dataset` button can disappear

Maybe we can do something like keep root dataset name and dataset name in path and replace the rest with ...
So in your case, `TANK/Sit/nostrud/incididunt/minim/in/laboris/aliquip/voluptate` would become `TANK/.../voluptate`

undsoft (changes requested):
1. When I have mouse connected, both vertical and horizontal scrollbars appear.
2. When page is loaded for the first time, I have two vertical scrollbars until I start scrolling. May be an artifact of `IxDetailsHeightDirective`. If too hard to fix, skip this item.
3. I thought we've agreed to only keep one path and remove path from Permissions.
4. I would suggest removing copy button. It doesn't work correctly (popover appears outside of visible area and this is not in scope for this PR.
5. Donut is still large, but weirdly gets smaller if I enter responsive testing. Then it gets smaller.

https://user-images.githubusercontent.com/1481270/186135157-b641b588-e72c-44cb-828e-1a5029b64cda.mov

dszidi:
1. Can't seem to reproduce scroll bars on my end
2. Adjusted life cycle hooks on IxDetailsHeightDirective. Seems to be better now
3. Removed path from Permissions card
4. Removed copy button. 
5. Made adjustments to Donut

undsoft (commented):
Horizontal scrolling is gone, donut can still be smaller, still have two vertical bars.
Let's continue with Devices page, I'd like to get this in for Beta1. We can improve later.

src/app/modules/common/topbar/topbar.component.ts
```diff
@@ -226,6 +232,34 @@ export class TopbarComponent implements OnInit, OnDestroy {
...
+      if (this.taskDialogRef) {
+        this.taskDialogRef.updatePosition(this.dialogPosition);
+      }
+    };
+
+    // Check width on page load
+    positioner(window.innerWidth < 768);
+
+    // Check width on resize
+    this.breakpoints.addEventListener('change', (state: { matches: any }) => positioner(state.matches));
```
undsoft, line 260: `: { matches: any }` can be removed.

undsoft (approved):
Mission accomplished.

PR #6968, opened 2022-08-11
NAS-117581 / 22.02.4 / NAS-117581: Fix Launch Docker Image button bugs

bvasilenko:
**Testing**

1. Go to **Apps** page  > **Launch Docker Image**
2. See bug description in the ticket


src/app/pages/applications/applications.component.ts
```diff
@@ -74,19 +74,10 @@ export class ApplicationsComponent implements OnInit, AfterViewInit {
...
-        case ApplicationTab.AvailableApps:
-          this.catalogTab.onToolbarAction(evt);
-          break;
-        case ApplicationTab.DockerImages:
-          this.dockerImagesTab.onToolbarAction(evt);
-      }
+      this.chartTab.onToolbarAction(evt);
+      this.catalogTab.onToolbarAction(evt);
+      this.manageCatalogTab.onToolbarAction(evt);
+      this.dockerImagesTab.onToolbarAction(evt);
```
AlexKarpov98, line 80: I think that it should apply an event exactly into a tab content when the tab is changed so a `switch-case` condition should be keep as original code.

A `disabled` property of the **`Launch Docker Image`** button depends on `isSelectedPool` state.
So I think that the solution is to set `isSelectedPool` state correctly when the page is loaded.

AlexKarpov98 (changes requested):
Need to change.

bvasilenko:
Thanks for your review, @AlexKarpov98  . I realized that I need to push more changes. 
I'm closing this, and once ready, I'll create a new PR.

PR #6971, opened 2022-08-11
NAS-117539 / 22.12 / NAS-117539: Select first row in Datasets on load

undsoft:
See ticket description.


denysbutenko (approved):
The issue with the inactive node on load is gone. Looks good!

PR #6974, opened 2022-08-12
NAS-117646 / None / NAS-117646: Reducing amount of any's

undsoft:
Code review should be enough.


src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
```diff
@@ -40,6 +41,12 @@ import {
 } from 'app/services';
 import { ModalService } from 'app/services/modal.service';
 
+type CloudSyncFormTask = Overwrite<CloudSyncTask, {
```
bvasilenko, line 44: `Overwrite` usage looks more like a hack. Wouldn't it be better to declare a class?

bvasilenko (approved):
Looks good overall. I've suggested an optional adjustment

src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
undsoft: Agree that it's a hack, but it's a hack because somebody put a different type in field. We will refactor this component at some point and this will go away.

PR #6975, opened 2022-08-13
NAS-117572 / 22.12 / NAS-117572: Handle empty states better on new storage pages.

AlexKarpov98:
**Testing:**
Storage, Datasets
It should show the empty card if there are no pools.

image


src/app/pages/datasets/components/dataset-management/dataset-management.component.html
```diff
@@ -1,78 +1,86 @@
...
-  <div class="table-container">
-    <ix-fake-progress-bar
-      class="loader-bar"
-      [loading]="isLoading$ | async"
-    ></ix-fake-progress-bar>
-
-    <div class="dataset-tree-search">
-      <ix-search-input [disabled]="!dataSource" (search)="onSearch($event)"></ix-search-input>
-    </div>
+<div fxLayout="column" class="material mat-card" [class.is-empty]="dataSource.data.length === 0" *ngIf="dataSource.data.length === 0; else datasetContainer">
```
undsoft, line 1: So, if the block only appears when `dataSource.data.length === 0`, then the `is-empty` class is always going to be there.
Let's put `*ngIf` as first attribute.

undsoft (approved):
Looks good, but I think it would look even better if we didn't have mat-card on the background similar to how dashboard looks:
Снимок экрана 2022-08-15 в 10 15 25

src/app/pages/datasets/components/dataset-management/dataset-management.component.html
AlexKarpov98: Fixed

PR #6976, opened 2022-08-14
NAS-117628 / 22.12 / NAS-117628: Dataset quota is not shown

AlexKarpov98:
**Testing:**
Datasets => Dataset Space Management card

- If `Quota for this dataset and all children` field is changed as not 0:

image

- If `Quota for this dataset and all children` field is changed as 0 and `Quota for this dataset` field is changed as not 0:

image


dszidi (approved):
LGTM

image

PR #6978, opened 2022-08-15
NAS-117126 / 22.02.4 / NAS-117126: Setting timezone for 'getTaskNextRun'

bvasilenko:
**Testing**

On the **Data Protection** page, at the **Periodic Snapshot Tasks**

**Next Run** column does now takes server's time zone into account.

1. Go to **System Settings > General > Localization** and see **Timezone** setting. By default it's set to "Los Angeles"
2. Go to **System Settings > Shell**  and type in `date` command, to see server's current time
3. Set a schedule **Add/Edit Periodic Snapshot** and note that this schedule is bound to the server's time. _Note that you can choose "Custom" schedule, and see explanation note in UI that it's in "Los Angeles" time zone._ **Example** If server's time were `00:52:00` as it was for me, then next run would be in **23 hours and 8 minutes**
4. See the **Next Run** column: it should take "Los Angeles" time zone into account. In my example from point 3 it should display ~23 hours.


denysbutenko (approved):
image

Looks good

PR #6984, opened 2022-08-16
NAS-117704 / 22.12 / NAS-117704: Enforce linting rules in CI

undsoft:
Code style will now be enforced on repo level.
Also changes output below to show statuses separately (parallelizes jobs) .
Code review only.


RehanY147 (approved):
Looks good

PR #6991, opened 2022-08-17
NAS-117700 / 22.12 / NAS-117700: Dashboard Config settings revert after save

AlexKarpov98:
**Testing:**
Dashboard => change Configure

After changing the Configure, the widget cards changed should not be reset or reverted without login again or refresh.


denysbutenko (approved):
Works well 👍🏻

PR #6993, opened 2022-08-18
NAS-117713 / 22.12 / NAS-117713: Specify allowlist explicitly when creating an API key

undsoft:
Testing info in the ticket.


bvasilenko (approved):
Now `allowlist` is sent

image

PR #6994, opened 2022-08-19
NAS-117718 / 22.12 / NAS-117718: No pools message flashing when switching to Storage

AlexKarpov98:
**Testing:**
Go to `Storage` page
See ticket description.


undsoft:
Does the same issue exist on Datasets?

dszidi (approved):
Code looks good and I don't see the empty state at all when loading a system with pools.

AlexKarpov98:
> Does the same issue exist on Datasets?

The same issue was caused when the browser is refreshed on the Datasets page(that is when datasetStore value is initialized).
I updated it like the Storage page.

PR #6995, opened 2022-08-20
NAS-117734 / 22.12 / NAS-117734: Handle nulls in disk temperatures

AlexKarpov98:
**Testing:**
Storage => Disk Health Card
The min, max or avg temperature value should show `No Data` instead of 0 if it returns `null` from the endpoint.


undsoft (changes requested):
```
InvalidPipeArgument: 'No Data is not a number' for pipe 'DecimalPipe'.
```

May be better to have that if condition in template.

AlexKarpov98:
> ```
> InvalidPipeArgument: 'No Data is not a number' for pipe 'DecimalPipe'.
> ```
> 
> May be better to have that if condition in template.

Fixed.

undsoft (approved):
Looks good
Снимок экрана 2022-08-22 в 12 00 42

PR #6996, opened 2022-08-20
NAS-117769 / 22.12 / NAS-117769: Support for multiple in ix-explorer

undsoft:
Adds support for replication datasets in ix-explorer and a couple of tests.
Testing: go to Data Protection -> click Test and follow instructions.


AlexKarpov98 (approved):
The multiple feature in ix-explorer works well.

I reviewed it on the Backup Credentials page. I couldn't see the "Test" button on the Data Protection page.
Screenshot_7

Also I tested it on another page, updating the ix-explorer template.
For example, that is in the Restore Replication Task dialog on the data-protection/replication page.
image

A comma is unnecessary in the first position of the multiple value, I think.

undsoft:
Thanks.
The issue on `Restore Replication Task` happens because there is a default empty string value in the form. This shouldn't be there for `multiple=true`.

PR #6997, opened 2022-08-21
NAS-117634 / 22.12 / Refactor bootenv-status.component

denysbutenko:
You can access this component by going to: System Settings → Boot → Actions → Boot Pool Status.


RehanY147 (changes requested):
The buttons need a bit of adjustment. And the Active column is empty. We should show some text there to explain.

image

denysbutenko:
> The buttons need a bit of adjustment. And the Active column is empty. We should show some text there to explain.
> 
> image

Good catch. I will fix it in the next PR #7000

denysbutenko:
The latest docker image includes the changes for `bootenv-list.component` from #7000

RehanY147 (approved):
Looks good.

PR #7003, opened 2022-08-22
NAS-117795 / 22.12 / Move replace-disk-dialog to storage2

denysbutenko:
Reduce dependencies from old storage module. Code review is enough


denysbutenko:
> The dialog is used in the new Storage but the component is only declared in the old Storage module.

Fixed

RehanY147 (approved):
Looks good

PR #7005, opened 2022-08-23
NAS-117073: fix: Pool Managers' dev type doesn't initially match the Estimated raw capacity for Safari

bvasilenko:
**What causes the bug?**

DOM gets updated, while Safari does not rerender the updated text. On the screenshot below compare "Mirror" in DOM vs. "Stripe" in Safari's viewport. On window resize, it renders correctly.

image

Here's what you need to replicate the bug:

1. Insert debugging code shown below into your `manager.component.ts` . This will simulate 2 disks with not empty `duplicate_serial`. (This is one of ways how user can see the bug.)

image

2. In Safari 15.x, click **Show disks with non-unique serial numbers**. Check 1st and 2nd disk, and click **->** arrow button. See video depicting these steps.

https://user-images.githubusercontent.com/20611516/186094627-bf8c2e3d-3dd6-48f3-8f77-e6cd064b6e65.mp4

**Testing**

Go through steps described above

### DIFF
--- a/src/app/pages/storage/volumes/manager/vdev/vdev.component.ts
+++ b/src/app/pages/storage/volumes/manager/vdev/vdev.component.ts
@@ -1,4 +1,5 @@
 import {
+  ChangeDetectorRef,
   Component,
   ElementRef,
   Input,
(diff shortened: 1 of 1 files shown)


undsoft:
Tried to test this. Couldn't reproduce the original issue (before the fix).
@dszidi If you cannot reproduce original issue as well, okay to fix as is. Extra `markForCheck` shouldn't do any harm.

dszidi (approved):
Was able to reproduce issue with provided debug code. 
Pool creation form behaves as described and passes test.
LGTM

PR #7008, opened 2022-08-23
NAS-117680 / 22.12 / NAS-117680: update ix-icon for dataset details

bvasilenko:
**Testing**

1. Datasets > choose a dataset. This should result in a change of an icon positioned "Details for" at the details section.
2. Refresh page, and choose another node of datasets tree. This should result in another change of an icon.

### DIFF
--- a/src/app/modules/ix-icon/ix-icon.component.ts
+++ b/src/app/modules/ix-icon/ix-icon.component.ts
@@ -1,6 +1,6 @@
 import {
   Attribute, ChangeDetectionStrategy, Component, ElementRef,
-  ErrorHandler, Inject, Input, OnInit, Optional, ViewEncapsulation,
+  ErrorHandler, Inject, Input, OnChanges, OnInit, Optional, ViewEncapsulation,
 } from '@angular/core';
 import {
   MatIcon, MatIconDefaultOptions, MatIconLocation, MatIconRegistry, MAT_ICON_DEFAULT_OPTIONS, MAT_ICON_LOCATION,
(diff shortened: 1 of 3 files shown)


src/app/pages/datasets/components/dataset-icon/dataset-icon.component.html
```diff
@@ -1,3 +1,4 @@
 <!-- TODO: https://jira.ixsystems.com/browse/NAS-116793 -->
 <!-- TODO: Remove me -->
-<ix-icon [name]="name"></ix-icon>
+<ix-icon *ngIf="isRoot" name="ix:dataset"></ix-icon>
```
undsoft, line 3: Feels like this should be within `name` getter.

src/app/modules/ix-icon/ix-icon.component.ts
```diff
@@ -70,15 +70,24 @@ export class IxIconComponent extends MatIcon implements OnInit {
...
-    if (this.iconName.startsWith('ix:')) {
-      this.svgIcon = this.iconName;
-    } else if (this.iconName.startsWith('mdi')) {
+    this.updateIcon(this.iconName);
+    super.ngOnInit();
+  }
+
+  ngOnChanges(): void {
+    this.updateIcon(this.iconName);
+  }
```
undsoft, line 79: Should parent `onChanges` be called as well?

undsoft (approved):
Icons change well, don't see any issues.

src/app/modules/ix-icon/ix-icon.component.ts
bvasilenko: Parent class has no `ngOnChanges` method.

PR #7013, opened 2022-08-24
NAS-117576 / 22.12 / NAS-117576: Optimizing calls on storage dashboard

undsoft:
Some calls were optimized, but you may still face an error if you have a lot of pools. A hack was introduced to make it less likely.

Testing:
* Check loading on storage dashboard for many pools.
* Check usage card.


dszidi (approved):
Code looks good. All usage cards are properly populated
LGTM

image

PR #7015, opened 2022-08-24
NAS-117803 / 22.12 / NAS-117803: Fixes empty dashboard on first login

undsoft:
Testing: 
1. In shell `midclt call user.pop_attribute 1 dashState`
2. Refresh page. 
3. Go to dashboard and test it.


AlexKarpov98 (approved):
Works well.
After the shell command the dashboard configuration is initialized.

PR #7018, opened 2022-08-25
NAS-117854 / 22.12 / NAS-117854: Add 'validate_host_path' field

bvasilenko:
**Testing**

On the **Apps** page, at the toolbar **Settings > Advanced Settings**

- New checkbox "Validate host path" saves/loads to a new field `validate_host_path` in `kubernetes.update`  (this field presents at the latest nightly build)

- When user unchecks this checkbox (in other words when it is `false`), UI shows a warning: "Disabling validation of host path can result in a data loss."


RehanY147 (changes requested):
Instead of showing a warning on submit, we should show a yellow warning underneath the field immediately when it is unchecked

bvasilenko:
@RehanY147 Now warning is displayed right below the UI element

RehanY147 (approved):
Looks good

PR #7020, opened 2022-08-26
Better error handling for the new Storage modules

RehanY147:


dszidi (changes requested):
I don't see any problems with the logic. All looks good.

Small nitpick. We're supposed to be avoiding shorthand parameter names. I've seen in quite a few places where the error handler starts with `(err) => {...}` where we should be using a more descriptive name. We should also be avoiding `(res) => {...}`

dszidi (approved):
LGTM

PR #7023, opened 2022-08-26
NAS-117372 :Add client side validation for app names

Raz-Dva:
App -> Available Applications -> Install application
The user should get an immediate message error while typing a name (field Application name)
Application name must have the following:
Lowercase alphanumeric characters can be specified
Name must start with an alphabetic character and can end with alphanumeric character
Hyphen '-' is allowed but not as the first or last character
 e.g abc123, abc, abcd-1232

### DIFF
--- a/src/app/modules/entity/entity-job/entity-job.component.html
+++ b/src/app/modules/entity/entity-job/entity-job.component.html
@@ -57,7 +57,7 @@
     </div>
   </div>
 
-  
+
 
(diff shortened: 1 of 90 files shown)


undsoft (approved):
Let's change message to
> Name must start with an alphabetic character and end with an alphanumeric character. Hyphen is allowed in the middle.

Othewise looks good.

PR #7024, opened 2022-08-26
NAS-117766: Extracting ix-label

undsoft:
Test ix form components.

### DIFF
--- a/src/app/modules/ix-forms/components/ix-input/ix-input.component.spec.ts
+++ b/src/app/modules/ix-forms/components/ix-input/ix-input.component.spec.ts
@@ -3,6 +3,7 @@ import { FormControl } from '@ngneat/reactive-forms';
 import { createHostFactory, Spectator } from '@ngneat/spectator/jest';
 import { MockComponent } from 'ng-mocks';
 import { IxErrorsComponent } from 'app/modules/ix-forms/components/ix-errors/ix-errors.component';
+import { IxLabelComponent } from 'app/modules/ix-forms/components/ix-label/ix-label.component';
 import { TooltipComponent } from 'app/modules/tooltip/tooltip.component';
 import { IxInputComponent } from './ix-input.component';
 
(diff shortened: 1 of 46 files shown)


src/app/modules/ix-forms/components/ix-input/ix-input.component.spec.ts
```diff
@@ -27,16 +29,22 @@ describe('IxInputComponent', () => {
...
   describe('rendering', () => {
-    it('renders a hint when it is provided', () => {
-      spectator.setInput('hint', 'Capital letters only');
+    it('renders a label and passes properties to it', () => {
+      spectator.setInput('label', 'New Password');
+      spectator.setInput('required', true);
+      spectator.setInput('tooltip', 'Minimum length is 8 characters.');
 
-      expect(spectator.query('mat-hint')).toHaveText('Capital letters only');
+      const label = spectator.query(IxLabelComponent);
```
RehanY147, line 37: This kind of testing feels a bit like cheating. We're just checking variable values. Not what's on the actual dom. There can be failure points between variables and dom still. Sort of like doing `this = that; if (this == that) ... `

src/app/modules/scheduler/components/scheduler/scheduler.harness.ts
```diff
@@ -22,8 +23,8 @@ export class SchedulerHarness extends ComponentHarness implements IxFormControlH
   getErrorText = getErrorText;
 
   async getLabelText(): Promise<string> {
-    const label = await this.locatorFor('label')();
-    return label.text({ exclude: '.required' });
+    const label = await this.locatorFor(IxLabelHarness)();
```
RehanY147, line 26: Missing the check for `!label`

RehanY147 (changes requested):

src/app/modules/ix-forms/components/ix-input/ix-input.component.spec.ts
undsoft: Not really, `spectator.query(IxLabelComponent)` will return nothing if component is not in the dom.
So here we check that ix-label is there and received correct inputs.
And we make sure that ix-label renders label correctly in its test file.

RehanY147 (approved):
LGTM

PR #7026, opened 2022-08-26
NAS-117836 / 22.12 / NAS-117836: Use new method for updating Isolated GPUs

bvasilenko:
**Testing**

On the **System settings > Advanced** page, at the **Isolated GPU Device(s)** section, after clicking **Settings** button, slide-in form **Isolated GPU PCI Id's** appears.

When saving changes, it should call `system.advanced.update_gpu_pci_ids` method, instead of `system.advanced.update`.


dszidi (approved):
Code looks good. The form sends the correct payload in testing.

image

PR #7027, opened 2022-08-27
NAS-117897 / 13.0 / NAS-117897: Showing controller in alerts

undsoft:
Testing: 
Find an HA machine in Core Jenkins (for example,  under HA API tests), see that controller is specified in Alerts.
Compare with a normal machine, where controller is not specified.


bvasilenko (approved):
Tested on tn-bhyve02.tn.ixsystems.net

Works good

image

PR #7030, opened 2022-08-29
NAS-117812 / 22.12 / Minor updates to storage pages

denysbutenko:
Changes:

- Add copy button for `Disk Serial` and `Dataset Path`
- Use snack bar service for copy button user feedback
- Update warning color on Pools Dashboard (triangles)
- Update offline color for the status column on the Devices page
- Fix text alignment on the Error dialog – `+ More info`
- Fix issue when closing `Import Pool` form triggered dashboard reloading

For testing, please check the original ticket

### DIFF
--- a/src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
+++ b/src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
@@ -62,12 +62,11 @@
         </span>
         <ix-copy-btn [text]="dataset.name"></ix-copy-btn>
       </div>
-      <div class="details-item">
+      <div class="details-item" *ngIf="!isLoading && hasComments">
         <span class="label">{{ 'Comments' | translate }}:</span>
-        <span *ngIf="!isLoading" class="value dataset-comments">
-          <ng-container *ngIf="dataset?.comments?.value; else notAvailable">
-            {{ dataset.comments.value }}
-          </ng-container>
+        <span class="value dataset-comments">
+          <span textLimiter threshold="128" [content]="dataset.comments.value" [popup]="false"></span>
+          <ix-copy-btn *ngIf="dataset.comments.value.length > 128" [text]="dataset.comments.value"></ix-copy-btn>
         </span>
       </div>
     </mat-card-content>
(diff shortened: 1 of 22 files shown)


src/app/pages/storage2/components/pools-dashboard/pools-dashboard.component.ts
```diff
@@ -86,8 +87,10 @@ export class PoolsDashboardComponent implements OnInit, AfterViewInit {
     this.store.loadDashboard();
 
     this.slideIn.onClose$
-      .pipe(untilDestroyed(this))
-      .subscribe(() => this.store.loadDashboard());
+      .pipe(
+        filter((value) => value.modalType === ImportPoolComponent && value.response === true),
```
undsoft, line 91: I don't think we need to do this. `onClose$` needs to be improved, but tying directly to `ImportPoolComponent` creates tightly coupled code here.

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
```diff
@@ -53,13 +53,22 @@ <h3 mat-card-title *ngIf="isZvol">{{ 'Zvol Details' | translate }}</h3>
...
           [matTooltip]="dataset.name"
           matTooltipPosition="above"
           class="value dataset-path">
           {{ dataset.name }}
         </span>
+        <ix-copy-btn [text]="dataset.name"></ix-copy-btn>
+      </div>
+      <div class="details-item">
+        <span class="label">{{ 'Comments' | translate }}:</span>
+        <span *ngIf="!isLoading" class="value dataset-comments">
```
undsoft, line 67: 1. Let's not show the row when comment is not available.
2. Long comment won't wrap.
3. It's weird, but adding a comment on parent will show the same comment on all children. I'm not sure if this is backend or not, but I don't think we should show parent comments like this.

undsoft (changes requested):

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
denysbutenko: image

1. Ok done
2. Fixed. If the text is greater than 128 symbols, it will be truncated, and copy button will be added at the end.
3. Good catch. The comments are a ZFS property and can be inherited. Added check for the source of comments

undsoft (approved):
Nice work.

PR #7031, opened 2022-08-29
NAS-117902 / 22.12 / NAS-117902: Fix form when show_if is '!='

bvasilenko:
## Testing

1. Pick any app, for example **jackett**
1. Go to **Services**
1. Change **Service Type** to **LoadBalancer**
   
   - Expected Result: It should show LoadBalancer IP, External IP’s, IP Family Policy and IP Families


RehanY147 (approved):
LGTM

PR #7032, opened 2022-08-29
NAS-117875 / 22.12 / NAS-117875: redirect user to Available Applications if pool is not chosen

bvasilenko:
## Testing

On **Apps** page, for the first time (when pool is not set), 

- Expected Result: user should be sent to the **Available Applications** tab and **Choose a pool for Apps** popup should be displayed


dszidi:
@undsoft do we want to completely prevent user from going to the "Installed Applications" tab until they set a pool? 
Seems a bit intrusive.

undsoft:
Yeah, maybe, but to be fair if they haven't set up a pool, they wont' have any installed applications.
If we can easily allow them to go to Installed applications after redirect, let's do it, otherwise it's okay to keep how it is now.

bvasilenko:
@dszidi 

Now **Choose a pool** opens only when user visits the **Apps** page, not when user switches to **Installed/available applications**

dszidi (approved):
Modal behaves as described above.
LGTM

PR #7035, opened 2022-08-30
NAS-117924 / 22.12 / Minor updates to dataset page and admin layout

denysbutenko:
Changes:

- Fix misaligned expand buttons on dataset tree
- Use more space for dataset name/path before truncating
- Add loading state for admin layout component. So we can avoid extra sidebar toggling when user preferences are loaded.
- Do not open mobile details when a user just landed on devices/dataset pages from the sidebar menu
- Use truenas icon for system dataset in Roles column
- More details available in the original ticket 

For testing, go to devices/dataset pages and look for issues or regressions


RehanY147 (changes requested):
UI gets stuck on this screen
image

denysbutenko:
> UI gets stuck on this screen image

Just tested with local and cloud VMs and can not reproduce it. I added 5 seconds timeout to load preferences otherwise, it will show UI with default settings.

RehanY147 (changes requested):
If you break the loadPreferences effect by replacing the `userId` in the middleware call with a string for example, the UI breaks in this case. Previously, you would get an error, but the UI is still usable. You can use a lot of it. But now, the UI gets stuck on the new loading page.

I can sometimes sign in but sometimes I get stuck on the new loading page with these errors on the console
image

RehanY147 (changes requested):
If I go to the datasets page from a different page, I still go to the details view on mobile screens.

RehanY147 (changes requested):
Icons should have tooltips to explain their meaning in the datasets tables

denysbutenko:
> If you break the loadPreferences effect by replacing the `userId` in the middleware call with a string for example, the UI breaks in this case. Previously, you would get an error, but the UI is still usable. You can use a lot of it. But now, the UI gets stuck on the new loading page.
> 
> I can sometimes sign in but sometimes I get stuck on the new loading page with these errors on the console image

Yeah, that was an issue with change detection. Should be resolved now

denysbutenko:
> If I go to the datasets page from a different page, I still go to the details view on mobile screens.

Yeah, since a user is redirected to the first node on page loading, it invokes opening mobile details. Reverted to the original behavior, let's revisit it later.

RehanY147 (changes requested):
The side-nav behaves weirdly when the screen is in tab mode (or when it has a shorter width on the browser). It opens in two steps some times. Once it only opens as icons. Then goes back. Then only opens as full form.

denysbutenko:
> The side-nav behaves weirdly when the screen is in tab mode (or when it has a shorter width on the browser). It opens in two steps some times. Once it only opens as icons. Then goes back. Then only opens as full form.

Nice catch 👍🏻 Fixed this regression

denysbutenko:
I added a fix for NAS-117959 since it's in the same area.
AR: If you set any other language from `en` and reload the page, you will see UI in English, but user preferences say it should be a different one.
ER: UI should be in language from user preferences

src/app/modules/common/layouts/admin-layout/admin-layout.component.ts
```diff
@@ -86,85 +84,87 @@ export class AdminLayoutComponent implements OnInit, AfterViewChecked, AfterView
...
 
-    this.isSidenavCollapsed = this.layoutService.isMenuCollapsed;
-
-    this.ws.authStatus.pipe(untilDestroyed(this)).subscribe((evt) => {
-      this.core.emit({ name: 'Authenticated', data: evt, sender: this });
+    combineLatest([
+      this.store$.pipe(waitForGeneralConfig),
+      this.store$.pipe(waitForPreferences),
+    ]).pipe(
+      timeout(10000),
```
RehanY147, line 146: This seems like an arbitrary amount of seconds. Request could just take longer than this. We shouldn't have to cater the requests specifically to adjust to the new loading screen. Plus, the new loading screen is needlessly blocking the user from content that they could be viewing sooner just because the prefs haven't loaded yet. I think we should remove this screen. @undsoft thoughts?

RehanY147 (changes requested):

RehanY147 (changes requested):

src/app/modules/common/layouts/admin-layout/admin-layout.component.ts
denysbutenko: Yeah, it seems better to keep as is

src/app/modules/common/layouts/admin-layout/admin-layout.component.ts
```diff
@@ -177,6 +177,20 @@ export class AdminLayoutComponent implements OnInit, AfterViewChecked, AfterView
       });
   }
 
+  patchScrollOnWindows(): void {
+    const navigationHold = document.getElementById('scroll-area');
+
+    // Allows for one-page-at-a-time scrolling in sidenav on Windows
+    if (window.navigator.platform.toLowerCase() === 'win32') {
+      navigationHold.addEventListener('wheel', (event) => {
+        // deltaY is 1 for page scrolling and 33.3 per line for regular scrolling; default is 100, or 3 lines at a time
```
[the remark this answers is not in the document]
denysbutenko: Good suggestion

denysbutenko:
Better to keep as is

PR #7037, opened 2022-08-31
NAS-117631 / 22.12 / NAS-117631: Rework to expansible/collapsible snapshots rows

bvasilenko:
**Details**

According to the ticket, the table is reworked:

  - details which was previously in the "extra columns" now moved to the expandable detail block under every row

  - for every row, when row expanded - now it gets additional details from `zfs.snapshot.query` with `id = ...` condition


src/app/pages/datasets/modules/snapshots/store/snapshot.effects.ts
```diff
@@ -25,7 +24,7 @@ export class SnapshotEffects {
     switchMap(() => this.store$.pipe(waitForPreferences)),
     switchMap((preferences) => {
       return this.ws.call('zfs.snapshot.query', [
-        snapshotExcludeBootQueryFilter,
```
undsoft, line 28: So what did this condition got removed?

undsoft (changes requested):
I don't think ticket called for removal of `Show extra columns` option. It's more like the version without extra columns should have details view.
We could also use a bit of loading indication when details view loads data. Right now it just flashes N/A, at the very least it should show empty space.

bvasilenko:
Now "Show extra columns" mode is restored, as well as `snapshotExcludeBootQueryFilter `

undsoft (approved):
There is an issue that probably existed before your changes, that would be good to fix, while we are here.
1. Go to Datasets, click on Manage Snapshots,
2. I'll be taken to a pre-filtered list of snapshots. 
3. If I then perform an action that results in snapshots being refreshed (such as deleting one), the filter will be re-applied, but only partially. There will be nothing in filter box.
If not easy to fix, this can be a separate ticket.

Other than that looks good.

bvasilenko:
The issue should be fixed now

undsoft (approved):

PR #7038, opened 2022-08-31
NAS-117867 / 22.12 / NAS-117867: Shows role card for VMs

bvasilenko:
**Testing**

Go to **Datasets**, select a dataset where VM is installed. Example:

   image

   - Expected result: **Roles** card appears at the right section, like this:

   
   image


dszidi (changes requested):
Code looks good and functions as expected. I think we could improve the text a little. The value field should be "This dataset is used by a VM" . In the link text we can also replace "Virtualization" with "VM". 

image

undsoft:
Can we tell user which VM uses the dataset?

bvasilenko:
@dszidi 
Now name of MV is displayed, and wording is improved

dszidi (approved):
Code looks good. We need a way to do better with line breaks but I'll make a separate ticket for that issue. Changes approved

PR #7039, opened 2022-08-31
NAS-117870 / 22.12 / NAS-117870: Properly handle change of an icon name at `ix-icon`

Raz-Dva:
Check for proper handling `@Input` 'name' in the ix-icon.
Icons should be correctly drawn when dynamically changing the input name.
For ligature icon different use cases
`<ix-icon>arrow_circle_right</ix-icon>
<ix-icon [name]="arrow_circle_right"></ix-icon>`
Don't mix these two options

https://user-images.githubusercontent.com/22006748/187682970-12466cbc-f1d0-4bc3-b3d2-e186ca3fe410.mp4

### DIFF
--- a/src/app/modules/ix-icon/ix-icon.component.scss
+++ b/src/app/modules/ix-icon/ix-icon.component.scss
@@ -13,8 +13,7 @@ $size: 24px;
     width: $size;
   }
 
-  &.mat-ligature-font::before,
-  &[data-mat-icon-type='svg']::before {
+  &.mat-ligature-font::before {
     display: inline-block;
     visibility: hidden;
     width: 0;
--- a/src/app/modules/ix-icon/ix-icon.component.scss
+++ b/src/app/modules/ix-icon/ix-icon.component.scss
@@ -13,7 +13,10 @@ $size: 24px;
     width: $size;
   }
 
-  &.mat-ligature-font::before {
-    content: attr(data-mat-icon-name);
+  &.mat-ligature-font::before,
+  &[data-mat-icon-type='svg']::before {
+    display: inline-block;
+    visibility: hidden;
+    width: 0;
   }
 }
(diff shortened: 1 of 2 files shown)


bvasilenko (commented):
I have an attempt to verify by rolling back d5fdb50867a3eb6ff7cf549c280eccc0fa6eecbf commit for getter `DatasetIconComponent.name`, and placing with a single line `<ix-icon [name]="name"></ix-icon>`

Actual result:
no improvement, e.g. bug still occurs when changing `name` property from `'ix:dataset'` to `'folder'` 
image

Expected result: `name` property can be dynamically changed to any correct icon name.

The purpose of `ix-icon` was to provide unified icon component. As alternative solution, split `ix-icon` into 2 separate components. One for "ix:" icons, and another for other icons.
@Raz-Dva @undsoft would you please share your thoughts of this?

undsoft:
I'd prefer to keep the sprit of ix-icon alive, i.e. one component to handle icons from different sources.

denysbutenko:
> Check for proper handling `@Input` 'name' in the ix-icon. Icons should be correctly drawn when dynamically changing the input name. For ligature icon different use cases `<ix-icon>arrow_circle_right</ix-icon> <ix-icon [name]="arrow_circle_right"></ix-icon>` Don't mix these two options

The main difference between the ligature approach is that `<ng-content></ng-content>` become available when `ngAfterContentInit()` is triggered, which means the icon will be initialized and then re-rendered. And using it like `<ix-icon name="icon_name">` allows us to initialize the proper icon from the beginning. I'll take a look at how we can improve it

Raz-Dva:
I also think that using one component to work with different types of icons is the right decision.

bvasilenko (approved):
Works good to me:

https://user-images.githubusercontent.com/20611516/188625061-59029d63-4679-4983-ba6f-0d91a7e659e8.mp4

PR #7041, opened 2022-08-31
NAS-113691 / 22.12 / NAS-113691: invalidate committed network changes, after user visits network settings

bvasilenko:
## Testing

1. Enter **Network** page, open settings of your network adapter, make a change
   
   image

    After you save your changes, you will see **unapplied changes** ui block with the **Test Changes** button and the input box to specify how long (in seconds) changes will be kept:
   
   image

   
2. Click **Test changes** button
    After you click it, you’ll see **temporarily applied changes** ui block, counting down the amount of seconds. 
    
   image

   - Normally, to apply your changes, you need to click the **Save Changes** button before this countdown reaches “0”, and if you do so, your changes will be persisted. If you won’t be quick enough, your changes will be unapplied, and you will be taken back to **unapplied changes** ui block.

   - To replicate the previously buggy scenario, ignore the normal scenario described above, and instead go to step 3

3. Before the countdown reaches “0”, open settings of your network adapter again, make another change and submit it.

   
    image

Expected result: after another change is submitted, 
- you should not be able to continue normal scenario:
   - **temporarily applied changes** ui block should not be kept visible
   - countdown should be stopped and nothing should happen after you wait for remaining amount of seconds
- instead, you should be taken back to **unapplied changes** ui block, from where you should be able to follow normal scenario

### DIFF
--- a/src/app/pages/network/network.component.ts
+++ b/src/app/pages/network/network.component.ts
@@ -216,9 +216,11 @@ export class NetworkComponent implements OnInit, OnDestroy {
 
     this.slideInService.onClose$.pipe(untilDestroyed(this)).subscribe(() => {
       this.staticRoutesTableConf.tableComponent.getData();
+      this.getInterfaces();
       this.checkInterfacePendingChanges();
     });
 
+    this.getInterfaces();
     this.checkInterfacePendingChanges();
     this.core
       .register({ observerClass: this, eventName: 'NetworkInterfacesChanged' })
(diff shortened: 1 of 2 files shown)


src/app/pages/network/network.component.ts
```diff
@@ -261,51 +263,67 @@ export class NetworkComponent implements OnInit, OnDestroy {
...
-          this.checkinWaiting = true;
-        } else {
-          this.checkinWaiting = false;
-          this.checkinRemaining = null;
-          if (this.checkinInterval) {
+      .toPromise();
+  }
+
+  private async cancelCommit(): Promise<void> {
+    await this.ws.call('interface.cancel_rollback')
```
RehanY147, line 299: All of these calls are missing error handlers

RehanY147 (approved):
Looks good

PR #7042, opened 2022-08-31
NAS-117464 / 22.12 / Fix `No active network interfaces` on network widget

denysbutenko:
For unknown reasons given network interface has `"link_state": "LINK_STATE_UNKNOWN"`

Before:
image
After:
image


undsoft (approved):
Code looks good, my only concern is that on the screenshot you've provided individual network widget says LINK STATE UP, which may not be true I suppose.

PR #7044, opened 2022-08-31
NAS-110795 / 22.12 / NAS-110795: Showing encryption checkbox in basic form

RehanY147:
To test
1. Create an encrypted pool
2. Try to create a dataset on that pool and disable encryption

Current Behavior: If you disable the `Inherit (encrypted)` field, the `Encryption` checkbox doesn't show up. So, you can only change how the dataset is encrypted. You can't disable the encryption altogether.

Expected Behavior: If you disable the `Inhert (encrypted)` field, the `Encryption` field should also show up with the rest of the fields. Disabling that field should help disable encryption on the newly created dataset. You should also see a warning when you disable encryption for the first time.


denysbutenko (approved):
image
image
image

Looks good!

PR #7047, opened 2022-09-02
NAS-117937 / 22.12 / NAS-117937: Refactor AlertConfigComponent to ix-forms

bvasilenko:


RehanY147 (changes requested):
The fields should have bit more margin from the card borders. Furthermore, the save and change category buttons are in the top bar. Our other forms don't work like that. We should put them inside the form. The category dropdown should be right next to the category name in the form. The save button should be underneath the form like in other ix-forms. The currently selected category should be a bit more obvious with maybe more font-size or a label like `Category: Applications` or something like that. With the drop down close to it.

bvasilenko:
@RehanY147 now all comments above should be addressed

denysbutenko:
@bvasilenko, wouldn't you mind renaming it to `AlertConfigFormComponent` so it would be easier to understand from the component name what it belongs to?

src/app/pages/system/alert-config/alert-config.component.ts
```diff
@@ -0,0 +1,122 @@
...
+  ]);
+
+  readonly policyOptions$ = this.ws.call('alert.list_policies').pipe(
+    map((policyList) => {
+      return policyList.map((policy) => ({ label: policy, value: policy }));
+    }),
+  );
+
+  get selectedCategory(): AlertCategory {
+    return this.categories.find((category) => this.selectedCategoryId === category.id);
```
RehanY147, line 47: UI fetches this value hundreds of times per second sometimes. Which means, we're running this check on this array thousands of times as you can see in this attachment. I add a console log `checking` inside the find and you can see how many times we're checking array items for no reason by the log counter. We should keep the selected category value in one variable and just return that variable to UI. Running a find method inside this getter is very inefficient. On checkbox value change, just update the variable value to the full category instead of just the `selectedCategoryId`.

https://user-images.githubusercontent.com/24204846/188987353-2b884d62-af3a-4479-bfe8-d06f5f75afea.mov

src/app/pages/system/alert-config/alert-config.component.html
```diff
@@ -0,0 +1,48 @@
+<mat-card class="form-card">
+  <mat-progress-bar *ngIf="isFormLoading" mode="indeterminate"></mat-progress-bar>
+  <div class="mat-title">
+    <div *ngIf="selectedCategory"> {{ 'Category' | translate }}: {{ selectedCategory.title }} </div>
```
RehanY147, line 4: Let's use a normal `ix-checkbox` with a label of `Category`. This doesn't look right. We can use a normal checkbox to present the currently selected category. I think that would look best.

RehanY147 (changes requested):

bvasilenko:
@denysbutenko 

>  wouldn't you mind renaming it to `AlertConfigFormComponent` so it would be easier to understand from the component name what it belongs to?

Sure. I've updated the name of the component

@RehanY147 

Now the bug is fixed, and the category selection is reworked

src/app/pages/system/alert-config/alert-config.component.html
RehanY147: I meant `ix-select` here. My bad.

RehanY147 (changes requested):
Code looks good. I meant something like this for the category dropdown.
image

denysbutenko:
> @denysbutenko
> 
> > wouldn't you mind renaming it to `AlertConfigFormComponent` so it would be easier to understand from the component name what it belongs to?
> 
> Sure. I've updated the name of the component

Thanks 👍🏻

bvasilenko:
@RehanY147 

> Code looks good. I meant something like this for the category dropdown.

Now the "Category" is a dropdown at the top of the form:

image

RehanY147 (approved):
Looks good

PR #7053, opened 2022-09-03
Added results dialog

RehanY147:
To test, create a snapshot (say 'testSnapshot') on dataset (say 'testDataset'). Now, run the following command on your shell `zfs hold keep testDataset@testsnapshot`. Now try to delete `testSnapshot`. Previously, UI would wrongly tell you that the operation was successful. Now, it should show you the errors it ran into. Also try deleting multiple snapshots together. The new dialog should show which snapshots faced errors when deleting and which snapshots were deleted successfully.

### DIFF
--- a/src/app/pages/storage/snapshots/snapshot-list/components/batch-delete-results/snapshots-batch-delete-results-dialog.component.html
+++ b/src/app/pages/storage/snapshots/snapshot-list/components/batch-delete-results/snapshots-batch-delete-results-dialog.component.html
@@ -1,0 +1,31 @@
+<h1 mat-dialog-title>Batch Operation Results</h1>
+<div mat-dialog-content>
+    <div *ngIf="errorsCount > 0">
+        <h4><strong>{{ errorsTitle  }}</strong></h4>
+        <ul>
+            <li *ngFor="let error of snapshotNamesWithErrors" class="item">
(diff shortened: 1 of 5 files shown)


src/app/pages/storage/snapshots/snapshot-list/components/batch-delete-results/snapshots-batch-delete-results-dialog.component.html
```diff
@@ -0,0 +1,31 @@
+<h1 mat-dialog-title>Batch Operation Results</h1>
+<div mat-dialog-content>
+    <div *ngIf="errorsCount > 0">
```
undsoft, line 3: Indentation doesn't look right

src/app/pages/storage/snapshots/snapshot-list/components/batch-delete-results/snapshots-batch-delete-results-dialog.component.html
```diff
@@ -0,0 +1,31 @@
+<h1 mat-dialog-title>Batch Operation Results</h1>
```
undsoft, line 1: Needs to be translated

undsoft (approved):
Works well.

PR #7054, opened 2022-09-03
NAS-118007 / 22.12 / NAS-118007: Removing BaseService

undsoft:
1. Removes resilvering indication on top bar.
2. Fixes temperatures in enclosure management on Minis.


dszidi (changes requested):
Confirmed with support, we really should keep the resilver prominent despite the fact that the task manager might also show resilver progress.

undsoft:
Brought resilvering indication back.

dszidi (changes requested):
Looks good. It does take a few seconds to get initial temperature readings. It would be good if that spinner stuck around until the temperatures are shown to the user. Then if there is a problem, put an error message in the empty space.

undsoft:
Changed the code a little bit to make first query immediately.

dszidi (approved):
LGTM

PR #7055, opened 2022-09-05
NAS-117953 / 22.12 / NAS-117953: Add 'serializeConfig' method

bvasilenko:
Fix for a bug when user clicks **Edit** and array fields are empty.

For testing, please see example in the ticket

### DIFF
--- a/src/app/services/app-schema.service.ts
+++ b/src/app/services/app-schema.service.ts
@@ -616,26 +616,44 @@ export class AppSchemaService {
     }) as HierarchicalObjectMap<ChartFormValue>[];
   }
 
-  serializeConfig(
-    config: { [key: string]: ChartFormValue } | ChartFormValue,
-    form: CustomUntypedFormGroup | FormGroup,
-  ): void {
+  /**
+   * Restores keys from form group
+   * @param config Object without keys. Example: { objectList: [{ nestedList: ['test4', 'test5'] }] }
+   * @param form Form group to restore keys from
+   * @returns Object with keys. Example: { objectList: [{ nestedList: ['test4', 'test5'] }] }
+   */
+  restoreKeysFromFormGroup(
+    config: HierarchicalObjectMap<ChartFormValue>,
+    form: FormGroup,
+  ): HierarchicalObjectMap<ChartFormValue> {
+    const newConfig = {} as HierarchicalObjectMap<ChartFormValue>;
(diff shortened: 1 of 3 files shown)


src/app/services/app-schema.service.ts
```diff
@@ -615,4 +615,27 @@ export class AppSchemaService {
...
+          valueConfig.forEach((valueItem, idxItem) => {
+            if (!_.isPlainObject(valueItem)) {
+              const keyItem = Object.keys(formConfig.value[idxItem])[0];
+              valueConfig[idxItem] = { [keyItem]: valueItem };
+            } else {
+              this.serializeConfig(valueItem, formConfig.controls[idxItem] as FormGroup);
+            }
+          });
+        } else if (_.isPlainObject(valueConfig)) {
+          this.serializeConfig(valueConfig, formConfig);
```
RehanY147, line 636: Not sure what this method is meant to do. Code needs to be more readable.

RehanY147 (changes requested):

bvasilenko:
@RehanY147 

> Shouldn't change the array while we're traversing it

Good one. Now it returns a modified value, and there's no mutable reference arguments

> Not sure what this method is meant to do. Code needs to be more readable.

Now I've renamed the method and added javadoc-style commentary

src/app/services/app-schema.service.ts
```diff
@@ -615,4 +615,45 @@ export class AppSchemaService {
       return this.serializeFormValue(listItem[Object.keys(listItem)[0]]);
     }) as HierarchicalObjectMap<ChartFormValue>[];
   }
+
+  /**
+   * Restores keys from form group
+   * @param config Object without keys. Example: { objectList: [{ nestedList: ['test4', 'test5'] }] }
+   * @param form Form group to restore keys from
```
RehanY147, line 622: form Form group

src/app/services/app-schema.service.ts
```diff
@@ -615,4 +615,45 @@ export class AppSchemaService {
...
+   * @returns Object with keys. Example: { objectList: [{ nestedList: ['test4', 'test5'] }] }
```
RehanY147, line 623: The example is the same as `@param config`

RehanY147 (approved):
Looks good. Nice!

PR #7057, opened 2022-09-05
NAS-118039: topbar cleanup 1

undsoft:
Testing: topbar, especially: user and power menus, jobs indicator, directory services indicator.


dszidi (approved):
Other than the import statement nitpick, the rest looks great. Topbar behaves as expected and no obvious regressions observed

image

PR #7058, opened 2022-09-05
NAS-117853 / 22.12 / NAS-117853: Not adding path when `create_zvol` flag is set to `true`

RehanY147:


bvasilenko (approved):
Works well for me

1. if `create_zvol: true`, then `path` not set
image

2. if `path` set, then `create_zvol` not set
image

undsoft:
Backport

PR #7070, opened 2022-09-10
NAS-111962 / 22.12 / NAS-111962: Not an interger error in Transfers field in Sync Cloud ta…

AlexKarpov98:
Not an interger error in Transfers field in Sync Cloud tax fix 
- form-input component updated

- uk.json was automatically added by scripts/validate_translations.js

Fixed:
Input event update for type=number, allowing any number: (1, -1, 1.5 and so on)
Setting input type=number to null if value cleared instead of an empty string
Validating input value with errors and clearing config error on value change (like updating element state)

Html:
More readable html


src/app/modules/entity/entity-form/components/form-input/form-input.component.ts
```diff
@@ -126,7 +141,25 @@ export class FormInputComponent implements Field {
...
-  onInput(): void {
+  onInput(event: Event): void {
+    const inputValue = (event as MessageEvent).data;
+
+    if (this.config.inputType === 'number') {
+      if (this.config.errors) {
+        this.config.errors = null;
+      }
+
+      if (!this.shouldShowResetInput() && !inputValue !== null) {
```
AlexKarpov98, line 152: shouldShowResetInput() shows if input is valid, so we can show reset button-icon

undsoft (approved):
No duplicate error is shown now and I can save the form.

PR #7074, opened 2022-09-12
NAS-117298 / None / NAS-117298: timemaching_quota added in smb-form

AlexKarpov98:


src/app/pages/sharing/smb/smb-form/smb-form.component.html
```diff
@@ -137,18 +143,28 @@
         ></ix-textarea>
       </ix-fieldset>
       <div class="form-actions">
+        {{ form.invalid ? 'inval' : 'val' }}
+        {{ isLoading ? 'loading' : 'noload' }}
```
undsoft, line 147: 😞

src/app/pages/sharing/smb/smb-form/smb-form.component.html
```diff
@@ -107,7 +111,9 @@
         ></ix-checkbox>
         <ix-checkbox
           formControlName="aapl_name_mangling"
-          [label]="helptextSharingSmb.placeholder_aapl_name_mangling | translate"
+          [label]="
+            helptextSharingSmb.placeholder_aapl_name_mangling | translate
```
undsoft, line 115: Weird layout.

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -321,53 +382,70 @@ export class SmbFormComponent implements OnInit {
...
+            (redirect) => {
+              this.isLoading = false;
+              this.cdr.markForCheck();
+              if (redirect) {
+                const sharePath: string = this.form.get('path').value;
+                const homeShare = this.form.get('home').value;
+                const datasetId = sharePath.replace('/mnt/', '');
+                const poolName = datasetId.split('/')[0];
+                this.router.navigate(
+                  ['/'].concat([
```
undsoft, line 414: I think previous layout was easier to read.

undsoft (changes requested):
According to ticket value for `timemachine_quota` is number of bytes, so it needs to use [formatter] to show it properly.

AlexKarpov98:
@undsoft - Updated PR

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -107,6 +139,7 @@ export class SmbFormComponent implements OnInit {
     hostsdeny: [[] as string[]],
     home: [false],
     timemachine: [false],
+    timemachine_quota: [],
```
undsoft, line 142: This probably results in type being not defined for this field. You should probably have something like:
```
timemachine_quota: [null as number]
```

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -300,24 +364,33 @@ export class SmbFormComponent implements OnInit {
...
+        }
+      });
   }
 
   submit(): void {
     this.isLoading = true;
     this.cdr.markForCheck();
     const smbShare = this.form.value;
+
+    if (!this.timemachineQuota) {
```
undsoft, line 388: Or just `smbShare.timemachine_quota` and getter can be removed.

undsoft (changes requested):
1. Looks like not sending a value doesn't remove it.
2. I think we should also not send it (clear value) if user unticks Time Machine.

AlexKarpov98:
@undsoft  - fixed

undsoft (changes requested):
No changes since my last review.

AlexKarpov98:
@undsoft Please check it now :)

undsoft (changes requested):
Same issues as during last review:
1. Once I set Time Machine Quota there is no way for me to remove it.
2. If I set Time Machine Quota, but then untick `Time Machine` checkbox, my expectation is that this removes quota when saved.

AlexKarpov98:
@undsoft 
Screen Shot 2022-09-19 at 23 38 59
We need to set 0 manually once the value is empty (updated PR)

undsoft (approved):
Works correctly now.

PR #7078, opened 2022-09-13
NAS-117805 / 22.12 / Adding warnings for unused disks with exported pools

RehanY147:
This PR aims to add warnings for wherever UI is dealing with unused disks that are currently also part of exported pools. Test the following places in the UI. 

- When creating a new pool or adding a disk to an existing pool, the ManagerComponent should show the following warnings for such disks
Name of the exported pool that's attached to the disk mentioned in the table like so
image
Also, when you click the checkbox to select this disk, you should see the warning like this
image
Select/unselect a bunch of disks to check that the warning is only shown for the correct disks
- On the new Storage Dashboard, when you click `Add to Pool` button on the unused disks card, the model opens up. On the model, disks that are attached to exported pools should be separate from other disks of similar size and properties with a warning underneath as in the screenshot.
image
- On the `Extend` and `Replace` buttons for `ZFS Info` cards on the new `Devices` page and `Replace` and `Attach` operations on the `Boot Pool Status` page, you see a modal where you can select a disk from a drop down. Disks with exported pools should have the exported pools name mentioned in the select. 
image
Also, if you select such a disk/vdev from the drop down, you should see a warning.
image
- On the `Disks` page, you should see the name of the exported pool in the `Pool` column of such disks with the text ` '(Exported)'` appended.
image
if you select to `Wipe` such a disk, you should see a confirmation dialog.
image


src/app/pages/storage/modules/disks/components/replace-disk-dialog/replace-disk-dialog.component.ts
```diff
@@ -21,15 +24,20 @@ export interface ReplaceDiskDialogData {
...
   form = this.formBuilder.group({
     replacement: ['', Validators.required],
     force: [false],
   });
 
+  unusedDisks: UnusedDisk[] = [];
+
   unusedDisks$ = this.ws.call('disk.get_unused').pipe(
+    tap((unusedDisks) => {
+      this.unusedDisks = unusedDisks;
```
undsoft, line 37: This is an unexpected side-effect - unusedDisks$ now has two responsibilities. 
You should remove both of these and a new loadUnusedDisks method in ngOnInit.

src/app/pages/storage/modules/disks/components/replace-disk-dialog/replace-disk-dialog.component.ts
```diff
@@ -44,8 +52,41 @@ export class ReplaceDiskDialogComponent {
...
+    this.setupExportedPoolWarningForUnusedDisks();
+  }
+
+  setupExportedPoolWarningForUnusedDisks(): void {
+    this.form.get('replacement').valueChanges.pipe(untilDestroyed(this)).subscribe(
+      this.warnAboutExportedPoolForUnusedDiskIfNeeded,
+    );
+  }
+
+  warnAboutExportedPoolForUnusedDiskIfNeeded = (diskIdentifier: string): void => {
```
undsoft, line 68: Same notes about extra functions, long names and template strings.

src/app/pages/system/bootenv/boot-pool-attach/boot-pool-attach-form.component.ts
```diff
@@ -19,22 +20,30 @@ import { DialogService, WebSocketService } from 'app/services';
...
   isFormLoading = false;
 
   form = this.fb.group({
     dev: ['', Validators.required],
     expand: [false],
   });
 
+  unusedDisks: UnusedDisk[] = [];
+
   dev = {
```
undsoft, line 33: Same.

undsoft (changes requested):
In manager selecting disk always shows warning despite the code that should prevent it. On the other hand selecting multiple disks via select all checkbox doesn't show a warning.

RehanY147:
The check to not show warning is there to avoid showing warning for a disk that was already selected. If the disk is unchecked and then checked again, showing the warning again.

undsoft (changes requested):
Replace disk dialog seems to be broken.

bvasilenko (approved):
Works good for me!

Just one minor observation: it's still possible to bypass the warning if user clicks **Suggest Layout** button:


https://user-images.githubusercontent.com/20611516/190899495-0a469d06-5a22-4511-9b6b-68bd80261fda.mp4

undsoft (approved):
👍🏻

PR #7080, opened 2022-09-13
NAS-118151 / 22.12 / NAS-118151: Hide Aliases section if DHCP/Autoconfiguration radio boxe…

undsoft:
…s are checked

Testing: as per ticket description.


AlexKarpov98 (approved):
Nice

PR #7081, opened 2022-09-14
NAS-110516 / 22.12 / NAS-110516: Add 'retention' to the details-row at snaphots page

bvasilenko:
**Testing**

On **Snapshots** table, new column "Retention" should be displayed at expanded details panel


src/app/interfaces/zfs-snapshot.interface.ts
```diff
@@ -10,7 +10,11 @@ export interface ZfsSnapshot {
     [property: string]: ZfsProperty<string | number | boolean | ApiTimestamp>;
     creation: ZfsProperty<ApiTimestamp>;
   };
-  retention: unknown;
+  retention?: {
+    datetime: ApiTimestamp;
+    source: string;
```
undsoft, line 15: Enum.

undsoft (changes requested):
Looks good, but I think we can improve messaging here. Seems there different states for this field and we should show this to the user:
* Will be automatically destroyed at datetime by periodic snapshot task
* Is planned to be automatically destroyed at datetime.
* "Will not be destroyed automatically" for null values.

bvasilenko:
@undsoft Now it has enum and improved wordings

undsoft (approved):
Looks good

PR #7082, opened 2022-09-14
NAS-117857 / 22.12 / Fix webui shell handling of long strings

Raz-Dva:
Check the rendering shell. (prompt: use the "mc" command)
When the menu is collapsed.
On all available breakpoints
Pay attention:
-  the line should not be wraps too early
- the line must not goes out of the border of the parent block
-  there should not be empty rows between lines

Check for a long string:

- Type long string, e.g. echo "some long string here"
- Press Enter.
- Press Up.
- Keep pressing Left to go back to the beginning of the string.


src/app/modules/terminal/components/terminal/terminal.component.ts
```diff
@@ -65,6 +73,26 @@ export class TerminalComponent implements OnInit, AfterViewInit, OnDestroy {
...
       });
     }
+
+    this.store$.pipe(
+      waitForPreferences,
+      filter((preferences) => Boolean(preferences.sidenavStatus)),
+      untilDestroyed(this),
+    ).subscribe(() => {
+      if (this.shellConnected) {
+        // 300ms waiting for sidenav width to be set
```
RehanY147, line 83: Better to store `300` in a properly named constant in the class. The variable name should be explanation of the value instead of a comment.

src/app/modules/terminal/components/terminal/terminal.component.ts
```diff
@@ -65,6 +73,26 @@ export class TerminalComponent implements OnInit, AfterViewInit, OnDestroy {
...
+        setTimeout(() => { this.resizeTerm(); }, 300);
+      }
+    });
+
+    this.core.register({
+      observerClass: this,
+      eventName: 'MediaChange',
+    }).pipe(untilDestroyed(this)).subscribe(() => {
+      if (this.shellConnected) {
+        setTimeout(() => { this.resizeTerm(); }, 300);
```
RehanY147, line 93: Same here

RehanY147 (approved):
Looks good

undsoft (approved):
Couldn't reproduce the issue as well.

src/app/modules/terminal/components/terminal/terminal.component.ts
Raz-Dva: ok :thumbsup:

PR #7091, opened 2022-09-16
NAS-117474 / 22.12 / Datasets table header sticky

AlexKarpov98:


AlexKarpov98:
@denysbutenko @undsoft - please have a look on these updates:

https://user-images.githubusercontent.com/22980553/190631060-f935beac-ecb2-4f48-8922-ae6c0137a6aa.mov

undsoft (changes requested):
Nice.
The only problem is that on some resolutions, when table has horizontal scrolling, header will be misaligned with columns.
Is this something we can fix?
Снимок экрана 2022-09-16 в 15 00 44

denysbutenko (changes requested):
the same issue

AlexKarpov98:
Could not find solution as well

denysbutenko (changes requested):
https://user-images.githubusercontent.com/351613/191278269-fd487226-24a3-433b-92fe-21742a741c6b.mp4

The dataset tree looks good now. Observed regression for dataset details panel. Also, the tree and details panel has different gaps at the bottom. It needs to be synced and probably reduced.

image

src/app/pages/datasets/components/dataset-management/dataset-management.component.scss
```diff
@@ -318,3 +311,8 @@ $padding-left: 8px;
     }
   }
 }
+
+ix-tree {
+  max-height: calc(100vh - 330px);
```
denysbutenko, line 316: What stands for 330px / How is it gained?

AlexKarpov98: I just calculated it visually, I think it will be good to refactor datasets blocks, to have same height via flex somehow.

AlexKarpov98: > What stands for 330px / How is it gained?
I just calculated it visually, I think it will be good to refactor datasets blocks, to have same height via flex somehow.

AlexKarpov98: if we like that approach I found, I'll work on refactoring

undsoft (changes requested):
Previous option was better. We wanted to keep page scrolling and not just have a smaller area to scroll.

AlexKarpov98:
@undsoft  - reverted to previous solution

AlexKarpov98:
@undsoft @denysbutenko 
It's done! 🥳 😎 ✅ 

https://user-images.githubusercontent.com/22980553/191559902-c13a45d6-17ad-4f12-be49-01e34514354b.mov

src/app/pages/datasets/components/dataset-management/dataset-management.component.ts
```diff
@@ -118,34 +122,89 @@ export class DatasetsManagementComponent implements OnInit, AfterViewInit {
...
-  onSearch(query: string): void {
-    this.dataSource.filter(query);
+  ngOnDestroy(): void {
+    this.subscription.unsubscribe();
   }
 
-  private setupTree(): void {
-    this.datasetStore.datasets$
-      .pipe(untilDestroyed(this))
+  listenForSystemdatasetConfig(): void {
```
undsoft, line 129: this is not really listen. More like `loadSystemDataset`

src/assets/styles/mixins/cards.scss
```diff
@@ -37,8 +37,6 @@
       flex-wrap: wrap;
       gap: $gap;
       height: 100%;
```
undsoft, line 39: Why were these numbers removed? This breaks card layout in details.

undsoft (changes requested):
Scrolling works perfectly now, nice work. 👍🏻 

Need to revert changes to card widths to fix them.

src/assets/styles/mixins/cards.scss
AlexKarpov98: returned

src/app/pages/datasets/components/dataset-management/dataset-management.component.ts
AlexKarpov98: should agree

undsoft (approved):
Weird enough both times when I was testing this PR, I had to remove `node_modules` and reinstall packages. Just doing `yarn` would result in compilation error, not sure why.

May be worth doing the same and committing changed `yarn.lock`.

But generally works well.

PR #7092, opened 2022-09-16
NAS-117986 / 22.12 / NAS-117986: fix DOM not updated properly in Safari for gauge chart

bvasilenko:
**Testing**

Check **CPU Widget**  in Safari. See ticket for details


undsoft (approved):
Double numbers are gone. 👍🏻

PR #7095, opened 2022-09-17
NAS-118169 / 22.12 / NAS-118169: Fixes SSH connection field when editing an rsync task form

undsoft:
Testing: create and edit rsync task in SSH mode and experiment with `Connect using` dropdown.


dszidi (commented):
Not sure what the problem is but I can't seem to get the authentication to happen during the "New SSH Connection" setup step. I created a new user on the target NAS and triple checked the password but it fails to generate a new key

undsoft:
@dszidi That's likely is not because of these changes. You can create credentials in some other way and create a ticket for the problem you're facing.

PR #7097, opened 2022-09-18
NAS-117481 / 22.12 / NAS-117481: Extracting some components from signin form

undsoft:
Testing: test affected components.
Truecommand status is easiest to test by mocking.


src/app/views/sessions/signin/true-command-status/true-command-status.component.html
```diff
@@ -0,0 +1,8 @@
+<div *ngIf="connectionState?.connected" class="truecommand" (click)="goToTrueCommand()">
+  <img
+    class="truecommand-logo"
+    src="../../../../../assets/images/truecommand/truecommand-logo-mark-full-color-rgb.svg"
+    alt="TrueCommand"
+  >
```
AlexKarpov98, line 6: let's self close the tag

undsoft: Closing tags like this was necessary in xhtml days and not necessary for html5.

AlexKarpov98: well, I'm old fashioned man then :D

AlexKarpov98 (approved):

PR #7098, opened 2022-09-18
NAS-113216 / 22.12 / NAS-113216: Hide 'acltype' and 'aclmode' when creating or editing first level dataset

bvasilenko:
**Summary**

Removed **ACL Type** and **Mode**,  if dataset is a **_root dataset_** , as suggested by the ticket's description

### DIFF
--- a/src/app/pages/datasets/components/dataset-form/dataset-form.component.ts
+++ b/src/app/pages/datasets/components/dataset-form/dataset-form.component.ts
@@ -624,7 +624,9 @@ export class DatasetFormComponent implements FormConfiguration {
             const value = event.event.value;
             if (value === DatasetAclType.Nfsv4) {
               aclModeFormControl.setValue(AclMode.Passthrough);
-              this.entityForm.setDisabled('aclmode', false, false);
+              if (!this.isRootDataset(this.parent)) {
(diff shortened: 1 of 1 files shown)


src/app/pages/datasets/components/dataset-form/dataset-form.component.ts
```diff
@@ -1452,6 +1454,10 @@ export class DatasetFormComponent implements FormConfiguration {
     return false;
   }
 
+  isRootDataset(path: string): boolean {
```
undsoft, line 1457: we have `isRootDataset` in `dataset.utils.ts` that you can use here.

undsoft (changes requested):
I think you misunderstood the task.
Root dataset is the first dataset in the hierarchy, it's the one that is created when you add a pool and can only be edited. It is this dataset that shouldn't have Acl Type and Acl Mode fields.
Right now this is not the case. I can still edit these fields on the root dataset, but cannot edit them on first child.

bvasilenko:
@undsoft Done! Now it uses `isRootDataset` in `dataset.utils.ts`

undsoft (approved):
Works well 👍🏻

PR #7099, opened 2022-09-19
NAS-117942 / 22.12 / NAS-117942: Cleaning up Storage module

undsoft:
Testing: very shallow regression testing in Storage should be enough.


RehanY147 (approved):
Looks good

PR #7101, opened 2022-09-20
NAS-118065 / 22.12 / NAS-118065: Add Extend button inside ZFS Info block

bvasilenko:
**Testing**

You should have a single disk pool, and an available disk

1. On **Storage > (Choose a single-disk pool) > Manage Devices** , select the single disk of the pool
2. Test the **Extend** button and the **Extend Dialog**


denysbutenko (approved):
Works well

PR #7108, opened 2022-09-21
updated conditional if statement for array

myoneflag:


undsoft (changes requested):
Hi,
Thank you for your contribution, but it won't be accepted in its current form.
Having "> 0" or direct comparisons is generally a matter of style and we use Airbnb Style Guide as base for our code style.
They [recommend to avoid](https://github.com/airbnb/javascript#comparison--shortcuts) shortcut comparison for numeric values.

PR #7109, opened 2022-09-21
NAS-118267 / 22.12 / Disallow mat-icon selector in styles

denysbutenko:
Reduce MatIcon usage in stylesheets.
These changes make styles compatible with `mat-icon` and `ix-icon` components. So migration from `<mat-icon>` to `<ix-icon>` be easier, and you don't need to worry about updating selectors anymore.


RehanY147 (approved):
Looks good. Icons seem to be working.

PR #7113, opened 2022-09-21
NAS-117958 / 22.12 / Nas 117958 - Fix chart colors on different themes

Raz-Dva:
Check Dataset Space Management (card). The colour swatches must match the colours on the chart
Check gauge chart on Storage page. 
Check Chart Colours on all themes.


bvasilenko (approved):
Looks good to me

PR #7114, opened 2022-09-21
NAS-118273 / 22.12 / NAS-118273: Extract some dialogs into separate components

undsoft:
Test affected components:
1. Stripping ACL (go to Datasets, Permission, Edit, Set ACL, Apply preset).
2. Rollback app (upgrade app to a newer version to get rollback option available).


src/app/pages/applications/chart-releases/chart-releases.component.ts
```diff
@@ -376,39 +359,20 @@ export class ChartReleasesComponent implements AfterViewInit, OnInit, OnDestroy
...
-    });
+  onRollback(name: string): void {
+    const chartRelease = this.chartItems[name];
+    this.mdDialog.open(ChartRollbackModalComponent, {
+      data: chartRelease,
+    })
+      .afterClosed()
+      .pipe(untilDestroyed(this))
+      .subscribe((wasRolledBack) => {
+        if (!wasRolledBack) {
```
AlexKarpov98, line 370: we can use RxJs `filter((wasRolledBack) => !!wasRolledBack)` 
and in .subscribe just use 
`this.refreshChartReleases();` 
without any checks

AlexKarpov98 (changes requested):

src/app/pages/applications/chart-releases/chart-releases.component.ts
undsoft: Yeah, I was going to and used `filter(Boolean)` in the past, but then I thought have it may be better to be more explicit.
With `filter(Boolean)` it's easy to miss it and not understand what's going on, with explicit variable it's clear what's happening.

AlexKarpov98 (approved):
looks good :)

PR #7117, opened 2022-09-21
NAS-118270 / 22.12 / NAS-118270: Seek and destroy

undsoft:
Removes storage module and some other unused code.


RehanY147 (changes requested):
Why are we removing the json translation files?

undsoft:
@RehanY147 We aren't, only some strings were removed.

RehanY147 (approved):
Looks good

PR #7118, opened 2022-09-22
NAS-117903 / 22.12 / NAS-117903: Set correct key sequence

bvasilenko:
**Testing**

On the **Reporting page**, click the **Zoom out** button for the CPU usage graph. Check that chart's labels and values aren't shuffled.

### DIFF
--- a/src/app/pages/reports-dashboard/components/report/report.component.ts
+++ b/src/app/pages/reports-dashboard/components/report/report.component.ts
@@ -6,6 +6,7 @@ import {
   OnDestroy,
   OnChanges,
   SimpleChanges,
+  OnInit,
 } from '@angular/core';
 import { NgForm } from '@angular/forms';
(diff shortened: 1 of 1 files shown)


undsoft (approved):
It works. 👍🏻

src/app/pages/reports-dashboard/components/report/report.component.ts
```diff
@@ -222,6 +223,13 @@ export class ReportComponent extends WidgetComponent implements AfterViewInit, O
     });
   }
 
+  async ngOnInit(): Promise<void> {
+    const zoom = this.zoomLevels[this.timeZoomIndex];
+    const rrdOptions = await this.convertTimespan(zoom.timespan);
```
undsoft, line 228: I'm surprised await works here. I would imagine that Angular doesn't wait for promises in ngOnInit to resolve.

PR #7120, opened 2022-09-22
NAS-117846 / 22.12 / Improving how we collect data for disks on `Pools Dashboard`

RehanY147:
In this PR, I have pulled out all of the middleware calls made to collect data for Pool Dashboard cards to one place. In the Pools Dashboard store. Previously, we were making new calls to fetch data for disks for every card individually. Which meant, the more pools, the more parallel calls. Which would lead to `Max concurrent calls` errors from middleware. Now regardless of the number of pools, we only make 3-4 calls. We collect data for all disks and than format it to individual pools locally. Which means we can avoid the concurrent calls errors.


src/app/interfaces/alert.interface.ts
```diff
@@ -3,7 +3,7 @@ import { AlertPolicy } from 'app/enums/alert-policy.enum';
 import { ApiTimestamp } from 'app/interfaces/api-date.interface';
 
 export interface Alert {
-  args: string;
+  args: string | { device: string; message: string };
```
undsoft, line 6: This is too specific as arguments can be all sort of things. Can we do `unknown` and type where it's being used?

PR #7124, opened 2022-09-23
NAS-118303 / 22.12 / New reasons in `failover.disabled.reasons`

yocalebo:
Added 3 new reasons in PR: https://github.com/truenas/middleware/pull/9904 so adding them here.


AlexKarpov98 (changes requested):
we also need to add new translations into all .json files

undsoft (approved):
That's fine, translations will get picked up on next PR.

@yocalebo 
If you do `yarn` next time, it should install git hooks that generate translation files when you commit.

yocalebo:
> That's fine, translations will get picked up on next PR.
> 
> @yocalebo If you do `yarn` next time, it should install git hooks that generate translation files when you commit.

great to know, thanks for the tip!

PR #7128, opened 2022-09-26
NAS-118234 / 22.12 / Nas 118234 - Prevented dialog modal from closing when manual update file is upload

Raz-Dva:
Disabled MatDialog closing with download process to prevent update file download from being cancelled.
Check the update process using the "Install Manual Update File".


undsoft (approved):
Dialog cannot be closed now.

PR #7129, opened 2022-09-26
NAS-118282: Search input fixes for datasets and generally

AlexKarpov98:
Fixed:
- Disabled state for dataset input (if no initial data is present)
- Handling initial data correctly (with action if not disabled and has initial value)
- Added debounceTime logic on `(input)` to improve performance
- Fixed dataset search logic to return only needed data
- Small refactoring

### DIFF
--- a/src/app/modules/common/search-input/search-input.component.ts
+++ b/src/app/modules/common/search-input/search-input.component.ts
@@ -7,34 +7,43 @@ import {
   Output,
   ViewChild,
   OnInit,
+  OnChanges,
+  SimpleChanges,
 } from '@angular/core';
+import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
+import { debounceTime, distinctUntilChanged, Subject } from 'rxjs';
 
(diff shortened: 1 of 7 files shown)


src/app/modules/common/search-input/search-input.component.ts
```diff
@@ -7,34 +7,43 @@ import {
...
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class SearchInputComponent implements OnInit {
+export class SearchInputComponent implements OnInit, OnChanges {
   @Input() disabled = false;
   @Input() value = '';
   @Output() search = new EventEmitter<string>();
 
-  @ViewChild('searchInput') input: ElementRef<HTMLInputElement>;
+  @ViewChild('ixSearchInput') input: ElementRef<HTMLInputElement>;
```
undsoft, line 28: `ix` prefix doesn't add value

undsoft (approved):
Search works well now 👍🏻

PR #7132, opened 2022-09-28
NAS-116558 / 22.12 / `special_small_block_size` dataset option is inheritable

undsoft:
Testing: Test `Metadata (Special) Small Block Size` in dataset add/edit form, especially `Inherit` option.


RehanY147 (approved):
We can remove the `blurSpecialSmallBlocks` method. The rest looks good.

bvasilenko (approved):
Works good to me. Tested "inherited" with N-levels of datasets

PR #7134, opened 2022-09-28
NAS-118334 / 22.12 / ScreenType enum added and used

AlexKarpov98:


AlexKarpov98:
> Looks good. We could just name the enum `ScreenType`. Just a suggestion.

yes, agree, since it's CamelCase name, we can use it (I just thought to make a bit different name cuz in some .ts files we have screenType variable but ScreenType will be uppercased as readonly enum value!

PR #7135, opened 2022-09-28
NAS-118278 / 22.12 / correctly name for SSH Public Key in user-form

Raz-Dva:
Check Credentials → Local Users → <user> → Edit → Authentication


RehanY147 (approved):
LGTM

PR #7136, opened 2022-09-28
NAS-118339 / 22.12 / Add links to snapshots

bvasilenko:
**Testing**

1. On **Data Protection** page at the **Period Snapshot Tasks** block  **Snapshots** button should be added to the footer
2. **Snapshots** button should also present on enlarged version of a page (when user clicks the heading of **Period Snapshot Tasks** block)


src/app/pages/data-protection/snapshot/snapshot-list/snapshot-list.component.ts
```diff
@@ -63,13 +63,22 @@ export class SnapshotListComponent implements EntityTableConfig<PeriodicSnapshot
     },
   };
 
+  customActions = [{
```
undsoft, line 66: Let's make this button non-primary.

undsoft (approved):
Looks good, but I think button in `Periodic Snapshot Tasks` should have more vertical margins.

PR #7139, opened 2022-09-29
NAS-118335: Improved spinners & some UI elements

AlexKarpov98:
Made spinners smaller (giant ones)
Improved radio group & tooltip icon positioning (centered)
Fixed slide toggle (bar color by default, not primary color, so it is a bit lighter that then circle)


denysbutenko (changes requested):
image

mat-slide-toggle looks worse than before

denysbutenko:
Seems like because of that
image

denysbutenko (approved):
Looks good

PR #7144, opened 2022-09-29
NAS-118352 / 22.12 / Fixed train

RehanY147:


src/app/pages/system/update/update.component.ts
```diff
@@ -114,9 +114,8 @@ export class UpdateComponent implements OnInit {
           this.checkable = true;
           this.fullTrainList = trains.trains;
 
-          // On page load, make sure we are working with train of the current OS
-          this.train = trains.current;
-          this.selectedTrain = trains.current;
+          this.train = trains.selected;
+          this.selectedTrain = trains.selected;
```
AlexKarpov98, line 118: @RehanY147 why do we need 2 same values in 2 different variables?

RehanY147: One is the value of the `ix-select` and one is to remember what the original selected train was so we don't show a warning when they switch back to that.

denysbutenko (approved):
image

select border color is wrong otherwise looks good

RehanY147:
NAS-118390 will handle the layout issues

PR #7150, opened 2022-09-30
NAS-118375 / 22.12 / Close mobile details after delete dataset

bvasilenko:


undsoft (changes requested):
Can we do this via store instead? Like have something that clears `selectedDatasetId`. 
Feels like there are now too many ways for components to communicate with each other.

bvasilenko:
@undsoft  Now `showMobileDetails` is in the store

undsoft (changes requested):
This feels kinda weird. Some presentational logic is in the component (media observers) and some is in the store now.
What if instead of having a separate property in the state, you do redirect and every time there is a redirect you close mobile details?

Also, would be nice to select parent dataset when child is deleted, not reset user to the root of the tree.

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.ts
```diff
@@ -64,6 +66,8 @@ export class DatasetDetailsCardComponent {
       .afterClosed()
       .pipe(filter(Boolean), untilDestroyed(this))
       .subscribe(() => {
+        this.router.navigate(['/datasets']);
```
AlexKarpov98, line 69: Let's navigate at the end of the .subscribe

src/app/pages/datasets/components/dataset-details-panel/dataset-details-panel.component.ts
```diff
@@ -1,5 +1,5 @@
 import {
-  ChangeDetectionStrategy, Component, Input, OnInit, EventEmitter, Output,
+  ChangeDetectionStrategy, Component, Input, OnInit,
```
AlexKarpov98, line 2: one line should be enough I think

src/app/pages/datasets/components/dataset-management/dataset-management.component.html
```diff
@@ -119,14 +119,13 @@
 
     <div
       class="details-container"
-      [class.details-container-mobile]="showMobileDetails"
+      [class.details-container-mobile]="showMobileDetails$ | async"
```
AlexKarpov98, line 122: please follow the ordering rules 
**ixDetailsHeight** should be higher than **[class.details-container-mobile]**

AlexKarpov98 (changes requested):

bvasilenko:
@AlexKarpov98 minor improvements done to the code.
@undsoft looks like I was going wrong direction. I've reverted and pushed another solution based on the router - now it looks much better to me.

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.ts
```diff
@@ -65,6 +67,10 @@ export class DatasetDetailsCardComponent {
       .pipe(filter(Boolean), untilDestroyed(this))
       .subscribe(() => {
         this.datasetStore.datasetUpdated();
+        this.datasetStore.selectedParentDataset$.pipe(first(), untilDestroyed(this)).subscribe((parent) => {
+          this.datasetStore.selectDatasetById(null);
```
undsoft, line 71: `selectDatasetById` doesn't seem to be necessary

undsoft (approved):
Works okay.

PR #7151, opened 2022-09-30
NAS-118325 / 22.12 / Add USB passthrough support in the UI

bvasilenko:
**Testing**

Use mockup JSON response for `vm.device.usb_passthrough_choices`

```
"result": {
  "usb_0_0": {
    "capability": {
      'product': "",
      'vendor': "",
      'product_id': "",
      'vendor_id': "",
      'bus': "",
      'device': "",
    }
    ,
    "available": true,
    "error": null,
    "": "",
  },
  "usb_0_1": {
    "capability": {
      'product': "",
      'vendor': "",
      'product_id': "",
      'vendor_id': "",
      'bus': "",
      'device': "",
    }
    ,
    "available": true,
    "error": null,
    "": "",
  },
}

```


undsoft (changes requested):
Cannot test at the moment due to middleware issue, but just looking at the code I think we should show something else in the dropdown to make it easier for user identify what USB device is going to be selected.
Something like product type and manufacturer, depending on the value.

bvasilenko:
@undsoft  Now dropdown labels display product and manufacturer, depending on the value

undsoft (approved):
Looks good.
Снимок экрана 2022-10-05 в 12 03 20

T3chArmy:
Thank you so much for adding this! My only suggestion from a user perspective, and just me being picky. Would be to add a colon between the device identifier(usb_1_7), and the device name/manufacturer.

For example: above screenshot would be "usb_1_7: Hub (ATEN International Co., Ltd)"

ServusNL:
Is the drive still mapped as: "device": "usb_3_1"? Resulting in the unusable situation that the usb_3_1 relations changes on every reboot.

undsoft:
Please direct additional issues to Jira.

PR #7152, opened 2022-09-30
NAS-117958 / 22.12 / Nas 117958 Fixed chart colors on different themes

Raz-Dva:
Check chart colors on different themes


Raz-Dva:
Created new issue  [NAS-118411](https://ixsystems.atlassian.net/browse/NAS-118411)

PR #7154, opened 2022-10-03
NAS-118327 / 22.12 / Restore Angular loading indication

denysbutenko:
Kapture 2022-10-03 at 13 23 211


undsoft (approved):
Nice, though maybe make flashing a little bit slower?

denysbutenko:
> Nice, though maybe make flashing a little bit slower?

Ok updated up to 2 seconds to be consistent with other similar animations

PR #7155, opened 2022-10-03
NAS-118411 / 22.12 / Added style colour for chart swatch

Raz-Dva:
Check Dataset Space Management (card). The colour swatches must match the colours on the chart


bvasilenko (approved):
LGTM

PR #7156, opened 2022-10-03
NAS-118101 / 22.12 / Minor code improvements for Datasets module

RehanY147:
Functions reordering and separation of responsibilities. Test for regressions.


src/app/pages/datasets/components/dataset-capacity-management-card/dataset-capacity-settings/dataset-capacity-settings.component.ts
```diff
@@ -24,29 +23,29 @@ const oneGb = 1024 * 1024 * 1024;
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class DatasetCapacitySettingsComponent {
-  readonly defaultWarning = 80;
-  readonly defaultCritical = 95;
+  readonly defaultReqQuotaWarning = 80;
```
undsoft, line 26: What's `reqquota`?

src/app/pages/datasets/components/dataset-quotas/dataset-quota-edit-form/dataset-quota-edit-form.component.ts
```diff
@@ -38,25 +40,43 @@ export class DatasetQuotaEditFormComponent {
...
-      : this.translate.instant(helptext.groups.data_quota.tooltip)
-        + this.translate.instant(globalHelptext.human_readable.suggestion_tooltip)
-        + this.translate.instant(' bytes.');
+      ? this.getUserDataQuotaTooltip()
+      : this.getGroupDataQuotaTooltip();
+  }
+
+  getUserDataQuotaTooltip(): string {
+    return this.translate.instant(helptext.users.data_quota.tooltip)
+    + this.translate.instant(globalHelptext.human_readable.suggestion_tooltip)
```
undsoft, line 70: Needs indentation.

src/app/pages/datasets/components/dataset-quotas/dataset-quota-edit-form/dataset-quota-edit-form.component.ts
```diff
@@ -38,25 +40,43 @@ export class DatasetQuotaEditFormComponent {
...
+      ? this.getUserDataQuotaLabel()
+      : this.getGroupDataQuotaLabel();
   }
+
+  getUserDataQuotaLabel(): string {
+    return this.translate.instant(helptext.users.data_quota.placeholder)
+    + this.translate.instant(globalHelptext.human_readable.suggestion_label);
+  }
+
+  getGroupDataQuotaLabel(): string {
```
undsoft, line 52: New methods are private.

src/app/pages/datasets/components/dataset-quotas/dataset-quota-edit-form/dataset-quota-edit-form.component.ts
```diff
@@ -122,19 +150,16 @@ export class DatasetQuotaEditFormComponent {
...
-        : of(true)
-    ).pipe(filter(Boolean), untilDestroyed(this)).subscribe(() => {
+    this.submit(values, payload);
+  }
+
+  submit(values: typeof this.form.value, payload: SetDatasetQuota[]): void {
+    let canSubmit$ = of(true);
+    if (this.isUnsettingQuota(values)) {
+      canSubmit$ = this.getConfirmation(values.name);
+    }
```
undsoft, line 160: Nice.

src/app/pages/datasets/components/dataset-quotas/dataset-quotas-grouplist/dataset-quotas-grouplist.component.ts
```diff
@@ -267,6 +263,19 @@ export class DatasetQuotasGrouplistComponent implements OnInit, AfterViewInit, O
     });
   }
 
+  getDeleteConfirmation(quota: DatasetQuota): Observable<boolean> {
```
undsoft, line 266: or `confirmDeletion`

src/app/pages/datasets/components/delete-dataset-dialog/delete-dataset-dialog.component.ts
```diff
@@ -56,53 +58,57 @@ export class DeleteDatasetDialogComponent implements OnInit {
...
+      tap(() => {
+        this.loader.close();
+        this.dialogRef.close(true);
+      }),
+      catchError(this.handleDeleteError.bind(this)),
+      untilDestroyed(this),
+    ).subscribe();
+  }
+
+  tryDeletingDataset(): Observable<boolean> {
```
undsoft, line 81: I think `try` is a bit misleading here. I mean I get that in context of `onDelete` method we are trying, but there is nothing that would explain `try` in the method itself. I'd go with just `deleteDataset`.

undsoft (changes requested):
1. It used to be that if you toggle switch on quotas lists and press Cancel the toggle went to the previous state. This is not the case anymore.
2. We should probably discuss order of methods as a team. My habit is to separate properties and methods and then sort them by private/public. This conflicts with an approach of having things together based on context.

RehanY147:
> We should probably discuss order of methods as a team. My habit is to separate properties and methods and then sort them by private/public. This conflicts with an approach of having things together based on context.
I generally agree with that. What I tried here was just put methods right after the first time they were used so the flow of the code is natural. That way you don't have to scroll around to find a method from where it was used.

undsoft (changes requested):
Toggle in user quotas still doesn't work correctly.
It switches and if you press Cancel, it stays in new position, even though it's incorrect.

undsoft (approved):
👍🏻

PR #7159, opened 2022-10-03
NAS-118420 / 22.12 / Extract user/group deletion dialog forms

undsoft:
Testing: test dialogs when Local User and Local Group are deleted.


AlexKarpov98 (approved):
LGTM, only 1 comment about empty lines at the end of the files.

PR #7161, opened 2022-10-04
NAS-112088 / 22.12 / Apps > Don't do validation on empty textboxes if they are not set required: true.

bvasilenko:
**Testing**

At **Apps** 
1. Pick an app which has a property with `type: int, min: ..., max: ...`  in the schema of it's configuration object. Most of such properties are port numbers.  For example, app **storj** property **Owner User ID**
2. Test validation for this property on the form for adding/editing the app


undsoft (approved):
Seems to be working correctly.
Any chance of having a test case for this?

PR #7162, opened 2022-10-04
NAS-118423 / 22.12 / Better error handling for Reports

denysbutenko:
image

Add error handler for Failed Export

For testing, use the host from the ticket description

### DIFF
--- a/src/app/pages/reports-dashboard/components/report/report.component.ts
+++ b/src/app/pages/reports-dashboard/components/report/report.component.ts
@@ -384,12 +384,17 @@ export class ReportComponent extends WidgetComponent implements OnInit, OnChange
   }
 
   handleError(evt: CoreEvent): void {
-    if (evt.data?.name === 'FetchingError' && evt.data?.data?.error === ReportingDatabaseError.InvalidTimestamp) {
-      const err = evt.data.data;
+    const err = evt.data.data;
+    if (evt.data?.data?.error === ReportingDatabaseError.FailedExport) {
       this.report.errorConf = {
         type: EmptyType.Errors,
-        large: false,
-        compact: false,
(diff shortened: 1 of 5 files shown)


src/app/pages/reports-dashboard/components/report/report.component.ts
```diff
@@ -384,12 +384,17 @@ export class ReportComponent extends WidgetComponent implements OnInit, OnChange
...
 
   handleError(evt: CoreEvent): void {
+    const err = evt.data.data;
+    if (evt.data?.data?.error === ReportingDatabaseError.FailedExport) {
+      this.report.errorConf = {
+        type: EmptyType.Errors,
+        title: this.translate.instant('Error getting chart data'),
+        message: err.reason,
+      };
+    }
```
RehanY147, line 394: Should we return here? Or create an object variable which holds the `errorConf` and assign it only once at the end of the method. Whichever seems a better option.

RehanY147 (commented):
The code looks good. But whatever was causing the error on that machine seems to be fixed because I don't see the error at all anymore.

denysbutenko:
> The code looks good. But whatever was causing the error on that machine seems to be fixed because I don't see the error at all anymore.

You could try to stop `rrdcached` service manually:

In WebUI Shell:
`systemctl stop rrdcached.service`

src/app/pages/reports-dashboard/components/report/report.component.ts
denysbutenko: Thanks for the suggestions, and sorry if the PR title is confusing. It is a bug fix to prevent breaking the UI in specific conditions—no plans for refactoring in this PR.

RehanY147 (approved):
Looks good

PR #7166, opened 2022-10-04
NAS-118359 / 22.12 / Update entities in docker-images store

bvasilenko:
**Testing**

On the **Apps** page double check that contents of **Docker Images** tab gets refreshed when user switches to other tag, and then switches back to **Docker Images** tab.


denysbutenko (approved):
Looks good to me

PR #7169, opened 2022-10-06
NAS-118447 / 22.12 / Fix: During Select an unused disk progress-spinner is not render

bvasilenko:
**Testing**

At **Storage → Manager Devices → (choose a disk) → Extend** a list of options should be loaded with indication of the progress.

### DIFF
--- a/src/app/modules/ix-forms/components/ix-select/ix-select.component.html
+++ b/src/app/modules/ix-forms/components/ix-select/ix-select.component.html
@@ -6,11 +6,11 @@
     [required]="required"
   ></ix-label>
 
-  <div class="input-container" [class.disabled]="isDisabled">
+  <div class="input-container" [class.disabled]="disabledState">
     <mat-select
       [(ngModel)]="value"
       [required]="required"
-      [disabled]="isDisabled"
+      [disabled]="disabledState"
       [multiple]="multiple"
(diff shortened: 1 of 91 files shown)


src/app/modules/ix-forms/components/ix-select/ix-select.component.ts
```diff
@@ -82,4 +82,12 @@ export class IxSelectComponent implements ControlValueAccessor, OnChanges {
     this.isDisabled = isDisabled;
     this.cdr.markForCheck();
   }
+
+  get disabledState(): boolean {
```
AlexKarpov98, line 86: I think it's better to keep using **is** prefix

Like **isDisabledState**

src/app/modules/ix-forms/components/ix-select/ix-select.component.html
```diff
@@ -38,7 +38,7 @@
       </ng-template>
     </mat-select>
     <mat-progress-spinner
-      *ngIf="isLoading"
+      *ngIf="loadingState"
```
AlexKarpov98, line 41: I think it's better to keep using **is** prefix

Like **isLoadingState**

AlexKarpov98 (approved):
Looks fine, but left some comments.

PR #7170, opened 2022-10-06
NAS-118490 / 22.12 / Extract strings from app routes for translations

denysbutenko:


undsoft (approved):
You also need to actually run translation in `RouterEffects`.

PR #7171, opened 2022-10-06
NAS-118471 / 22.12 / delete odd label

bvasilenko:


undsoft (changes requested):
Well, weird label is gone, but what about this checkbox behaviour:

https://user-images.githubusercontent.com/1481270/194343617-faa70e58-728f-4a14-9ec5-d24ae4e33ea0.mov

bvasilenko:
@undsoft Now the bug is fixed

undsoft (approved):
Explorer works correctly now. Can we add a test case?

PR #7172, opened 2022-10-06
NAS-118443 / 22.12 / Filter disks when 'disk' or 'device' is null

bvasilenko:
**Testing**

You should have a pool with 2 or more drives. Remove one drive from a pool in order to simulate replacement of the drive. 

Now, **Storage > tank > Topology > Manage Devices** should display the remaining drives correctly.


darkfiberiru:
Do we have something stopping this from getting merged.

undsoft (approved):
Looks good. I was able to open Devices page for a missing disk.

PR #7173, opened 2022-10-06
NAS-118466 / 22.12 / Root Path MNT

AlexKarpov98:


undsoft (changes requested):
Are we expecting more values here? Currently it's just a constant.

```
export const mntPath
```

AlexKarpov98:
> export const mntPath

I do agree, just thought it will be good to handle enum, with extending possibility :)
But CONST works!

undsoft (approved):
Looks good 👍🏻

PR #7174, opened 2022-10-06
NAS-118499 / 22.12 / Extracting some of VM dialogs into separate components

undsoft:
Test the following actions on the VMs page:
- Stopping
- Cloning
- Deleting
- Opening a display device (VM must be on).
- Opening a display device when there are multiple display devices in a VM
- Opening a display device when it is password protected.

### DIFF
--- a/src/app/pages/vm/vm-list/delete-vm-dialog/delete-vm-dialog.component.html
+++ b/src/app/pages/vm/vm-list/delete-vm-dialog/delete-vm-dialog.component.html
@@ -1,0 +1,36 @@
+<h1 mat-dialog-title>{{ 'Delete Virtual Machine' | translate }}</h1>
+<div mat-dialog-content>
+  <form [formGroup]="form">
+    <ix-checkbox
+      formControlName="zvols"
+      [label]="'Delete Virtual Machine Data?' | translate"
+      [tooltip]="helptext.delete_dialog.zvolsTooltip | translate"
+    ></ix-checkbox>
+    <ix-checkbox
+      formControlName="force"
+      [label]="'Force Delete?' | translate"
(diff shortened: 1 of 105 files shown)


denysbutenko (changes requested):
I was redirected `/ui/vm` after opening a display device when VM has a single or multiple display devices. Tested on m40 instance

src/app/pages/vm/vm-list/delete-vm-dialog/delete-vm-dialog.component.html
```diff
@@ -0,0 +1,36 @@
...
+      [tooltip]="helptext.delete_dialog.zvolsTooltip | translate"
+    ></ix-checkbox>
+    <ix-checkbox
+      formControlName="force"
+      [label]="'Force Delete?' | translate"
+      [tooltip]="helptext.delete_dialog.forceTooltip | translate"
+    ></ix-checkbox>
+    <ix-input
+      formControlName="confirmName"
+      [label]="'Enter {name} below to confirm.' | translate: { name: vm.name }"
```
denysbutenko, line 16: There is a similar string that may have translations:
`"Enter {value} below to confirm."`

undsoft:
@denysbutenko I think url redirection is an artefact of localhost testing. If you enable 'Preserve Log' on Network tab in Chrome inspector and [enable Auto-open DevTools for Popups](https://stackoverflow.com/questions/12212504/automatically-open-chrome-developer-tools-when-new-tab-new-window-is-opened), you can catch it:
Снимок экрана 2022-10-10 в 11 36 02

denysbutenko (approved):
Looks good 👍🏻

PR #7175, opened 2022-10-06
NAS-118464 / 22.12 / Fixes saving Small Block Size on basic edit

undsoft:
Test how `Metadata (Special) Small Block Size` gets saved on the dataset form when dataset is edited or added.
Previous error occurred when dataset was added in Basic mode.


bvasilenko (approved):
Tested **Basic settings** with N-level nesting

PR #7178, opened 2022-10-07
NAS-118454 / 22.12-BETA.2 / ACL manager after dataset creation fix

AlexKarpov98:
Now breadcrumbs are fixed and navigation works.
Updated /mnt string to CONST aswell in this commit. (all matches now)


AlexKarpov98:
@undsoft I'am sorry for pushing "mnt" in this commit

src/app/services/route-parts/route-parts.service.ts
```diff
@@ -39,6 +39,15 @@ export class RoutePartsService {
       }
 
       const { title, breadcrumb, disabled } = child.snapshot.data;
+
+      if (title === 'Datasets') {
```
undsoft, line 43: This is pretty bad. You are hardcoding specific strings in a service that should not know anything about datasets or storage.

src/app/pages/sharing/smb/smb-form/smb-form.component.spec.ts
```diff
@@ -419,13 +420,9 @@ describe('SmbFormComponent', () => {
...
-    const homeShare = await (await loader.getHarness(
-      IxCheckboxHarness.with({ label: formLabels.home }),
-    )).getValue();
-    const datasetId = sharePath.replace('/mnt/', '');
-    const poolName = datasetId.split('/')[0];
+    const datasetId = sharePath.replace(`${mntPath}/`, '');
     expect(spectator.inject(Router).navigate).toHaveBeenCalledWith(['/'].concat(
-      ['storage', 'id', poolName, 'dataset', 'acl', datasetId],
-    ), { queryParams: { homeShare } });
+      ['datasets', datasetId, 'permissions', 'acl'],
```
undsoft, line 425: I believe `homeShare` needs to be preserved.

undsoft (changes requested):
Redirect mostly works okay, just need to bring back the `homeShare` param.

I don't know if replacing `/mnt` with a constant provides value in tests. I mean, it's used in tests as specific example of a path. It is not meant to be easily changeable. It's like extracting "John" into a constant.
We can keep this change I suppose.

undsoft (changes requested):
I found an issue. It's probably not a regression as I don't see anything in the code that would cause it, but on the other hand I can only reproduce it on localhost.
1. Connect to m40.
2. Go to shares -> click Edit permissions next to `x` SMB share.
3. If you don't see the error, reload and keep reloading until the error appears.

Would be good to fix it here if not too hard. Otherwise a new ticket can be created.

undsoft:
@AlexKarpov98 
Reproduction steps I've given you were probably incorrect. Try creating a SMB share with Home Share enabled.

https://user-images.githubusercontent.com/1481270/195047862-0ff69f2a-2ba1-414a-87b1-fa0834e473e9.mov

AlexKarpov98:
Gonna make another PR to **release/22.12-BETA.2**

PR #7179, opened 2022-10-07
NAS-116557 / 22.12 / Applying dataset name and nesting limits

RehanY147:


src/app/pages/datasets/components/dataset-form/dataset-form.component.ts
```diff
@@ -1063,6 +1067,22 @@ export class DatasetFormComponent implements FormConfiguration {
       this.pk = this.parent;
       this.isNew = true;
       entityForm.formGroup.controls['parent'].setValue(this.parent);
+      entityForm.formGroup.controls['name'].setValidators(datasetNameTooLong(this.parent));
```
AlexKarpov98, line 1070: let's separate if () block from others with empty line

src/app/pages/datasets/components/dataset-form/name-length-validation.ts
```diff
@@ -0,0 +1,34 @@
+import { UntypedFormControl, ValidatorFn } from '@angular/forms';
+import { DefaultValidationError } from 'app/enums/default-validation-error.enum';
+
+export function datasetNameTooLong(parentPath: string): ValidatorFn {
+  const maxLengthAllowed = 200;
```
AlexKarpov98, line 5: can we be dynamic here and don't hardcode the 200 value?

src/app/pages/datasets/components/dataset-form/name-length-validation.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!control.parent) {
+      return null;
+    }
+
+    // Initializing the validator.
+    if (!thisControl) {
+      thisControl = control;
+    }
+
+    if (thisControl.value === '' || thisControl.value === undefined) {
```
AlexKarpov98, line 18: I think
**!control.value** should work instead of 2 checks

src/app/pages/datasets/components/dataset-form/name-length-validation.ts
```diff
@@ -0,0 +1,34 @@
...
+      return null;
+    }
+
+    if (!parentPath) {
```
AlexKarpov98, line 22: I think we can merge

src/app/pages/datasets/components/dataset-form/name-length-validation.ts
```diff
@@ -0,0 +1,34 @@
+import { UntypedFormControl, ValidatorFn } from '@angular/forms';
+import { DefaultValidationError } from 'app/enums/default-validation-error.enum';
+
+export function datasetNameTooLong(parentPath: string): ValidatorFn {
+  const maxLengthAllowed = 200;
+  let thisControl: UntypedFormControl;
+
+  return function datasetNameTooLongValidate(control: UntypedFormControl) {
+    if (!control.parent) {
```
AlexKarpov98, line 9: can we merge 3 if blocks in one?

**if (!parentPath || !control.value || !control.parent) {
  return null
}**

src/app/pages/storage/components/manager/manager.component.html
```diff
@@ -29,6 +29,7 @@
                   [formControl]="nameControl"
                   (input)="checkPoolName()"
                 />
+                <p class="error" *ngIf="nameControl.errors?.maxlength">Max length allowed for pool name is 50</p>
```
AlexKarpov98, line 32: how do we i18n translate this string?

src/app/pages/storage/components/manager/manager.component.ts
```diff
@@ -545,6 +545,9 @@ export class ManagerComponent implements OnInit, AfterViewInit {
     if (this.isNew && !this.nameControl.value) {
       return false;
     }
+    if (this.nameControl.value.length > 50) {
```
AlexKarpov98, line 548: merge these **3 if blocks** please, they do the same thing

AlexKarpov98 (changes requested):
please have a look at my suggestions.

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.scss
```diff
@@ -34,3 +34,7 @@
   height: 20px;
   width: 35%;
 }
+
+::ng-deep .custom-tooltip {
```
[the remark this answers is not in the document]
AlexKarpov98: we can use **::ng-deep :host .tooltip {}**

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
```diff
@@ -65,6 +65,7 @@ <h3 mat-card-title *ngIf="isZvol">{{ 'Zvol Details' | translate }}</h3>
           matTooltipPosition="above"
           class="value dataset-path"
           [matTooltip]="dataset.name"
```
denysbutenko, line 67: Perhaps a tooltip delay would make sense if there is no way to detect when the tooltip is necessary or not

src/app/pages/datasets/components/dataset-form/dataset-form.component.ts
```diff
@@ -130,7 +131,7 @@ export class DatasetFormComponent implements FormConfiguration {
       label: true,
       config: [
         {
-          type: 'input',
+          type: 'textarea',
```
denysbutenko, line 134: image

image

`Add Dataset` and `Edit Dataset` forms look inconsistently

denysbutenko (changes requested):
When trying to create a new dataset getting this validation error:
image
image

dszidi (commented):
The validation message and tooltip text could be improved as well. Users will most likely not know why this seemingly arbitrary number is the limit and they might not make the connection that the parent path is part of that limit

image

src/app/pages/datasets/components/dataset-form/dataset-form.component.ts
RehanY147: I'm not sure I understand the issue. Can you elaborate please?

src/app/pages/datasets/components/dataset-form/name-length-validation.ts
RehanY147: This is the limit for dataset name. The validator is also very specific as there's logic in here that can't be applied to other generic fields with length limit validations. So, I don't think we need to make this generic. If in the future, the dataset limit changes, we can just edit this one value and it'll still work.

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
RehanY147: @denysbutenko Can you elaborate a bit? When don't you need the tooltip?

src/app/pages/datasets/components/dataset-form/dataset-form.component.ts
AlexKarpov98: @RehanY147 

const ...

if ()

So empty line between variables and if statements

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
denysbutenko: I mean cases when the dataset name is not trimmed and fits into cell

denysbutenko: Initially, my assumption was that we are talking about the dataset tree tooltips and not `dataset-details-card`. Sorry for the confusion. That's not related to this ticket. The `dataset-details-card` looks good and works well 👍🏻

src/app/pages/datasets/components/dataset-form/dataset-form.component.ts
denysbutenko: image

This example is when the dataset name reaches the limit. Perhaps three lines in the textarea would be enough in most cases and help to save some vertical space, especially in the `Add dataset` form with two textarea fields. Does it make sense?

denysbutenko:
https://user-images.githubusercontent.com/351613/196139358-fe9cdc21-d3ef-4abd-9088-0e28df8a6865.mp4

After clicking the `Add dataset` button, the' Action Not Possible' dialog looks confusing. We may hide the button like in other scenarios. Does it make sense?

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
```diff
@@ -65,6 +65,7 @@ <h3 mat-card-title *ngIf="isZvol">{{ 'Zvol Details' | translate }}</h3>
...
+          [matTooltipClass]="['tooltip-name']"
```
AlexKarpov98, line 68: I didn't mean to call it .tooltip-name

it's a placeholder, please call it **dataset-path-tooltip**

src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.scss
```diff
@@ -34,3 +34,7 @@
   height: 20px;
   width: 35%;
 }
+
+::ng-deep :host .tooltip-name {
```
AlexKarpov98, line 38: **dataset-path-tooltip**

src/app/pages/storage/components/manager/manager.component.ts
```diff
@@ -540,7 +540,7 @@ export class ManagerComponent implements OnInit, AfterViewInit {
   }
 
   canSave(): boolean {
-    if (this.isNew && !this.nameControl.value) {
+    if (this.isNew && (!this.nameControl.value || this.nameControl.value.length > 50)) {
```
AlexKarpov98, line 543: should we disable the form if existing (not new) form nameControl.value.length > 50 ??

currently it is like
new form has name limit of 50
existing form does not have any limit

AlexKarpov98 (changes requested):

src/app/pages/storage/components/manager/manager.component.ts
RehanY147: On the existing form, you can't change name so the validation doesn't matter there.

AlexKarpov98 (commented):
from my side it is fine, but let's wait for other reviewers updates as well.

AlexKarpov98 (approved):
from my side it is fine, but let's wait for other reviewers updates as well.

dszidi (commented):
The tooltip help text looks good. 

image


The validation message under the field still needs to be more helpful though. 
Let change it to "The combined length of the 'Name' and 'Parent path' field values cannot exceed {{value}}"

RehanY147:
@dszidi The user can only edit this 'Name' field though. If we show a message like `The combined length of the 'Name' and 'Parent path' field values cannot exceed 200`, then they have to count the characters in the parent path. We should give them a simple number that they can keep track of.

dszidi (approved):
LGTM

denysbutenko (approved):
Looks good 👍🏻

PR #7180, opened 2022-10-07
NAS-118514 / 22.12 / Removing DocsReplace and other services

undsoft:
* Removes DocsReplace service.
* Extracts Delete Extent into a dialog.
* Linter rule for not using `window` directly.

Testing: test deletion of an extent in iSCSI.


RehanY147 (approved):
Looks good

PR #7184, opened 2022-10-10
NAS-118510: Redirect URL fix after manual address change

AlexKarpov98:
Have a look at hove to test it:
https://ixsystems.atlassian.net/browse/NAS-118510

video:
https://api.media.atlassian.com/file/cd9c9356-4e33-4469-a0b1-1b8fe17d2c57/artifact/video_1280.mp4/binary?client=931169de-affd-4506-a342-cd6186521b1e&collection=&max-age=2592000&token=eyJhbGciOiJIUzI1NiJ9.eyJpc3MiOiI5MzExNjlkZS1hZmZkLTQ1MDYtYTM0Mi1jZDYxODY1MjFiMWUiLCJhY2Nlc3MiOnsidXJuOmZpbGVzdG9yZTpmaWxlOmNkOWM5MzU2LTRlMzMtNDQ2OS1hMGIxLTFiOGZlMTdkMmM1NyI6WyJyZWFkIl19LCJleHAiOjE2NjU0MDY4NDYsIm5iZiI6MTY2NTQwNTk0Nn0.XAtMfTPtZhyXg2xj4p8aIDkP5jEn0J92qHlCCSU6aVw


undsoft:
Haven't tested the PR, but my questions are:
1. Why were we saving `currentUrl` in the first place? Doesn't something break now like reconnection after disconnect is handled?
2. Where is `currentUrl` saved?

AlexKarpov98:
@undsoft 

1. sessionStorage.currentUrl handled previous URL, which we were redirected to if try to load new page. Didn't notice any breaks. Not sure why we hold data in sessionStorage..

2. ws.redirectUrl -> handles it.

this.ws.redirectUrl = state.url;

AlexKarpov98:
@undsoft can you test PR as well?
If you will not find something, we can merge, cuz I didn't catch any problems.

undsoft (changes requested):
Yeah, so it no longer goes back to the correct page if websocket disconnects.
With that said, I think we should remove this session thing and handle this redirection in some other way.

undsoft (approved):
`sessionStorage` may actually be better here, because now if you open two tabs and both of them get disconnected, once you reconnect both of them will go to the same place.
Otherwise works well.

AlexKarpov98:
> `sessionStorage` may actually be better here, because now if you open two tabs and both of them get disconnected, once you reconnect both of them will go to the same place. Otherwise works well.

I agree, gonna use sessionStorage one :)

PR #7190, opened 2022-10-10
NAS-118526 / 22.12 / Partially enabling no-restricted-syntax

undsoft:
for-in loops are not recommended by Airbnb style guide.

Testing: general regression testing.


RehanY147 (approved):
Looks good. Nothing looks broken.

PR #7196, opened 2022-10-11
NAS-116562 / 22.12 / Added paginator for data vdevs

RehanY147:
I've left a piece of code in there to replicate unused disks a 1000 times in the pool creation manager for simplicity. That should be removed before merging.

To test, create/remove vdevs with different configurations and frequencies. Remove/add disks to vdevs and try duplication with the repeat button. UI should get stuck. All the info presented should be accurate such as vdevs, disks in vdevs, which page you're on, what items you're seeing on the paginator label etc.


dszidi (commented):
After a quick test the duplication functionality looks good. A few observations. 

If I choose an even number of duplicates, the paged output will display a single vdev which seems unintuitive. I think it would be nicer if the pager starts with the first page after generating duplicates.

Also the vdev item template could use some improvement. I think the way things are presented currently doesn't emphasize where the vdev item starts and ends correctly.

It wouldn't take much effort to make this a little more visually concise with something like this perhaps

image

@undsoft What do you think?

undsoft (changes requested):
Starting with first page makes sense to me as long as it doesn't break something else.
As for testing, I've added a vdev, used repeat, removed some of vdevs and tried adding Cache Vdev from the dropdown. This basically broke the page.

RehanY147:
@dszidi The paginator is meant to stay on the page where you hit duplication. So, if you created some vdevs and got to the second page and then hit "Repeat", after the duplication, the paginator will still be at the second page.

undsoft (changes requested):
1. If I go to first page of vdevs and remove the first one, it jumps to next page.

https://user-images.githubusercontent.com/1481270/196139971-ee34152c-e69a-4958-a769-a701e0137318.mov

2. Clicking Reset layout doesn't reset pagination.
3. Repeat button seems to work by repeating first vdev, which may get confusing to the user if user is on page 4. Let's rename the button to "Repeat first Vdev".

undsoft (changes requested):
1. Resetting layout if you are on last page doesn't reset pagination.
2. Deleting all entries if you are on the last page doesn't take me to a previous page. I just see paginator and an empty page.
3. Validation for Create button and corresponding warnings only take into account current page. I can get different warnings and different states for Create button based on the page I'm currently at.
4. If I delete all items on last page and then go on previous page, the Create button is disabled and total raw data capacity is reported as 0.

undsoft (approved):
Don't forget to remove test code before merging.

PR #7201, opened 2022-10-13
NAS-118546 / 13.0 / Disable some cloud sync buckets

undsoft:
1. Add a Storj cloudsync provider.
2. Go to Cloudsync form and select newly created provider.
3. You may need to mock to see that items in select may become disabled based on `Enabled` property.


denysbutenko (approved):
image

Looks good

PR #7203, opened 2022-10-13
NAS-117941 / 22.12-RC.1 / Error when going to datasets after removing all pools

AlexKarpov98:
To share service in 2 different module, we can make service global as I did via **providedIn: 'root',**
or create a separate Shared Module which we will use in 2 components.

And I had 2 solutions 

--- 

1. Inside of the dataset-details-panel.component are some components which take dataset ID and make request
Since after deletion we still have datasets in store, it tries to take the ID and load some data, but this dataset is no longer existing so that’s why we see error modals.
So I put skeleton to wait for datasets to be loaded, so we know if selected dataset exists

2. After pool deletion, we can clear DATAsets store (but need to plug datasets store module to Storage)

**Selected approach number #2.**


undsoft (approved):
Works well 👍🏻

PR #7204, opened 2022-10-13
NAS-118546 / 22.12 / Allowing some of cloudsync buckets to be disabled

undsoft:
1. Add a Storj cloudsync provider.
2. Go to Cloudsync form and select newly created provider.
3. You may need to mock to see that items in select may become disabled based on Enabled property.


denysbutenko (approved):
image

`demo-bucket` is disabled for 22.12, and it was enabled when tested for 13.0 not sure if that's an issue. Looks good to me

PR #7209, opened 2022-10-14
NAS-118505 / 13.0 / Added r50bm

RehanY147:
Check out ticket description for testing

### DIFF
--- a/src/app/core/classes/hardware/r50bm.ts
+++ b/src/app/core/classes/hardware/r50bm.ts
@@ -1,0 +1,111 @@
+import { Container, Texture, Sprite } from 'pixi.js';
+import { DriveTray } from './drivetray';
+import { Chassis } from './chassis';
+import { ChassisView, Position, LayoutGenerator } from './chassis-view';
+
+export class R50BM extends Chassis {
+  constructor(rearChassis = false) {
+    super();
+    this.model = 'r50bm';
+
+    this.front = new ChassisView();
+    this.front.container = new PIXI.Container();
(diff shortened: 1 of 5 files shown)


undsoft (changes requested):
1. When I click on REAR in enclosure UI management, the image is ginormous.
2. I've created a pool using `nvd` disks. I don't know if they should be shown here, but they aren't:
Снимок экрана 2022-10-17 в 11 04 23

RehanY147:
The nvme disks are shown on the rear @undsoft

undsoft (changes requested):
As mentioned on Slack, when I've created a pool with all disks and then went to enclosure management, I could still two unused disks and no disks on the rear for some reason.

RehanY147:
@undsoft I've solved the issue of unassigned pool. but the unassigned vdev issue still persists on certain disks. this looks like the unassigned vdev is separate issue triggered by older code. I created this ticket for it. https://ixsystems.atlassian.net/browse/NAS-118658

RehanY147:
@undsoft the enclosure has `has_slot_status` property `false`. Which means rear disks are not "identifiable". Need more clarification from someone from middleware.

RehanY147:
It also seems like in `disk.query` response, the first 48 disks (that show up on the front view) have enclosure info like `number` and `slot` but the last four (that appear on the rear view) have `enclosure` property set to `null` in the `disk.query` response. Which is why UI can't identify those four disks to be part of any enclosure so they get excluded from selection querying.

undsoft (approved):

undsoft:
backport

undsoft:
backport

PR #7214, opened 2022-10-17
NAS-118470 / 22.12-RC.1 / Multi-select styles are broken

AlexKarpov98:
Fixed multi-select styles

Screen Shot 2022-10-18 at 13 47 33


undsoft (changes requested):
This is how old multiselect used to look like (and how entity multiselect still looks like):
Снимок экрана 2022-10-18 в 11 28 17

AlexKarpov98:
@dszidi - please check the screenshot about, what do you think about such selection UI?

dszidi (changes requested):
Cosmetically it look fine to me. What I'm missing though is the option for "Select All". I think this should be baked in but optional. Alternatively maybe have it come into play by default if there are more than 3 options present.

For example ix-select is being used in reporting and there could potentially be hundreds or thousands of disks as well as quite a few report types to choose from. Let's not make users have to click so much

undsoft:
@dszidi That may be useful, but this is outside of scope for this PR.

AlexKarpov98:
@undsoft @dszidi 
So currently we are only looking for UI view.
If visually it is fine, please approve PR and I will merge it, and for "Select ALL" case, we can create a NEW ticket eventually.

dszidi (approved):
In that case, LGTM
I'll make a separate ticket about the "Select All" case

PR #7216, opened 2022-10-17
NAS-118443 / 22.12 / display removed disks in Storage Devices; restore 'Replace' menu item for every disk shown

bvasilenko:
**Testing**

1. Have a Mirrored pool of 2 drives
1. Remove a drive from that pool. And by remove, I mean physically pull the drive. Pool does show degraded.
1. Then attempt to replace the drive by going to **Storage > Pool > Topology > Manage Devices**.


undsoft (approved):
Functionality works, but UI could be improved. I will merge this PR now to unblock BETA2 release and will create another ticket to fix it.

PR #7219, opened 2022-10-18
NAS-118404 / 22.12-RC.1 / Dataset view, a Role Icon has double tooltip

AlexKarpov98:
Removed titles from **SVG** elements.


RehanY147 (approved):
Looks good

PR #7220, opened 2022-10-18
NAS-118541 / 22.12-RC.1 / Progress bar overflows jobs dialog

AlexKarpov98:
Problem was with jobs-panel.component manual **400px** width passed as props.


denysbutenko (approved):
👍🏻

PR #7221, opened 2022-10-18
NAS-118443 / 22.12-BETA.2 / display removed disks in Storage Devices; restore 'Replac…

undsoft:
…e' menu item for every disk shown


undsoft:
This was missing from previous backport for some reason. Will merge without review.

PR #7223, opened 2022-10-18
NAS-118517 / 22.12 / fix: CalledProcessError dialog appears when I open /ui/storage page

bvasilenko:


bvasilenko:
@undsoft I've updated the code to filter out particular error (CalledProcessError)

undsoft (changes requested):
That's not what I meant. There are several network requests in this file. Some of these requests may fail with `CalledProcessError` and this will be a valid error that we may want to show to user.
Instead we only want to silence errors for getting disk temperatures.

undsoft (approved):
Looks good.

PR #7226, opened 2022-10-19
NAS-118557 / 22.12-RC.1 / Replication: Only Same as Source and None retention policies can be used with Naming regex

AlexKarpov98:
Go to http://localhost:4200/data-protection

Replications Tasks -> ADD 

And try the described case for Wizard and Advanced Form

When creating a replication task (both using wizard and advanced form) we need to forbid using CUSTOM retention policy if Naming regex is used


undsoft (approved):
Appears to be working correctly.

PR #7228, opened 2022-10-19
NAS-118628 / 22.12 / Rework UI for replacing an unavailable disk

bvasilenko:
**Testing**

See PR https://github.com/truenas/webui/pull/7216 as it's a rework of it


RehanY147 (approved):
Looks good. Disk gets replaced.

PR #7229, opened 2022-10-20
NAS-118545 / 22.12-RC.1 / Make Dataset Space Management chart smaller

AlexKarpov98:
So, in result the chart & card are more tight and smaller now.

Before:
Screen Shot 2022-10-20 at 09 12 02

After: 
Screen Shot 2022-10-21 at 13 08 57

### DIFF
--- a/src/app/pages/datasets/components/dataset-capacity-management-card/dataset-capacity-management-card.component.scss
+++ b/src/app/pages/datasets/components/dataset-capacity-management-card/dataset-capacity-management-card.component.scss
@@ -11,7 +11,6 @@
 
     .details-item {
       margin: 20px 0 0;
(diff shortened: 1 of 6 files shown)


undsoft (changes requested):
Would there be enough space to try to always use horizontal layout?

undsoft (approved):
Generally looks okay.
On some resolutions however legend may get too close to the right border - need additional margins and can reduce margins on the left between legend and donut.
Снимок экрана 2022-10-21 в 11 30 40

AlexKarpov98:
@undsoft  - please have a look at it now :)

undsoft (approved):
Looks good.

PR #7230, opened 2022-10-20
NAS-117236 / 22.12-RC.1 / Check for keyboard support on new storage pages

AlexKarpov98:
checked accessibility via keyboard, made sure everything highlights and responds to user keyboard actions.

Screen Shot 2022-10-20 at 13 01 00


undsoft (changes requested):
Cannot select device in Devices, cannot select a dataset in Datasets.

AlexKarpov98:
Fixed hovering effect for **tree-item**
Screen Shot 2022-10-21 at 15 19 12

And added **enter press** support to view tree-item's data.

src/app/pages/datasets/components/dataset-management/dataset-management.component.html
```diff
@@ -86,6 +87,7 @@
                 routerLinkActive="selected"
                 [routerLink]="['/datasets', dataset.id]"
                 (click)="openMobileDetails()"
+                (keydown.enter)="openMobileDetails(['/datasets', dataset.id])"
```
undsoft, line 90: It looks like you can unify all of this: 
1. Remove routerLink directive.
2. Have the same handler for (click) and (keydown.enter).
3. Call it something else like `onRowSelected(dataset: Dataset)`

undsoft (approved):
If I am in a smaller screen and have mobile layout, it seems like focus stays in the table and it's hard to start using elements in the mobile view (I have to Tab through the whole table without seeing anything happening).
See if there is an easy fix for this. If there is no simple fix, you can merge without it.

AlexKarpov98:
@undsoft - have a look at it now please :)

undsoft (approved):
There is a button to expand/collapse rows. It is no longer possible to expand row on mobile without going to mobile details.
That is if you have a collapsed row and you want to select a child, it's not going to be easy.

Otherwise works okay.

AlexKarpov98:
tested and now works correctly

undsoft (approved):
Works okay.
I would do
```
(keydown.enter)="$event.stopPropagation()"
```
on button instead of looking at element tag. Less fragile.

PR #7231, opened 2022-10-20
NAS-118668 / 22.12 / Removing usages of `any`

undsoft:
Code review should be enough. You can also test some places where code was changed.


bvasilenko (approved):
LGTM

PR #7238, opened 2022-10-21
NAS-118677 / 22.12 / Adding jest eslint rules

undsoft:
Code review should be enough.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -288,6 +288,16 @@ module.exports = {
       }
     },
     {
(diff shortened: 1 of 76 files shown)


AlexKarpov98 (approved):
Looks good, I think it will be better to fix these warnings:(at the bottom of PR)

Screen Shot 2022-10-24 at 11 29 37

undsoft:
These are false warnings unfortunately.

PR #7239, opened 2022-10-21
NAS-118683 / 22.12 / Extracting dialog components

undsoft:
Testing: test affected dialogs.


src/app/modules/truecommand/components/truecommand-connect-modal/truecommand-connect-modal.component.html
```diff
@@ -0,0 +1,74 @@
+<h1
```
undsoft, line 1: This component is a rename, not a new file.

PR #7242, opened 2022-10-24
NAS-113762 / 22.12 / Fixed missing toolbar in Mozilla

Raz-Dva:
Go to System Settings -> Services  and run this command in below in the dev console 
`window.document.evaluate('//tr[contains(.,"WebDAV")]', document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue.scrollIntoView(true)`
Expected: toolbar should not disappear or change its position

image


undsoft (approved):
Appears to be working correctly with no obvious regressions.

PR #7246, opened 2022-10-25
NAS-118520 / 22.12 / Fix 'Keyboard map' label bug

bvasilenko:
**Testing**

Tweak your box config (or hard-code json response) for **system.general.config** so it returns property `kbdmap` with a value `"us.unix"` set to it, to replicate user's box settings, which he got stuck at after the upgrade.

Go to **Settings > General** and make sure that UI is not broken, and you're able to set keyboard mapping to some other value via the UI.


undsoft (approved):
Thanks, unexpected value no longer breaks the page.

PR #7252, opened 2022-10-26
NAS-118505 / 22.12 / Nas 118505 22.12 alt

RehanY147:
Added new enclosure system


undsoft (approved):
Retested on a core system. Everything appears to be in order.

PR #7253, opened 2022-10-26
NAS-118661 / 22.12 / Remove any in CloudsyncFormComponent

AlexKarpov98:


src/app/interfaces/cloud-sync-task.interface.ts
```diff
@@ -13,14 +13,29 @@ export interface CloudCredential {
 
 export interface BwLimit {
   time: string;
-  bandwidth: number;
+  bandwidth: string | number;
+}
+
+export interface CloudSyncTaskForm extends CloudSyncTask {
```
undsoft, line 19: You can introduce a local type alias instead:
```
CloudsyncFormComponent['form']['value']
```

undsoft (changes requested):
We should make this match middleware responses.
I.e. `bandwidth` in `BwLimit` is not a `string` according to API docs
`bwlimit` is not a `string | string[]`
if `credentials` are a number in query and are not during create, this should be reflected in corresponding interfaces.

This will give you errors in `prepareData`, but the problem with that function is that it mutates input argument. It's better to rewrite it in a way that doesn't mutate input, but forms a new object instead.

src/app/interfaces/cloud-sync-task.interface.ts
```diff
@@ -44,7 +44,12 @@ export interface CloudSyncTask {
   create_empty_src_dirs: boolean;
 }
 
-export type CloudSyncTaskUpdate = Omit<CloudSyncTask, 'id' | 'job' | 'locked'>;
+export type CloudSyncTaskRequestPart = Omit<CloudSyncTask, 'id' | 'job' | 'locked' | 'credentials' | 'encryption_salt' | 'args' | 'filename_encryption' | 'bwlimit'>;
```
undsoft, line 47: Doesn't seem to be used anywhere else. Either incorporate it in definition of `CloudSyncTaskUpdate` (preferred) or remove `export`.

src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
```diff
@@ -572,92 +580,75 @@ export class CloudsyncFormComponent {
...
-    }
-    if (value.fast_list !== undefined) {
-      attributes.fast_list = value.fast_list;
-      delete value.fast_list;
-    }
-    if (value.chunk_size !== undefined) {
-      attributes.chunk_size = value.chunk_size;
-      delete value.chunk_size;
-    }
+    atrributesToFill.forEach((name: FormAttribute) => {
```
undsoft, line 644: You can delete `FormAttribute` type and add `as const` to `attributesToFill`.

undsoft (changes requested):
When encryption is enabled, it looks like password is not being sent - I get validation error about it.

undsoft (changes requested):
Same issue as in last review - when encryption is enabled password is not sent.

undsoft (approved):
Compared API payloads before and after change in different scenarios. They appear to be the same.

PR #7254, opened 2022-10-26
NAS-118612 / 22.12 / Linter for attribute order in html 1

undsoft:
Enforces order of attributes on html elements.
Code review is enough.


denysbutenko (approved):
Looks good 👍🏻

PR #7255, opened 2022-10-26
NAS-115669 / 22.12 / After upgrade the Standby controller card didn't update until Screen refresh/relog in

AlexKarpov98:
Download the manual update file
In the UI:
System->Update
Click Install Manual Update File
Set Include Password Seed and click Save Configuration
Set the Memory Device as the File temporary location
Upload the file using the Choose File button.
Click Apply Update.


src/app/interfaces/events/ha-status-event.interface.ts
```diff
@@ -1,6 +1,7 @@
 import { FailoverDisabledReason } from 'app/enums/failover-disabled-reason.enum';
+import { HaStatusText } from 'app/enums/ha-status-text.enum';
 
 export interface HaStatus {
-  status: 'HA Enabled' | 'HA Disabled';
+  status: HaStatusText;
```
undsoft, line 5: Can't we just have a boolean flag here? 😭

AlexKarpov98: @undsoft 
I've just refactored, not sure about this status property :)

undsoft: Please add a TODO line to make this a boolean variable.

undsoft (approved):
I don't think there is a way for me to test this with failover upgrade, but the code makes sense.
Let's merge it and ask Release Engineering to check it next time they test this part of system.

PR #7259, opened 2022-10-27
NAS-118612 / 22.12 / Lint attribute order

undsoft:
Adds linter for order of attributes. Code review is enough.

Also changed markuplint not to `--fix` automatically because then it doesn't report errors it couldn't fix correctly.


undsoft:
Found a better solution with eslint. Will change markuplint to eslint in next PR, but rest of the fixes stand.

undsoft:
Nevermind, fixed it in current PR.

RehanY147 (approved):
Looks good

PR #7260, opened 2022-10-27
NAS-118768 / 22.12 / Directive to show UI elements on nightlies only

undsoft:
Testing: there is a test button on main dashboard that should only be visible on nightlies.


bvasilenko (approved):
Code looks good to me. Tested with mocks

PR #7262, opened 2022-10-28
NAS-118777 / 22.12 / Default cloudsync credential name to provider name

undsoft:
Go to Credentials -> Backup credentials. 
When adding a new cloud credential the order of fields is different and name will default to provider name until user interacts with Name control.


denysbutenko (changes requested):
It doesn't work on the initial load. I see two empty fields, provider options are loaded, and provider input is set to the `Storj iX` value, but the `Name` field is still empty. If I change the provider it works well.

undsoft:
Fixed.

denysbutenko (approved):
Looks good 👍🏻

PR #7265, opened 2022-10-30
NAS-118801 / 22.12 / Removing any

undsoft:
Testing: there are some places where runtime code and not only times were affected (e.g. portal list, disk list). These can be tested.


dszidi (approved):
Code looks good. Tested Disks page and no regressions found
LGTM

PR #7266, opened 2022-10-31
NAS-118773 / 22.12 / Added dialog to download all keys

RehanY147:
Create an encrypted pool. Then add a dataset under that pool that is encrypted but with a different encryption (not inherited). On the datasets page, if you select such a root dataset of the pool, you should see a button "Export All Keys" on the Zfs Encryption card which should let you download all the keys of the encrypted datasets on that pool.


undsoft (changes requested):
It works, but I think we should change the logic of how it works.
There is no need to load keys every time user goes to Datasets.
They are only needed when user tried to export all keys.
You could show Export All Keys when this is a root dataset and it is encrypted. Then send network request and save keys when button is pressed.

undsoft (approved):
Works well, I can export all keys now.

PR #7268, opened 2022-10-31
NAS-118669 / 22.12 / Extract dialogs from ChartReleasesComponent

bvasilenko:
**Testing**

Optionally, you can use mockup response of `chart.release.pod_console_choices`:

```
{
  "storj2-68bf795df7-fb56g": [
    "storj"
  ]
}

```


RehanY147 (approved):
Looks good

PR #7271, opened 2022-11-01
NAS-118066 / 22.12 / Nas 118066 Fixed: UI isn't updating/properly showing snapshots

Raz-Dva:
Go to Datasets. Create new snapshots for zvol and dataset using button "Create Snapshot" in the card "Data Protection".
The "Total shots" field should display the actual number of snapshots.
Go to page Snapshots and check it out. 
Expected: that clicking Manage Snapshots would show the snapshots.

image


src/app/pages/datasets/components/data-protection-card/data-protection-card.component.ts
```diff
@@ -13,15 +18,39 @@ import { IxSlideInService } from 'app/services/ix-slide-in.service';
...
+      }),
+      untilDestroyed(this),
+    ).subscribe((res) => {
+      this.dataset = { ...this.dataset, snapshot_count: res.length };
+      this.isLoading = false;
+      this.cdr.markForCheck();
+    });
+  }
+
+  @Input() dataset: DatasetDetails;
```
AlexKarpov98, line 47: let's move it at the top before `readonly console = console;`
and add empty line

src/app/pages/datasets/components/data-protection-card/data-protection-card.component.ts
```diff
@@ -13,15 +18,39 @@ import { IxSlideInService } from 'app/services/ix-slide-in.service';
...
   styleUrls: ['./data-protection-card.component.scss'],
 })
-export class DataProtectionCardComponent implements OnInit {
+export class DataProtectionCardComponent implements OnInit, OnChanges {
   readonly console = console;
-  @Input() dataset: DatasetDetails;
-
+  changeEvent$: Subject<string> = new Subject();
+  isLoading = false;
   constructor(
```
AlexKarpov98, line 25: add empty line before constructor

AlexKarpov98 (approved):
Looks good, please fix the comments above.

Some questions left: 
Current behavior for viewing **root dataset snapshots** is like: view all snapshots for all datasets instead only for root dataset. 
example: I see 2 snapshots text for root dataset, but when click "Manage snapshots" I see all snapshots for all datasets.

For non-root elements, it is working fine.

Raz-Dva:
> 

I think that in this case, we can to expand the filter's capabilities to display a list of snapshots for only one dataset (without child snapshots)
I discussed this with Eugene. So far, this is not necessary.

Raz-Dva:
I think that in this case, we can to expand the filter's capabilities to display a list of snapshots for only one dataset (without child snapshots) I discussed this with Evgeny @undsoft . So far, this is not necessary.

PR #7272, opened 2022-11-01
NAS-118602 / 22.12 / Add ability for users to fix clock

bvasilenko:
**Testing**

1. On your VM box shift the clock so it differs from your host machine clock
2. Check System Info widget: warning should be displayed
3. Click "Synchronize time" button and check that the time was updated


dszidi (commented):
After I hit the refresh icon in the warning, I got a traceback

image

dszidi (approved):
Updated my VM and all works as expected. LGTM

PR #7273, opened 2022-11-02
NAS-118818 / 22.12 / Subscribing to disk events

RehanY147:
Repeating the process of reloading disks when `disk.query` event from middleware gets fired.


undsoft (changes requested):
Subscribing to `disks.query` will not send you a list of all disks. I believe it will instead return you a single disk that was changed.

undsoft (changes requested):
Enclosure page doesn't open - missing @UntilDestroy.
After fixing that, I've tried to simulate removal of a disk like this:
```
zpool status -v
zpool offline pool one-of-the-disks
```
Nothing changed on disk status page in real time, but if I reload the page I can see updated status on some of the disks.

Could be that my testing methodology is flawed, but then it would be good to ask for middleware's help on how to simulate disk removal.

undsoft (approved):
Tested DiskUpdate service on a disk-list, where it works correctly.

PR #7274, opened 2022-11-02
NAS-118847 / 22.12 / add new failover.disabled.reasons

yocalebo:
See https://github.com/truenas/middleware/pull/10052 for details


RehanY147 (approved):
Looks good

PR #7275, opened 2022-11-02
NAS-118830 / 22.12 / Localhost redirects to remote machine when Api Keys page is reloaded

AlexKarpov98:
proxy redirect happens for **/api/** only path and is not touching **/api***


undsoft (approved):
Page reloads correctly now.

PR #7281, opened 2022-11-03
NAS-115979 / 22.12 / The Web UI isn't updating During/after Upgrading

AlexKarpov98:


src/app/store/system-config/system-config.effects.ts
```diff
@@ -29,8 +30,14 @@ export class SystemConfigEffects {
     }),
   ));
 
+  advancedConfigUpdated$ = createEffect(() => this.actions$.pipe(
+    ofType(advancedConfigUpdated),
+    map(() => this.window.location.reload()),
+  ), { dispatch: false });
```
undsoft, line 36: I don't think I like the idea of reloading the page on every advanced setting changed. There are a lot of settings that don't require this.
You can keep it in one place where it is needed, but a // TODO comment needs to be added as this is a hack that would need to be fixed later.

src/app/store/system-info/system-info.effects.ts
```diff
@@ -77,21 +78,6 @@ export class SystemInfoEffects {
     }),
   ));
 
-  subscribeToHa = createEffect(() => this.actions$.pipe(
```
undsoft, line 80: Why remove this part?

undsoft (changes requested):
I've initiated failover, logged in and waited for card on dashboard to update. It never updated.

src/app/store/system-info/system-info.effects.ts
AlexKarpov98: @undsoft  - as I've seen it's not used in the APP

undsoft: I think was used - it created a subscription to HA updates.

src/app/store/system-config/system-config.effects.ts
AlexKarpov98: actually it's only if few different places, one for advancedConfig updated and some fore different routes where we trigger /reboot or /failover.
I'll add comment to all the places, they are independent

undsoft: `advancedConfigUpdated` is dispatched whenever any of the advanced settings forms is edited. There is no need to reload the whole UI when I say change something in Console form.

undsoft (changes requested):
Failover card updates correctly. What are you trying to solve with reloading UI after advanced settings are updated?

PR #7286, opened 2022-11-07
NAS-118900 / 22.12 / Improve type checking for appLet directive

denysbutenko:
Add `ngTemplateContextGuard` for `*appLet` directive to return the correct type in templates.

image
image


AlexKarpov98 (approved):
works good, nice work.
Screen Shot 2022-11-08 at 15 41 42

PR #7291, opened 2022-11-08
NAS-118867 / 22.12 / Using the selected version

RehanY147:
Previously, we were deleting version attribute. Using the correct value now. New installation of an older version of an app should install the selected version now instead of just always installing the latest version.


undsoft (approved):
Confirmed that selected version is being sent in websocket query.

PR #7292, opened 2022-11-08
NAS-118857 / 22.12 / Using job instead of call

RehanY147:
Changing the encryption type from key to passphrase of datasets should show correct info on the datasets page afterwards without having to reload.


AlexKarpov98 (approved):
Tested, status is updating correctly, realtime.
👍 
Screen Shot 2022-11-08 at 15 33 53

AlexKarpov98:
@RehanY147  - fix test cases please ;)

PR #7293, opened 2022-11-08
NAS-118868 / 22.12 / Apps UI goes into a back and forth loop between tabs.

AlexKarpov98:
Testing: I was able to reproduce when started recording a video, so that computer slows down a bit and eaten more RAM.


RehanY147 (approved):
Looks good

Ornias1993:
Wonderfull @AlexKarpov98 I've worked with this portion before and was quite amazed that it was kinda convoluted thinking "this must be causing issues somewhere" 👍

PR #7295, opened 2022-11-08
NAS-113218 / 22.12 / Enclosure UI stage left Vdev should show name not type

AlexKarpov98:
Vdev now has a **name** property and we display it.
On MINI | M40 | other systems
Screen Shot 2022-11-09 at 01 16 51


dszidi (commented):
The name value should be all caps. eg. MIRROR-1

image

darkfiberiru (approved):
Looks great glad too see this got added

PR #7301, opened 2022-11-10
NAS-117214 / 22.12 / Garbled output on Chrome on MacOS

AlexKarpov98:
Not sure how to detect this bug, but probably there were some issues with font, so updated logic:
we have a default monospace font: "Courier" which will work in any browser.

Screen Shot 2022-11-10 at 15 48 05


undsoft:
I'm not sure it's a font issue, looks more like an encoding issue. Maybe see if terminal library has any options for that, or see if switching browser/Truenas languages helps reproduce it.

AlexKarpov98:
> I'm not sure it's a font issue, looks more like an encoding issue. Maybe see if terminal library has any options for that, or see if switching browser/Truenas languages helps reproduce it.

was not able to reproduce it at any language

undsoft:
@AlexKarpov98 I think we should investigate this more. I'll ask user some questions.

undsoft:
I've tried this fix on user's computer. It didn't work, so I'll close the PR.

PR #7302, opened 2022-11-10
NAS-116577 / 23.10 / NAS-116577: Improve Handled unknown error in sentry

AlexKarpov98:
Updated one package to avoid this error: **[(...).transition is not a function]**

Found a way to extend the default error handler, 
and will return **JSON.stringify(error)** if no matches found for error handler to avoid **"Handled unknown error. "**


undsoft (changes requested):
1. Second argument in `extractor` is the `defaultExtractor`. You call the default extractor and then if it fails, do something else.
2. It would be good to try to extract error message, not just put in JSON, as it may be huge with stacktrace and other data.
See `EntityUtils.handleWsError` and `EntityUtils.errorReport`.

AlexKarpov98:
would be good to check how much new default messages will we get 
`iXsystems [sentry] default handled error`

undsoft (changes requested):
`error.message` is already extracted by default extractor. Not sure what `error.name` is.

You need to use custom extractor to extract error messages that we may get from middleware. You can look at `WebsocketError` and `Job` for relevant fields and at `EntityUtils` for how errors are currently being processed.

undsoft:
Bump.

src/app/helpers/sentry-custom-extractor.helper.ts
```diff
@@ -0,0 +1,39 @@
...
+  errorCandidate: WebsocketError | Job | Error | any,
+  defaultExtractor: (error: unknown) => unknown,
+): unknown => {
+  const defaultExtractorResults = defaultExtractor(errorCandidate);
+
+  if (defaultExtractorResults) {
+    return defaultExtractorResults;
+  }
+
+  const errorReportValue = errorReport(errorCandidate as WebsocketError);
```
undsoft, line 25: `as WebsocketError | Job`

src/app/helpers/sentry-custom-extractor.helper.ts
```diff
@@ -0,0 +1,39 @@
+import ErrorStackParser from 'error-stack-parser';
+import { Job } from 'app/interfaces/job.interface';
+import { WebsocketError } from 'app/interfaces/websocket-error.interface';
+
+const errorReport = (err: WebsocketError | Job): string | void => {
```
undsoft, line 5: Isn't really `errorReport`, morel like `extractApiError`

undsoft (approved):
Looks good.

undsoft:
Would be good to fix comments as well. @AlexKarpov98

AlexKarpov98:
@undsoft, yes, sorry, auto push triggered after approval.
Changes are here now: https://github.com/truenas/webui/pull/7403

PR #7310, opened 2022-11-13
NAS-118899 / 22.12 / Refactoring VM Edit Form 1

undsoft:
Most changes are code only.
You can test VMs -> Add -> CPU and Memory -> product of threads, cores and VCPUs shouldn't exist some limit that system has.


dszidi (approved):
Code looks good. Tested CPU validation and it seems to work fine.

image

PR #7316, opened 2022-11-14
NAS-118548 / 22.12 / Hide menu items for unavailable reports

denysbutenko:
Improvements for Reports Toolbar to reduce empty screens on the Reporting page

- Hide `Disk Temperature` from `Disk Metrics` when no temperature data
- Hide `UPS` menu item when no UPS
- Hide `Target` menu item when no graphs


dszidi (approved):
Tested on a VM. Disk temperatures, UPS and Target categories were hidden as expected. Did not encounter any regressions. LGTM

image


image

PR #7321, opened 2022-11-15
NAS-119001 / 22.12 / Preselect first option for rollback

denysbutenko:
Select the first available version from options in App Rollback dialog


undsoft (approved):
Looks good.

PR #7322, opened 2022-11-15
NAS-118938 / 22.12 / Added loader and corrected ids

RehanY147:
Go to the NFS List page by going to the Shares Dashboard and then clicking on the table's title for NFS Shares. On this page, previously, clicking the checkbox of any row affected only the first row. That should be fixed now. Also, user shouldn't be able to click the checkbox multiple times while the first action is still loading so I've added a loading state. Make sure it works ok.


denysbutenko (approved):
Kapture 2022-11-15 at 18 00 33

Works well.

PR #7327, opened 2022-11-16
NAS-118937 / 22.12 / Add dispatch `loadHaStatus()`

bvasilenko:
**Summary**

Forgotten call `dispatch()` at `loadHaStatus()`. In my previous PR I didn't notice that it wasn't called after user was logged in. 

**Testing**

On new **Datasets** page for the **Import Data** button:

- For regular server the button should be displayed because it is not HA. 

- If you have (or simulate) HA server, then the button should not be displayed.


src/app/pages/datasets/components/dataset-management/dataset-management.component.ts
```diff
@@ -306,8 +307,9 @@ export class DatasetsManagementComponent implements OnInit, AfterViewInit, OnDes
   }
 
   loadHaStatus(): void {
-    this.store$.select(selectHaStatus).pipe(untilDestroyed(this)).subscribe((haStatus) => {
-      this.hasHa = !!(haStatus && haStatus.hasHa);
+    this.store$.dispatch(loadHaStatus());
+    this.store$.select(selectHaStatus).pipe(filter(Boolean), untilDestroyed(this)).subscribe((haStatus) => {
```
denysbutenko, line 311:
```diff
- *ngIf="showImportData"
+ *ngIf="showImportButton$ | async"
```

```diff
- hasHa: boolean;
- get showImportData(): boolean {}
- loadHaStatus(): void {}
+  /**
+   * Hidden on HA systems.
+   * Issues: fenced reservations and the potential to cause a kernel panic, as well as +an alert being raised.
+   */
+  showImportButton$: Observable<boolean> = this.store$
+    .select(selectHaStatus)
+    .pipe(
+      filter(Boolean),
+      map((haStatus) => !haStatus.hasHa),
+    );
```

bvasilenko:
> Technically speaking this action should already be dispatched: `admin-layout` dispatches `adminUiInitialized` this loads system info in `SystemInfoEffects` and if system has an HA license `loadHaStatus` is dispatched.
> 
> Please check what's going on there, maybe there is a mistake there.

@undsoft 
Ok, now the `adminUiInitialized` will trigger the `loadHaStatus`. And I've rolled back direct call to `dispatch(loadHaStatus)`.

src/app/store/system-info/system-info.effects.ts
```diff
@@ -64,7 +64,7 @@ export class SystemInfoEffects {
   ));
 
   loadHaStatus = createEffect(() => this.actions$.pipe(
-    ofType(loadHaStatus, haSettingsUpdated, passiveNodeReplaced),
+    ofType(loadHaStatus, haSettingsUpdated, passiveNodeReplaced, adminUiInitialized),
```
undsoft, line 67: If we are always going to load HA status on admin ui init, we should remove `loadHaStatus` being dispatched a couple lines above.

undsoft (changes requested):
So, I've tried it with one of the machines from Jenkins that doesn't have HA and I don't see the button.

bvasilenko:
> So, I've tried it with one of the machines from Jenkins that doesn't have HA and I don't see the button.

@undsoft 
Ah, I finally noticed the missing condition checks, and this explains why it was working for one box, but not for the other. Now it should work correctly on any box.

undsoft (approved):
In the future we probably need to streamline how we work with HA status in one central store, this is quite complicated.
But for the purposes of this ticket, the button is there and works okay.

PR #7328, opened 2022-11-16
NAS-119025 / 22.12 / Errors not shown when ACME certificate is not created

undsoft:
Testing: 
1. In Certificates add an ACME DNS Authenticator
2. Add a CSR. 
3. Once CSR is added, press `Create ACME Certificate`. 
4. Submit form.
You should be able to see errors now.


denysbutenko (approved):
image

It works as described.

PR #7329, opened 2022-11-16
NAS-119000 / 22.12 / Fixing update dialog for a single app

undsoft:
See in the bottom here how to create a fake version to update to: https://ixsystems.atlassian.net/wiki/spaces/IE/pages/332038543/Environments+Licenses

Testing: upgrade a single app.


denysbutenko (approved):
Nice catch 👍🏻

PR #7332, opened 2022-11-16
NAS-118826 / 22.12 / `extra` in ExportDisconnectModalComponent

AlexKarpov98:
we should update Job interface and remove anys from ExportDisconnectModalComponent.


src/app/interfaces/job.interface.ts
```diff
@@ -8,6 +8,10 @@ export interface Job<R = unknown, A = unknown[]> {
   arguments: A;
   description: string;
   error: string;
+  extra?: {
```
undsoft, line 11: I would imagine that `extra` can be a lot of things, not just these two. Better to type it as `Record<string, unknown>`.

undsoft (approved):

PR #7337, opened 2022-11-17
NAS-117990 / 22.12 / Service running toggle state incorrect after canceling

AlexKarpov98:


undsoft (approved):
That's an unusual way of resetting a column, but I suppose it's okay. Toggle works okay now.

PR #7338, opened 2022-11-17
NAS-118366 / 22.12 / Replace Epilogue with Titillium Web for header fonts

dszidi:


undsoft (changes requested):
The font itself looks fine, but there needs to be a `@font-face` defined somewhere.
I would also try to remove other uses of `Epilogue` and remove the font completely.

undsoft (approved):
Looks good and consistent across browsers I have.

PR #7339, opened 2022-11-17
NAS-119046 / 22.12 / Remove redundant header

dszidi:
From Ticket Description...

> This "Unused Resources" header doesn’t seem to add much considering that at the moment there will only be one card that might be in this area. We should remove it.
> 
> 
image


RehanY147 (approved):
👍🏽

PR #7341, opened 2022-11-17
NAS-115632 / 22.12 / Fix SMB list & table on dashboard when isClustered

bvasilenko:
**Testing**

1. On **Sharing** page, when system is in clustering mode, then the page should display a different path, and for each table row particular actions should be disabled.

   **Note:** If your VM does not support clustering mode, then you can hard-code `True` response for **cluster.utils.is_clustered** request.

2. For **Sharing > SMB** page same logic of condition checking, path displaying and disabling certain actions should be applied.


undsoft (changes requested):
1. You should also update which column is used. You didn't actually change it to `path_local`.
2. I think users won't notice the tooltip. I would do something else, like add a dialog when user tried clicking on the row, or a tooltip on the row if possible.

undsoft (changes requested):
1. One thing that I forgot to mention in description is that we still need to be able to edit filesystem acl.
2. I think we can permanently update path to path local even when system is not in clustered mode.

bvasilenko:
> 1. One thing that I forgot to mention in description is that we still need to be able to edit filesystem acl.
> 2. I think we can permanently update path to path local even when system is not in clustered mode.

Done!

undsoft (changes requested):
As per previous comment. I should still be able to edit Filesystem ACL on clustered systems.

AlexKarpov98 (approved):
code looks good

undsoft (approved):
Thanks, looks good.

PR #7342, opened 2022-11-18
NAS-119027 / 22.12 / Extracting dialog in Pool Manager

undsoft:
Also fixes URL in ACL editor.
Testing: test dialog that appears after Repeat button in pressed when creating a pool.


denysbutenko (approved):
image

Looks good 👍🏻

PR #7344, opened 2022-11-18
NAS-119022 / 22.12 / Fix sysctl editing

undsoft:
Go to System Settings -> Advanced and add an item to sysctl.
You can use `vm.laptop_mode` for variable name.
Then edit the entry that was created.


bvasilenko (approved):
Editing works good, when using correct values (numeric)

However, the input field allows to enter invalid value, and displays the error afterwards.

image


If possible, let's add a validation here.

undsoft:
Don't think I can do validation on client here.

PR #7346, opened 2022-11-18
AUTO-69 / 22.12 / Adding test to create AD SMB share on HA system and testing it works

ericbsd:
I added the xpaths module to source xpaths to make it easier to update UI test xpath
I replaced all xpath that is in test_NAS_T0964.py in the other test files with the xpaths module.
I reverted the disk wipe step code because of a DOM issue.
I fixed some minor issues.


q5sys (approved):
LGTM

PR #7351, opened 2022-11-21
NAS-118904 / 23.10 / Reducing boiler plate code for empty tables

RehanY147:


src/app/pages/account/users/user-list/user-list.component.html
```diff
@@ -14,101 +14,89 @@
...
-          </td>
-        </ng-container>
+<mat-card>
+  <mat-card-content>
+    <table
+      ix-table
+      matSort
+      matSortDisableClear="true"
+      ix-table-empty
+      [emptyConfig]="(isLoading$ | async) ? loadingConfig : (emptyOrErrorConfig$ | async)"
```
undsoft, line 24: It would be good to find a way to reduce boilerplate for having those configs in component file as well.

src/app/pages/account/groups/group-list/group-list.component.spec.ts
```diff
@@ -115,10 +115,10 @@ describe('GroupListComponent', () => {
       expect(snapshots).toEqual([]);
     });
 
-    const table = await loader.getHarness(IxTableHarness);
-    const text = await table.getCellTextByIndex();
+    spectator.detectChanges();
 
-    expect(text).toEqual([['Can not retrieve response']]);
+    const emptyTitle = spectator.query('.empty-title');
```
undsoft, line 120: Would be good to have a harness here, so that `.empty-title` class is hidden away.

undsoft (changes requested):
1. VMs page gives me an error.
2. I don't know if you meant to show "No records" when filtering the table, but currently nothing is shown (which is okay too).

RehanY147:
@undsoft The vms page issue isn't exclusive to this PR. It's on `master`.

src/app/modules/ix-tables/services/empty.service.ts
```diff
@@ -23,6 +25,12 @@ export class EmptyService {
     title: this.translate.instant('Can not retrieve response'),
   };
 
+  private readonly emptySearchResultsConfig: EmptyConfig = {
+    title: this.translate.instant('No such items'),
```
undsoft, line 29: `No such items` will get extracted into translation files and people will translate it, though it is not used.

src/app/modules/ix-tables/services/empty.service.ts
```diff
@@ -31,9 +39,16 @@ export class EmptyService {
         return this.loadingConfig;
       case EmptyType.Errors:
         return this.errorConfig;
+      case EmptyType.NoSearchResults:
+        return {
+          ...this.emptySearchResultsConfig,
+          title: this.translate.instant('No matching {title} found', { title }),
```
undsoft, line 45: Here and below we should just stick with generic message, similar to `No matching records found`.
This would work for English, but may not work in other languages, because word in `{title}` may need to be of different form when inside of `No matching {title} found` sentence.

undsoft (approved):
Okay for now, additional improvements to come later when we hopefully come up with better tables.

PR #7356, opened 2022-11-21
NAS-119081 / 23.10 / Do not disallow failover when system versions mismatch

AlexKarpov98:


src/app/pages/dashboard/components/widget-sys-info/widget-sys-info.component.html
```diff
@@ -247,11 +288,21 @@ <h3>
                   <mat-list-item *ngIf="uptimeString">
                     <strong>{{ 'Uptime' | translate }}:</strong>
                     <span>
-                      {{ '{uptimeString} as of {dateTime}' | translate: { uptimeString, dateTime } }}
+                      {{
+                        '{uptimeString} as of {dateTime}'
```
undsoft, line 292: I kinda like pipes on the same line. Same goes for ternary conditions.

src/app/pages/dashboard/components/widget-sys-info/widget-sys-info.component.html
```diff
@@ -247,11 +288,21 @@ <h3>
...
-                  <mat-list-item *ngIf="showTimeDiffWarning" class="time-warning">
-                    <mat-icon class="icon-warning" [matTooltip]="timeDiffWarning">error</mat-icon>
+                  <mat-list-item
+                    *ngIf="showTimeDiffWarning"
+                    class="time-warning"
+                  >
+                    <mat-icon
+                      class="icon-warning"
+                      [matTooltip]="timeDiffWarning"
+                      >error</mat-icon
```
undsoft, line 304: Indentation is off.

undsoft (commented):
Don't have a machine with update to test. Will test once machine is available.

src/app/pages/dashboard/components/widget-sys-info/widget-sys-info.component.html
AlexKarpov98: sorry, had prettier ON

undsoft:
The issue here is not just with widget on main dashboard, but also with `checkUpgradePending` in `topbar.component.ts`.
When you update one of the controllers and failover, it is supposed to call `failover.upgrade_pending` to finish the update.
However this call is behind a check for failover disabled reasons. You will get `MismatchVersions` and `failover.upgrade_pending` will never get called.

undsoft (changes requested):
There is a problem now:
1. I start the update on controller A.
2. Controller B is being updated and rebooted.
3. Once controller B is rebooted, UI will receive new HA status and for some time will show a dialog about updating second controller even before previous dialog goes away.
4. The rest works correctly including showing dialog after rebooting first controller.

AlexKarpov98:
@undsoft  - can you check again please?

src/app/store/ha-info/ha-info.effects.ts
```diff
@@ -42,7 +43,10 @@ export class HaInfoEffects {
   loadUpgradePendingState = createEffect(() => this.actions$.pipe(
     ofType(haStatusLoaded),
     mergeMap(({ haStatus }) => {
-      if (haStatus.hasHa && haStatus.reasons.length === 0) {
+      if (
+        (haStatus.hasHa && haStatus.reasons.length === 0)
+        || (haStatus.reasons?.includes(FailoverDisabledReason.MismatchVersions))
```
undsoft, line 48: The check is not correct. We only want to allow this proceed when MismatchVersion is the only disabled reason, not one of them.

AlexKarpov98: gonna fix

undsoft (changes requested):
Still doesn't work. You can connect to our m40 to reproduce the issue.

undsoft (approved):
Was able to finish update with some hiccups.

PR #7359, opened 2022-11-21
NAS-119058 / 22.12 / UPS Service "Port or Hostname" logic fix

AlexKarpov98:
in v22.02.4 
Custom value was allowed in the input

did the same for Scale, check it visually

Added new prop for ix-combobox
**[allowCustomValues] ==> boolean**

telegram-cloud-photo-size-2-5467726597873321187-y


src/app/modules/ix-forms/components/ix-combobox/ix-combobox.component.ts
```diff
@@ -33,18 +33,19 @@ export class IxComboboxComponent implements ControlValueAccessor, OnInit {
   @Input() hint: string;
   @Input() required: boolean;
   @Input() tooltip: string;
-  options: Option[] = [];
+  @Input() allowCustomValues = false;
```
undsoft, line 36: `allowCustomValue`

undsoft (changes requested):
1. Placeholder in input says 'Search'. Let's change it to 'Search or enter value'
2. Once you add custom value, save the form and edit this field, custom value will be shown in autocomplete options. I would not expect to see it there.
3. Harness for ix-combobox needs to be updated.
4. Unit tests for ix-combobox need a case for entering custom value.
5. It's a good idea to add unit test covering this to UPS form as well.

src/app/modules/ix-forms/components/ix-combobox/ix-combobox.harness.ts
```diff
@@ -42,6 +44,12 @@ export class IxComboboxHarness extends ComponentHarness implements IxFormControl
     await harness.selectOption({ text: optionLabel });
   }
 
+  async writeCustomValue(text: string): Promise<void> {
+    const input = await this.getMatInputHarness();
+    await input.setValue(text);
+    return (await input.host()).dispatchEvent('change');
```
undsoft, line 50: I don't see why we need to dispatch `change` here. It works without it, and ix-combobox doesn't seem to use it.

src/app/modules/ix-forms/components/ix-combobox/ix-combobox.component.ts
```diff
@@ -124,13 +124,13 @@ export class IxComboboxComponent implements ControlValueAccessor, OnInit {
            * if we haven't found the correct option in the list of options fetched so far. The assumption
            * is that the correct option exists in one of the following pages of list of options
            */
-          this.options.push({ label: this.value as string, value: this.value });
-          this.selectedOption = { ...this.options.find((option: Option) => option.value === this.value) };
-          if (this.selectedOption) {
+          this.selectedOption = { label: this.value as string, value: this.value };
```
undsoft, line 127: I believe there was a reason why this was added. Something about having multiple pages of suggestions and having currently selected option on the page 3.
You can check git blame for ticket.
I don't think this functionality needs to be removed for non-custom values.

undsoft (changes requested):

src/app/modules/ix-forms/components/ix-combobox/ix-combobox.component.spec.ts
```diff
@@ -68,10 +68,23 @@ describe('IxComboboxComponent', () => {
...
       expect(spectator.query('input')).toHaveValue('new value');
     });
 
+    it('form control value is set to custom value if [allowCustomValue] enabled', () => {
+      spectator.setInput('allowCustomValue', true);
+      spectator.typeInElement('/my-custom-1', 'input');
+      expect(formControl.value).toBe('/my-custom-1');
+    });
+
+    it('form control value is set to null if [allowCustomValue] disabled', () => {
```
undsoft, line 82: if [allowCustomValue] is disabled and user types custom value.

src/app/pages/services/components/service-ups/service-ups.component.spec.ts
```diff
@@ -142,4 +143,22 @@ describe('ServiceUpsComponent', () => {
...
+    const portSelect = await loader.getHarness(IxComboboxHarness.with({ label: 'Port or Hostname' }));
+
+    await portSelect.writeCustomValue('/my-custom-port');
+
+    const portSelectValue = await portSelect.getValue();
+
+    const saveButton = await loader.getHarness(MatButtonHarness.with({ text: 'Save' }));
+    await saveButton.click();
+
+    const formValue = await form.getValues();
```
undsoft, line 159: This still is not very different from what you had before.
Should be something like:
```
// Fill form
// Press save
// expect(ws.call).toHaveBeenCalledWith('ups.update', [expect.objectContaining({
  port: '/my-custom-port',
}]);
```

src/app/modules/ix-forms/components/ix-combobox/ix-combobox.component.ts
```diff
@@ -116,21 +118,25 @@ export class IxComboboxComponent implements ControlValueAccessor, OnInit {
...
           /**
            * We are adding a custom fake option here so we can show the current value of the control even
            * if we haven't found the correct option in the list of options fetched so far. The assumption
            * is that the correct option exists in one of the following pages of list of options
            */
-          this.options.push({ label: this.value as string, value: this.value });
-          this.selectedOption = { ...this.options.find((option: Option) => option.value === this.value) };
-          if (this.selectedOption) {
-            this.filterChanged$.next('');
+          if (!isCalledOnInit) {
```
undsoft, line 129: Why not just `if(!this.allowCustomValue) {`? It is much easier and the end result seems like what we need.

undsoft (approved):
The field works okay, but there are some code comments.

PR #7362, opened 2022-11-22
NAS-119039 / 22.12 / Nas 119039 - Fixed breadcrumb on Pod Shell/Logs applications

Raz-Dva:
Go to Apps -> Installed Applications.
Click on the three dots menu in the right corner of the application card. Choose Shell or Logs.
Use breadcrumb to return to Applications in the top left. 

image


src/app/modules/page-header/page-header.module.ts
```diff
@@ -3,6 +3,8 @@ import { NgModule } from '@angular/core';
 import { FlexModule } from '@angular/flex-layout';
 import { RouterModule } from '@angular/router';
 import { TranslateModule } from '@ngx-translate/core';
+import { CommonDirectivesModule } from 'app/directives/common/common-directives.module';
```
AlexKarpov98, line 6: why do we need `CommonDirectivesModule`?

src/app/modules/page-header/page-header.module.ts
```diff
@@ -11,11 +13,14 @@ import { PageTitleHeaderComponent } from 'app/modules/page-header/page-title-hea
...
     RouterModule,
     TranslateModule,
+    CommonDirectivesModule,
   ],
   declarations: [
+    BreadcrumbComponent,
     PageTitleHeaderComponent,
   ],
   exports: [
+    BreadcrumbComponent,
```
AlexKarpov98, line 23: do we really need to export `BreadcrumbComponent`?

AlexKarpov98 (approved):
This is working, answer questions please as well.

undsoft (changes requested):
It's weird that breadcrumb component which is inside of `layout` folder is not part of `layout` module.
I would try to remove breadcrumb module. If that is not possible, you need to move the component to a different folder.

src/app/modules/page-header/page-header.module.ts
Raz-Dva: We are using the PageHeaderModule to import into the LayoutModule and EntityModule and in this case we need to migrate the BreadcrumbComponent to PageHeaderModule otherwise it will not be available in the PageTitleHeaderComponen template

undsoft (approved):
Looks good.

PR #7363, opened 2022-11-22
NAS-119103 / 22.12 / Highlight degraded vdevs and other minor fixes

undsoft:
Changes:
* Degraded pools are now shown in orange.
* Hit area on back button in mobile details increased.
* In some places WidgetUtils replaced with filesize.


undsoft:
@dszidi Can't say that orange looks particularly great, but so far nothing looks too good.

dszidi (commented):
@undsoft ^^

undsoft:
Ok, fn-theme-orange still has white text, but fn-theme-yellow looks okaish.

dszidi (commented):
We should stick with orange IMO because that is the status color we're using in other places in the UI. (Storage Dashboard, data protection task status etc)

We should be able to change the text color in `.fn-theme-orange` by modifying the `orange-txt` css var in `src/assets/styles/scss-imports/cssvars.scss`

AlexKarpov98 (commented):
Code looks good to me (UI not checked yet)

denysbutenko (approved):
image

Looks good to me

undsoft:
@dszidi I think I'm going to keep it yellow for now. Other changes seem to be more intrusive and we can tweak it later as needed.

PR #7365, opened 2022-11-22
NAS-119062 / 22.12 / Refresh storage screen after successful import of a pool

bvasilenko:
Testing: see video in the ticket


dszidi (approved):
Page successfully refreshes and displays all pools after pool import

PR #7366, opened 2022-11-22
NAS-119098 / 22.12 / Remove inline styles on body

dszidi:
Inline styles on body was causing problems on non default themes.

To test, switch to various themes and make sure the correct --bg1 color value is being applied to `<body>`.
The most obvious will be the lighter themes like "Paper" and "iX Blue"


src/index.html
```diff
@@ -26,7 +26,7 @@
     }
   </style>
 </head>
-<body class="ix-dark" style="--bg1: #1e1e1e; background-color: var(--bg1)">
```
denysbutenko, line 29: It was added for initial loading until styles and scripts were loaded. Perhaps we may need to find a less hacky approach to achieve this result or just remove it.

The difference could be shown in this video, `localhost:8080` is `NAS-119098` changes, and `localhost:4200` is `master` branch:

https://user-images.githubusercontent.com/351613/203682727-83679950-86e4-4977-a8b9-36384b68b52e.mp4

undsoft (changes requested):
Yeah, it would be cool to keep the dark background for initial load and that flashing loading indicator.
We could apply theme styles to body or some container tag once theme has been loaded.

dszidi:
I think the issue here might be that the `--bg1` is being declared on the body element whereas the theme puts it on `:root {}`. It would explain why bg1 is never right when you change themes. Let me try setting --bg1 another way.

undsoft (approved):
This works, thanks.

PR #7368, opened 2022-11-22
Using `path_local` where applicable

RehanY147:
Test that there are no redirects to `DatasetAclEditorComponent` that don't pass the full path (including `/mnt/`) in the query params. Also, whenever editing acls of shares, the value used for full path should be the `path_local` attribute now passed by middleware.


anodos325 (commented):
I tested against my home cluster and was able to edit the filesystem ACL under my SMB share with these changes.

src/app/pages/datasets/modules/permissions/containers/permissions-card/permissions-card.component.html
```diff
@@ -2,7 +2,7 @@
   <mat-card-header>
     <h3 mat-card-title>{{ 'Permissions' | translate }}</h3>
 
-    <a *ngIf="canEditPermissions" mat-button [routerLink]="editPermissionsUrl">
+    <a *ngIf="canEditPermissions" mat-button (click)="redirectToEditPermissions()">
```
undsoft, line 5: Previous way was a bit better, because you could open link in new tab.

undsoft (changes requested):
1. If you go to an ACL editor, there is sometimes a `Permissions Editor` button visible there. It now has broken url.
2. When ACL is being saved, let's also change `Updating Dataset ACL` text to `Updating ACL`
~3. Strip ACL button appears to be not working correctly. I believe it used to take user back after acl was stripped. It doesn't seem to be doing anything.~

AlexKarpov98 (commented):
code changes looks fine to me.

AlexKarpov98 (approved):
code changes looks fine to me.

undsoft:
> 3. Strip ACL button appears to be not working correctly. I believe it used to take user back after acl was stripped. It doesn't seem to be doing anything.
This part can be skipped as it is broken in master too. I'll fix it as part of another ticket.

src/app/pages/datasets/components/dataset-form/dataset-form.component.ts
```diff
@@ -1679,7 +1679,11 @@ export class DatasetFormComponent implements FormConfiguration {
                 if (confirmed) {
                   this.ws.call('filesystem.getacl', [parentPath]).pipe(untilDestroyed(this)).subscribe({
                     next: () => {
-                      this.router.navigate(['/', 'datasets', restPostResp.name, 'permissions', 'acl']);
+                      this.router.navigate(['/', 'datasets', 'acl', 'edit'], {
```
undsoft, line 1682: This will fail if one of my datasets is named `mnt`.

PR #7369, opened 2022-11-23
NAS-119060 / 22.12 / Installed apps not showing

AlexKarpov98:


src/app/pages/applications/chart-releases/chart-releases.component.ts
```diff
@@ -217,16 +217,27 @@ export class ChartReleasesComponent implements AfterViewInit, OnInit, OnDestroy
...
 
-              charts.forEach((chart) => {
-                chart.selected = false;
-                this.chartItems.set(chart.name, chart);
+                this.filterChartItems();
               });
-
-              this.filterChartItems();
-            });
+            } catch {
```
RehanY147, line 238: Try/catch looks redundant since we're catching the error now.

src/app/pages/applications/chart-releases/chart-releases.component.ts
```diff
@@ -217,16 +217,27 @@ export class ChartReleasesComponent implements AfterViewInit, OnInit, OnDestroy
...
             this.showLoadStatus(EmptyType.Errors);
           } else {
-            this.appService.getChartReleases().pipe(untilDestroyed(this)).subscribe((charts) => {
-              this.chartItems.clear();
+            try {
+              this.appService.getChartReleases().pipe(
+                catchError(() => of(this.showLoadStatus(EmptyType.Errors))),
+                untilDestroyed(this),
+              ).subscribe((charts) => {
+                if (!charts) {
```
RehanY147, line 225: This could also just mean there are no charts to show instead of having an error.

RehanY147 (changes requested):

src/app/pages/applications/chart-releases/chart-releases.component.ts
AlexKarpov98: if charts will be an empty array, that won't cause issues 
If it's null or undefined -> then it's error probably?

src/app/pages/applications/chart-releases/chart-releases.component.ts
AlexKarpov98: yeah, I put it to catch manual **throw new Error()**, will remove

undsoft (approved):
Error status is being shown.

PR #7373, opened 2022-11-24
NAS-119051 / 22.12 / Fix `as any` casts

bvasilenko:
**Summary**

1st part of this ticket is a problem on middleware side.

2nd part is removing `any` casts, and it's done here:


undsoft:
Soo, we're saying that we expect `passphrase2` field to be present in `certificate.query`?
That's kinda weird, let's wait for middleware confirmation that they expect those fields there before merging.

AlexKarpov98 (approved):
`as any` cast is removed now

PR #7374, opened 2022-11-24
NAS-119102 / 22.12 / Allow specifying remote NAS administrator username

bvasilenko:
**Testing**

1. On **Credentials -> Backup Credentials** page, 

2. Click **SSH Connections -> Add** to open **New SSH Connection** form, 

3. At this form, please fill in the new text input field **Admin username**

Expected result: 
- by default field contains `root`
- when filled field's value is sent to middleware


undsoft (approved):
The field is there and it is being sent.

I think it could use a better tooltip, it is not clear to me as a user as to 1) what the difference between `username` and `admin username` fields, or 2) why `admin username` is needed to be specified for the connection.

@themylogin Can you provide an insight into this?

PR #7376, opened 2022-11-24
NAS-119042 / 22.12 / Fixing various inconsistencies with ACL editor

undsoft:
Fixes:
* Stripping ACL would not redirect user back to datasets.
* ACL editor would remember last opened ACL and therefore Select Preset ACL prompt was shown inconsistently.


undsoft:
@AlexKarpov98 I remember that there was a reason why `initialState` was replaced here: https://github.com/truenas/webui/pull/7199, but even with `initialState` it appears to work okay and I cannot reproduce the original issue in that ticket.

denysbutenko (approved):
Looks good

AlexKarpov98 (approved):
👍

PR #7378, opened 2022-11-25
NAS-118993 / 23.10 / Update WebUI for quota API changes

bvasilenko:
**Summary**

The change relies on the PR https://github.com/truenas/middleware/pull/10073 . Make sure you have latest nightly build installed on your box before testing.

**Testing**

1. Go to **Datasets > Choose a dataset > Manage Quota**
2. For each quota in the list, make sure that
   - usage percentage is calculated and displayed correctly
   - N/A is displayed when quota has undefined amount


src/app/pages/datasets/components/dataset-quotas/dataset-quotas-userlist/dataset-quotas-userlist.component.ts
```diff
@@ -116,7 +116,12 @@ export class DatasetQuotasUserlistComponent implements OnInit, AfterViewInit, On
       case 'used_percent':
         return `${Math.round(row[field] * 100) / 100}%`;
       case 'obj_used_percent':
-        return `${Math.round(row[field] * 100) / 100}%`;
+        if (row.obj_quota) {
+          return `${Math.round(row.obj_used / row.obj_quota * 100) / 100}%`;
+        }
+        return this.translate.instant('N/A');
```
undsoft, line 122: I would show dash `–` instead of N/A. N/A doesn't immediately sounds like no quota, and I may instead think that quota was not loaded or something.

undsoft (commented):
Didn't do proper testing, as I don't see a machine with backend changes available.

However, looking at quota forms - there is this behaviour that when you set quota to 0 it is treated as removing a quota. Is this still true?

undsoft (changes requested):
On latest nightlies there is an error when opening quota list.

src/app/pages/datasets/components/dataset-quotas/dataset-quotas-userlist/dataset-quotas-userlist.component.ts
bvasilenko: if quota is set as 0 then it will remove the quota

bvasilenko:
Error is fixed now

undsoft (changes requested):
Using TrueNAS-SCALE-23.10-MASTER-20221207-015232

https://user-images.githubusercontent.com/1481270/206194563-f2f3d723-e4bd-4ae4-8be8-3a1b146fb4ea.mov

bvasilenko:
I verified this with the updated backend.
Everything works correctly now, no more changes needed

undsoft (changes requested):
I don't see any changes when I test on latest nightly in Jenkins.

bvasilenko:
> I don't see any changes when I test on latest nightly in Jenkins.

I see, it got broken again. Now I've pushed the last missing part, and now it should be fine

undsoft (changes requested):
1. If I have an existing data quota and then I edit it and press (x) in the Data field, I will not be able to save it.
I will still be able to save if I manually type 0 in the field.
2. There seems to be a confirmation box that is shown when you set both quotas to 0 (effectively deleting a quota).
This confirmation is inconsistent and is not shown when you use (x) to clear fields.

undsoft (approved):
Looks good 👍🏻

PR #7379, opened 2022-11-25
NAS-119096 / 22.12 / Set `Certificate` to null when `Enable TLS` is false

bvasilenko:
**Testing**

1. Go to **System Settings > Services**
2. Click **Configure** button next to **FTP** service, then click **Advanced settings**
4. Tick **Enable TLS**, choose a certificate and click **Save**
5. Return to advanced settings, and untick the **Enable TLS** and click **Save**
- Expected result:  selected certificate submitted as `null` to middleware, erasing previously selected certificate


undsoft (approved):
Specific scenario described in PR works okay.

PR #7386, opened 2022-11-28
NAS-118367 / 23.10 / Add Tree Virtual Scroll View

denysbutenko:
Work in progress.

Known issues:

- [x] Initial dataset loading is broken
- [x] Indentation needs to be fixed
- [x] DataSource changes do not trigger ChangeDetection in the virtual scroll view.
- [x] The `ix-dataset-node` needs a dataset object
- [x] Fix change detection on filtering
- [x] Unresolved `any`

~Avoid `ix-` prefix in file names.~ Not in this PR.

To test virtual scrolling, you can use a host from the ticket and visit the datasets page.


src/app/modules/ix-tree/ix-tree-flattener.ts
```diff
@@ -0,0 +1,73 @@
+import { TreeControl } from '@angular/cdk/tree';
+import { Observable, take } from 'rxjs';
+
+export class IxTreeFlattener<T, F, K = F> {
```
undsoft, line 4: Generally speaking we've used `ix` prefixes to distinguish new things from things that may already be in the system with the same name. 
No need to use them for things that don't have existing equivalents.

undsoft (commented):
Code looks okay so far, but the functionality is broken for me on datasets.

src/app/modules/ix-tree/ix-tree-flattener.ts
denysbutenko: Yeah makes sense. I'll revisit it before merging.

denysbutenko:
> Code looks okay so far, but the functionality is broken for me on datasets.


https://user-images.githubusercontent.com/351613/204704567-b1748e70-dad5-4ee5-8ee9-2e6b55bca2de.mp4

undsoft (changes requested):
Generally looks good, but here some of the issues I've noticed:
1. Sometimes dataset view doesn't occupy the whole available space.
Снимок экрана 2022-12-06 в 14 35 38

2. When I scroll down the list sometimes jumps.

https://user-images.githubusercontent.com/1481270/205927188-4e2c4849-4f5d-433e-96e4-d8c7a57610a0.mov

3. Is dragging using scrollbar supposed to work? I mean, it does scroll, but user experience is not always great.

4. And probably the biggest one - is that it kinda breaks the whole page scrolling that we had reducing usable space significantly. Would be good to keep that if possible, like on Github perhaps when you view PR files.
If this is too complicated, we could consider redesigning the page.

denysbutenko:
> Generally looks good, but here some of the issues I've noticed:
> 
> 1. Sometimes dataset view doesn't occupy the whole available space.
> 
> Снимок экрана 2022-12-06 в 14 35 38
> 
> 2. When I scroll down the list sometimes jumps.
> 
>  2022-12-06.14.37.10.mov 
> 3. Is dragging using scrollbar supposed to work? I mean, it does scroll, but user experience is not always great.
> 4. And probably the biggest one - is that it kinda breaks the whole page scrolling that we had reducing usable space significantly. Would be good to keep that if possible, like on Github perhaps when you view PR files.
>    If this is too complicated, we could consider redesigning the page.

The first and fourth points are related to styling and should be easy to fix. The second and third points are performance issues, so using CdkPortal to inject the ix-dataset-node component on the fly might help improve things. I planned to make performance optimizations in a [separate ticket](https://ixsystems.atlassian.net/browse/NAS-119276).

src/app/modules/ix-tree/ix-tree-flattener.ts
denysbutenko: Reference to the ticket - https://ixsystems.atlassian.net/browse/NAS-119275

denysbutenko:
I just made some updates:

- Fixed the node height to match the defined. It causes additional scrolling issues.
- Added a button to scroll to the top of the tree

I'm still improving the page scrolling and rendering for dataset-node content.

denysbutenko:
> 1. Sometimes dataset view doesn't occupy the whole available space.

It has been fixed.

> 2. When I scroll down the list sometimes jumps.

Yeah, I see it. I couldn't identify a reason yet, keeps observation.

> 3. Is dragging using scrollbar supposed to work? I mean, it does scroll, but user experience is not always great.

I saw the same behavior on the official material demos. Perhaps, it is by design.

> 4. And probably the biggest one - is that it kinda breaks the whole page scrolling that we had reducing usable space significantly. Would be good to keep that if possible, like on Github perhaps when you view PR files.
> If this is too complicated, we could consider redesigning the page.

Yes, that's a good point. Unfortunately, two scrolls do not work together. I changed the layout  (at least for demo purposes) to fill the available height for the tree.


Recent changes:

- Fixed more cases when ix-dataset-node has a greater height than expected which causes scrolling issues.
- Injecting the `ix-dataset-node` component using `cdkPortalOutlet`
- Use `DEFAULT_SCROLL_TIME` and `DEFAULT_RESIZE_TIME` constants for debounce in scrolling/resizing

denysbutenko:
> The tree is broken after merging the master branch. Looking into this 👀

The issue is fixed. @undsoft

denysbutenko:
New TODO:
- [ ] Adapt flat API response to existing code
- [ ] Improve UX on scrolling (reduce jumping and unnecessary re-renders)

denysbutenko:
The code is irrelevant to the flat data structure and can't be applicable anymore.

undsoft (changes requested):
It still jumps around. It feels like it first correctly scrolls, but then adds too many elements on top or something like that, which causes items to jump down.
It's easier to see if you:
1. Scroll down in the middle of the list.
2. Put finger in the middle of the tree view.
3. Focus on one of the datasets on top of the list and quickly scroll it down, so that it's on the same level as your finger.
4. Observe that datasets jump after you finish scrolling like too many items were added on top. As a user I'd rather see updates less often (more gray), but with things where I expect them.

In the video you can see items jumping twice.


https://user-images.githubusercontent.com/1481270/208694749-a3c54f78-bcb5-4ff7-bc95-c00b668b1bd1.MOV

denysbutenko:
> It still jumps around. It feels like it first correctly scrolls, but then adds too many elements on top or something like that, which causes items to jump down. It's easier to see if you:
> 
> 1. Scroll down in the middle of the list.
> 2. Put finger in the middle of the tree view.
> 3. Focus on one of the datasets on top of the list and quickly scroll it down, so that it's on the same level as your finger.
> 4. Observe that datasets jump after you finish scrolling like too many items were added on top. As a user I'd rather see updates less often (more gray), but with things where I expect them.
> 
> In the video you can see items jumping twice.
> 
>  IMG_0817.MOV

Yeah, I can reproduce this. While scrolling, `virtualScrollView.renderedRangeStream` emits multiple values, some of which are wrong. I am looking for a way to filter out wrong values or get a cause for why that happens.

denysbutenko:
It seems the most annoying scrolling glitches are gone.

https://user-images.githubusercontent.com/351613/209055149-70704eae-0dbc-4295-92d8-418cad483d77.mp4

undsoft (changes requested):
Yes, scrolling is much better (though I still had an occasional jump once or twice). It's good enough to keep it this way for now.

Let's finish remaining tasks and get this merged.
Collapsing/expanding doesn't work for me (don't know if it should) and "go up" arrow could use some background.

There is also an excessive amount of re-rendring that is being done, but we can solve it later. I've created another ticket for it https://ixsystems.atlassian.net/browse/NAS-119576

denysbutenko:
New TODO:

- [ ] Complete changes for the flat data structure [NAS-119467](https://ixsystems.atlassian.net/browse/NAS-119467)

denysbutenko:
@undsoft 

> Yes, scrolling is much better (though I still had an occasional jump once or twice). It's good enough to keep it this way for now.

Ok, let's keep observations.

> Let's finish remaining tasks and get this merged. Collapsing/expanding doesn't work for me (don't know if it should) and "go up" arrow could use some background.

Fixed the issue with collapsing/expanding nodes. Also, checking for other regressions [Known: filtering issue].


> There is also an excessive amount of re-rendring that is being done, but we can solve it later. I've created another ticket for it https://ixsystems.atlassian.net/browse/NAS-119576


Re-rendering would be a good ticket for improvements. I did a quick test, and it seems `cdk-portal-outlet` approach does not give any performance benefits over a trivial approach in this case. So, I switched it back for now.

undsoft (approved):
Generally looks good enough to merge.
Some additional issues that we may want to fix either now or in a separate ticket:
1. Scroll up icon shouldn't appear when the list is relatively short. It should only appear if I have scrolled considerable distance.
2. If there is a dataset with a long name, columns may jump when I scroll. This is kinda okay, though I'd prefer them to stay the same if the list is not too long.
Even if column width change, table header should also get updated.

https://user-images.githubusercontent.com/1481270/209977858-b2f6d632-da35-4da5-ac6a-0ba2d37978ae.mov

denysbutenko:
backport

denysbutenko:
backport

PR #7388, opened 2022-11-28
NAS-118439 / 13.0 / Cast `vlan` value ti integer or null

bvasilenko:
**Testing**

You should test with an M60 box, for example m60-100.dc1.ixsystems.net

On **Network** page, at **IPMI** block:

1. Click an item from IPMI block to open it for editing
1. Set **VLAN ID** to any integer
1. Save configuration
1. Erase **VLAN ID** integer
1. Save configuration
   - Expected result:  VLAN ID should be erased without an error "Is not integer value" mentioned in the ticket

### DIFF
--- a/src/app/pages/network/ipmi/ipmi.component.ts
+++ b/src/app/pages/network/ipmi/ipmi.component.ts
@@ -254,6 +254,7 @@ export class IPMIComponent {
   }
 
   customSubmit(payload) {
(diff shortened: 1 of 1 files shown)


denysbutenko:
It seems the `M60` is unavailable. Attempted to use core cloud VM, and there are no mentions of IPMI in Network pages. I'll do another round review later.

denysbutenko (approved):
The changes looks okay

PR #7390, opened 2022-11-28
NAS-119167 / 23.10 / Module preloading

undsoft:
Enables this https://angular.io/guide/lazy-loading-ngmodules#preloading-modules for better UX after login.


RehanY147 (approved):
Looks good

PR #7394, opened 2022-11-29
NAS-118754 - Fixed dashboard missing widgets when vlan is configured

Raz-Dva:
Go to Network -> Interfaces.
Create new VLAN.
Return to dashboard and check widgets.


denysbutenko (approved):
The vlan interface appears on the main dashboard

image

undsoft:
Confirmed fix using user's data.

PR #7396, opened 2022-11-29
NAS-119185 / 22.12.1 / Adding loading indicator to permissions editor

undsoft:
Also centers logo.

### DIFF
--- a/src/app/modules/layout/components/admin-layout/admin-layout.component.scss
+++ b/src/app/modules/layout/components/admin-layout/admin-layout.component.scss
@@ -36,6 +36,7 @@
   line-height: 64px;
   max-width: var(--sidenav-width); /*15rem;*/
   overflow: hidden;
+  padding-right: 6px;
   position: fixed;
   text-align: center;
   top: 0;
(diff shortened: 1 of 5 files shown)


src/app/modules/layout/components/admin-layout/admin-layout.component.scss
```diff
@@ -36,6 +36,7 @@
   line-height: 64px;
   max-width: var(--sidenav-width); /*15rem;*/
   overflow: hidden;
+  padding-right: 6px;
```
dszidi, line 39: No need for padding as logo is already centered by way of the `justify-content: center` rule.

image

dszidi (commented):
Other than the css issue, code looks good and loader works well.

image

src/app/modules/layout/components/admin-layout/admin-layout.component.scss
undsoft: It's centered programatically, not visually (easier to see on light themes).

dszidi: I see what you mean. If we want to correct this, we should also consider that the logo appears elsewhere in the UI as well. I'll approve and leave it up to you whether we want to include this here or do something more global on another ticket

dszidi (approved):
LGTM

undsoft:
Ok, I'll merge as is and then we can improve on it later.

PR #7400, opened 2022-11-30
NAS-110800 / 23.10 / Poor communication of what elements can be interacted with

AlexKarpov98:
Didn't catch any problem with having `:pointer` issues.
But catched UI problems, and fixed them 
- VM list was broken (white screen)
- Disks search input was not aligned and looked bad


undsoft:
It's better not to re-use tickets with completely unrelated fixes.
Please recreate using https://ixsystems.atlassian.net/browse/NAS-119205 and targeting different branches.

PR #7404, opened 2022-12-01
NAS-119210 / 23.10 / It's truenas.com, not truenas.org

themylogin:


undsoft (approved):
Weird that it used to work.
FYI 13.0 still uses freenas.org

themylogin:
@undsoft I broke it my previous PR, replaced "freenas" with "truenas" but forgot to replace "org" with "com" :(

PR #7410, opened 2022-12-01
NAS-116966 / 23.10 / Hide `togglesmart` and `smartoptions` for unsupported SMART

bvasilenko:
**Testing**

On **Storage > Disks** page,

Expand any of the disks, there is going to be a checkbox to **Enable SMART**. 

- **Actual result:** The checkbox will be there even if device doesn't support SMART as returned by `smart.test.disk_choices`


denysbutenko (approved):
Looks good

PR #7420, opened 2022-12-02
NAS-118879 / 23.10 / New Google Drive cloud sync task field: acknowledge_abuse

AlexKarpov98:
Current look:
Screen Shot 2022-12-02 at 17 31 24


bvasilenko (approved):
Works as expected:

image

image

PR #7425, opened 2022-12-05
NAS-119262 / 23.10 / Update setLanguageFromBrowser logic to get data from LocalStorage if set WS is not connected.

AlexKarpov98:
Now selected language correctly appears on all pages!


denysbutenko (changes requested):
The language key won't appear in local storage until the form is submitted. Perhaps we could save the language key when general settings are loaded.

denysbutenko (approved):
Looks good 👍🏻

PR #7443, opened 2022-12-07
NAS-117481 / 23.10 / Refactoring sign-in form

undsoft:
Testing: everything in sign-in form, including failover, setting root password, autocomplete, OTP password.


bvasilenko (changes requested):
@undsoft 

Noticed that the call `system.product_type` is not being executed. This call inserts the product name in the copyright text, which looks missing

image

bvasilenko (approved):
Ok, now it looks good to me

PR #7444, opened 2022-12-08
NAS-119277 / 23.10 / Add dump currently active sessions

bvasilenko:
**Testing**

Go to **System > Advanced Settings** page and check the newly added block


RehanY147 (changes requested):
Unable to compile

bvasilenko:
@RehanY147 Fixed build errors, and merged conflicts, so now it should be good to review

RehanY147 (changes requested):
image
Getting this error

bvasilenko:
@RehanY147 I've pushed potential fix, please try again

RehanY147 (changes requested):
Clicking "terminate other sessions" doesn't remove all the session entries from the table. It removes a couple sometimes. Sometimes removes less or more. But after clicking that button, there are 3-4 session entries present on the table. Also no loading state for the table after you click the button. No indication that something is happening. Also, the "View more" button and "collapse" button are not showing up correctly. There is a scenario possible where the button is showing but clicking and toggling between the two states does nothing on the table.

bvasilenko:
Blocked by middleware

bvasilenko:
@RehanY147 Now the issue is fixed

RehanY147 (changes requested):
If I open the UI on two different browsers and sign in, then choose to terminate sessions from one of them, the other one reloads but also gets logged in automatically without needing credentials. Don't think that's supposed to happen.

bvasilenko:
@RehanY147 

Middleware got the bug fixed, so now it should be testable

RehanY147 (changes requested):
Not working properly. I click the "terminate other sessions" button and it removes all items from the table except one. Then if I log in on a different browser and reload the original page, I see a bunch of new items when I should only be seeing two.

bvasilenko:
Cannot test, because of a duplicate of the `WebSocketService` service has been added in 23.10, and two connections are opened at once, respectively, two sessions are created. Once this is fixed, this PR will be unblocked

RehanY147 (approved):

bvasilenko:
We're merging this, keeping in mind that to function properly it needs the 2nd `WebSocketService` to be removed in future

PR #7448, opened 2022-12-08
NAS-119297 / 23.10 / Enabling no-interpolation-in-attributes 1

undsoft:
Code review is enough.

### DIFF
--- a/src/app/pages/storage/components/manager/manager.component.html
+++ b/src/app/pages/storage/components/manager/manager.component.html
@@ -69,7 +69,7 @@
                 >
                   <mat-option
                     *ngFor="let option of encryptionAlgorithmOptions"
-                    id="encryption_algorithm_option_{{ option.label }}"
+                    [id]="'encryption_algorithm_option_' + option.label"
                     [value]="option.value"
                   >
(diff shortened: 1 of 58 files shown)


src/app/pages/storage/components/manager/manager.component.html
```diff
@@ -515,7 +515,7 @@ <h4 id="avail-disks-title">{{ 'Available Disks' | translate }}</h4>
                   <mat-checkbox
                     *ngIf="disks[rowIndex]"
                     color="primary"
-                    id="pool-manager__disks-{{ disks[rowIndex].devname }}"
+                    [id]="'pool-manager__disks-' + disks[rowIndex].devname"
```
AlexKarpov98, line 518: `[id]` should go after `name=""`

AlexKarpov98 (approved):
Good job, only 1 thing noticed 😄

src/app/pages/storage/components/manager/manager.component.html
undsoft: Will resolve in next PR.

PR #7459, opened 2022-12-10
NAS-119279 / 23.10 / Missing an option to promote dataset

undsoft:
Requires fresh middleware.

Go to Snapshots, take snapshot of a dataset, clone this snapshot to a dataset. Go to this new dataset and use Promote button.


AlexKarpov98 (approved):
works good, we can add feedback for user when clicking promote (loading indication)
Screen Shot 2022-12-12 at 15 06 08

Screen Shot 2022-12-12 at 15 06 03

undsoft:
> Loading indication

True, maybe later.

PR #7461, opened 2022-12-12
NAS-117633 / 23.10 / Improving event subscriptions

RehanY147:


undsoft (changes requested):
I think we can still simply syntax.
Current approach would require a Subject every time you want to use subscription. If you take a look at something like `AlertEffects`, it would add quite a bit of boilerplate there for both subscribing and unsubscribing.
Perhaps we could just return an observable from `newSub` and unsubscribe from middleware when there are no more subscriptions to this observable.

RehanY147:
This ticket requires https://ixsystems.atlassian.net/browse/NAS-119462 to be done before it can be tested 100%. It works generally but fails when there's a case of removing an event using the deprecated `cleared` flag from middleware.

Test for regressions to even subscriptions in different places. Also check testing methodology.

src/app/services/ws.service.ts
```diff
@@ -5,18 +5,26 @@ import { UUID } from 'angular2-uuid';
...
-import { ApiEventDirectory } from 'app/interfaces/api-event-directory.interface';
-import { ApiEvent, IncomingWebsocketMessage } from 'app/interfaces/api-message.interface';
+import { ApiEventDirectory, LogsEventDirectory } from 'app/interfaces/api-event-directory.interface';
+import {
+  ApiEvent, IncomingWebsocketMessage, ResultMessage,
+} from 'app/interfaces/api-message.interface';
 import { LoginParams } from 'app/interfaces/auth.interface';
 import { Job } from 'app/interfaces/job.interface';
 
+// type ApiSubscriptionDirectory = ValuesType<ApiEventDirectory | LogsEventDirectory>['response'];
```
undsoft, line 26: Dead comment.

src/app/services/ws.service.ts
```diff
@@ -108,67 +116,81 @@ export class WebSocketService {
...
-      return;
+      return null;
     }
+  }
 
-    if ('error' in data && data.error.error === 13 /** Not Authenticated */) {
-      return this.socket.close(); // will trigger onClose which handles redirection
-    }
+  hasAuthError(data: IncomingWebsocketMessage): boolean {
+    return 'error' in data && data.error.error === 13;
```
undsoft, line 161: Don't forget to change error code here to what's in master.

src/app/services/ws.service.ts
```diff
@@ -211,59 +221,63 @@ export class WebSocketService {
...
+   * and will unsubscribe from the middleware endpoint on its own
+   * when no subscriptions are active
    */
-  sub<T = any>(api: string, subscriptionId?: string): Observable<T> {
-    const nom = api.replace('.', '_'); // Avoid weird behavior
-    if (!this.pendingSubs[nom]) {
-      this.pendingSubs[nom] = {
-        observers: {},
-      };
+  newSub<K extends keyof ApiEventDirectory>(apiMethod: K | string): Observable<ApiEvent<ApiEventDirectory[K]['response'] & LogsEventDirectory[K]['response']>> {
```
undsoft, line 229: newSub can be renamed back to `subscribe` since there are no *old* subs anyway.

undsoft (changes requested):
Something doesn't work with events, not sure if middleware or UI.
On a fresh instance from Jenkins, I don't seem to be getting any events whether old or new.
On m40, old code has events coming from middleware, but new one doesn't.

I am also happy to abandon this PR if we are going to get the same functionality in WebsocketService2.

RehanY147:
Implementing these changes in `WebSocketService2`

PR #7470, opened 2022-12-15
NAS-119340 / 23.10 / Refactor `AppSchemaService.addFormControls`

AlexKarpov98:


undsoft (approved):
Compared a number of apps with previous implementation. Payloads appear to be identical.
Nice work!

PR #7474, opened 2022-12-16
NAS-119437 / 23.10 / Use `ErrnoMixin.ENOTAUTHENTICATED` to redirect to the login page, not…

themylogin:
… `errno.EACCES` as it is a legitimate error code for many methods


bvasilenko (approved):
Works good for me

PR #7483, opened 2022-12-19
NAS-119049 / 22.12.1 / Nas-119049 - Fixed LDAP slide-in is sometimes blank

Raz-Dva:
Go to Credentials ->  Configure Active Directory 

image


undsoft (changes requested):
I don't really understand what these changes do or how they should help. Can you explain?

Raz-Dva:
If click a row in Сertificate Signing Requests and create an ACME certificate(button), the slide is blank. 
The reason for this 
`  onCreateAcmePressed(): void {
    this.slideInService.close();
    const acmeForm = this.slideInService.open(CertificateAcmeAddComponent);
  }`
The code in setTimeout (`this.slideInBody.clear()`) will run after the component instance is created in the `open` method

src/app/modules/ix-forms/components/ix-slide-in/ix-slide-in.component.ts
```diff
@@ -67,19 +69,24 @@ export class IxSlideInComponent implements OnInit, OnDestroy {
 
   closeSlideIn(): void {
     this.isSlideInOpen = false;
-
-    setTimeout(() => {
+    this.clearBodyExpected = true;
```
undsoft, line 72: Let's rename to `wasBodyCleared` instead.

src/app/modules/ix-forms/components/ix-slide-in/ix-slide-in.component.ts
```diff
@@ -29,6 +29,8 @@ export class IxSlideInComponent implements OnInit, OnDestroy {
   isSlideInOpen = false;
   wide = false;
   private element: HTMLElement;
+  timeOutOfClear: any;
```
undsoft, line 32: There is a `Timeout` type in our project for this.

undsoft (approved):
I now see what you're trying to do. Make sense and good catch.
It's a good idea to also clear timeout if it's set in `ngOnDestroy`.

PR #7485, opened 2022-12-19
NAS-119482: Make TranslateService.instant more type safe

undsoft:
`instant` in original `TranslateService` supports different styles of arguments and is typed to return `any`.
Since this is not something we need, this updates the typing.
Code review is enough.

If you have errors about JIT compilation, try removing `.angular` and `node_modules`.


denysbutenko (approved):
How it was before:
image

How it is now:
image

The changes look good to me.

AlexKarpov98 (approved):

PR #7489, opened 2022-12-21
NAS-118001 / 23.10 / Refactor Dataset Unlock Form to ix-forms

bvasilenko:
**Testing**

On **Storage** page,

All UX elements at **Dataset Unlock Form** should function properly


src/app/helptext/storage/volumes/datasets/dataset-unlock.ts
```diff
@@ -5,7 +5,7 @@ export default {
   dataset_unlock_title: T('Unlock Datasets'),
   unlock_key_file_placeholder: T('Unlock with Key file'),
   unlock_key_file_tooltip: T('Use an exported encryption key file to unlock datasets.'),
-  unlock_children_placeholder: T('Unlock Child Encrypted Roots'),
```
undsoft, line 8: Don't think we need to rename this.

src/app/pages/datasets/modules/encryption/components/dataset-unlock/dataset-unlock.component.html
```diff
@@ -0,0 +1,89 @@
...
+        [label]="helptext.unlock_key_file_placeholder | translate"
+        [tooltip]="helptext.unlock_key_file_tooltip | translate"
+        [options]="useFileOptions$"
+      ></ix-radio-group>
+    
+      <ng-container *ngIf="useFile">
+        <ix-checkbox
+          formControlName="unlock_children"
+          [label]="helptext.unlock_children_placeholder | translate"
+          [tooltip]="helptext.unlock_children_tooltip"
```
undsoft, line 15: | translate

src/app/pages/datasets/modules/encryption/components/dataset-unlock/dataset-unlock.component.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    spectator = createComponent();
+    loader = TestbedHarnessEnvironment.loader(spectator.fixture);
+  });
+
+  it('saves when set key from file', async () => {
+    const file = fakeFile('key.txt');
+
+    const fileInput = await loader.getHarness(IxFileInputHarness.with({ label: 'Upload Key file' }));
+    await fileInput.setValue([file]);
+    // expect(spectator.component.form.value).toBe('');
```
undsoft, line 85: unnecessary comment

src/app/pages/datasets/modules/encryption/components/dataset-unlock/dataset-unlock.component.ts
```diff
@@ -1,227 +1,96 @@
...
-} from 'app/pages/datasets/modules/encryption/components/dataset-unlock/dataset-unlock-form-values.interface';
+import { DatasetUnlockFormValues } from 'app/pages/datasets/modules/encryption/components/dataset-unlock/dataset-unlock-form-values.interface';
 import { UnlockDialogComponent } from 'app/pages/datasets/modules/encryption/components/unlock-dialog/unlock-dialog.component';
-import { DialogService } from 'app/services';
+import { DialogService, WebSocketService } from 'app/services';
 
 @UntilDestroy()
 @Component({
-  template: '<ix-entity-form [conf]="this"></ix-entity-form>',
+  selector: 'ix-dataset-unlock',
```
undsoft, line 23: selector can be removed

src/app/pages/datasets/modules/encryption/components/dataset-unlock/dataset-unlock-form-values.interface.ts
```diff
@@ -1,12 +1,14 @@
 export interface DatasetUnlockFormValues {
```
undsoft, line 1: This whole interface shouldn't be necessary - forms are typed now.

src/app/pages/datasets/modules/encryption/components/dataset-unlock/dataset-unlock.component.ts
```diff
@@ -277,118 +143,119 @@ export class DatasetUnlockComponent implements FormConfiguration {
...
+  };
 
-    this.keyFileControl.valueChanges.pipe(untilDestroyed(this)).subscribe({
-      next: (hideKeyDatasets: boolean) => {
-        for (let i = 0; i < this.datasets.controls.length; i++) {
-          const datasetControls = (this.datasets.controls[i] as UntypedFormGroup).controls;
-          const controls = listFields[i];
-          const keyConfig = _.find(controls, { name: 'key' });
-          const nameTextConfig = _.find(controls, { name: 'name_text' });
+  unlockSubmit(payload: DatasetUnlockParams): void {
```
undsoft, line 152: Would be good to clean up those methods, e.g.:
1. UnlockDialogComponent can be using dialogRef to communicate that it has been called instead of relying on `parent`.
2. `unlockSubmit` and `onSubmit` have similar names.
3. Sounds like UnlockDialogComponent can be renamed to UnlockSummaryDialog or something similar.
4. `unlockSubmit` is too fat. There is some code that can be extracted into separate methods, like opening an Unlock summary dialog or parsing the summary.

undsoft (changes requested):
Functional comments:
1. If I unlock a password protected dataset, should I even have an option to select a key? I thought keys only applied to importing a pool.
2. `Unlock child encrypted roots` affects how form is submitted even when manual mode is used, but checkbox is not shown. I can uncheck `Unlock child encrypted roots` and then switch to manual and get unexpected results.
3. I've tried to import a key encrypted pool. I've selected a json file with keys, but I got a middleware error that json file is invalid.
4. Error message about this was shown twice

bvasilenko:
@undsoft 

Here's a few requests for details of the functional testing:

1. 

>If I unlock a password protected dataset, should I even have an option to select a key? I thought keys only applied to importing a pool.

Explanation: for datasets it depends whether the mode is manual/automatic
   - In manual mode, the password/key_file is selected depending on the type
   - In automatic mode all you need to supply is the file

2. 

>`Unlock child encrypted roots` affects how form is submitted even when manual mode is used, but checkbox is not shown. I can uncheck `Unlock child encrypted roots` and then switch to manual and get unexpected results.

Looks like has been resolved, at least I can't replicated this bug. Would you please re-test?

3.

>I've tried to import a key encrypted pool. I've selected a json file with keys, but I got a middleware error that json file is invalid.

Could you please share a JSON file which causes an error?

undsoft (changes requested):
Waiting for changes as discussed on Slack.

bvasilenko:
@undsoft ok, now requested changes should be done

undsoft (changes requested):
1. I've created a dataset with key encryption.
2. Then I've exported pool.
3. Imported pool and tried to unlock a dataset.
I get `Input file must be a valid JSON file`. I can see an XHR request, but it appears to have empty file:
Снимок экрана 2023-01-19 в 11 07 56
My json file looks like this:

```
{"second_pool/key": "982b31a8068b6cccdaf99226de17b2e564d04ce8f769633cd52d0f213b0b54af"}
```

bvasilenko:
@undsoft webUI now sends proper contents of the key file

undsoft (approved):
Given the following set of datasets:

```
pool (key encrypted)
- pass
-- no encryption
--- keychild
```

If I lock `pass` dataset and then try to unlock it and then select to provide keys manually, I am required to provide key for `keychild` even if I don't want to unlock it/don't know the key.
We should either always show the checkbox to unlock children, even in manual mode, or not have field be required.

Otherwise looks good, all other scenarios worked okay for me.

PR #7494, opened 2022-12-22
NAS-119412 / 23.10 / Nas-119412 Removed popup after getting logs received

Raz-Dva:
Go to Applications -> Choose Logs -> Pod Logs
Click Reconnect for getting logs


RehanY147 (commented):
Not sure how to test for this. Need more info. @Raz-Dva

RehanY147 (changes requested):
We should show a different message if the call is failing or show a loading indicator if the call is taking too long to respond. Showing nothing is not a better solution than the original bug.

Raz-Dva:
Added text and spinner.

spinnerLG

RehanY147 (changes requested):
If we get no logs, meaning there's no strings returned from the middleware, we should show muted text `<Empty>` or `<No logs>`.

RehanY147 (changes requested):
https://user-images.githubusercontent.com/24204846/211510859-803d31d6-e92a-4275-94c1-7622d375e3cf.mov

I still see a blank page when there's no logs to show. What I meant was that not only show <Empty> or <No logs> when the request is loading but until you get an actual string from middleware to show in the blank page, we want to show the user that the ui didn't crash. There's just nothing to show. So, until middleware responds with at least one character/line of logs, show <Empty> in the blank page.

src/app/pages/apps-old/pod-logs/pod-logs.component.html
```diff
@@ -32,6 +32,10 @@
     class="pod-logs"
     [ngStyle]="{ 'font-size': fontSize + 'px' }"
   >
+    <div *ngIf="areLoadingPodLogs" class="log-row">
+      <span  class="no-logs">{{ 'No logs' }}</span>
```
RehanY147, line 36: You should instead do something like
`<div *ngIf="isLoadingPodLogs">...show loading spinner....</div> <div *ngIf="!isLoadingPodLogs"> <div *ngIf="!podLogs.length">{{ 'No logs yet' | translate }}</div><div *ngIf="podLogs.length">...show logs...</div></div>`

src/app/pages/apps-old/pod-logs/pod-logs.component.ts
```diff
@@ -43,6 +43,7 @@ export class PodLogsComponent implements OnInit, AfterViewInit, OnDestroy {
   protected tailLines = 500;
   podLogSubscriptionId: string = null;
   podLogSubName = '';
+  areLoadingPodLogs = false;
```
RehanY147, line 46: `isLoadingPodLogs`

RehanY147 (changes requested):

RehanY147 (approved):
`No logs to show` or `No logs yet` would be a better message. But looks good.

PR #7500, opened 2022-12-23
NAS-119552 / 23.10 / Migrate mat-dialog to MDC

AlexKarpov98:


undsoft (changes requested):
The following dialogs do not look good:
1.  Top bar -> Truecommand
2. Top bar -> Truecommand -> connect
3. Storage -> Unassigned disks -> Add to Pool.

There may be others, those are the ones I've tested.

AlexKarpov98:
@undsoft - Fixed these issues and checked other modals (rechecked)

undsoft (changes requested):
At least following dialogs look broken:
* User menu -> About.
* System Settings -> Cron Job -> Add -> Schedule -> Custom.
* Trucommand -> Connect (there is now a scrollbar on macos when mouse is connected).

undsoft (changes requested):
When job is running, minimize icon needs to be on top.
Снимок экрана 2022-12-28 в 10 29 24

When creating an encrypted pool, buttons are weird (rounded corners on one and no background on another).
Снимок экрана 2022-12-28 в 10 30 05

undsoft (approved):
Looks good.

PR #7501, opened 2022-12-23
NAS-117048 / 23.10 / WebsocketService2

RehanY147:
Open the Datasets page to test this service at runtime. Also, you have to change the username password provided in the `auth.login` call. 

This is meant to test how a `WebsocketService` implemented with `rxjs/websocketSubject` would work in our project.


src/app/services/ws2.service.ts
```diff
@@ -0,0 +1,103 @@
...
+    });
+    this.ping();
+  }
+
+  ping(): void {
+    this.subject$.next({ msg: OutgoingApiMessageType.Ping, id: UUID.UUID() });
+    setTimeout(() => this.ping(), 20 * 1000);
+  }
+
+  setupCall<K extends ApiMethod>(method: K, params?: ApiDirectory[K]['params']): Observable<ApiDirectory[K]['response']> {
```
undsoft, line 47: Let's just name it `call`.

src/app/services/ws2.service.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.subject$.subscribe({
+      complete: () => {
+        this.router.navigate(['/sessions/signin']);
+      },
+    });
+    this.ping();
+  }
+
+  ping(): void {
+    this.subject$.next({ msg: OutgoingApiMessageType.Ping, id: UUID.UUID() });
```
undsoft, line 43: You use `send` method in other places.

src/app/services/ws2.service.ts
```diff
@@ -0,0 +1,103 @@
...
+    setTimeout(() => this.ping(), 20 * 1000);
```
undsoft, line 44: Hard to imagine where this would be an issue, but we should store timeout and clear it on destroy just for sake of good practice.

src/app/services/ws2.service.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  hasAuthError(data: IncomingWebsocketMessage): boolean {
+    return 'error' in data && data.error.error === 207;
+  }
+
+  send(payload: unknown): void {
+    this.subject$.next(payload);
+  }
+
+  setupSubscription<K extends keyof ApiEventDirectory>(name: K | string): Observable<ApiEvent<ApiEventDirectory[K]['response'] & LogsEventDirectory[K]['response']>> {
```
undsoft, line 78: I think it may be a good idea to provide `subscribeToLogs` with separate typing, so that people would get errors if there were trying to subscribe to non-existent api event endpoint.

undsoft, line 78: Also, let's just call it `subscribe`.

undsoft (approved):
Very nice. Much cleaner and easier to understand.

undsoft:
Added another ticket for future improvements https://ixsystems.atlassian.net/browse/NAS-119575

RehanY147:
r

PR #7507, opened 2022-12-26
NAS-119589 / 23.10 / Enable no-explicit-any rule

undsoft:
Code review is enough.
Remaining places are hard to fix without refactoring whole things, so I've mostly just added exceptions.


dszidi (approved):
LGTM

PR #7512, opened 2022-12-27
NAS-119580 / 23.10 / Update MatCard & MatMenu Angular component to MDC

AlexKarpov98:


undsoft (changes requested):
* Any chance of us removing `!important` from root css styles so that we don't have to add exceptions in components?
* Margins on buttons in Dataset details cards are too big now.
* Data Protection -> Add Cloud Sync Task margins are too large. Other forms too.
* Local users has incorrect margins.
* System Settings -> General, Advanced and Services

undsoft (approved):
* Overall font on cards is smaller than what it used to be. This is easy to see on Storage dashboard. We should standartize on font sizes, but for now let's keep it the way it was.
* Datasets -> Permissions card has wider margins now.
* Credentials -> Directory Services -> Kerberos Settings card doesn't have lines between items.

Other things look good.

PR #7451, opened 2022-12-09
NAS-119131 / 23.10 / VM Details are not updated after form submit, page reload helps

AlexKarpov98:

PR #7513, opened 2022-12-27
NAS-119607 / 23.10 / Storage -> Disks if one disk details toggled, all the disks are toggled as well

AlexKarpov98:
Go to disks and check that once you open details for one disk, it's only opened for one disk.
As well check other expandable table rows, like `S.M.A.R.T. Test Results..`
Screenshot 2022-12-27 at 18 09 19

### DIFF
--- a/src/app/modules/entity/entity-table/entity-table.component.html
+++ b/src/app/modules/entity/entity-table/entity-table.component.html
@@ -166,7 +166,7 @@
               (mouseout)="onHover($event, false)"
             >
               <ng-container *ngIf="hasDetails() === true">
-                <ng-container *ngIf="element?.id === expandedElement?.id">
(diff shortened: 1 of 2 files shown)


undsoft (changes requested):
There is a `getRowIdentifier` that you may be able to use.
I would also like to understand what changed that broke this.

RehanY147 (changes requested):
We can use `UUID.UUID()` instead to get a unique id. `Math.random` isn't always going to be unique by design even if rare

AlexKarpov98:
> There is a `getRowIdentifier` that you may be able to use. I would also like to understand what changed that broke this.

as I see there was simply no `ID` field is some places and in code we have `item.id` checks.

undsoft (approved):
Alex explained that the issue appeared because of changes here https://github.com/truenas/webui/pull/7451/files

RehanY147 (approved):

PR #7528, opened 2022-12-30
NAS-119663 / 23.10 / Create common form-actions component

undsoft:
Testing: test some of the the affected components.


undsoft:
@dszidi Bump.

PR #7531, opened 2022-12-30
NAS-119668: Correctly report vdev width

dszidi:
To test:

Create each configuration and check the reported width in the Storage Dashboard topology card.

- Two raidz1 vdevs with 5 disks in each vdev. Topology card should say "5 wide"
- Two mirror vdevs with 2 disks in each vdev. Topology card should say "2 wide"
- Two mirrors. One made with two disks and another with 3 disks. Topology card should say "Mixed Width VDEVs"

### DIFF
--- a/src/app/core/testing/utils/mock-storage-generator.utils.ts
+++ b/src/app/core/testing/utils/mock-storage-generator.utils.ts
@@ -31,6 +31,14 @@ export interface MockTopology {
   disks: StorageDashboardDisk[];
 }
 
+export interface AddTopologyOptions {
+  scenario: MockStorageScenario;
+  layout: TopologyItemType;
+  diskSize: number;
+  width: number;
+  repeats: number;
+}
+
 export class MockStorageGenerator {
   poolState: PoolInstance;
   disks: StorageDashboardDisk[];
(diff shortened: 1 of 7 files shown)


src/app/pages/storage/components/dashboard-pool/topology-card/topology-card.component.spec.ts
```diff
@@ -35,31 +35,45 @@ describe('TopologyCardComponent', () => {
             data: [
               {
                 type: CreateVdevLayout.Raidz1,
-                children: [{ type: 'DISK', disk: 'sda' }, { type: 'DISK', disk: 'sdb' }, { type: 'DISK', disk: 'sdc' }],
+                children: [
+                  { type: 'DISK', disk: 'sda' },
```
undsoft, line 39: While we are here, can we please change 'DISK' to TopologyItemType.Disk?

undsoft (approved):
Looks good.
Can we extract this code in some utility function and add more unit tests for every scenario?
It may come in handy in new pool creation wizard.

src/app/pages/storage/components/dashboard-pool/topology-card/topology-card.component.ts
```diff
@@ -20,7 +23,7 @@ interface TopologyState {
 }
 
 const notAssignedDev = 'VDEVs not assigned';
```
undsoft, line 25: Feels weird comparing to a string down below in code. Should be something else and string itself should be translated.

src/app/services/storage.service.ts
```diff
@@ -386,4 +397,149 @@ export class StorageService {
...
+      vdevTypes.add(vdev.type);
+    });
+    return vdevTypes;
+  }
+
+  isMixedVdevType(vdevTypes: Set<string>): boolean {
+    return vdevTypes.size > 1;
+  }
+
+  validateVdevs(category: PoolTopologyCategory,
```
undsoft, line 487: I believe you mentioned that there were some scenarios where we may want to compare things between different categories. In this case this method may not work.
You may want to make other methods that are not intended to be used externally `private` and extensively test public methods.

src/app/core/testing/utils/mock-storage-generator.utils.ts
```diff
@@ -0,0 +1,381 @@
...
+      status: 'ONLINE',
+      topology: {
+        data: [],
+        special: [],
+        log: [],
+        spare: [],
+        cache: [],
+        dedup: [],
+      },
+    } as unknown as PoolInstance;
```
undsoft, line 59: unknown appears to be unnecessary and you can use enum for `status`.

src/app/core/testing/utils/mock-storage-generator.utils.ts
```diff
@@ -0,0 +1,381 @@
...
+
+    const disks: StorageDashboardDisk[] = [];
+
+    return { poolState: pool, disks };
+  }
+
+  addDataTopology(scenario: MockStorageScenario,
```
undsoft, line 66: If you have more than 3 arguments, it's better to just have one argument called `options`.

src/app/core/testing/utils/mock-storage-generator.utils.ts
```diff
@@ -0,0 +1,381 @@
...
+      case TopologyItemType.Raidz1:
+      case TopologyItemType.Raidz:
+      case TopologyItemType.Mirror: {
+        const vdevDisks: TopologyDisk[] = [];
+        for (let i = 0; i < width; i++) {
+          const topologyDisk: unknown = this.generateTopologyDisk();
+          vdevDisks.push(topologyDisk as TopologyDisk);
+        }
+
+        const vdev: unknown = {
```
undsoft, line 268: `const vdev = {} as Vdev.`
`unknown` will swallow errors, like if you mistype one of the fields.

Same in other places in this file.

undsoft (changes requested):
Approach looks good so far. 👍🏻

src/app/services/storage.service.ts
dszidi: In those scenarios we specifically compare to the Data vdevs. Case is covered at line 537 where we check for `TopologyWarning.RedundancyMismatch`. We only do this for Metadata and Dedup VDEVs

src/app/enums/v-dev-type.enum.ts
```diff
@@ -1,13 +1,17 @@
...
   Mirror = 'MIRROR',
   Spare = 'SPARE',
   Log = 'LOG',
   Missing = 'MISSING',
   Root = 'ROOT',
   File = 'FILE',
   Raidz = 'RAIDZ',
+  Raidz1 = 'RAIDZ1',
+  Raidz2 = 'RAIDZ2',
+  Raidz3 = 'RAIDZ3',
```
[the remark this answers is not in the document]
dszidi: Yep

`pool.query` via midclt...


image

Actual payload received on Storage Dashboard...


image

undsoft (approved):
I haven't checked all the specific scenarios, but overall code looks good. Nice work.

undsoft:
backport

undsoft:
backport

PR #7533, opened 2023-01-01
NAS-119477 / 23.10 / Manual disk selection layout

RehanY147:


undsoft (changes requested):
Needs more CSS work to make it look like in the designs.

undsoft (changes requested):
It's closer, but there are some things that can be improved.
I would recommend this extension https://chrome.google.com/webstore/detail/perfectpixel-by-welldonec/dkaagdgjmgdmbnecmcefdhjekcoceebi?hl=en
You can add designs, set scale to something like 0.54 and compare different elements.
Common elements like inputs and buttons don't have to be exactly 1 to 1, but new elements can be closer to designs.

Among other things to fix:
* Dark backdrop is missing.
* Enclosure icon is too big.
* Disk icon is too wide.
* Vdev corners shouldn't be rounded.
* Colors look different?
* etc.

dszidi:
For reference here is what the mockup had for the disk icons.

image
The background colors are off for the `.identifier-container` and `.size-container` The color used in the mockup is #575757

Instead of listing all the things that are off with the icon, I think you might save yourself a lot of hassle if you just used the svg as the disk icon component template. [Angular allows this](https://angular.io/guide/svg-in-templates) and you can use your usual interpolations and `ngIf` type tricks in the markup.

There isn't an svg of just the disk icon component yet in the webui designs repo. I'll add it real quick and give you a link in slack

src/app/pages/storage/modules/pool-manager/components/disk-icon/disk-icon.component.ts
```diff
@@ -0,0 +1,11 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'ix-disk-icon',
+  templateUrl: './disk-icon.component.svg',
+  styleUrls: ['./disk-icon.component.scss'],
+})
+export class DiskIconComponent {
+    @Input() identifier: string;
```
undsoft, line 9: Indentation is off, also needs `ChangeDetection.OnPush`.

src/app/pages/storage/modules/pool-manager/components/disk-info/disk-info.component.html
```diff
@@ -0,0 +1,26 @@
+<div class="container">
+  <div class="row">
+    <div class="left">
+      <div class="item">
+        <strong class="title">Slot:</strong> {{ slot }}
```
undsoft, line 5: Let's add translate pipe now.

src/app/pages/storage/modules/pool-manager/components/disk-info/disk-info.component.ts
```diff
@@ -0,0 +1,13 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'ix-disk-info',
+  templateUrl: './disk-info.component.html',
+  styleUrls: ['./disk-info.component.scss'],
+})
+export class DiskInfoComponent {
+    @Input() slot: number;
```
undsoft, line 9: Same about indentation and `OnPush`.

undsoft (approved):
Looks pretty good. I would try larger width and smaller text for `RaidZ2 / 24 Tib Est. Usable Capacity` text, because at least on my resolution vdevs and text wraps.

PR #7534, opened 2023-01-02
NAS-119616 / 23.10 / Changing from method to subject

RehanY147:


dszidi (approved):
Code looks good and editing google photos credential works as expected
LGTM

PR #7541, opened 2023-01-04
NAS-119670 / 23.10 / IPMI is missing when page is reloaded

AlexKarpov98:
IPMI should be available now on page reload:

Screenshot 2023-01-04 at 17 24 19


RehanY147 (approved):
Looks good

PR #7544, opened 2023-01-05
NAS-119604 / 23.10 / Design flaw: If the error log is too large, it will be displayed beyond the log window

AlexKarpov98:
Now it scrolls by the max-height of `300px`:

Screenshot 2023-01-05 at 14 55 26


undsoft:
Lines are not readable, then should probably be force-wrapped.

bvasilenko (approved):
Works good for me!

PR #7546, opened 2023-01-06
NAS-119757 / 23.10 / Category selector in alerts looks too much like multiselect

undsoft:
Before:
Снимок экрана 2023-01-06 в 19 01 07

After:
Снимок экрана 2023-01-06 в 19 00 08


AlexKarpov98 (approved):
I agree, approved 👍

PR #7553, opened 2023-01-09
Fix variant of the 256-bit NIST ECC curve

iainnicol:
First, this is consistent with the other two NIST curves in the codebase, which use the random ‘r’ variant of the curves, not the Koblitz ‘k' variants. Second, attempting to request a Lets Encrypt certificate with the previous type of secp256k1 fails, with the error message “Error parsing certificate request: x509: unsupported elliptic curve”. Third, the CAB forum says the 256-bit curve must be secp256r1; see section 7.1.3.1.2 of <https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.8.6.pdf>.

### DIFF
--- a/src/app/enums/ec-curve.enum.ts
+++ b/src/app/enums/ec-curve.enum.ts
@@ -2,6 +2,6 @@ export enum EcCurve {
   BrainpoolP512R1 = 'BrainpoolP512R1',
   BrainpoolP384R1 = 'BrainpoolP384R1',
   BrainpoolP256R1 = 'BrainpoolP256R1',
-  Secp256K1 = 'SECP256K1',
+  Secp256R1 = 'SECP256R1',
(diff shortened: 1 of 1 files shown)


iainnicol:
I should have linked them earlier, but I just had a related PR approved for the middleware:

https://github.com/truenas/middleware/pull/10369

This PR here in the webui is less important - AFAICT the code is unused anyway. But perhaps you can share the Jira ticket number, or perhaps not. Dunno.

yocalebo:
This should be merged into Cobia only. We merged the relevant middleware PR here: https://github.com/truenas/middleware/pull/10369

undsoft:
You are correct, this is not in fact used. I'll remove this enum in future code to avoid confused and will close this PR.
Thanks.

PR #7558, opened 2023-01-10
NAS-119743 / 23.10 / System -> General -> Support -> Production system checkbox and modal bug

AlexKarpov98:
How it works now:

https://user-images.githubusercontent.com/22980553/211558909-6489a766-6de9-4d87-8014-634d13ec0d77.mov

Previously if we cancelled, checkbox was marked as checked and didn't get back to default state (unchecked)
Like that:
https://user-images.githubusercontent.com/22980553/211559206-f94aeb14-8137-4067-b302-5b1b1189b61d.mov

### DIFF
--- a/src/app/pages/system/general-settings/support/support.component.ts
+++ b/src/app/pages/system/general-settings/support/support.component.ts
@@ -8,7 +8,7 @@ import { Store } from '@ngrx/store';
 import { TranslateService } from '@ngx-translate/core';
 import _ from 'lodash';
 import { Observable, of, switchMap } from 'rxjs';
-import { filter, map } from 'rxjs/operators';
+import { filter, tap } from 'rxjs/operators';
 import { helptextSystemSupport as helptext } from 'app/helptext/system/support';
 import { SnackbarService } from 'app/modules/snackbar/services/snackbar.service';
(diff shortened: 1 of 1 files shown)


src/app/pages/system/general-settings/support/support.component.ts
```diff
@@ -150,10 +150,19 @@ export class SupportComponent implements OnInit {
...
 
   updateProductionStatus(event: MatCheckboxChange): void {
-    let request$: Observable<boolean | SetProductionStatusDialogResult> = of(true);
+    let request$: Observable<boolean | SetProductionStatusDialogResult> = of(false);
     if (event.checked) {
       request$ = request$.pipe(
-        switchMap(() => this.matDialog.open(SetProductionStatusDialogComponent).afterClosed()),
+        switchMap(() => this.matDialog.open(SetProductionStatusDialogComponent).afterClosed().pipe(
+          map((confirmed) => {
+            if (confirmed) {
```
denysbutenko, line 158: The `map` operator seems unnecessary here since we don't change the resulting output of the stream, so the `tap` operator may fit better to produce side effects. If there was an intention to use especially `map,` it is okay too.
```
tap((confirmed) => { 
  if (!confirmed) { 
    this.isProduction = false; 
    this.cdr.markForCheck();
  }
})
```

denysbutenko (approved):
Looks good 👍🏻

PR #7559, opened 2023-01-10
NAS-119793 / 23.10 / Sidebar menu closes if I hit ESC button

AlexKarpov98:
Menu should be closed only by clicking the hamburger menu.

### DIFF
--- a/src/app/modules/layout/components/admin-layout/admin-layout.component.html
+++ b/src/app/modules/layout/components/admin-layout/admin-layout.component.html
@@ -14,6 +14,7 @@
     [fxFlex]="sidenavWidth"
     [opened]="isSidenavOpen"
     [mode]="sidenavMode"
+    [disableClose]="true"
     (closedStart)="onMenuClosed()"
   >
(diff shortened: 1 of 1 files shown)


denysbutenko (changes requested):
I agree that we need to handle hitting `Esc` on the sidebar menu, and we probably need to keep the possibility to close the menu by clicking on the backdrop. @undsoft, what do you think about it?

AlexKarpov98:
> I agree that we need to handle hitting `Esc` on the sidebar menu, and we probably need to keep the possibility to close the menu by clicking on the backdrop. @undsoft, what do you think about it?

I need to check that functionality, I forgot that we also have backdrop.
It was easy quick solution :D

undsoft:
I'd rather keep closing via Esc if no easy solution exists. It's not a big issue.

AlexKarpov98:
Solution:
Disable `Esc` close only if `sidenavMode !== 'over'` so then we see this and `Esc` is needed.

Screenshot 2023-01-11 at 23 39 23

denysbutenko (approved):
Works well 👍🏻

PR #7561, opened 2023-01-11
NAS-119812 / 23.10 / No way to manipulate cron tasks in UI

AlexKarpov98:


undsoft (changes requested):
I don't really think this is necessary. You can already access cron jobs from Advanced Settings.

AlexKarpov98:
closed due to @undsoft comment.
`I don't really think this is necessary. You can already access cron jobs from Advanced Settings.`

PR #7567, opened 2023-01-12
NAS-119716 / 23.10 / Switching a drop down selection, does not "drop" values that are now hidden

AlexKarpov98:
`
    if (this.isNew || !_.get(this.config, keys)) {
      _.unset(data, keys);
    }
`

it was unnecessary check because all we needed to do -> just remove field and not check if it has value, if it came inside this method:

`
    deleteField$.pipe(untilDestroyed(this)).subscribe({
      next: (fieldToBeDeleted) => {
`

that means it should be deleted anyways! 😄


bvasilenko (approved):
Looks good to me

PR #7573, opened 2023-01-13
NAS-119591 / 23.10 / [SCALE/Apps]: `default` does not work in `list` types

AlexKarpov98:
Go to apps:
try to install `docspell`, there you should see pre-created array item:
`default: ["127.0.0.1"]`

Screenshot 2023-01-13 at 13 10 09

This  block `label: IPs` has one element inside of a card.

So when middleware wants to pre-create element, it set's it as as `string[]` 
`default: ["127.0.0.1", "1.1.1.1"]` -> will result in 2 pre-created cards

image

There might be multiple elements inside of a card.
In this case middleware sets it as `object[]`

Screenshot 2023-01-13 at 13 12 32

example:
```
[
  {
    name: 'preselected name 1',
    value: 'preselected value 1,
  },
  {
    name: 'preselected name 2',
    value: 'preselected value 2,
  }
]
```
In this case 2 cards will be created.

### DIFF
--- a/src/app/interfaces/chart-release.interface.ts
+++ b/src/app/interfaces/chart-release.interface.ts
@@ -136,7 +136,7 @@ export interface ChartSchemaNodeConf {
   type: ChartSchemaType;
   attrs?: ChartSchemaNode[];
   items?: ChartSchemaNode[];
-  default?: unknown;
+  default?: unknown | unknown[];
   enum?: ChartSchemaEnum[];
   required?: boolean;
   value?: string;
--- a/src/app/interfaces/dynamic-form-schema.interface.ts
+++ b/src/app/interfaces/dynamic-form-schema.interface.ts
@@ -1,6 +1,7 @@
 import { UntypedFormArray } from '@angular/forms';
 import { Observable } from 'rxjs';
 import { DynamicFormSchemaType } from 'app/enums/dynamic-form-schema-type.enum';
+import { ChartSchemaNode } from 'app/interfaces/chart-release.interface';
 import { Option } from 'app/interfaces/option.interface';
 import { TreeNodeProvider } from 'app/modules/ix-forms/components/ix-explorer/tree-node-provider.interface';
 
@@ -62,7 +63,8 @@ export interface DynamicFormSchemaIpaddr extends DynamicFormSchemaBase {
 export interface DynamicFormSchemaList extends DynamicFormSchemaBase {
   type: DynamicFormSchemaType.List;
   items?: DynamicFormSchemaNode[];
-  itemsSchema?: unknown[];
+  itemsSchema?: ChartSchemaNode[];
+  default?: unknown[];
 }
 
 export interface DynamicFormSchemaDict extends DynamicFormSchemaBase {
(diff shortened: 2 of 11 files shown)


denysbutenko (approved):
image

Two fields appeared. I don't know if having two same addresses is the intended behavior.

AlexKarpov98:
@denysbutenko 
please share a schema where you tested, so I can tell if there 2 fields should appear, If it's not a question, then please merge PR. (maybe someone updated schema for testing)

denysbutenko:
> @denysbutenko please share a schema where you tested, so I can tell if there 2 fields should appear, If it's not a question, then please merge PR. (maybe someone updated schema for testing)

Well. I set up the docker container against the latest nightly testing build. In the UI - I added the `truecharts` catalog using the `incubator` train. I then installed the `docspell` app without modifications. As a result, the install form shows a single `IP Entry` block, and the Edit form shows two `IP Entry` blocks with the same IP address.

AlexKarpov98:
@denysbutenko 
then it's a bug, we don't need to set default data for EDIT state, correct?

denysbutenko:
> @denysbutenko then it's a bug, we don't need to set default data for EDIT state, correct?

I don't know.

AlexKarpov98:
@denysbutenko 
I need to make this fix.
Screenshot 2023-01-17 at 11 49 29

AlexKarpov98:
@denysbutenko - check it again please

Ornias1993:
> @denysbutenko I need to make this fix. Screenshot 2023-01-17 at 11 49 29

Edgecase:
If an App is updated and a new field is added, there is no old data to fetch and it should load the default anyway.
Is this taken into account @AlexKarpov98 ?

AlexKarpov98:
@Ornias1993 
> Edgecase: If an App is updated and a new field is added, there is no old data to fetch and it should load the default anyway. Is this taken into account @AlexKarpov98 ?

no, (it currently only adds default data in `creation` mode) but it's not good in the case when I want to remove all cards for example, why would there default setup appear then? (If I want an empty array)

Ornias1993:
> @Ornias1993
> 
> > Edgecase: If an App is updated and a new field is added, there is no old data to fetch and it should load the default anyway. Is this taken into account @AlexKarpov98 ?
> 
> no, (it currently only adds default data in `creation` mode only) but it's not good in the case when I want to remove all cards for example, why would there default setup appear then? (If I want an empty array)

If an app is updated, fields can be added with an update.
Those fields should NOT be empty, they should be filled with the defaults on update.

AlexKarpov98:
I didn't touch current implementation, only added defaults in `creation mode`

AlexKarpov98:
@Ornias1993 previously there were no implementation for `default` field for lists at all 😃

Ornias1993:
> @Ornias1993 previously there were no implementation for `default` field for lists at all 😃

Okey, so this creation/edit mode is limited to the lists and will not affect other fields?

Ornias1993:
> I didn't touch current implementation, only added support for defaults in `creation mode` @Ornias1993

Which means it's borderline unusable, because Apps cannot be updated to use it at all.

undsoft:
@Ornias1993 So if the app is updated, how would new default value be saved if user doesn't know to go and edit-save the app? There is a difference between putting something as a default in the form and this value actually being saved.
I feels like there should be some other middleware mechanism for this and then default values during edits won't be a problem.

Ornias1993:
> @Ornias1993 So if the app is updated, how would new default value be saved if user doesn't know to go and edit-save the app? There is a difference between putting something as a default in the form and this value actually being saved. I feels like there should be some other middleware mechanism for this and then default values during edits won't be a problem.

Ahh yes, indeed... you're correct the default should be applied on the middleware level.

PR #7574, opened 2023-01-13
NAS-119512 / 23.10 / Fixed theme is lost when session times out

Raz-Dva:
Change the colour theme.  After the session has expired(or do this manually) check the applying of the theme.

### DIFF
--- a/src/app/store/preferences/preferences.effects.ts
+++ b/src/app/store/preferences/preferences.effects.ts
@@ -13,7 +13,7 @@ import {
   autoRefreshReportsToggled,
   builtinGroupsToggled,
   builtinUsersToggled, guiFormSubmitted, localizationFormSubmitted,
-  preferencesLoaded, preferredColumnsUpdated, clearPreferencesState,
+  preferencesLoaded, preferredColumnsUpdated,
   themeNotFound,
   updateRebootAfterManualUpdate,
 } from 'app/store/preferences/preferences.actions';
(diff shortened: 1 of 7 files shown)


src/app/store/preferences/preferences.effects.ts
```diff
@@ -28,6 +28,7 @@ export class PreferencesEffects {
   loadPreferences$ = createEffect(() => this.actions$.pipe(
     ofType(adminUiInitialized),
     mergeMap(() => {
+      this.store$.dispatch(clearPreferencesState());
```
undsoft, line 31: I think you can achieve the same by simply setting initial state on `adminUiInitialized`.

src/app/store/preferences/preferences.reducer.ts
```diff
@@ -36,6 +36,7 @@ export const preferencesReducer = createReducer(
   on(noDashboardStateFound, (state) => ({ ...state, dashboardState: null })),
   on(preferencesLoaded, (state, { preferences }) => ({ ...state, preferences, areLoaded: true })),
   on(noPreferencesFound, (state) => ({ ...state, preferences: defaultPreferences, areLoaded: true })),
+  on(clearPreferencesState, (state) => ({ ...state, ...initialState })),
```
undsoft, line 39: or simply `() => initialState`

undsoft (changes requested):
I get
```
TypeError: menuInfo is not iterable
    at AdminLayoutComponent.toggleMenu 
admin-layout.component.ts:195:31
```
when I click on System Settings menu item.

undsoft (approved):
Looks good, theme is no longer lost.

PR #7578, opened 2023-01-13
TE-153 / 22.12.1 / Fixing AD netbios, system dataset and deleting dataset test issues 

ericbsd:
Test results at https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Bluefin/job/UI%20Tests%20-%20TrueNAS%20SCALE%20Bluefin%20-%20Nightly/57/

### DIFF
--- a/tests/bdd/xpaths.py
+++ b/tests/bdd/xpaths.py
@@ -74,7 +74,7 @@ class addZvol:
 
 class advanced:
     title = '//h1[contains(.,"Advanced")]'
-    systemDatasetPool_card = '//h3[text()="System Dataset Pool"]'
+    systemDatasetPool_card = '//h3[text()="Storage"]'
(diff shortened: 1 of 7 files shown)


tests/bdd/xpaths.py
```diff
@@ -74,7 +75,7 @@ class addZvol:
 
 class advanced:
     title = '//h1[contains(.,"Advanced")]'
-    systemDatasetPool_card = '//h3[text()="System Dataset Pool"]'
+    systemDatasetPool_card = '//h3[text()="Storage"]'
```
jordan-mcgillivray005, line 78: Is it possible there is a more robust selector than the text? Not sure how often the text will be changing if the future so it may not be an issue.

alexandrabain: We have an open discussion on the future setup with Dev right now. This is acceptable for now

alexandrabain (approved):

PR #7583, opened 2023-01-16
NAS-119844 / 23.10 / NVMe drives show "Enable S.M.A.R.T." checkbox unchecked even if `togglesmart` API field is true

AlexKarpov98:
main was merged here:
https://github.com/truenas/webui/pull/7591
now it's a cleanup PR


src/app/pages/storage/modules/disks/components/disk-form/disk-form.component.ts
```diff
@@ -64,7 +63,11 @@ export class DiskFormComponent implements OnInit {
 
   setFormDisk(disk: Disk): void {
```
RehanY147, line 64: Instead of adding a null check here, we can avoid it by not passing null when calling this method in the disk-list component for example. Let's find other places where this method is used and make sure that null isn't passed for this values.

RehanY147 (changes requested):

src/app/pages/storage/modules/disks/components/disk-form/disk-form.component.ts
AlexKarpov98: I think it’s better to keep my fix, because other places just call this method, so It’s better to fix it in single place rather that in all places like: 
editForm.setFormDisk(this.disk)

AlexKarpov98: @RehanY147

denysbutenko (approved):
No objections.

RehanY147 (approved):

PR #7588, opened 2023-01-17
NAS-119735 / 23.10 / Restore functionality

bvasilenko:
**Testing**

Go to **Services > NFS** and click **Add SPN** button


src/app/pages/services/components/service-nfs/service-nfs.component.spec.ts
```diff
@@ -46,6 +47,11 @@ describe('ServiceNfsComponent', () => {
           '192.168.1.119': '192.168.1.119',
         }),
         mockCall('nfs.update'),
+        mockCall('kerberos.keytab.has_nfs_principal', false),
+        mockCall('directoryservices.get_state', {
+          activedirectory: DirectoryServiceState.Healthy,
+          ldap: DirectoryServiceState.Disabled,
+        }),
```
undsoft, line 54: Also needs a test case to check that field appears when conditions are met and that it opens the dialog.

undsoft (approved):
Had to mock some endpoints, but generally looks okay.

PR #7592, opened 2023-01-17
NAS-119792 / 23.10 / Changing user password doesn't work on non-root users

AlexKarpov98:
Change password should be visible if user has `{ local: true }` prop.


undsoft (changes requested):
As per discussion on Slack there is additional scope of work for this ticket. I'll add additional discussions in the ticket.

src/app/services/ws.service.ts
```diff
@@ -350,4 +354,19 @@ export class WebSocketService {
       this.window.location.reload();
     });
   }
+
+  private getLoggedInUserInformation(): void {
```
undsoft, line 358: All this logic doesn't really belong to websocket service, but it's okay for now as this service is going away anyway.

src/app/services/ws.service.ts
```diff
@@ -350,4 +354,19 @@ export class WebSocketService {
...
+    let authMeUser: DsUncachedUser;
+
+    this.call('auth.me').pipe(
+      switchMap((loggedInUser: DsUncachedUser) => {
+        authMeUser = loggedInUser;
```
undsoft, line 363: No point in having this variable and assignment.

src/app/services/ws.service.ts
```diff
@@ -350,4 +354,19 @@ export class WebSocketService {
...
+        return this.call('user.query', [[['uid', '=', authMeUser.pw_uid]]]);
```
undsoft, line 364: If this is not a local user, you'll get an empty array in result and will end up spreading undefined.

undsoft (approved):
Overall looks good, I was able to change user password.

src/app/services/ws.service.ts
AlexKarpov98: we need to keep data from `auth.me` and then I merge it with `user.query` data.

src/app/services/ws.service.ts
AlexKarpov98: included this case

src/app/services/ws.service.ts
AlexKarpov98: we can create separate store for it if needed.

AlexKarpov98: I thought if we handle `loginCallback` , `loginWithToken` and other things in here, so it will be OK to get user data here.

PR #7599, opened 2023-01-18
NAS-119879 / 23.10 / Unable to set or delete an ACL preset

undsoft:
Test by creating an acl preset and removing it.


bvasilenko (approved):
Code looks good

PR #7602, opened 2023-01-19
NAS-119917 / 23.10 / Margins on some dialogs is incorrect

undsoft:
Testing: check affected dialogs.

Before:
Снимок экрана 2023-01-19 в 11 56 44


bvasilenko (approved):
Looks good to me

image

image

image

PR #7605, opened 2023-01-19
NAS-119303 / 23.10 / Experimenting with test id attributes

undsoft:
This is a temporary experiment to see if this is something release engineering can use.
Code review is enough. Accepting comments and suggestions on how to make this better/easier to use.


AlexKarpov98 (approved):
I like it 👍

PR #7607, opened 2023-01-19
NAS-119923: Replacing usages of websocket

undsoft:
Mostly auto-replaces usages of websocket with websocket2.


RehanY147 (approved):
Looks good

PR #7611, opened 2023-01-20
NAS-119211 / 22.12.2 / DisksUpdateService also returns a list of updated disks

Raz-Dva:
Subscribing to DisksUpdateService allows to get an updated disk or the entire list of already updated disks. 
Added debounceTime to avoid too many concurrent calls to the server.


undsoft:
@Raz-Dva Is this a blind fix?

Raz-Dva:
> @Raz-Dva Is this a blind fix?

My guess is that when a drive is removed, `subscribe('disk.query')` will emit an event on each disk (the user has 50 of them). Each subscriber additionally makes a `call('disk.query')` request, which will result in an error from the server _"Maximum number of concurrent calls (20) has exceeded."_ Therefore, the Enclosure View page does not update the list of disks after a drive is removed.

undsoft:
Best to wait for Damian to test as he may be able to reproduce the issue.

dszidi (commented):
I don't see anything problematic with the code and on my own test system enclosure ui updates as expected on removal of drive as well as on insertion of drive. I should mention that this functionality also works on the recent bluefin release that is installed on my machine. So I wasn't able to see the actual problematic behavior reported.

Perhaps the problem only presents itself on larger systems with many drives?

dszidi (approved):

undsoft:
Let's merge and ask to reproduce the issue again.

PR #7615, opened 2023-01-22
NAS-119968 / 23.10 / Docker images page doesn't work

undsoft:
Testing: see that docker images tab in Apps works.


denysbutenko (approved):
image

Looks good.

PR #7619, opened 2023-01-23
NAS-119973 / 23.10 / Clear loading indicator in signin form on error

undsoft:
Testing. Do something in code to make one of the token generation requests fail.


bvasilenko (approved):
Works good

PR #7625, opened 2023-01-24
NAS-119980 / 23.10 / Missing icon in dataset details header

AlexKarpov98:
Testing:
- Switch to mobile view
- Open dataset
Screenshot 2023-01-24 at 15 21 15

- If icon is there, go to Storage and then go back.

- Open same dataset 

-- 
As well test devices for the same logic
Screenshot 2023-01-24 at 15 41 03


bvasilenko (approved):
Works good for me

image

PR #7627, opened 2023-01-24
NAS-119385 / 23.10 / Add Certificate wizard refactoring

undsoft:
Testing: go to certificates and test the wizard that appears when you press Add.

Several scenarios are available. You can create a certificate and key for import using these instructions https://stackoverflow.com/questions/10175812/how-to-generate-a-self-signed-ssl-certificate-using-openssl

Import behaviour was reworked because it didn't make sense to send private key and passphrases when CSR was used. They were not in fact sent even in previous implementation.


bvasilenko (approved):
Tested with the following:

```
openssl req -newkey rsa:2048 -keyout PRIVATEKEY.key -out MYCSR.csr
openssl req -x509 -new -nodes -key PRIVATEKEY.key -sha256 -days 1825 -out CERTIFICATE.pem
```

PR #7628, opened 2023-01-25
NAS-119996 / 23.10 / Fixed value to GiB unit

RehanY147:
Check that no matter what value you enter into the `Swap Size` field, it's treated as value in GiB unit. Check that when you see the value in the `Advanced Settings -> Storage` section, it's also presented as GiBs


undsoft (changes requested):
I don't think we need to add obscure arguments to core functionality to make this happen.
I'd rather just remove memory formatter from the field, make it a type = number and add `(in Gbs)` to label.

PR #7633, opened 2023-01-25
NAS-119431 / 22.12.2 / Removed check

RehanY147:


undsoft (changes requested):
Let's ask middleware to add user's schema to one of the apps, so that we can test this.

PR #7637, opened 2023-01-26
NAS-119969 / 23.10 / Replaced `WebsocketService` with `WebsocketService2`, `WebsocketManagerService`, and `AuthService`

RehanY147:
Code review
Check that wherever logic had to be updated because of the new service, nothing was missed
Check for regressions
Check that logic changed to fix tests isn't missing anything

`NonAuthApiMethod` and `NonAuthApiDirectory` is created to protect the developer from making authentication calls directly. Authentication logic should be constrained to `AuthService` and hidden behind methods.

`WebsocketManagerService` is meant to separate the call and subscription logic from the websocket connection management and maintenance.


src/app/services/auth/auth.service.ts
```diff
@@ -1,26 +1,166 @@
...
+      id: uuid,
+      msg: IncomingApiMessageType.Method,
+      method: 'auth.generate_token',
+      params: [tokenLiftime],
+    };
+    this.wsManager.send(payload);
+    return this.getFilteredWebsocketResponse<string>(uuid);
+  }
+
+  generateTokenWithDefaultLifetime(): Observable<string> {
```
undsoft, line 107: This method is not called anywhere.

src/app/services/auth/auth.service.ts
```diff
@@ -1,26 +1,166 @@
...
+
+  getFilteredWebsocketResponse<T>(uuid: string): Observable<T> {
+    return this.wsManager.websocketSubject$.pipe(
+      filter((data: IncomingWebsocketMessage) => data.msg === IncomingApiMessageType.Result && data.id === uuid),
+      map((data: ResultMessage<T>) => data.result),
+      take(1),
+    );
+  }
+
+  generateToken(tokenLiftime: number): Observable<string> {
```
undsoft, line 90: Typo in `Lifttime`

src/app/services/ws-manager.service.ts
```diff
@@ -17,8 +18,13 @@ import { ApiEvent, IncomingWebsocketMessage } from 'app/interfaces/api-message.i
 })
 export class WebsocketManagerService {
```
undsoft, line 19: As discussed let's rename to `WebsocketConnection`.

undsoft (changes requested):

src/app/services/auth/auth.service.ts
RehanY147: It is on line 112 of `TerminalComponent`

src/app/services/auth/auth.service.ts
```diff
@@ -1,26 +1,152 @@
...
+    this.wsManager.send(payload);
+    return this.getFilteredWebsocketResponse<void>(uuid).pipe(tap(() => {
+      this.isLoggedIn = false;
+      this.isLoggedIn$.next(false);
+    }));
+  }
 
-    this.window.sessionStorage.setItem('redirectUrl', state.url);
-    this.router.navigate(['/sessions/signin']);
+  private getLoggedInUserInformation(): void {
```
undsoft, line 118: Wrong line. This method is unused and it makes `Change Password` in user menu hidden.

src/app/interfaces/api-message.interface.ts
```diff
@@ -29,7 +29,7 @@ export interface ApiEvent<T = unknown> {
   id: number;
   msg: IncomingApiMessageType.Changed | IncomingApiMessageType.Added | IncomingApiMessageType.Removed;
   // true when item is fully removed from the collection
-  // TODO: Find usages and guard against directly in WebsocketService. Only rely on 'removed' ApiEventMessage
+  // TODO: Find usages and guard against directly in WebsocketService2. Only rely on 'removed' ApiEventMessage
   /**
    * @deprecated
    */
```
undsoft, line 35: In the future we would need to do something about this.

undsoft (changes requested):
I got an error about product type upon reloading the page after logging in.

Снимок экрана 2023-01-30 в 13 39 49

undsoft (approved):
Don't see any other regressions. 👍🏻

PR #7638, opened 2023-01-26
NAS-119865 / 23.10 / Add App Detail View Component

denysbutenko:
The basic layout for the new App Detail View. The static data will be replaced continuously.

For testing, you can open the app detail view page `/apps/installed/plex` or `/apps/available/plex` for example, or use other app names.


undsoft:
@denysbutenko I would suggest finishing this up to a point and marking it Ready for review, so that somebody reviews this. Additional improvements can come in another PR.

bvasilenko (approved):
Looks good to me

PR #7640, opened 2023-01-26
NAS-120011 / 23.10 / Do not check for failover.upgrade_pending on non-HA systems

AlexKarpov98:


undsoft (approved):
Appears to be working correctly (at least on our end). I had some other issues with middleware.

PR #7644, opened 2023-01-27
NAS-119773: removed validation pattern for ssh keypair name

Raz-Dva:
Go to Credentials -> Backup-credentials
Change name of existing SSH Keypairs. `Validators.pattern` for this field was removed.
Probably the validator was added by mistake during the refactoring of the component, because it was not in the old code


AlexKarpov98 (approved):
👍

PR #7646, opened 2023-01-27
NAS-120073 / 23.10 / Banning form.get()

undsoft:
Native angular implementation actually doesn't check for anything when you do `form.get('something')`.

Code review is enough.


AlexKarpov98 (approved):
👍

PR #7648, opened 2023-01-27
NAS-119896 / 23.10 / Legacy checkbox -> MatLegacyCheckboxModule to MatCheckboxModule

AlexKarpov98:


src/app/modules/ix-forms/components/ix-checkbox/ix-checkbox.spec.ts
```diff
@@ -26,11 +26,6 @@ describe('IxCheckboxComponent', () => {
       spectator.component.setDisabledState(true);
       expect(spectator.component.isDisabled).toBeTruthy();
     });
-    it('when called with false, input is not disabled', () => {
```
RehanY147, line 29: Why did we remove this test case?

src/assets/styles/other/_material-reduction.scss
```diff
@@ -923,3 +922,7 @@ ix-nested-tree-node > div {
 body .mat-mdc-option {
   min-height: 36px;
 }
+
+body .mat-mdc-checkbox label {
```
RehanY147, line 926: We shouldn't add generic styles at the end of the pile to fix problems. It might be possible to figure out specific instances of the css that's causing issues and remove that. This will force generic styles anywhere in the UI where you want to use the checkbox. Which means more styles will have to be added to this file.

RehanY147 (changes requested):
Inconsistent top and bottom margins. Also, checkbox isn't left aligned. There's some margin/padding to it.
image

src/assets/styles/other/_material-reduction.scss
AlexKarpov98: This is default **mdc** checkbox style which is different from our previous implementation, so that's why we need this fix. And there are no css variable to tweak that, so we need to keep it.

RehanY147 (approved):
Looks good

PR #7655, opened 2023-01-29
NAS-120047 / 22.12-BETA.2 / UI issue on login window with 2FA enabled

AlexKarpov98:
on MASTER, it works fine, problem is only on 22.12 (Bluefin) version.

Screenshot 2023-01-30 at 00 17 16


undsoft (changes requested):
Wrong target branch. 22.12.2 is `stable/bluefin`.

AlexKarpov98:
Will be managed here: https://github.com/truenas/webui/pull/7657

PR #7667, opened 2023-01-31
NAS-113146 / 23.10 / Improve how logout timeout works

bvasilenko:
**Testing**

Here's the scenario:

- After login, a timeout is started for the value set as the lifetime of the token. 

- When an activity is performed, the timeout is reset to zero and starts counting again. 

- If the timeout has expired, a dialog is shown offering to continue the session (a certain time is shown). If the session continuation is confirmed, the timeout is restarted. 

- In case of inactivity or refusal, the value of the token is cleared and a redirect to the login page takes place.

**Note:** you may want to temporarily adjust the code, so the inactivity period would be ~10 seconds


src/app/pages/system/advanced/advanced-settings.component.ts
```diff
@@ -446,6 +450,16 @@ export class AdvancedSettingsComponent implements OnInit, AfterViewInit {
           title: helptextSystemAdvanced.fieldset_sessions,
           tableConf: this.sessionsTableConf,
         },
+        {
+          title: helptextSystemAdvanced.fieldset_token,
+          id: AdvancedCardId.Token,
+          items: [
+            {
+              label: this.translate.instant('Token Lifetime'),
+              value: this.window.localStorage.getItem('lifetime'),
```
undsoft, line 459: I think having this setting in local storage is going to be confusing for the user. We have user preferences (see `preferences.actions.ts`) that you can use here.

src/app/pages/system/advanced/advanced-settings.component.ts
```diff
@@ -446,6 +450,16 @@ export class AdvancedSettingsComponent implements OnInit, AfterViewInit {
...
+            },
+          ],
+        },
```
undsoft, line 462: Also I'd prefer to avoid having a new separate section for just 1 item. Can we combine it with `Sessions` card?
Perhaps something like this:
```
Sessions
Token Lifetime: 300 seconds          [Configure]

Active sessions
- Active sessions table -
[ Terminate other sessions ]

```

src/app/services/token-lifetime.service.ts
```diff
@@ -0,0 +1,69 @@
...
+import { WebSocketService } from 'app/services';
+import { DialogService } from 'app/services/dialog.service';
+
+@UntilDestroy()
+@Injectable({
+  providedIn: 'root',
+})
+export class TokenLifetimeService {
+  protected timeoutToken: Timeout;
+  protected timeoutLogout: Timeout;
```
undsoft, line 15: Confusing names.

src/app/services/token-lifetime.service.ts
```diff
@@ -0,0 +1,69 @@
...
+      this.addListeners();
+    }
+
+    this.stop();
+    const lifetime = Number(this.window.localStorage.getItem('lifetime'));
+    this.timeoutToken = setTimeout(() => {
+      this.timeoutLogout = setTimeout(() => this.ws.logout(), 5000);
+      this.dialogService.confirm({
+        title: this.translate.instant('Logout'),
+        message: this.translate.instant('Are you sure you want to logout?'),
```
undsoft, line 35: I think the messages should be something like this:
> It looks like your session has been inactive for more than x minutes.
> For security reasons we will log you out at <time>.

or 
> For security reasons we will log you out in <active countdown of x> seconds.

And one button: `Extend session`, which resets the whole thing.

src/app/services/token-lifetime.service.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  start(addListeners = false): void {
+    if (addListeners) {
+      this.addListeners();
+    }
+
+    this.stop();
+    const lifetime = Number(this.window.localStorage.getItem('lifetime'));
+    this.timeoutToken = setTimeout(() => {
+      this.timeoutLogout = setTimeout(() => this.ws.logout(), 5000);
```
undsoft, line 32: 5 seconds is not too much time. We should do something like 30 seconds.

src/app/services/token-lifetime.service.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (removeListeners) {
+      this.removeListeners();
+    }
+    if (this.timeoutToken) {
+      clearTimeout(this.timeoutToken);
+    }
+  }
+
+  addListeners(): void {
+    this.window.addEventListener('mouseover', (() => this.start()));
```
undsoft, line 61: I think we should also have `EntityJobComponent` pause session expiration while it is open to account for scenarios when user is staring at the screen waiting for a long job to complete.

src/app/pages/system/advanced/token-settings/token-settings.component.html
```diff
@@ -0,0 +1,31 @@
...
+></ix-modal-header>
+<mat-card>
+  <mat-card-content>
+    <form class="ix-form-container" [formGroup]="form" (submit)="onSubmit()">
+      <ix-fieldset>
+        <ix-input
+          formControlName="token_lifetime"
+          type="number"
+          [label]="'Token Lifetime' | translate"
+          [tooltip]="'Stands for Time To Live of the token, in seconds. The token will be invalidated if the connection has been inactive for a time greater than this.' | translate"
```
undsoft, line 13: Let's change tooltip to:
> Time in seconds after which current user session will be disconnected. Interacting with UI extends the session.

undsoft (changes requested):
Looks good so far, but I will do proper testing after code adjustments.
One scenario to consider though. What would happen if I have two tabs and I'm only interacting with one. Would I be logged out on inactive tab or would inactive tab also log out my current session?

bvasilenko:
@undsoft  Ok, the remarks have been addressed

src/app/pages/system/advanced/token-settings/token-settings.component.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  });
+
+  it('updates lifetime when save is pressed', async () => {
+    const form = await loader.getHarness(IxFormHarness);
+    await form.fillForm({
+      'Token Lifetime': '60',
+    });
+
+    const saveButton = await loader.getHarness(MatButtonHarness.with({ text: 'Save' }));
+    await saveButton.click();
```
undsoft, line 63: Need to actually check that something is being saved. You can spy on `store.dispatch`.

src/app/pages/system/advanced/token-settings/token-settings.component.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  ngOnInit(): void {
+    this.store$.select(selectPreferences).pipe(untilDestroyed(this)).subscribe((preferences) => {
+      this.form.controls.token_lifetime.setValue(preferences.lifetime);
+      this.cdr.markForCheck();
+    });
+  }
+
+  onSubmit(): void {
+    this.window.localStorage.setItem('lifetime', this.form.value.token_lifetime.toString());
```
undsoft, line 41: Why do we need to use localstorage now if you have this value in store?

src/app/pages/system/advanced/token-settings/token-settings.component.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.store$.dispatch(lifetimeTokenUpdated({ lifetime: this.form.value.token_lifetime }));
+    this.store$.dispatch(advancedConfigUpdated());
```
undsoft, line 43: Don't think we need to dispatch `advancedConfigUpdated` here.

src/app/services/token-lifetime.service.ts
```diff
@@ -0,0 +1,94 @@
...
+      if (jobs.length) {
+        this.stop(true);
+      } else if (this.jobsCount) {
+        this.start(true);
+      }
+      this.jobsCount = jobs.length;
+    });
+  }
+
+  start(addListeners = false): void {
```
undsoft, line 45: 1. This method is a bit hard to read, can we break it apart?
2. Can we refactor to not have an argument? I.e. check if listeners are already set, or add new separate methods called `pause` and `resume`?

src/app/services/token-lifetime.service.ts
```diff
@@ -0,0 +1,94 @@
...
+    private dialogService: DialogService,
+    private translate: TranslateService,
+    private ws: WebSocketService,
+    private appStore$: Store<AppState>,
+    private jobStore$: Store<JobSlice>,
+    @Inject(WINDOW) private window: Window,
+  ) {
+    this.startBound = this.start.bind(this);
+
+    this.jobStore$.select(selectJobsPanelSlice).pipe(untilDestroyed(this)).subscribe((jobs) => {
```
undsoft, line 35: I meant looking at whether `EntityJobComponent` is currently being shown.
Just looking at jobs state will stop counter even if you have jobs in the background, which is not intended.

src/app/views/sessions/signin/store/signin.store.ts
```diff
@@ -191,7 +192,8 @@ export class SigninStore extends ComponentStore<SigninState> {
   }
 
   private authenticateWithToken(): Observable<unknown> {
-    return this.ws.call('auth.generate_token', [this.tokenLifetime]).pipe(
+    const tokenLifetime = Number(this.window.localStorage.getItem('lifetime')) || this.tokenLifetimeDefault;
+    return this.ws.call('auth.generate_token', [tokenLifetime]).pipe(
```
undsoft, line 196: I think `tokenLifetime` doesn't matter too much here, as AFAIK any message, including pings will extend its duration.
It's worth checking if this is true and if true, remove localstorage part from here.

undsoft (changes requested):

undsoft:
Also the comment about putting setting in Sessions card (as opposed to creating a new card) was not addressed.

bvasilenko:
All the requested changes should be addressed now

bvasilenko:
Seems like the issues above occurred after updates made to `master` branch. Now it should work as expected

undsoft (approved):
Looks good.

PR #7679, opened 2023-02-01
NAS-119887 / 23.10 / Fixed bug and added validation

RehanY147:
Type something invalid in the name field like 'iSCSI' and you should see the validation errors immediately instead of when you press the submit button on the wizard form.

Also this PR should also fix the `controller is undefined` bug on this wizard that I found. Just opening the form will give you that error in the console window on `master`


AlexKarpov98 (commented):
Code looks good, I could not check name validation on m40 when typed invalid data and as well it was saved in the end with name ` iSCSI` with no problems.

AlexKarpov98 (approved):
It works, I typed incorrect data and seen the error message. 👍 
Tested on latest nightly release.
http://10.238.238.199/ui/sharing/iscsi/wizard

Screenshot 2023-02-06 at 11 15 10

PR #7680, opened 2023-02-01
NAS-120182 / 23.10 / Fixed error throwing my new ws2

RehanY147:
The new `WebsocketConnectionService` wasn't throwing errors properly. It should do that now.

Also, the new method called in the constructor of `SystemGeneralService` is there because not putting it there is causing a race condition of some sort where the signin page keeps loading indefinitely and you can't login. Refreshing the signin page a few times should reproduce that bug.


undsoft (changes requested):
Looks like main dashboard doesn't load on machines in Jenkins. If I go to dashboard and then back, then it loads, but not if first page loaded is main dashboard.

undsoft (approved):
Looks good, appears to be working.

PR #7681, opened 2023-02-02
NAS-120123 / 23.10 / System Settings > Services > S.M.A.R.T > UI flags fields as required on tab through

AlexKarpov98:


denysbutenko (changes requested):
image

The validation error is gone. However, input has missed the clear button when the value is `0`. Also, is a unit test applicable for this change?

denysbutenko (approved):
Perfect 👍🏻

PR #7683, opened 2023-02-02
NAS-120076 / 23.10 / Unable to edit an App

AlexKarpov98:
There was a problem where for some reason **checkIsValidSchedule** received `undefined` or so and caused error.


RehanY147 (changes requested):
The way you've described the problem, it makes me think we should investigate why `undefined` was received. That may be another related bug that needs fixing.

AlexKarpov98:
@RehanY147 
added check for non empty value, so it will not send null or undefined anymore.
As well added **console.error()** statement, to see errors in console for Apps.

RehanY147 (approved):

PR #7686, opened 2023-02-03
NAS-120045 / 23.10 / One Controller's System Serial Number is listed twice and the other is missing in System Settings/General/Support

AlexKarpov98:


denysbutenko (approved):
image

checked on m40 - it looks good.

denysbutenko (changes requested):
Oops. I picked the wrong branch for testing. 

So, I only see two serial numbers in `System General -> Support` and one serial on `Dashboard -> Widget Sys Info`

image

denysbutenko (approved):
Well. Checked it one more time with the correct branch and system.

image
image

It matches expectations.

PR #7687, opened 2023-02-04
NAS-NULL: Translate "scrub" to "清理"

JoeyWNK:
Refer to [ORACLE doc](https://docs.oracle.com/cd/E26926_01/html/E25826/gbbwa.html)
Update words "scrub" to "清理", to Avoid miss understanding on "scrub" as "校验" or "擦除"


AlexKarpov98 (approved):
👍

PR #7694, opened 2023-02-06
NAS-120107 / 23.10 / Similar Apps on App Detail View

denysbutenko:
Shows the first four catalog apps as `Similar Apps` to fulfill the emptiness in the layout.

For testing, you can open the `Apps (WIP) -> Available -> Choose Any App` and check the similar apps on the app detail view.


AlexKarpov98 (approved):
It works:
Screenshot 2023-02-10 at 15 38 44

PR #7701, opened 2023-02-06
NAS-119941 / 22.12.1 / Drives disappear when "Suggest Layout" is used in Create …

undsoft:
Testing: in create pool click Suggest Layout.
If you don't see Suggest Layout, you may need to remove:
```
*ngIf="suggestableDisks.length <= 8"
```


dszidi (approved):
Code looks good and disks are correctly appearing in the listed vdev
LGTM

AlexKarpov98 (commented):
It works good, approved.

PR #7708, opened 2023-02-07
NAS-120245 / 13.0 / UI does not respect user's choice for update file upload …

undsoft:
Testing: see ticket.


dszidi (approved):
Code looks good. Tested on VM and "destination" param is now included in the payload

image

PR #7709, opened 2023-02-07
NAS-120213 / 13.0 / Refresh UI after an update

undsoft:
Testing is complicated and not fully possible. I would just suggest inspecting code.


dszidi (approved):
Looks good

PR #7714, opened 2023-02-08
NAS-120115 / 23.10 / Add url string formatter

denysbutenko:
Details in [the ticket](https://ixsystems.atlassian.net/browse/NAS-120115). For testing, go to `Credentials -> Backup Credentials` and check the `TrueNAS URL` field.


undsoft (changes requested):
Formatters format value from formControl to present it to the user in different form. What you need is a parser to alter value in formControl itself. Currently, while user sees `http://`, actual value submitted doesn't have it.

Let's also check other places where `url` field is used.

src/app/modules/ix-forms/services/ix-formatter.service.ts
```diff
@@ -193,4 +197,16 @@ export class IxFormatterService {
     }
     return (1024 ** (this.iecUnits.indexOf(unitStr)));
   };
+
+  stringAsUrlFormatting = (value: string): string => {
```
undsoft, line 201: This doesn't look correct. Just having a parser should be enough.
I think there is a mistake in `ix-input`, where it doesn't update visible value after it has been parsed.
Please update `ix-input` instead and add a test case to it.

undsoft (changes requested):

src/app/modules/ix-forms/services/ix-formatter.service.ts
denysbutenko: It makes sense. Thanks for bringing this to my attention 👍🏻

undsoft (changes requested):
Whatever parser is doing probably should only happen on blur.
Right now memory inputs are broken (see Bandwidth fields in FTP service settings).

denysbutenko:
> Whatever parser is doing probably should only happen on blur. Right now memory inputs are broken (see Bandwidth fields in FTP service settings).

Good catch. Fixed.

undsoft (approved):
Looks good 👍🏻

denysbutenko:
I stuck with adding this case in unit tests. Reverting the added check.

PR #7731, opened 2023-02-09
NAS-120279 / 23.10 / NAS-117934: Generate token periodically

RehanY147:
Ticket description has an explanation of the bug


undsoft:
Please note that we've recently added a feature where we log user out after certain period of inactivity. https://github.com/truenas/webui/pull/7667
We need to make sure during testing that that functionality still works.

undsoft (changes requested):
Does not build.

dszidi (changes requested):
Would be nice if there was a spec file for auth.service that makes sure things like the token regeneration works.

RehanY147:
@undsoft It's building for me. I can login and use the UI normally. Can you help me reproduce that issue?

RehanY147:
@dszidi NAS-120307 will add tests for the AuthService

src/app/services/auth/auth.service.ts
```diff
@@ -65,6 +68,16 @@ export class AuthService {
     }));
   }
 
+  setupPeriodicTokenGeneration(): void {
+    if (!this.generateTokenSubscription || this.generateTokenSubscription.closed) {
+      this.generateTokenSubscription = timer(0, 290 * 1000).pipe(
```
undsoft, line 73: Let's extract 300 to a variable to explain that this is a default middleware token duration.

src/app/services/auth/auth.service.ts
```diff
@@ -65,6 +68,16 @@ export class AuthService {
...
+        switchMap(() => this.generateToken()),
+        tap((token) => this.token2 = token),
+        untilDestroyed(this),
+      ).subscribe();
+    }
+  }
```
undsoft, line 79: I don't see any code that would clear this subscription. Wouldn't logging out and logging-in again result in two timers?

undsoft (changes requested):

undsoft (approved):
Looks good, nice work.
One thing we may want to do is explicitly set timeout duration when generating a token, just so that it's definitely 300 and not some other value that middleware may change in the future.

dszidi (approved):
Code looks good. Tested and token gets regenerated without any duplicates

RehanY147:
Please review again. Found a bug where if you were to redirect directly to a component which immediately needed the token, the token wouldn't be ready and the component would break.

The new added commit makes it so that when logged in using credentials, token is generated together with the login request. If you want to check the logic, successful login call emits `true` to `isLoggedIn$` which triggers the subscription in the `AuthService` constructor where we call the method to setup token generation. But in the original login method, after issuing `true` to `isLoggedIn$`, it is now waiting on `authToken$` which will emit when eventually we get a token from middleware and only then will the login method complete and the login request complete basically. I explained the logic because I want that reviewed as well.

undsoft (approved):
Still appears to be working correctly.

PR #7733, opened 2023-02-09
NAS-120083 / 22.12.2 / Fixed calls wherever token is used

RehanY147:
The ticket bug was caused by the fact that the token generated by the call `auth.generate_token` expires after 300 seconds. So, at the time of using that token, if it has expired, middleware will give a 403 error.

This PR aims to call `auth.generate_token` to generate a fresh token every time it is used in such a way.


undsoft (approved):
Tested by adjusting token duration in signin.component. Couldn't reproduce the issue.

PR #7737, opened 2023-02-10
NAS-119783 / 13.0 / NAS-119783 - Fixed error "[schedule] A dict was expected" for create a replication task

Raz-Dva:
Create a replication task. 
Set the state to true for the Schedule checkbox.
Edit the task and change the Schedule state to false. 
There should be no errors after saving.

image

### DIFF
--- a/src/app/pages/task-calendar/replication/replication-form/replication-form.component.ts
+++ b/src/app/pages/task-calendar/replication/replication-form/replication-form.component.ts
@@ -818,6 +818,7 @@ export class ReplicationFormComponent {
           name: 'schedule',
           placeholder: helptext.schedule_placeholder,
           tooltip: helptext.schedule_tooltip,
+          value: false,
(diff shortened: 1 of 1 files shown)


bvasilenko (changes requested):
Wasn's able to test.
Hitting a bug:
1. `schedule` property is missing from the query payload after filling *Replication task form and submitting it,
2. when editing existing Replication task and setting **Schedule** to true, then the following error is displayed:
image

bvasilenko:
> Wasn's able to test. Hitting a bug:
> 
> 1. `schedule` property is missing from the query payload after filling *Replication task form and submitting it,
> 2. when editing existing Replication task and setting **Schedule** to true, then the following error is displayed:
> 
> image

This is not actual, due to my mistake during testing.

bvasilenko (approved):
Works for me

PR #7739, opened 2023-02-10
NAS-120264 / 23.10 / Dataset names don't have to begin with an alphanumeric character.

AlexKarpov98:


denysbutenko (changes requested):
The title and description of the ticket need to be more consistent.

image

image

The `name` field accepts non-alphanumeric and alphanumeric values at the beginning. It depends on the goal of the ticket.

AlexKarpov98:
@denysbutenko as an examiner this was sent to me
image

denysbutenko (approved):
image

It works as described. Other non-alphanumeric characters are still allowed.

PR #7748, opened 2023-02-13
NAS-119955 / 23.10 / Network speed mismatch between Key-Max-Mean-Min values

AlexKarpov98:


RehanY147 (changes requested):
Seeing `/s` instead of a meaningful unit. Should be fixed.

RehanY147 (approved):
Looks good

PR #7749, opened 2023-02-13
NAS-120260 / 22.12.2 / Removed mixed disks warning

RehanY147:
Removed the mixed disks warning as the logic was wrong according to middleware

### DIFF
--- a/src/app/enums/v-dev-type.enum.ts
+++ b/src/app/enums/v-dev-type.enum.ts
@@ -27,7 +27,6 @@ export enum CreateVdevLayout {
 export enum TopologyWarning {
   MixedVdevLayout = 'Mixed VDEV types',
   MixedVdevCapacity = 'Mixed VDEV Capacities',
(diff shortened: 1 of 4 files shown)


undsoft:
Eh, it was my understanding that the warning in this specific case was wrong (and needs to be fixed), not that it shouldn't be there at all.

undsoft (changes requested):
See comment.

RehanY147:
@undsoft I talked to Caleb and he said it's best to remove this warning as the logic is all wrong and UI shouldn't decide if mixed disks capacity exists.

RehanY147:
@undsoft I also explained the logic UI had for showing this warning and he still wanted it removed.

undsoft (changes requested):
See ticket comment.

RehanY147:
new PR https://github.com/truenas/webui/pull/7804

PR #7756, opened 2023-02-13
NAS-120326 / 23.10 / Snapshots -> Show Extra columns slide-toggle is not showing correct status

AlexKarpov98:


bvasilenko (approved):
Works good

PR #7763, opened 2023-02-14
NAS-120239 / 23.10 / Remove time sync feature

undsoft:
Unfortunately feature had to be removed. Testing: see that the feature is not there :)


RehanY147 (approved):
Looks good

PR #7726, opened 2023-02-09
NAS-120181 / 23.10 / Scale branding in bottom left and copyright are too small to be readable.

AlexKarpov98:

PR #7765, opened 2023-02-14
NAS-120181: [BACKPORTED] Scale branding in bottom left and copyright are too small to be readable.

AlexKarpov98:


undsoft (approved):
Backported from https://github.com/truenas/webui/pull/7726

PR #7769, opened 2023-02-14
NAS-120264 / 22.12.2 / [BACKPORTED] Dataset names don't have to begin with an alphanumeric character.

AlexKarpov98:


undsoft (approved):
Backported from https://github.com/truenas/webui/pull/7739

PR #7779, opened 2023-02-16
NAS-120349 / 23.10 / Fix stdout tooltip in cron form

denysbutenko:
A small fix. Code review should be enough.


AlexKarpov98 (approved):
👍

PR #7780, opened 2023-02-16
NAS-120240 / 23.10 / Update job object when an error occurs

bvasilenko:
**Testing**

Try replacing a disk and check the validation error mentioned in the ticket


AlexKarpov98 (approved):
👍

PR #7782, opened 2023-02-16
TE-118 / 23.10 / backporting "Adding HA iSCSI on failover test cases" from #7781

ericbsd:


tests/bdd/xpaths.py
```diff
@@ -418,6 +461,10 @@ class progress:
 
 class services:
     title = '//h1[text()="Services"]'
+    iscsi_Service_Checkbox = '//tr[contains(.,"iSCSI")]//mat-checkbox'
```
alexandrabain, line 464: are we able to concatenate the strings i.e. use title in the later strings so if this ever changes we only have to fix it in one place?

markbainix (approved):

alexandrabain (approved):

PR #7786, opened 2023-02-16
NAS-120369 / 23.10 / Rename WebsocketService2 to WebsocketService

undsoft:
Code review is enough.


AlexKarpov98 (approved):
Code looks good! 👍

PR #7788, opened 2023-02-17
NAS-120052 / 23.10 / Migrate Input, Select, Autocomplete, FormField to MDC (related)

AlexKarpov98:


src/app/modules/ix-forms/classes/simple-async-combobox-provider.ts
```diff
@@ -31,5 +31,7 @@ export class SimpleAsyncComboboxProvider implements IxComboboxProvider {
   }
 
   private options: Option[];
-  constructor(private options$: Observable<Option[]>) { }
+  constructor(private options$: Observable<Option[]>) {
+    options$.pipe(take(1)).subscribe((value) => this.options = value);
```
undsoft, line 35: This is incorrect, this class is not designed to load things on its own. If you are concerned about combobox constantly spinning in some forms, then this is another bug: https://ixsystems.atlassian.net/browse/NAS-120116

undsoft (changes requested):
Some minor things look different than what they used to:
- checkbox is now bigger
- icons in filters (like in /sharing/smb) are larger now
- when you hover over a password field (for example in user form) there is a small x icon appears in corner, which we didn't have before.
- when adding a certificate authority, when you go step 3 "subject alternate names" has misaligned text
- add certificate -> extra constraints -> extended key usages. Usages dropdown has misaligned tooltips.

undsoft (changes requested):
Other comments not addressed.

undsoft (approved):
Tooltips in select in Add Certificate -> Extra Constraints -> Key Usage also need to be aligned.
Otherwise looks good.

AlexKarpov98:
Screenshot 2023-02-24 at 13 29 07
now it's fixed

PR #7793, opened 2023-02-18
NAS-120101 / 23.10 / Replace `job status` with `replication task status`

bvasilenko:
**Testing**

On **Data Protection** page at **Replication Tasks** block,

1. Create and run replication task, watch how it's status displayed
2. Use mockup in code to simulate `WAITING` status (see ticket how)


AlexKarpov98 (approved):
State value is now showing correctly. 👍

PR #7795, opened 2023-02-19
NAS-120348 / 23.10 / Refresh dashboard on pool changes

denysbutenko:
For testing, open the main dashboard and focus on the Storage widget in the first tab, open the storage page in the second tab, and focus on pool status. Pick a pool with multiple disks, then make one disk offline in the third tab. Check if data is synced across tabs.

### DIFF
--- a/src/app/pages/storage/stores/pools-dashboard-store.service.ts
+++ b/src/app/pages/storage/stores/pools-dashboard-store.service.ts
@@ -50,6 +50,7 @@ export class PoolsDashboardStore extends ComponentStore<PoolsDashboardState> {
     private sorter: StorageService,
   ) {
     super(initialState);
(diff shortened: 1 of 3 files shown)


src/app/pages/storage/stores/pools-dashboard-store.service.ts
```diff
@@ -50,6 +50,7 @@ export class PoolsDashboardStore extends ComponentStore<PoolsDashboardState> {
     private sorter: StorageService,
   ) {
     super(initialState);
+    this.subscribeToPoolsUpdate();
```
undsoft, line 53: I think it may be better to call this explicitly from outside (or call something like `poolDashboardEntered` that would call `loadDashboard` and this.

PR #7797, opened 2023-02-20
NAS-120357 / 23.10 / Work on sidebar in pool creation wizard

undsoft:
Adds a bunch of various improvements in new pool creation.
Testing: see sidebar in manual selection modal. It is worth checking on a system with enclosure. Find a system in dctrack or message me.

Known issues: enclosure groups do not open on first click.


RehanY147 (approved):
Looks good. Although, if there's no disks to add to the pool, the dialog shouldn't open. But I think that check can be placed higher in the chain of events. Maybe the user shouldn't get to the pool manager page if there's no disks unused.

PR #7798, opened 2023-02-20
NAS-120358 / 23.10 / Error message not shown when incorrect password is used o…

undsoft:
See ticket.


dszidi (approved):
My VM doesn't actually exhibit the loader behavior described in the ticket. With that said, code in PR looks fine and login still works as expected when testing branch.

PR #7799, opened 2023-02-20
NAS-120413 / 23.10 / Enable consistent-return

undsoft:
Code review is enough.

### DIFF
--- a/src/app/pages/system/view-enclosure/classes/system-profiler.ts
+++ b/src/app/pages/system/view-enclosure/classes/system-profiler.ts
@@ -366,7 +366,9 @@ export class SystemProfiler {
   }
 
   rawCapacity(): number {
-    if (!this.diskData || this.diskData.length === 0) { return; }
+    if (!this.diskData || this.diskData.length === 0) {
+      return undefined;
+    }
     let capacity = 0;
     this.diskData.forEach((disk) => {
       if (disk.vdev && disk.vdev.topology === 'data') {
--- a/src/app/services/route-parts/route-parts.service.ts
+++ b/src/app/services/route-parts/route-parts.service.ts
@@ -54,6 +54,8 @@ export class RoutePartsService {
 
       return this.generateRouteParts(child, url, routeParts);
     }
+
+    return undefined;
   }
 
   get routeParts(): RoutePart[] {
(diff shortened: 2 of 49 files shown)


src/app/services/schema/app-schema.service.ts
```diff
@@ -239,7 +239,7 @@ export class AppSchemaService {
         const fieldSchema = _.find(schema?.questions, { variable: key });
 
         if (fieldSchema?.schema?.null) {
-          return null;
+          return;
```
AlexKarpov98, line 242: in this case we need to return `null`.
**return;** -> will end up with **undefined** but we need `null` exactly I assume, because field is **nullable** and we should be able to submit **null**, but not sure about _undefined_.

src/app/services/route-parts/route-parts.service.ts
```diff
@@ -54,6 +54,8 @@ export class RoutePartsService {
 
       return this.generateRouteParts(child, url, routeParts);
     }
+
+    return undefined;
```
AlexKarpov98, line 58: simple **return;** 
should make the same undefined in the end, isn't it?

src/app/pages/system/view-enclosure/classes/system-profiler.ts
```diff
@@ -366,7 +366,9 @@ export class SystemProfiler {
   }
 
   rawCapacity(): number {
-    if (!this.diskData || this.diskData.length === 0) { return; }
+    if (!this.diskData || this.diskData.length === 0) {
+      return undefined;
```
AlexKarpov98, line 370: and here **return;**

AlexKarpov98 (approved):
Overall Looks good. 
Only few questions on **return undefined**, check my comments.

And for **null** to be returned (one comment)

src/app/services/schema/app-schema.service.ts
undsoft: This is within `forEach`. It doesn't return anywhere.

src/app/services/route-parts/route-parts.service.ts
undsoft: Because the way linter rule works is that it should either have all return values specified explicitly, or all returns be just `return` with no value. In here there is a `return this.generateRouteParts(child, url, routeParts);` above that may get returned sometimes.

PR #7803, opened 2023-02-21
NAS-120266 / 23.10 / Dont unsub when dialog is destroyed

RehanY147:
The issue was that we open `StopVmDialogComponent` with `MatDialog` and when the call is made to middleware, all of the subscriptions are set to be destroyed when `StopVmDialogComponent` is destroyed which happens immediately because we are done with that dialog and now we only need the job dialog. So, updated the code to keep the subscriptions alive until the parent component is destroyed.

### DIFF
--- a/src/app/pages/vm/vm-list/stop-vm-dialog/stop-vm-dialog.component.ts
+++ b/src/app/pages/vm/vm-list/stop-vm-dialog/stop-vm-dialog.component.ts
@@ -8,6 +8,7 @@ import { TranslateService } from '@ngx-translate/core';
 import helptext from 'app/helptext/vm/vm-list';
 import { VirtualMachine } from 'app/interfaces/virtual-machine.interface';
 import { EntityJobComponent } from 'app/modules/entity/entity-job/entity-job.component';
+import { VmListComponent } from 'app/pages/vm/vm-list/vm-list.component';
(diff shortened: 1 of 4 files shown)


src/app/pages/vm/vm-list/stop-vm-dialog/stop-vm-dialog.component.ts
```diff
@@ -34,20 +35,20 @@ export class StopVmDialogComponent {
...
       },
     );
-    jobDialogRef.componentInstance.setCall('vm.stop', [this.vm.id, {
+    jobDialogRef.componentInstance.setCall('vm.stop', [this.data.vm.id, {
       force: false,
       force_after_timeout: this.forceAfterTimeoutCheckbox.value,
     }]);
     jobDialogRef.componentInstance.submit();
-    jobDialogRef.componentInstance.success.pipe(untilDestroyed(this)).subscribe(() => {
+    jobDialogRef.componentInstance.success.pipe(untilDestroyed(this.data.parent)).subscribe(() => {
```
bvasilenko, line 47: Ideally, we should move this subscription into the parent `VmListComponent`, however I'm not sure if this is easily achievable

bvasilenko (approved):
Works good to me. I've also shared an idea of improvement

PR #7804, opened 2023-02-21
NAS-120260 / 22.12.2 / Dont use data of disk that doesn't exist

RehanY147:


src/app/services/storage.service.ts
```diff
@@ -463,8 +463,10 @@ export class StorageService {
       const vdevDiskCapacities = new Set<number>(); // There should only be one value
       if (vdev.children.length) {
         vdev.children.forEach((child) => {
-          const diskSize = disks?.find((disk) => disk.name === child.disk)?.size;
-          vdevDiskCapacities.add(diskSize);
+          const disk2 = disks?.find((disk) => disk.name === child.disk);
```
dszidi, line 466: Just a nitpick. "disk2" doesn't seem like an appropriate variable name. Other than that, LGTM

dszidi (approved):

PR #7809, opened 2023-02-23
NAS-120116 / 23.10 / Fix options loading in combobox

denysbutenko:


undsoft (approved):
This resolves the issue with comboboxes and VM wizard.

PR #7817, opened 2023-02-23
NAS-120444 / 13.0 / Passwords don't match error when they do

undsoft:
Testing: see ticket + test some of affected validators.


dszidi (approved):
The code looks fine to me and during testing I didn't experience any erroneous validation errors while creating encrypted datasets with passphrase option.

undsoft:
backport

PR #7818, opened 2023-02-23
NAS-120466 / 23.10 / Better disabled state for search input

denysbutenko:
For testing, please check `ix-input-search` when it is disabled, e.g., while loading.


src/app/modules/common/search-input/search-input.component.ts
```diff
@@ -21,6 +22,10 @@ import { debounceTime, distinctUntilChanged, Subject } from 'rxjs';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class SearchInputComponent implements OnInit, OnChanges {
+  @HostBinding('class.disabled')
+  get isDisabledParentClass(): boolean {
+    return this.disabled;
+  }
   @Input() disabled = false;
```
undsoft, line 29: I think you can just add `@HostBinding` directly to `disabled` input.

PR #7820, opened 2023-02-23
NAS-120387 / 23.10 / When loading currently running jobs for task manager, exclude the ones that have `transient: true`

AlexKarpov98:
New **Boolean** field was added to the API.
Check this:
https://github.com/truenas/middleware/pull/10714/files


denysbutenko (approved):
👍🏻

PR #7822, opened 2023-02-24
NAS-120090 / 23.10 / Implement the Apps Installation page

bvasilenko:
**Testing**

Please test the new Install App page


dszidi (commented):
@bvasilenko It's off to a great start. There is however an issue that needs to be addressed. The sticky card on the right with the glossary needs to be altered for screen widths less than 960px.

I would suggest reducing the contents to just the search bar and placing it above the form. It should never overlap the form and should also stick to the top so it is always visible. Here are a couple of screens to help illustrate what I'm looking for.

image

image

bvasilenko:
@dszidi Ok, should be fixed now

dszidi (approved):
LGTM

PR #7824, opened 2023-02-24
NAS-120435: Button-field for type in "Reporting" disapears when pressing "Reporting" again

AlexKarpov98:


undsoft (changes requested):
I don't know if this entirely correct. What if I'm going from `/apps` to `/apps/something`?

undsoft (approved):
Works correctly.

PR #7828, opened 2023-02-24
NAS-120494 / 23.10 / Make page header match width of content below

dszidi:
Test: Check to make sure bottom border of page header element lines up with content below


bvasilenko (approved):
Looks good to me

image

PR #7829, opened 2023-02-25
NAS-120503 / 23.10 / Fixed potential race condition

RehanY147:


undsoft (commented):
Are you sure there will be a race condition? I mean it feels the code between sending a message and filtering for incoming messages is synchronous.

RehanY147:
> Are you sure there will be a race condition? I mean it feels the code between sending a message and filtering for incoming messages is synchronous.

There was a race condition in the original PR. Then it got fixed but I feel like the fix still leaves a tiny fraction of a chance for a race condition. Also, previously, the old WebsocketService also did things like this PR. Currently, the code is in two separate steps, step 1 - send a request. step 2- start listening for responses. This PR makes that process 1 step. Starting listening to websocket responses AND send a request.
I feel like this is similar to state updates in react/redux if you're relying on the old state. Very little chance that it'll be a race condition, but it's still a chance.

undsoft:
I would suggest using a debugger to check whether race condition is possible here. In mind my it should send a message, then synchonously switchMap in the same call stack. It should not be possible to for ws$ to emit anything before switchMap statement is reached.

undsoft:
To rewrite it to pseudocode:
```
function call(): void {
  this.wsManager.sendMessage();
  this.ws$.subscribe(() => {});
}
```
It should not be possible for anything to happen between those two statements.

RehanY147:
After discussion, agreed that the race condition is no longer an issue after a recent PR.

PR #7830, opened 2023-02-25
NAS-120307 / 23.10 / Added tests for `AuthService`

RehanY147:
The tests check for all scenarios when you login via either username/password or with a token. Also checks scenarios for logout.

This PR also fixes the race condition potential in AuthService that was fixed in WebsocketService by https://github.com/truenas/webui/pull/7829

### DIFF
--- a/src/app/services/auth/auth.service.spec.ts
+++ b/src/app/services/auth/auth.service.spec.ts
@@ -1,0 +1,59 @@
+import { createServiceFactory, SpectatorService } from '@ngneat/spectator';
+import { of } from 'rxjs';
+import { TestScheduler } from 'rxjs/testing';
+import { IncomingApiMessageType } from 'app/enums/api-message-type.enum';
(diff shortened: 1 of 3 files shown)


src/app/services/auth/auth.service.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+    strategyStub = new StorageStrategyStub(LocalStorageStrategy.strategyName);
+    spectator = createService();
+
+    testScheduler = new TestScheduler((actual, expected) => {
+      return expect(actual).toEqual(expected);
+    });
+  });
+
+  describe('Login', () => {
+    it('initalizes auth session with triggers and token with username/password login', () => {
```
undsoft, line 70: Typo in initalizes

undsoft (approved):
Tests looks good, as for potential condition, I'd like to see proof that it is actually an issue. In my mind the code will be executed synchronously.

PR #7831, opened 2023-02-25
NAS-120157 / 22.12.2 / Added the new field

RehanY147:
The new field is selected if the user wants to create a new directory in their home directory with their username.
The tooltip should be checked. Also, a warning should be shown that the permissions will change on the home directory based on the following conditions.
`user.create`  call without `home_create` should warn
`user.update` call without `home_create` and with home_mode set to non-NULL should warn
`user.update` call with new path without `home_create` should warn (regardless of `home_mode`)


src/app/pages/account/users/user-form/user-form.component.ts
```diff
@@ -116,6 +120,37 @@ export class UserFormComponent {
...
   readonly groupProvider = new SimpleAsyncComboboxProvider(this.groupOptions$);
 
+  get homeCreateWarning(): string {
+    const homeCreate = this.form.get('home_create').value;
+    const home = this.form.get('home').value;
+    const homeMode = this.form.get('home_mode').value;
+    if (this.isNewUser) {
+      if (!homeCreate && home !== '/nonexistent') {
+        return this.translate.instant(
+          'With this configuration, Home Directory under path {path} will not be created.',
```
anodos325, line 130: Maybe:
"With this configuration, the existing directory {path} will be used a home directory without creating a new directory for the user."

RehanY147: Thanks! I'll make the change.

AlexKarpov98:
@RehanY147 
Screenshot 2023-02-28 at 15 11 38

I didn't get any warning on CREATE user with `Create Home Directory` unchecked.

RehanY147:
@AlexKarpov98 You have to select a home directory path that's not `non-existent` default value. Can you confirm if you did that please?

AlexKarpov98 (approved):
Seems like it's working. But I see a lot of errors in console, maybe some problem on latest master?
Screenshot 2023-03-01 at 12 03 33

RehanY147:
@AlexKarpov98 Yes. That doesn't seem related to this PR.

PR #7833, opened 2023-02-27
NAS-120376 / 23.10 / Storage - Manage Devices incorrectly presents the "EXTEND button for member disks in RAIDZ VDEVs"

AlexKarpov98:


src/app/pages/storage/modules/devices/components/zfs-info-card/zfs-info-card.component.ts
```diff
@@ -35,6 +33,13 @@ export class ZfsInfoCardComponent {
     return this.topologyItem.type === TopologyItemType.Mirror;
   }
 
+  get isRAIDZParent(): boolean {
```
undsoft, line 36: isRaidzParent

RehanY147 (approved):
Looks good. But Evgeny's comments should be addressed.

PR #7839, opened 2023-02-27
NAS-120381 / 22.12.2 / Calculate slot when no disk is present

dszidi:
Testing requires physical access to hardware
- Enclosure UI visualization should update when disk is added or removed (could take up to 10 seconds or so)
- Verify that the color coding on drivetrays is correct and matches physical location of disk

Do this on both MINI products as well as Enterprise models

### DIFF
--- a/src/app/pages/system/view-enclosure/classes/chassis-view.ts
+++ b/src/app/pages/system/view-enclosure/classes/chassis-view.ts
@@ -288,16 +288,19 @@ export class ChassisView {
   }
 
   colorDriveTray(options: ChangeDriveTrayOptions): void {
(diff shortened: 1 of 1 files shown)


AlexKarpov98:
@dszidi I'm not able to test this on physical  hardware.

dszidi:
> @dszidi I'm not able to test this on physical hardware. We should assign someone with needed access.

Yep. I unassigned you

AlexKarpov98 (approved):
👍

dszidi:
@AlexKarpov98 as per @undsoft 's instruction, no testing required. Just code review

PR #7843, opened 2023-02-28
NAS-120489 / 23.10 / [SCALE/Apps]: Setting  on a  causes it to be locked even at install time.

AlexKarpov98:


dszidi (commented):
Code looks fine to me. For testing do we simply ensure the name field is available at install and locked at edit?

AlexKarpov98:
> Code looks fine to me. For testing do we simply ensure the name field is available at install and locked at edit?

@dszidi - yes, absolutely.

dszidi (approved):
In that case LGTM

PR #7845, opened 2023-02-28
NAS-120530 / 23.10 / Remove Auxiliary Params from SMB Service and SMB Shares

AlexKarpov98:


undsoft (approved):
Looks good, auxiliary params are gone.

PR #7847, opened 2023-03-01
NAS-120469 / 23.10 / Kerberos Realms Screen Action Dropdown has Two Add Options

AlexKarpov98:
Go to `credentials/directory-services`  then `directoryservice/kerberosrealms`
In header there should be only one `Add` button
Screenshot 2023-03-01 at 13 06 24


bvasilenko (approved):
Works good for me

PR #7849, opened 2023-03-01
NAS-120510 / 23.10 / Nextcloud cron settings were not recognized during installation.

AlexKarpov98:


dszidi (changes requested):
Code looks fine so far.
We should really be validating the cron syntax in the form field though.

dszidi (approved):

dszidi:
After some discussion it was decided to address the field validation in another ticket. LGTM

PR #7853, opened 2023-03-02
NAS-117934 / 23.10 / New ErrorHandlerService

RehanY147:


src/app/app.module.ts
```diff
@@ -131,6 +132,10 @@ import { RoutePartsService } from './services/route-parts/route-parts.service';
     IxSlideInService,
     IxFileUploadService,
     DisksUpdateService,
+    {
+      provide: ErrorHandler,
+      useClass: ErrorHandlerService,
+    },
```
undsoft, line 138: Feels like this would be overridden just a couple lines below.

src/app/modules/common/dialog/multi-error-dialog/error-template/error-template.component.ts
```diff
@@ -0,0 +1,77 @@
...
+  selector: 'ix-error-template',
+  templateUrl: './error-template.component.html',
+  styleUrls: ['./error-template.component.scss'],
+})
+export class ErrorTemplateComponent {
+  @ViewChild('errorMessageWrapper') errorMessageWrapper: ElementRef;
+  @ViewChild('errorTitle') errorTitle: ElementRef;
+  @ViewChild('errorMdContent') errorMdContent: ElementRef;
+  @ViewChild('errorBtPanel') errorBtPanel: ElementRef;
+  @ViewChild('errorBtText') errorBtText: ElementRef;
```
undsoft, line 22: You can use `ElementRef<HtmlElement>` to avoid specifying type below.

src/app/services/error-handler.service.ts
```diff
@@ -0,0 +1,217 @@
...
+    if (this.isTypeOfJobError(error)) {
+      const errors = this.parseJobError(error);
+      this.dialog.error(errors);
+      return;
+    }
+    console.error('Error', error);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  isTypeOfWebsocketError(obj: any): boolean {
```
undsoft, line 35: Can be changed to:
```
isTypeOfWebsocketError(error: unknown): error is WebsocketError {
    return typeof error === 'object'
      && 'error' in error
      && 'extra' in error
      && 'reason' in error
      && 'trace' in error;
```

src/app/services/error-handler.service.ts
```diff
@@ -0,0 +1,217 @@
...
+            title: this.translate.instant('Error'),
+            message: item,
+          });
+        });
+      }
+    });
+    return errors;
+  }
+
+  parseHttpError(error: HttpErrorResponse): ErrorReport | ErrorReport[] {
```
undsoft, line 182: Unused?

src/app/services/error-handler.service.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  parseWsError(error: WebsocketError): ErrorReport | ErrorReport[] {
+    return this.parseErrorOrJob(error);
+  }
+
+  reportJobError(failedJob: Job): Observable<boolean> {
+    return this.dialog.error(this.parseJobError(failedJob));
+  }
+
+  parseJobError(failedJob: Job): ErrorReport | ErrorReport[] {
```
undsoft, line 113: Maybe error parsing needs to be extracted somewhere to clear up this file a little bit.
Also see `sentryCustomExtractor`, we already have some error parsing there. It may make sense to combine these two somehow.

src/app/services/error-handler.service.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  reportWsError(error: WebsocketError): Observable<boolean> {
+    return this.dialog.error(this.parseWsError(error));
+  }
+
+  parseWsError(error: WebsocketError): ErrorReport | ErrorReport[] {
+    return this.parseErrorOrJob(error);
+  }
+
+  reportJobError(failedJob: Job): Observable<boolean> {
```
undsoft, line 109: `parseWsError` is the same as `parseErrorOrJob`, so why not just rename the latter?
Also since both `reportWsError` and `reportJobError` essentially do the same thing, we don't really need the distinction.

undsoft (commented):
Reviewed the code only, will do some testing later.

src/app/services/error-handler.service.ts
undsoft: Still unused.

src/app/services/error-handler.service.ts
```diff
@@ -0,0 +1,249 @@
...
+import { Observable } from 'rxjs';
+import { ResponseErrorType } from 'app/enums/response-error-type.enum';
+import { sentryCustomExceptionExtraction } from 'app/helpers/error-parser.helper';
+import { ErrorReport } from 'app/interfaces/error-report.interface';
+import { Job } from 'app/interfaces/job.interface';
+import { WebsocketError } from 'app/interfaces/websocket-error.interface';
+import { DialogService } from 'app/services';
+
+@Injectable()
+export class ErrorHandlerService implements ErrorHandler {
```
undsoft, line 15: This class mixes a lot of responsibilities:
* Default Angular error handler (Sentry reporting).
* Parsing errors.
* Showing form errors.
* Reporting generic errors.

It feels like these should be 4 separate things, not one.

undsoft (changes requested):
Cannot test in `manager.component.ts`. I just get `No provider for ErrorHandlerService!`.

src/app/services/error-handler.service.ts
```diff
@@ -0,0 +1,249 @@
...
+  dialogService: DialogService;
+  translateService: TranslateService;
+  get translate(): TranslateService {
```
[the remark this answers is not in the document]
RehanY147: I'll address the rest of the comments. But the reason for this is that ErrorHandler is supposed to be ready before any other general services because all code naturally depends on ErrorHandler to be ready. However, adding dependencies in the ErrorHandler means now ErrorHandler is dependent on those services which results in a circular dependency. That's why, we get services instances lazily with the custom injection.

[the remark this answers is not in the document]
RehanY147: Also, the service won't necessarily be available when the `ErrorHandlerService` service is constructed.

RehanY147:
@undsoft `parseHttpError` method is expected to be used later on when we fully convert to the new errorhandler. Replacing usages all over the UI is a tedious task so first I want to confirm how exactly we want to go about using the new errorhandler. Once we've decided on that and tested it, then we can replace old code with the new one. I'll do refactoring as well. Please review how we want to use the new error handler.
Also, regarding your last comment, I can't reproduce that issue. The error handler is picking up the errors for me. Can you test again? Or provide steps to reproduce.

undsoft (changes requested):
Okay, I don't get injector error for some reason, but I get another error when trying to create a pool with incorrect name: `Cannot add property extra, object is not extensible`.

With regards to parseHttpError - my question is whether we would ever need this functionality, since we are mostly relying on websocket connection.

undsoft:
Disregard my previous comment about it being not necessary to parse http error codes. Sometimes it's necessary, like when  getting an error when uploading a file. I just got a similar error in `IxFileUploadService.onUploading$` because of incorrect path. Default error handler should be able to handle HttpErrorResponse as well.

src/app/services/error-handler.service.ts
```diff
@@ -0,0 +1,261 @@
...
+      && 'exc_info' in obj);
+  }
+
+  /**
+   * @param error
+   * @param formGroup
+   * @param fieldsMap Overrides backend field names with frontend field names.
+   * TODO: See if second `string` in fieldsMap can be typed to key of formGroup.
+   */
+  handleWsFormError(
```
undsoft, line 75: Will we keep this error handler here or will it belong to a separate FormErrorHandlerService. I feel like latter may be preferrable.

undsoft (changes requested):
Same error as last time: `Cannot add property extra, object is not extensible`

RehanY147:
It's showing the error dialog now

https://user-images.githubusercontent.com/24204846/227049429-f707a2e9-97f4-4a1b-8ac0-afe4fdcab44e.mov

undsoft (approved):
Okay, looks good. Let's merge it and start testing it in action.

PR #7857, opened 2023-03-03
Validate iSCSI Target Name field

parrotmac:
When creating a new iSCSI target the name field has some rules that aren't reflected until you reach the end of the wizard at which point you may find that the name was invalid. While these rules are probably known by someone familiar with iSCSI, it's not necessarily user-friendly as-is.

I was trying to find a way to show a more helpful validation error when the validation rule was violated, but gave up and instead added the specificity to the tooltip. Also, happy to port this to the latest!

Screenshot 2023-03-02 at 10 07 54 PM


RehanY147 (approved):
Looks good. Thank you.

PR #7858, opened 2023-03-03
NAS-120537 / 23.10 / If S.M.A.R.T. test status is ABORTED, do not treat it as a pool health issue

bvasilenko:
**Testing**

Run SMART test for a disk. Abort the test by rebooting, while the test is still running

Most probably VM box won't allow to run SMART test, so please use mockup: simulate an aborted test putting `'ABORTED'` into the `status` field.

**Expected Result:**

A warning icon should not be displayed on **Storage** page at **Disk Health** block

### DIFF
--- a/src/app/pages/storage/components/dashboard-pool/disk-health-card/disk-health-card.component.ts
+++ b/src/app/pages/storage/components/dashboard-pool/disk-health-card/disk-health-card.component.ts
@@ -76,7 +76,7 @@ export class DiskHealthCardComponent implements OnInit, OnChanges {
   }
 
   get iconType(): PoolCardIconType {
-    if (this.diskState.alerts || this.diskState.smartTests) {
+    if (this.diskState.smartTests) {
       return PoolCardIconType.Warn;
     }
     return PoolCardIconType.Safe;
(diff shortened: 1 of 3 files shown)


src/app/pages/storage/components/dashboard-pool/disk-health-card/disk-health-card.component.ts
```diff
@@ -76,7 +76,7 @@ export class DiskHealthCardComponent implements OnInit, OnChanges {
   }
 
   get iconType(): PoolCardIconType {
-    if (this.diskState.alerts || this.diskState.smartTests) {
+    if (this.diskState.smartTests) {
```
dszidi, line 79: By suppressing the icon for all alerts, it also won't show when there are temperature related alerts.
We should probably be trying to filter the test status instead

dszidi (changes requested):
See comment above

src/app/pages/storage/components/dashboard-pool/disk-health-card/disk-health-card.component.ts
bvasilenko: I totally agree, thanks for noticing that

bvasilenko:
@dszidi WebUI code will work fine without any change to it's logic. The `ABORTED` statuses will be filtered out by an existing `.filter` call at `getProcessedDisks` method.

undsoft:
@bvasilenko So why was this issue reported?

bvasilenko:
> @bvasilenko So why was this issue reported?

The issue was reported [here](https://ixsystems.atlassian.net/browse/NAS-120498), and fixed [here](https://github.com/truenas/middleware/pull/10764) in the middleware. I don't believe the issue persisted after this PR was merged.

PR #7859, opened 2023-03-03
NAS-120492 / 23.10 / [SCALE/Apps]: Initial value of a dropdown is null, but after picking something and reverting to empty is no longer null

AlexKarpov98:
For testing you can use machine from ticket.

Go to apps -> Available Apps -> search for **minio**
click **Install** 

Under **Minio Configuration** configuration there is:

Test Certificate. 
Once you select a certificate, there should appear `Test cert` string field.
And if you select `No certificate` then `Test cert` string input field should disappear.

Screenshot 2023-03-03 at 11 29 13

Or contact @stavros-k (he knows how to setup on any machine)


undsoft (approved):
Looks good.

PR #7860, opened 2023-03-03
NAS-120580 / 23.10 / Apps cron input with type "string" no validation on cron field

AlexKarpov98:
For testing you can use: 
mini3xlp-101.dc1.ixsystems.net (machine)
creds: **root / abcd1234** 

Go to apps -> Available Apps -> search for **nextcloud**
click **Install** 

Under **CronJob configuration** configuration there is:

Schedule of a type string.
Try to put some invalid cron expression and you should see validation error.

Screenshot 2023-03-03 at 12 37 54

Or contact @stavros-k (he knows how to setup on any machine)

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
     "chart.js": "~2.9.4",
     "core-js": "~3.6.4",
     "cron-parser": "~4.2.1",
-    "cron-validator": "~1.3.1",
     "croner": "~4.2.3",
     "cronstrue": "~1.113.0",
     "d3": "~5.16.0",
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "chart.js": "~2.9.4",
     "core-js": "~3.6.4",
     "cron-parser": "~4.2.1",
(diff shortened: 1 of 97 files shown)


package.json
```diff
@@ -89,6 +89,7 @@
     "chart.js": "~2.9.4",
     "core-js": "~3.6.4",
     "cron-parser": "~4.2.1",
+    "cron-validator": "~1.3.1",
```
undsoft, line 92: We can keep it, but it was possible to use `cron-parser` to check if cron is valid. See `CrontabPartValidatorService`.

undsoft (approved):
Looks good, validation works.

package.json
AlexKarpov98: I'd like to keep cron-validator, since all it is only responsible for is - cron validation, returning `:boolean`, which I used in other places as well.

AlexKarpov98: @undsoft - I removed `cron-validator`

undsoft (approved):
Looks good.

PR #7861, opened 2023-03-03
NAS-120438   Limiting Web Login and API access for security hardening

bvasilenko:
**Testing**

1. Go to **Advanced settings** and setup IP filtering using newly added UI block

   **WARNING:** be careful not to loose access to your remote box! Make sure you have some other machine, which has a different IP address, before going to the next step.

2. Block your machine from accessing the WebUI

Expected Result:

1. When attempting to view any page of WebUI , you should see "Your IP is restricted" error message like shown below

image

2. UI for setting up IP rules should function well. Try saving/loading with different input (being careful not to loose the connection!)


AlexKarpov98:
Screenshot 2023-03-03 at 16 08 09

Form missing some paddings, isn't it?

AlexKarpov98:
We need to manage IP address validation, currently I can type anything I want and it counts valid:
Screenshot 2023-03-03 at 16 10 09

AlexKarpov98 (changes requested):
Please review my comments above. 🔝 

As well I think about the UI changes.

Since we are limited visiting any page, I infinitely see 
`Connecting to TrueNAS ... Make sure the TrueNAS system is powered on and connected to the network.`

Screenshot 2023-03-03 at 16 20 30

SO, maybe it's better to just override the current visible window to show current modal content as current viewport (window) content instead of modal?

End result like this: (seeing only on whole window)

autodraw 03_03_2023

AlexKarpov98:
@undsoft  @dszidi - please let me know your thoughts as well regarding my latest comment.

undsoft:
I agree, would be nice to have something better than the ordinary modal.

AlexKarpov98 (approved):
Looks good to me. 👍 

https://user-images.githubusercontent.com/22980553/223564342-52598cb1-ae4e-429e-b501-ee34294eb270.mov

Screenshot 2023-03-08 at 00 10 34

PR #7867, opened 2023-03-06
NAS-120560 / 23.10 / UI should not treat catalog.items as a job

AlexKarpov98:


undsoft (changes requested):
Typing is not correct. It still seems to return `CatalogItems`, and `as unknown` casts are unnecessary.

AlexKarpov98:
@undsoft, yep, you are right.

undsoft (approved):
Looks good.

PR #7868, opened 2023-03-06
NAS-120626: Adapt to disk.get_unused changes

undsoft:
Also improves on `setup_env.js`.
For testing: setup_env can be tested.


AlexKarpov98 (approved):
Screenshot 2023-03-07 at 15 33 55

It worked for me.

PR #7875, opened 2023-03-07
NAS-120361 / 22.12.2 / Mini R Enclosure View should not have Identify Drive button

bvasilenko:
See ticket, use machine http://minir-101.dc1.ixsystems.net


undsoft (approved):
Identify button is not there.

PR #7876, opened 2023-03-07
(NAS-120533) UI does not allow for EXTEND operations on single-device special VDEVs

chrisperedun:
added OR conditions for canExtendDisk function to present Extend option for single-disk Dedup and Special VDEVs


AlexKarpov98:
We'll continue here: https://github.com/truenas/webui/pull/7883/files

PR #7880, opened 2023-03-08
NAS-120606 / 22.12.2 / Sudo checkbox in replication wizard

undsoft:
https://user-images.githubusercontent.com/1481270/223810303-9dd47e64-0447-41b7-be47-52544a81ae76.mov


src/app/helptext/data-protection/replication/replication-wizard.ts
```diff
@@ -48,6 +48,8 @@ export default {
  <code>/zvol1</code> after <i>dataset1</i> will create <i>zvol1</i> for \
  snapshot storage.'),
 
+  sudo_warning: T(`Selected SSH connection uses non-root user. Would you like to use passwordless sudo with <i>/usr/sbin/zfs</i> commands?
```
themylogin, line 51: We should say "Would you like to use sudo with <i>/usr/sbin/zfs</i> commands? Passwordless `sudo` must be enabled on the remote system" because this option does not enable passwordless sudo on the remote system (like it might seem from this dialog test

themylogin (changes requested):

themylogin (approved):

PR #7883, opened 2023-03-09
NAS-120533 / 23.10 / UI does not allow for EXTEND operations on single-device special VDEVs

AlexKarpov98:
We checked the logic together with **Chris Peredun**.
I think code review is enough.

### DIFF
--- a/src/app/pages/storage/modules/devices/stores/devices-store.service.ts
+++ b/src/app/pages/storage/modules/devices/stores/devices-store.service.ts
@@ -78,7 +78,7 @@ export class DevicesStore extends ComponentStore<DevicesState> {
         return this.ws.call('pool.query', [[['id', '=', poolId]]]).pipe(
           switchMap((pools) => {
             // TODO: Handle pool not found.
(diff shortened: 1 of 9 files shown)


src/app/pages/storage/modules/devices/stores/devices-store.service.ts
```diff
@@ -78,7 +78,7 @@ export class DevicesStore extends ComponentStore<DevicesState> {
         return this.ws.call('pool.query', [[['id', '=', poolId]]]).pipe(
           switchMap((pools) => {
             // TODO: Handle pool not found.
-            return this.ws.call('disk.query', [[['pool', '=', pools[0].name]], { extra: { pools: true } }]).pipe(
+            return this.ws.call('disk.query', [[['pool', '=', pools?.[0]?.name]], { extra: { pools: true } }]).pipe(
```
RehanY147, line 81: If this data can be falsy, we should check beforehand. Otherwise, we will just make an invalid request to middleware which is guaranteed to fail.

RehanY147 (changes requested):

RehanY147 (approved):

PR #7884, opened 2023-03-09
NAS-119863: Load the "Available Apps" dashboard

AlexKarpov98:
In order to check this you need to connect to one of the latest nightlies.
like this `10.238.238.172` machine.

Go Apps WIP -> Available 

Screenshot 2023-03-09 at 14 02 42

You gonna see apps split by category.
Regarding the designs we have: 
`Recommended Apps` and `New & Updated Apps`
And we should show Recommented apps only if there are so (flagged with `recommended: true`)

We should show max 6 elements and by clicking view more for these 2 categories -> show all items.

If we are watching custom category and there are more then 6 items, View should trigger form filtering and submit to such categories page.
Currently this logic is not implemented and out of scope of this task.

Screenshot 2023-03-09 at 14 04 58

### DIFF
--- a/src/app/pages/apps/components/app-detail-view/app-detail-view.component.html
+++ b/src/app/pages/apps/components/app-detail-view/app-detail-view.component.html
@@ -12,7 +12,7 @@
         <a
           mat-button
           color="primary"
-          [ixTest]="[app.name, 'install']"
+          [ixTest]="[app?.name, 'install']"
           [disabled]="!app" (click)="onInstallButtonPressed()"
         >{{ 'Install' | translate }}</a>
       </div>
--- a/src/app/pages/apps/components/available-apps/available-apps.component.html
+++ b/src/app/pages/apps/components/available-apps/available-apps.component.html
@@ -11,35 +11,28 @@
 
 <ix-available-apps-header></ix-available-apps-header>
 
-<div class="available-apps-list">
-  <ng-container *ngTemplateOutlet="appsSection; context: {
-    title: 'Recommended Apps',
-    apps: recommendedApps
-  }"></ng-container>
+<div *ngIf="apps.length" class="available-apps-list">
+  <ng-container *ngFor="let appSection of appSections; trackBy: trackByAppSectionTitle">
+    <section *ngIf="appSection.apps$ | async as apps">
+      <ng-container *ngIf="apps.length">
+        <h3 class="apps-section-title">
+          {{ appSection.title | translate }}
+        </h3>
 
-  <ng-container *ngTemplateOutlet="appsSection; context: {
-    title: 'New & Updated Apps',
-    apps: newAndUpdatedApps
-  }"></ng-container>
+        <div class="apps">
+          <ix-app-card
+            *ngFor="let app of apps; trackBy: trackByAppId"
+            [app]="app"
+            [routerLink]="['/apps', 'available', app.name]"
+          ></ix-app-card>
+        </div>
+
+        <p *ngIf="apps.length <= sliceAmount" class="view-all" (click)="appSection.onViewMore()">
+          <b>
+            {{ 'View All' | translate }} {{ appSection.title | translate }}
+          </b>
+        </p>
+      </ng-container>
+    </section>
+  </ng-container>
 </div>
-
-<ng-template #appsSection let-title="title" let-apps="apps">
-  <section *ngIf="apps.length">
-    <h3 class="apps-section-title">
-      {{ title | translate }}
-    </h3>
-    <div class="apps">
-      <ix-app-card
-        *ngFor="let app of apps; trackBy: trackByAppId"
-        [app]="app"
-        [routerLink]="['/apps', 'available', app.name]"
-      ></ix-app-card>
-    </div>
-    <p class="view-all">
-      <b>
-        {{ 'View All' | translate }}
-        {{ title | translate }}
-      </b>
-    </p>
-  </section>
-</ng-template>
--- a/src/app/pages/apps/components/available-apps/available-apps.component.html
+++ b/src/app/pages/apps/components/available-apps/available-apps.component.html
@@ -11,10 +11,35 @@
 
 <ix-available-apps-header></ix-available-apps-header>
 
-<div class="apps">
-  <ix-app-card
-    *ngFor="let app of apps; trackBy: trackByAppId"
-    [app]="app"
-    [routerLink]="['/apps', 'available', app.name]"
-  ></ix-app-card>
+<div class="available-apps-list">
+  <ng-container *ngTemplateOutlet="appsSection; context: {
+    title: 'Recommended Apps',
+    apps: recommendedApps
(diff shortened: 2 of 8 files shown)


dszidi (commented):
When I click "View all from `<category-name>`"  it will show all from category but it also still shows the other category summaries underneath. 

image

This should not be the case. If the user clicks the "View all Networking" button, the only thing on the page should be apps from that category. 

Beyond that, the grouping of app cards by category and such seems to be in working order.
Well done

src/app/pages/apps/components/app-detail-view/app-detail-view.component.html
```diff
@@ -12,7 +12,7 @@
         <a
           mat-button
           color="primary"
-          [ixTest]="[app.name, 'install']"
+          [ixTest]="[app?.name, 'install']"
           [disabled]="!app" (click)="onInstallButtonPressed()"
```
RehanY147, line 16: We want to avoid using method/variable names prefixed with `on`

src/app/pages/apps/components/available-apps/available-apps.component.html
```diff
@@ -11,10 +11,28 @@
...
+
+        <div class="apps">
+          <ix-app-card
+            *ngFor="let app of apps; trackBy: trackByAppId"
+            [app]="app"
+            [routerLink]="['/apps', 'available', app.name]"
+          ></ix-app-card>
+        </div>
+
+        <p *ngIf="appSection.totalApps > sliceAmount" class="view-all" (click)="appSection.onViewMore()">
```
RehanY147, line 30: `onViewMore` -> `viewMore` or `fetchMore`

src/app/pages/apps/components/available-apps/available-apps.component.ts
```diff
@@ -66,16 +81,65 @@ export class AvailableAppsComponent implements OnInit, AfterViewInit {
...
-      catalogToAppsTransform(),
-      untilDestroyed(this),
-    ).subscribe((apps) => {
-      this.apps = apps;
-      this.loader.close();
-      this.cdr.markForCheck();
+
+    combineLatest([
+      this.appService.getAllCatalogs().pipe(
+        catalogToAppsTransform(),
```
RehanY147, line 93: I think `mapCatalogsToApps` would be a better name for this method

src/app/pages/apps/components/available-apps/available-apps.component.ts
```diff
@@ -66,16 +81,65 @@ export class AvailableAppsComponent implements OnInit, AfterViewInit {
...
+        this.loader.close();
+        this.cdr.markForCheck();
+      });
+  }
+
+  private setupApps(apps: CatalogApp[], appCategories: string[]): void {
+    this.apps = apps;
+
+    this.allRecommendedApps = this.apps.filter((app) => app.recommended);
+    this.allNewAndUpdatedApps = this.apps
```
RehanY147, line 109: Wouldn't this just sort all apps in an order. Meaning all apps are technically "new and updated". Just further down the list. We should have a threshold. Like maybe apps last updated this week or this month. Or apps with versions newer than what the user has installed or something similar. Figuring this out maybe more difficult. We can discuss and create a new ticket if it deviates too much from the scope of this one.

src/app/pages/apps/components/available-apps/available-apps.component.ts
```diff
@@ -66,16 +81,65 @@ export class AvailableAppsComponent implements OnInit, AfterViewInit {
...
+      .sort((a, b) => new Date(a.last_update).getTime() - new Date(b.last_update).getTime());
+
+    this.recommendedApps$.next(this.allRecommendedApps.slice(0, this.sliceAmount));
+    this.newAndUpdatedApps$.next(this.allNewAndUpdatedApps.slice(0, this.sliceAmount));
+
+    this.appSections.push(
+      {
+        title: 'Recommended Apps',
```
RehanY147, line 117: Translate

src/app/pages/apps/components/available-apps/available-apps.component.ts
```diff
@@ -66,16 +81,65 @@ export class AvailableAppsComponent implements OnInit, AfterViewInit {
...
+        apps$: this.recommendedApps$,
+        totalApps: this.allNewAndUpdatedApps.length,
+        onViewMore: () => this.recommendedApps$.next(this.allRecommendedApps),
+      },
+      {
+        title: 'New & Updated Apps',
```
RehanY147, line 123: Translate

RehanY147 (changes requested):
Can't access the machine linked and on the latest nightlie, I get an "app service not available" error from middleware.

RehanY147 (changes requested):
Please note that there's a new call available `app.available` which will list down all of the apps and their data in one call. Furthermore, this call can be used to filter data based on some parameters. We want to use this call instead of `catalog.query`.

src/app/pages/apps/components/available-apps/available-apps.component.ts
AlexKarpov98: maybe we should show latest updated apps with 6 in limit with no possibility to show more?

I think if we want to improve it we can do it in terms of new ticket.

RehanY147 (changes requested):
We want to keep the UI consistent so showing the remaining apps for the first two categories on the same page and possibly showing all other categories different is not the way to go. Let's just wait for the next ticket to figure out how these buttons should behave for all categories.

src/app/pages/apps/components/available-apps/available-apps.component.ts
```diff
@@ -66,16 +82,61 @@ export class AvailableAppsComponent implements OnInit, AfterViewInit {
...
+    this.allRecommendedApps = this.apps.filter((app) => app.recommended);
+    this.allNewAndUpdatedApps = this.apps
+      .sort((a, b) => new Date(a.last_update).getTime() - new Date(b.last_update).getTime());
+
+    this.recommendedApps$.next(this.allRecommendedApps.slice(0, this.sliceAmount));
+    this.newAndUpdatedApps$.next(this.allNewAndUpdatedApps.slice(0, this.sliceAmount));
+
+    this.appSections.push(
+      {
+        title: this.translate.instant('Recommended Apps'),
```
RehanY147, line 113: Let's add a `category` field in these objects to easily identify which categories this section to belong to. For recommended apps use `recommended` and for new and updated apps use `new` or something similar. In the future when we implement the filter, we can add these two options as checkboxes to only show those apps on the filtered results if the use wants to filter by those. We can decide this after discussion.

RehanY147 (changes requested):

RehanY147 (approved):
This looks good for now. We can create a new ticket to handle those concerns.

AlexKarpov98:
> This looks good for now. We can create a new ticket to handle those concerns.

Can you create these tickets, since you know what is needed ?

PR #7887, opened 2023-03-09
NAS-120669 / 22.12.2 / Fixing an error with HA status

undsoft:
This just fixes a console error, but probably does nothing to fix error in the ticket. Code review is enough, will continue testing.


AlexKarpov98 (approved):
is ok

PR #7890, opened 2023-03-10
NAS-120471 / 23.10 / Refactor VM wizard

undsoft:
Testing: test VM form.


dszidi (commented):
I'm having trouble with the file uploader

image

undsoft:
@dszidi There is a problem with how we handle http errors right now. This should get fixed in another ticket. Please try a different path instead of /mnt.

dszidi (approved):
Successfully created a VM. Validators all seem to work and I didn't notice any regressions

PR #7895, opened 2023-03-13
App Details Panel and Cards

denysbutenko:
You can check the new apps page and select any installed apps for testing.


dszidi (commented):
Off to a good start. Cards are present and update on table row click as expected.
The application card is missing the Admin Portal button and the Update button but otherwise looks ok.

The Containers card looks good but the shell action needs to be moved. We don't invoke a shell for the app as each app can have multiple containers. For this reason we should remove the "Shell" button from the card header and provide one for each individual pod list items. I can adjust the existing mockup to reflect that change if need be.

The History card looks good and I'm assuming the endpoint for Notes is not ready yet? Those should be provided by middleware

denysbutenko:
> Off to a good start. Cards are present and update on table row click as expected. The application card is missing the Admin Portal button and the Update button but otherwise looks ok.
 
The `Web/Admin Portal` buttons are coming from the `app.portals` property and appear dynamically. 
The `Update` button relies on `app.update_available` property. I going to sync the condition with the old apps module. Thanks for pointing that out.

> The Containers card looks good but the shell action needs to be moved. We don't invoke a shell for the app as each app can have multiple containers. For this reason we should remove the "Shell" button from the card header and provide one for each individual pod list items. I can adjust the existing mockup to reflect that change if need be.

It makes sense, agreed.

> The History card looks good and I'm assuming the endpoint for Notes is not ready yet? Those should be provided by middleware

Yeah, I'll replace static content when new data is available.

denysbutenko:
@dszidi ok, I've addressed the feedback. The `Shell` and `View Logs` buttons will appear as icon buttons with a tooltip. Although the dialogs require further work to accept pods outside, we will make sure to update them at a later time. I am open to suggestions about changing icons or the approach itself if needed.

Preview of changes:
image

denysbutenko:
> We should remove the underlines from the icons
> 
> image

Oh yeah, it was removed.

> > The Update button relies on app.update_available property. I going to sync the condition with the old apps module.
> 
> Will this be done for this PR or as a separate ticket item? Currently the update button is still missing

The condition was updated to include both parameters - `update_available` and `container_images_update_available`. Perhaps, I misunderstood the requirements. The initial assumption was to show the button when an update is available and hide it when there are no updates. However, I made changes to make the button visible all time and disabled it when there were no updates.

dszidi (changes requested):
> ... Perhaps, I misunderstood the requirements. The initial assumption was to show the button when an update is available and hide it when there are no updates. However, I made changes to make the button visible all time and disabled it when there were no updates.

Apologies. You didn't actually misunderstand the requirements. While testing I mistakenly thought an update was available for the app I was looking at and the button wasn't showing. On closer inspection it turns out there wasn't an update and your button was doing the right thing. No need to show the update button if there isn't one available.

Beyond that, the rest looks good to me

image

denysbutenko:
> > ... Perhaps, I misunderstood the requirements. The initial assumption was to show the button when an update is available and hide it when there are no updates. However, I made changes to make the button visible all time and disabled it when there were no updates.
> 
> Apologies. You didn't actually misunderstand the requirements. While testing I mistakenly thought an update was available for the app I was looking at and the button wasn't showing. On closer inspection it turns out there wasn't an update and your button was doing the right thing. No need to show the update button if there isn't one available.
> 
> Beyond that, the rest looks good to me
> 
> image

Ok, no problem. Fixed that.

dszidi (commented):
After having updated Plex using bulk actions, the app details card still shows the update button. When I press it gives an error dialog

image

denysbutenko:
> After having updated Plex using bulk actions, the app details card still shows the update button. When I press it gives an error dialog
> 
> image

Good catch, @dszidi. The dialog invokes refreshing the data on close, but catching the finishing update event is tricky. It should not be an issue when the app data store is implemented. I'll keep that in mind for future improvements.

dszidi (approved):
LGTM

PR #7901, opened 2023-03-14
NAS-120988 / 23.10 / Fix the installed badge on the new AppCard

AlexKarpov98:


undsoft (approved):
Looks good, checked via code.

PR #7911, opened 2023-03-15
NAS-121042 / 23.10 / Refactoring certificate authority wizard

undsoft:
Test CA wizard.
I've moved some controls to different steps to make my life easier.


denysbutenko (approved):
image

CA Wizard looks good and works well.

Two things worth mentioning:

1. The `Name` field needs a duplicate name validator. 
2. The `Email` field needs to be synced with middleware rules, e.g. `test@test` is valid for UI but not for middleware.

PR #7917, opened 2023-03-16
NAS-121043 / 23.10 / webUI is showing incorrect warning

AlexKarpov98:
Testing:
Credentials -> Local Users:
Try to update user 
IF Home Directory is `/nonexistent` we don't need to show warning modal and simply update user. 

Screenshot 2023-03-16 at 12 26 25

Screenshot 2023-03-16 at 12 28 31

Once we changed it (Home Directory) -> Modal should warn us.
Screenshot 2023-03-16 at 12 27 44


bvasilenko (approved):
Works good to me!

PR #7919, opened 2023-03-16
NAS-121066 / 23.10 / add to LOC_FAILOVER_ONGOING HA disabled reasons

yocalebo:
See https://github.com/truenas/middleware/pull/10877 for details. This should, hopefully, ease confusion when end-user initiates a manual failover and is presented the login page. Instead of showing all reasons why HA is unhealthy (which is expected since user initiated the failover process), it'll give a more concise message to the end-user.


undsoft (approved):
We cool.

yocalebo:
The test failed to retreive from codecov website with 500 internal server error so seems unrelated to my changes. I'm merging.

undsoft:
backport

PR #7927, opened 2023-03-16
NAS-121045 / 23.10 / Refactoring CSR wizard

undsoft:
Testing: compare new CSR wizard with old CSR wizard.
You can find instructions on how to generate a CSR online.


RehanY147 (approved):
The slide-in width seems shorter than what the norm is. Otherwise, looks Ok.

PR #7929, opened 2023-03-17
NAS-121007 / 23.10 / Improve selected app state on Installed Apps

denysbutenko:
For testing, go to the installed apps page and select any installed app - check the URL and page title. It must reflect the chosen app like in datasets.


bvasilenko (approved):
Works good for me

PR #7930, opened 2023-03-17
NAS-119754 / 23.10 / CPU Mode Displays on Edit VM Form for All CPU Mode Options

AlexKarpov98:
For testing you can use any machine where you can set up VM.
I used this one:
**mini3xlp-101.dc1.ixsystems.net**


denysbutenko (approved):
Changing `CPU Mode` from `Custom` to either `Host` option removes the `CPU Model` field in the VM form.

PR #7934, opened 2023-03-17
NAS-121054 / 22.12.2 / Hide identify button for R30

dszidi:
To test go to enclosure UI and click on a disk. Underneath the drive icon there should not be an "Identify Drive" button under the disk icon. Look in the ticket comments for a test machine url


undsoft (approved):
Looks good.

PR #7947, opened 2023-03-20
NAS-121107 / 23.10 / Remove high CPU usage indicator

undsoft:
Some people complained about it and to be fair high CPU usage is not an error.


bvasilenko (approved):
LGTM!

PR #7950, opened 2023-03-20
NAS-120655: Errors in authorized hosts are not displayed in Web UI during NFS share create or edit

AlexKarpov98:
Check `/sharing/nfs` route.

Click ADD 

And add invalid hosts -> submit.

You should see validation message.

Screenshot 2023-03-20 at 14 39 32

As well I fixed network input: 

Screenshot 2023-03-20 at 14 45 51


denysbutenko (approved):
image

image

The validation works as described.

PR #7960, opened 2023-03-21
NAS-121116 / 23.10 / Advanced Options on Replication Page dumps user back to Replication Page

AlexKarpov98:
Follow steps in ticket for testing.


undsoft (approved):
Great success!

PR #7967, opened 2023-03-22
NAS-121128 / 23.10 / Reporting > CPU - chart only ever shows in %

AlexKarpov98:
decided:
get rid of cpu_in_percentage and graphite_separateinstances from UI side.

Screenshot 2023-03-22 at 15 46 15

**! NOTE !**
Screenshot 2023-03-22 at 15 56 57


bvasilenko (approved):
Looks and works good

image


image

PR #7968, opened 2023-03-22
TE-333 / 22.12.2 / fixed User test fixed Apps test while loop and removed machinaris app test

ericbsd:
machinaris got  removed from the default Apps


tests/bdd/reusableSeleniumCode.py
```diff
@@ -170,6 +170,35 @@ def Verify_The_Dashboard(driver):
...
+
+        timeout = time.time() + 30
+        while time.time() < timeout:
+            assert wait_on_element(driver, 10, '//span[contains(.,"Refresh Events")]', 'clickable')
+            driver.find_element_by_xpath('//span[contains(.,"Refresh Events")]').click()
+            # make sure Please wait pop up is gone before continuing.
+            if wait_on_element(driver, 3, xpaths.popup.please_Wait):
+                assert wait_on_element_disappear(driver, 10, xpaths.popup.please_Wait)
+
+            if is_element_present(driver, f'//div[contains(text(),"Started container {app}")]'):
```
alexandrabain, line 191: Need to have a delay function in here so it doesn't continuously poll without a break. needs to be  1 second

alexandrabain (approved):

PR #7977, opened 2023-03-23
NAS-121105: Load state not working on the User create/edit form

AlexKarpov98:


denysbutenko (approved):
When I confirm that dialog, the save button on the original form becomes disabled, and I see a blue loading bar at the top. When saving is unsuccessful, the save button is still disabled, and a validation error appears below the field. It works as expected.

PR #7987, opened 2023-03-24
Update System Info after changing Localization Settings

bvasilenko:
**Testing**

Please see issue replication steps, described in the ticket


AlexKarpov98 (approved):
Tested with timezone changing -> Dashboard `System information` card showed correct time.

PR #7989, opened 2023-03-24
NAS-121124 / 23.10 / Terminal is not resized properly when shell reconnects

AlexKarpov98:
For testing Follow ticket steps.


src/app/modules/terminal/components/terminal/terminal.component.ts
```diff
@@ -105,14 +105,17 @@ export class TerminalComponent implements OnInit, AfterViewInit, OnDestroy {
...
   initShell(): void {
-    this.authService.authToken$.pipe(
-      tap((token) => {
-        this.initializeWebShell(token);
-        this.shellService.shellOutput.pipe(untilDestroyed(this)).subscribe(() => {});
-        this.initializeTerminal();
-      }),
-      untilDestroyed(this),
-    ).subscribe();
+    setTimeout(() => {
```
RehanY147, line 108: Why is this needed? We should avoid this kind of simple workarounds.

RehanY147 (changes requested):

src/app/modules/terminal/components/terminal/terminal.component.ts
AlexKarpov98: yeah, sorry

RehanY147 (changes requested):
Still seeing this issue. Made sure PR was up to date.

RehanY147:
https://user-images.githubusercontent.com/24204846/228993915-2a67a5fc-639d-4c0c-b185-ab4d0c959e5f.mov

AlexKarpov98:
@RehanY147 - which machine do you use?
I used m40 here, please check.
Screen Recording 2023-03-31 at 13.31.35.mov

RehanY147 (changes requested):
I can still recreate this issue but also, there's another issue that's in this PR but not in master. When I logout intentionally with the button in the UI, I can't use any other username other than `root` to log back in. The `Login` button stays disabled and only enables for `root` username. There are other usernames now which can login like `admin`.

https://user-images.githubusercontent.com/24204846/229638448-53e37261-eaa4-48ec-a87e-dc1dd1743d27.mov

src/app/services/auth/auth.service.ts
```diff
@@ -37,9 +37,10 @@ export class AuthService {
    */
   readonly tokenRegenerationTimeMillis = 290 * 1000;
 
-  private readonly latestTokenGenerated$ = new ReplaySubject<string>(1);
+  private readonly latestTokenGenerated$ = new Subject<string>();
+
   get authToken$(): Observable<string> {
-    return this.latestTokenGenerated$.asObservable();
+    return of(this.token);
```
RehanY147, line 43: The `ReplaySubject` was used for if you stay on this shell page for longer than 5 minutes without doing anything, the token needs to be refreshed. This will only emit the token once when it's used and subsequently, the token won't be emitted and shell will be expired. We have to come up with another solution.

RehanY147 (changes requested):
Please confirm that with this change, as mentioned below, wait on the shell for 5-10 minutes, try to use the shell after that and see if it still works.

src/app/services/auth/auth.service.ts
AlexKarpov98: You mean we don't need to see this modal?
Screenshot 2023-04-05 at 13 00 39

AlexKarpov98: @RehanY147

AlexKarpov98:
I was not able to find a correct solution, so returning this ticket to the queue.

PR #7993, opened 2023-03-26
NAS-120029 / 23.10 / Added multichannel field to form

RehanY147:


AlexKarpov98 (approved):
Field was successfully added. 👍

PR #7995, opened 2023-03-27
NAS-121106 / 23.10 / More improvements for the installed apps

denysbutenko:
To do:

- [x] The color of the status tiles should match the designs (e.g., yellow doesn’t match)
   - [x] yellow
   - [x] green
   - [x] red
- [x] Don’t see the catalog name under the app name
- [x] “Action” column name missing
- [x] The table can take more width on the page as only single column app cards are required
- [x] Font-size differences in text like app name from the designs
- [x] The bulk-actions button and button text doesn’t match the designs
- [x] If no apps are to be updated, the updated all button at the top right should be disabled
- [x] Add a better empty state for when there are no apps installed
- [x] Show the app status in the `Update` column
- [x] Move `ix-app-row` styles to the appropriate file
- [x] Update secondary page title style to match other headers
- [x] Select the first app on page load


denysbutenko:
Better empty states:

Kapture 2023-03-27 at 21 59 06

Kapture 2023-03-27 at 22 11 17

image

image

Note: The search still needs to be implemented.

src/app/pages/apps/components/installed-apps/app-containers-card/app-containers-card.component.ts
```diff
@@ -0,0 +1,102 @@
...
+  constructor(
+    private appService: ApplicationsService,
+    private cdr: ChangeDetectorRef,
+    private matDialog: MatDialog,
+    private router: Router,
+    private translate: TranslateService,
+  ) {}
+
+  ngOnInit(): void {
+    this.getResources();
```
RehanY147, line 36: This is not needed. `ngOnChanges` will trigger for the first time already.

RehanY147 (commented):
Little confused, Is this PR ready for review?

denysbutenko:
@RehanY147 The branch has been rebased, and the PR is cleaned up for a simple review.

RehanY147 (changes requested):
We can give these status tiles some more margin.
image

Also, column names and their values don't seem perfectly aligned. We should make them aligned vertically.

Furthermore, the app name text font is larger than column headers. It seems so in the designs as well but in the designs, the app name text has muted grey color. Should talk to Damian on what's the best way forward. As is, the app name takes attention away from the header.

The items under the `Action` column aren't center aligned. They are center aligned in the designs.

In the designs, the table title and the App Info title have more top-margin than what's there now. When you select a row and bulk actions show, the text and drop down seem too close to the top divider. We should have a bit more separation there if not exactly not like the designs.

The text of Bulk Actions has a larger font than the designs. As does the text inside the dropdown it feels like.

We can improve the message on empty state where it says `click here to see the TrueNAS catalog to get started`. We should say, `click here to see all the available apps to get started`.

Selected app should have a "selected" state to show that it's selected. Some indication like a highlighter that shows this row is selected.

denysbutenko:
> We can give these status tiles some more margin. image

Do you mean to add some vertical space between columns?

> Also, column names and their values don't seem perfectly aligned. We should make them aligned vertically.
> 

Some regression in this branch. It was fixed.

> Furthermore, the app name text font is larger than column headers. It seems so in the designs as well but in the designs, the app name text has muted grey color. Should talk to Damian on what's the best way forward. As is, the app name takes attention away from the header.
> 

That's true. There are many differences in fonts, sizes, and colors between design mockups and the default theme in WebUI. For example,

- `h1` font-size is `35px` in mockups and `30.478px` in webui.
- `h2` font-size is `24.19px` in mockups and `25.40px` in webui.
- Side Menu items is `14px` in mockups and `12.25px` in webui.

As for colors, I believe we should use existing variables for theme compatibility.
The font size requires a complex solution and can be handled in a separate task.

> The items under the `Action` column aren't center aligned. They are center aligned in the designs.
> 

I made changes to align the text and buttons in the last column by center.

> In the designs, the table title and the App Info title have more top-margin than what's there now. When you select a row and bulk actions show, the text and drop down seem too close to the top divider. We should have a bit more separation there if not exactly not like the designs.
> 

Done 👍🏻 

> The text of Bulk Actions has a larger font than the designs. As does the text inside the dropdown it feels like.
> 

It was the default font size of the button. The font size was decreased for this element only.

> We can improve the message on empty state where it says `click here to see the TrueNAS catalog to get started`. We should say, `click here to see all the available apps to get started`.

Thanks for the suggestion. The text has been updated.

> Selected app should have a "selected" state to show that it's selected. Some indication like a highlighter that shows this row is selected.

The regression was fixed.

RehanY147:
> Do you mean to add some vertical space between columns?
No, I mean horizontal margins.

> As for colors, I believe we should use existing variables for theme compatibility.
I'm not recommending that we change any colors or add any custom colors. I'm saying that app name having larger font-size only works if the text color is muted. Since, we don't have a muted color for text that is supported by themes, we shouldn't have app name font-size larger than table headers. It should be at max the same size. Or smaller. We can discuss with @dszidi

RehanY147 (changes requested):
Add a bit of horizontal margin to the `Action` column from the right border.

I think we can reduce the margins around the app name column as app names aren't going to be that long. We can add a hover effect to show the name so if it is too long, we can show it separately.
Also I think the margin is now a bit too much around the state tiles column. We want the margin there, just we can reduce it a little bit.
Please note that these column values don't look center aligned. They look left aligned. We want them to be center aligned.
image

When the screen is shortened, the apps info cards disappear and don't reappear when you select a row. Also, the columns are misaligned again.
image

RehanY147:
@denysbutenko if you want, we can merge some changes and you can create a new ticket for the remaining changes so you can avoid merge conflicts as the other PRs get merged into `master`.

denysbutenko:
>>Do you mean to add some vertical space between columns?

> No, I mean horizontal margins.

Ok, it was changed.

> I'm not recommending that we change any colors or add any custom colors. I'm saying that app name having larger font-size only works if the text color is muted. Since, we don't have a muted color for text that is supported by themes, we shouldn't have app name font-size larger than table headers. It should be at max the same size. Or smaller. We can discuss with @dszidi 

I reverted the font size to default values – `12.25px` for the app name. And keep `10px` for the catalog.


> Add a bit of horizontal margin to the `Action` column from the right border.
>

Ok, it was changed.

> I think we can reduce the margins around the app name column as app names aren't going to be that long. We can add a hover effect to show the name so if it is too long, we can show it separately.

The name column fills available space depending on other columns. If you want to make it fixed, please let me know more details.

> Also I think the margin is now a bit too much around the state tiles column. We want the margin there, just we can reduce it a little bit. 

Ok, it changed. This column preserves a space for max possible width when the status is deploying. If you have another value in mind, please let me know.

> Please note that these column values don't look center aligned. They look left aligned. We want them to be center aligned. image
> 

They left aligned exactly like in the design. This sounds like a new requirement to me. Let's return to this when we get real values for these columns.
image

> When the screen is shortened, the apps info cards disappear and don't reappear when you select a row. Also, the columns are misaligned again. image

Yes, the styles were partially copied from another place, and the mobile details view needs to be implemented. Could you create a ticket for this?

@RehanY147 I'm good to finish the remaining changes in this PR. As for new requirements, it would be better to split the work. Thanks for the feedback.

RehanY147 (approved):
Looks good

PR #7997, opened 2023-03-27
NAS-119556 / 22.12.3 / Hidden button when not needed

RehanY147:


undsoft (approved):
Looks good. I think having a root certificate needs to be also included in condition in the server form.

PR #7998, opened 2023-03-28
NAS-120550 / 23.10 / Exclude unnecessary connection via WebSocket

bvasilenko:
Two timers were simultaneously running to reconnect at `onClose`. I added an additional check for the existence of a reconnect attempt. The unnecessary connection via WebSocket does not occur now. Reloading the page every minute is now resolved


undsoft (approved):
It appears to be better - UI doesn't constantly reload when session times out.

PR #7999, opened 2023-03-28
NAS-121183 / 23.10 / Hide `Enable Atime` for Zvol

denysbutenko:
From the original ticket:
> Since Enable Atime does not really apply to Zvol datasets, that line should be removed from the "Zvol Details" section of the following TrueNAS SCALE page

For testing, go to the datasets page and check if it does meet requirements


AlexKarpov98 (approved):
Works as described.

PR #8003, opened 2023-03-28
NAS-121218 / 23.10 / Datasets table header is not sticked with rows

AlexKarpov98:
Before:
https://user-images.githubusercontent.com/22980553/228266752-53e41bd9-1e93-4119-9629-b3da301a9883.mov

After:
https://user-images.githubusercontent.com/22980553/228266851-0a81a26f-5a42-4f70-9b12-6f46061452cb.mov


RehanY147 (approved):
Headers scroll with the table now. Nice!

PR #8010, opened 2023-03-29
NAS-121060 / 23.10 / Difficult to click on the username field of the login page

AlexKarpov98:
problem was that `.cdk-overlay-pane` was taking place over input with z-index: 1000


RehanY147 (approved):
Looks good. Can't reproduce the issue.

PR #8011, opened 2023-03-29
NAS-120691 / 23.10 / Add component store to enclosure ui

dszidi:
Replaces old system.profiler with component store. During refactor other improvements have been added and are listed below

- empty slots are now clickable
- slot selections will display both the vdev type and vdev name
- store has a spec file with a couple of basic tests (this can be expanded later)
- storage generator now also generates enclosure data
- storage generator makes use of templates to provide accurate simulated middleware responses
- various CSS fixes
- Improved the Elements button. (It is now hidden if only disks view is available)
- replaced spinner with itemized loader for extreme systems where middleware might take a while to gather all the data

Please test this against the following machines/scenarios

- Mini (both the mini R and non-rackmount versions)
- M50/M60 rear drives are visualized (these report rear bays as separate chassis but store is merging them into one enclosure object for consistency)
- systems with shelves
- Check that loader works on extreme systems

TODO:
- Responding to disk.query update events when disks are added/removed *** DONE
- Cleaning up any leftover comments **** Mostly done. Still have some siblings related items (requires access to a real MINI-R)


src/app/pages/system/view-enclosure/stores/enclosure-store.service.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+      mockProvider(WebSocketService, {
+        subscribe: jest.fn(() => websocketSubscription$),
+      }),
+      mockProvider(DialogService),
+    ],
+  });
+
+  beforeEach(() => {
+    spectator = createService();
+    testScheduler = getTestScheduler();
```
undsoft, line 41: You only need test scheduler if you want to test observables with `expectObservable` or something similar.
Otherwise you can just use `mockWebsocket` with. `mockCall` like in other tests.

src/app/pages/system/view-enclosure/stores/enclosure-store.service.ts
```diff
@@ -0,0 +1,481 @@
...
+      enclosureViews: this.getEnclosureViewsData().pipe(
+        this.patchStateWithEnclosureViewsData(),
+      ),
+      selectedEnclosure: this.getSelectedEnclosure().pipe(
+        this.patchStateWithSelectedEnclosure(),
+      ),
+    });
+  }
+
+  getSelectedEnclosure(): Observable<number> {
```
undsoft, line 115: Unnecessary. Just just be part of initial state.

src/app/pages/system/view-enclosure/stores/enclosure-store.service.ts
```diff
@@ -0,0 +1,481 @@
...
+  });
+
+  updateState(): Observable<{
+    enclosures: Enclosure[];
+    pools: Pool[];
+    disks: Disk[];
+    enclosureViews: EnclosureView[];
+  }> {
+    return forkJoin({
+      enclosures: this.getEnclosures().pipe(
```
undsoft, line 97: I would just simplify to private `getEnclosures` that would have necessary logic to update state/log errors in it.

src/app/pages/system/view-enclosure/stores/enclosure-store.service.ts
```diff
@@ -0,0 +1,481 @@
...
+          disk: enclosureDisk,
+          isSelected: false,
+          enclosure: enclosure.number,
+          slot: slotSource.slot,
+          slotStatus: slotSource.status,
+          fault: slotSource.fault,
+          identify: slotSource.identify,
+          pool: pool ? pool.name : null,
+          vdev: pool ? topologyInfo.vdev : null,
+          topologyStatus: enclosureDisk ? 'AVAILABLE' : null,
```
undsoft, line 259: Enum?

undsoft (changes requested):
There
Some things that I've noticed on our m40.
1. Selecting `sdp` doesn't show status (previously shown Available).
2. If you click anywhere on black background before selecting a disk, and then try to select first blue disk, it won't be selected.

Would be good to clean up unused code before I continue reviewing further.

undsoft (approved):
Looks good.

src/app/core/testing/utils/mock-storage-generator.utils.ts
```diff
@@ -528,10 +551,113 @@ export class MockStorageGenerator {
...
+    }
+
+    return chassis;
+  }
+
+  private populateEnclosures(mockEnclosures: MockEnclosure[], dispersal: EnclosureDispersalStrategy): Enclosure[] {
+    let populated: Enclosure[] = [];
+
+    if (dispersal === EnclosureDispersalStrategy.Min) {
+      //
```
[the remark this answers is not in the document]
dszidi: Yes. I wanted to add dispersal support as a feature for storage generator but it seems out of scope for this refactor. Can I leave it in with a TODO comment or should I remove it?

PR #8012, opened 2023-03-30
NAS-121171 / 23.10 / Using the new `ErrorHandlerService`

RehanY147:


src/app/modules/entity/utils.ts
```diff
@@ -24,152 +17,156 @@ export interface FlattenedData extends Record<string, unknown> {
 /**
  * TODO: Likely outdated.
  */
-interface HttpError {
-  code: number;
-  error: {
-    [field: string]: string[];
-  };
-}
+// interface HttpError {
```
dszidi, line 20: We should delete these if not in use

dszidi (changes requested):
The code looks reasonable. The only thing I would change is
- ditch the commented out EntityUtils code.
- Add some spec files to test make sure the ErrorHandlerService produces desired behavior.

RehanY147:
Tests will be added by https://ixsystems.atlassian.net/browse/NAS-121305

dszidi (approved):
Tested by triggering the dialog after init on snapshots list page. 
LGTM

image

PR #8014, opened 2023-03-30
NAS-121176 / 23.10 / Do not allow to terminate current session

denysbutenko:
**Description:**
> Now: 
> 1. I go to Advanced Settings.
> 2. Click terminate next to my current session.
> 3. Get a warning saying that current session cannot be terminated.
> 
> Instead we should communicate this to the user earlier. We should show Terminate icon for current session as disabled with a tooltip saying that current session cannot be terminated.

**Testing:**

Go to `Settings -> Advanced` take a look on the `Sessions` card and check if the button to terminate current session is disabled. Make sure others sessions can be terminated.

### DIFF
--- a/src/app/modules/entity/table/table.component.html
+++ b/src/app/modules/entity/table/table.component.html
@@ -209,7 +209,7 @@
                     <button
                       mat-icon-button
                       [ixTest]="['row-action', _tableConf.name, rowAction.name]"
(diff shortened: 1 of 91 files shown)


AlexKarpov98 (approved):
works well, `terminate session button` stays disabled for current session.

PR #8015, opened 2023-03-30
NAS-120337 / 23.10 / Redirect to newly created dataset or zvol

denysbutenko:
**Description**
> Change so that when dataset or zvol is created, user is redirected to it.
> Also add a success toast with “Switched to new dataset {name}” or “Switched to new zvol {name}”

**Testing**

1. Go to the Datasets page
2. Create a new zvol/dataset
3. Check desired behavior


bvasilenko (approved):
Works good for me!

PR #8016, opened 2023-03-30
NAS-121249 / 23.10 / Fix update and removal in stores

bvasilenko:
**Testing**

There are subscribers that modify application data, depending on the value of the 'msg' field.

For example, for **'zfs.snapshot.query'**:
image

There are 11 such places:
image

So, please verify these subscribers for 7 endpoints:
1. **'alert.list'**
2. **'core.get_jobs'**
3. **'group.query'**
4. **'user.query'**
5. **'container.image.query'**
6. **'pool.query'**
7. **'zfs.snapshot.query'**

For example, for **'zfs.snapshot.query'**:

https://user-images.githubusercontent.com/20611516/228942907-1e300116-e184-4395-b499-5a6ddc4977e7.mp4

For **'group.query'**:

https://user-images.githubusercontent.com/20611516/228942956-4df977cb-6d65-4a3f-b968-4db2c5d8a3df.mp4

PR #8017, opened 2023-03-30
NAS-121249 / 22.12.3 / Fix update and removal in stores

bvasilenko:
**Testing**

See original PR: https://github.com/truenas/webui/pull/8016

### DIFF
--- a/src/app/interfaces/api-event-directory.interface.ts
+++ b/src/app/interfaces/api-event-directory.interface.ts
@@ -1,8 +1,9 @@
 import { FailoverDisabledReason } from 'app/enums/failover-disabled-reason.enum';
 import { Alert } from 'app/interfaces/alert.interface';
 import { ChartRelease } from 'app/interfaces/chart-release.interface';
-import { PullContainerImageResponse, PullContainerImageParams } from 'app/interfaces/container-image.interface';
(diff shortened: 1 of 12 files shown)


undsoft (changes requested):
I get
```
instrument.js:109 TypeError: Cannot read properties of undefined (reading 'id')
    at alert.reducer.ts:50:15
```
when I dismiss an alert using one of the Bluefin machine from Jenkins.

bvasilenko:
@undsoft Now the bug should be fixed

undsoft (changes requested):
Deleting a user or a group does not refresh the table.

bvasilenko:
@undsoft Ok, now it should be fixed

undsoft (changes requested):
I get
```
TypeError: Cannot read properties of undefined (reading 'name')
    at snapshot.reducer.ts:35:18
```
when I attempt to delete a snapshot.

bvasilenko:
@undsoft 

I kindly apologize for back and forth.

After thorough testing on bluefin version of the middleware, it turned out that none of the fixes made here are applicable to the bluefin: when deleting, middleware responds with 2 messages - `removed` and `changed` with the `cleared` field (which is obsolete in post-bluefin versions)

So no need to backport these... I'm closing this PR now.

PR #8019, opened 2023-03-31
NAS-121193 / 23.10 / Split GeneralSettingsComponent into separate components

AlexKarpov98:


src/app/pages/system/general-settings/support/support-card/support-card.component.html
```diff
@@ -0,0 +1,7 @@
+<mat-card class="system-card" gdArea="system-card">
+  <mat-toolbar-row class="title-bar">
+    <h3>{{ supportTitle | translate }}</h3>
+  </mat-toolbar-row>
+
+  <ix-support></ix-support>
```
undsoft, line 6: Feels like you can move `ix-support` here. No real value in this wrapper component.

undsoft (changes requested):
Layout on smaller resolutions used to be different.
Otherwise looks good.

undsoft (approved):
Layout on certain resolutions is incorrect. General settings do not follow the same layout as in advanced settings – sometimes support card occupies full width, while other cards turn into columns.

Снимок экрана 2023-04-02 в 10 51 19

Otherwise looks okay.

PR #8020, opened 2023-03-31
NAS-121248 / 23.10 / Hotfixes login form

undsoft:
See that you can login again on latest nightlies.


denysbutenko (approved):
It works well. Thanks!

PR #8022, opened 2023-04-01
NAS-121215 / 22.12.3 / Reduce memory leak in `ix-entity-table` component

denysbutenko:
**Description**
> Adding many cloud sync tasks causes the UI to become slower and slower. Eventually every click takes 30 seconds for the UI to recover. System is not under any load or shows any spike in CPU during this event.

**The Root Cause**

The Cloud Sync page uses the deprecated `ix-entity-table` component with the rare `asyncView` param. When a user interacts with the cloud sync form and when the form is closed, new data is fetched, creating an interval that invokes fetching data every 10 seconds. One form close - one new interval. After data fetching, it gathers transforming, which is a heavy operation if you have more than 15 entries. That causes the main thread to be locked, and UI is unresponsive until the page gets refreshed.

**How to reproduce the issue?**

1. Go to the full cloud sync list page (path `/data-protection/cloudsync`)
2. Click on the `Add` button
3. Press the `Escape` key or click the `Close` button. Alternatively, Create a new task.
4. Repeat step 3 at least five times or more.
5. Now you may notice that UI is unresponsive

**Testing** 
Please make sure the UI is still working smoothly after creating more than 5 cloud sync tasks.

Note: This is a fix for Bluefin. For Cobia, we need to get rid of `ix-entity-table` usage. For more context, please check the original ticket.

### DIFF
--- a/src/app/modules/entity/entity-table/entity-table.component.ts
+++ b/src/app/modules/entity/entity-table/entity-table.component.ts
@@ -550,13 +550,8 @@ export class EntityTableComponent<Row extends SomeRow = SomeRow> implements OnIn
       this.getFunction = EMPTY;
     }
 
-    if (this.conf.callGetFunction) {
-      this.conf.callGetFunction(this);
-    } else {
-      this.callGetFunction();
-    }
-
-    if (this.asyncView) {
+    if (this.asyncView && !this.interval) {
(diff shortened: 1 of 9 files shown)


RehanY147 (approved):
Nice work. Looks good.

PR #8023, opened 2023-04-03
NAS-121222 / 23.10 / Update `shell` field

bvasilenko:
## **Testing**

At the **Credentials > Users > Add User** and **Edit User**, 

On the **Add/Edit user** form, 

When value of **Primary Groups** or **Auxiliary Groups** is changed

**Expected result:**

In the **Shell** dropdown input box, the list of options is should be refreshed from the API call `user.shell_choices`

### DIFF
--- a/src/app/pages/account/users/user-form/user-form.component.ts
+++ b/src/app/pages/account/users/user-form/user-form.component.ts
@@ -11,7 +11,7 @@ import {
   combineLatest, from, Observable, of, Subscription,
 } from 'rxjs';
 import {
-  filter, map, switchMap,
+  filter, map, switchMap, take,
 } from 'rxjs/operators';
 import { allCommands } from 'app/constants/all-commands.constant';
 import { choicesToOptions } from 'app/helpers/options.helper';
@@ -119,7 +119,7 @@ export class UserFormComponent {
   readonly groupOptions$ = this.ws.call('group.query').pipe(
     map((groups) => groups.map((group) => ({ label: group.group, value: group.id }))),
   );
-  shellOptions$: Observable<Option[]>;
+  shellOptions$: Observable<Option[]> = of([]);
   readonly treeNodeProvider = this.filesystemService.getFilesystemNodeProvider();
   readonly groupProvider = new SimpleAsyncComboboxProvider(this.groupOptions$);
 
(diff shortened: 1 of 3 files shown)


denysbutenko (changes requested):
When I made changes to the `Auxiliary Groups` it invoked two API requests for every change. See the screen recording.

https://user-images.githubusercontent.com/351613/229686247-f64a216e-dc06-4363-8e58-b05033e8e76a.mp4

bvasilenko:
@denysbutenko now this issue should be fixed

src/app/pages/account/users/user-form/user-form.component.ts
```diff
@@ -435,4 +443,15 @@ export class UserFormComponent {
 
     return username.toLocaleLowerCase();
   }
+
+  private updateShellOptions(group: number, groups: number[]): void {
+    let ids: number[] = [];
```
denysbutenko, line 448: Alternatively, `new Set()` can fit better here to avoid duplicate values when adding the group id below.

src/app/pages/account/users/user-form/user-form.component.ts
```diff
@@ -435,4 +443,15 @@ export class UserFormComponent {
...
+    ids = ids.concat(groups);
+    if (group && !ids.find((id) => id === group)) {
+      ids.push(group);
+    }
+    this.ws.call('user.shell_choices', [ids]).pipe(choicesToOptions(), untilDestroyed(this)).subscribe((options) => {
```
denysbutenko, line 453: It may be worth adding `take(1)` since every change in the `group` and `groups` fields invokes `updateShellOptions`, which creates a new observable until the form is closed. It seems that's not critical. Up to you.

denysbutenko (changes requested):
Getting error in the console:
`TypeError: Cannot read properties of undefined (reading 'pipe')
    at M.setFirstShellOption (user-form.component.ts:413:24)`

bvasilenko:
@denysbutenko the requested changes should be addressed now

denysbutenko (changes requested):
The TypeError has been resolved! However, it seems that there may still be an issue with the `setFirstShellOption` function. When I tried to create a new user, the Shell value was empty.

bvasilenko:
@denysbutenko now the requested changes are done

denysbutenko (approved):
It works well from a user perspective.

Although it makes many calls to the `user.shell_choices` endpoint under the hood.

Add form:
image

Edit form:
image

PR #8028, opened 2023-04-03
NAS-121299 / 23.10 / NAS-121293: Button missing to edit app info on the installed apps page

AlexKarpov98:


RehanY147 (commented):
The install form doesn't work yet. Need that to be fixed to confirm that the edit button works. I'll add the blocking ticket.

undsoft:
@RehanY147 @AlexKarpov98 Are we still waiting for something here?

AlexKarpov98:
> @RehanY147 @AlexKarpov98 Are we still waiting for something here?

I assume we could merge it.

RehanY147:
Was waiting for the store ticket. We were fetching all apps from one catalog and train with a constant value. That didn't work and we couldn't go from details to install form like that. That's fixed in the store ticket.

RehanY147 (changes requested):
You need to have the catalog and train from the details page to the chart wizard/form to fetch the correct app schema etc.

RehanY147:
You should merge `master` to get the latest code fixes

RehanY147 (approved):
There should be a test case to test the button click that it loads the right URL with the right params.

PR #8031, opened 2023-04-04
NAS-121297 / 23.10 / Filter doesn't work on the installed apps table

bvasilenko:
**Testing**

On **Apps > Installed** page, test that filter is working


dszidi (approved):
LGTM

PR #8034, opened 2023-04-05
NAS-121342 / 22.12.3 / Fix rear NVME drives on M Series

dszidi:
Test against an M Series machine with the following conditions:

- Rear drives are NVME
- Rear drives are not assigned to a pool
- no console errors when switching to rear view

Check ticket description for an M Series url


RehanY147 (approved):
Looks good

PR #8036, opened 2023-04-05
NAS-121290 / 23.10 / Improvements for the apps available filter

bvasilenko:
**Testing:**

On the **Apps > Available** page:

1. Verify that the filter width looks good
2. Test that checkboxes and radiobuttons now match the design, provided [here](https://github.com/truenas/webui/discussions/7502), under **Search & Inventory** heading
3. Verify that progress bar looks properly when filter is applied

### DIFF
--- a/src/app/pages/apps/components/available-apps/available-apps-header/available-apps-header.component.spec.ts
+++ b/src/app/pages/apps/components/available-apps/available-apps-header/available-apps-header.component.spec.ts
@@ -4,7 +4,6 @@ import { ReactiveFormsModule } from '@angular/forms';
 import { MatButtonHarness } from '@angular/material/button/testing';
 import { MatChipInputHarness } from '@angular/material/chips/testing';
 import { MatInputHarness } from '@angular/material/input/testing';
-import { MatSelectHarness } from '@angular/material/select/testing';
 import { SpectatorHost } from '@ngneat/spectator';
 import { createHostFactory } from '@ngneat/spectator/jest';
 import { mockWebsocket, mockCall } from 'app/core/testing/utils/mock-websocket.utils';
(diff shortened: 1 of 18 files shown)


src/app/pages/apps/components/available-apps/available-apps-header/available-apps-header.component.spec.ts
```diff
@@ -82,28 +76,6 @@ describe('AvailableAppsHeaderComponent', () => {
     expect(changeSearch).toHaveBeenLastCalledWith('search string');
   });
 
-  it('emits (filters) when user selects catalogs', async () => {
```
undsoft, line 85: We still need tests.

undsoft (changes requested):
I really don't know if having an `altView` is a good idea. I understand the desire to avoid duplicating code, but having this setting makes API of the component kinda messy. Harnesses would need to be adjusted, you'd need to have another set of tests for this layout. And currently it's used in one place.
I would rather we create a new set of components, maybe even in separate module or within apps module for now.

UI feedback:
1. The filter is now too wide. I think there should some maximum width. If you are on ultra-wide you'll get a very long, inconvenient input. Also numbers on the right need some room to breathe between them and filters.
2. Need to change cursor on new controls.
3. Checkmarks are larger than in designs.
4. In designs there are no "empty" checkmarks.
5. Catalogs and Sort labels have an underline and have different font.

bvasilenko:
@undsoft 
Ok, now these components extracted to a custom module

src/app/pages/apps/components/available-apps/available-apps-header/available-apps-header.component.spec.ts
```diff
@@ -62,8 +63,8 @@ describe('AvailableAppsHeaderComponent', () => {
     await filtersButton.click();
 
     searchInput = await loader.getHarness(MatInputHarness.with({ placeholder: 'Search' }));
-    catalogsSelect = (await loader.getAllHarnesses(MatSelectHarness))[0];
-    sortSelect = (await loader.getAllHarnesses(MatSelectHarness))[1];
+    catalogsItems = spectator.queryAll('.catalogs .item');
+    sortItems = spectator.queryAll('.sort .item');
```
undsoft, line 67: This is fine, but it would be easier to work with the component in tests if you were to create a harness for `filter-select-list`.

undsoft (approved):
Looks good.

PR #8040, opened 2023-04-06
Added tests for ErrorHandlerService

RehanY147:


src/app/services/error-handler.service.ts
```diff
@@ -136,6 +100,54 @@ export class ErrorHandlerService implements ErrorHandler {
     };
   }
 
+  private parseJobWithArrayExtra(errorJob: Job): ErrorReport | ErrorReport[] {
```
denysbutenko, line 103: It looks like the method always returns an array

denysbutenko (approved):
Looks good.

PR #8044, opened 2023-04-07
NAS-121375 / 23.10 / The search input has value despite resetting outside of the component

AlexKarpov98:


bvasilenko (approved):
Works ok for me

PR #8046, opened 2023-04-08
NAS-121307 / 23.10 / Use the new endpoint to fetch the latest updated apps

RehanY147:
This PR does two things
- Uses the new endpoint to fetch data for the "New and Updated" apps category on the Available Apps dashboard.
- Make it so that when user clicks on the "View All [Category Name] Apps" button on the Available Apps dashboard for any category, the filter is reset and then only that category is applied as a filter.

Known issue:
The filter inputs values don't get updated when you click on that "View All" button. That should be handled by https://ixsystems.atlassian.net/browse/NAS-121189 when the new component store is implemented which will also handle filtering apps and applying filter values.


undsoft (approved):
Looks good.

PR #8050, opened 2023-04-10
NAS-121296 / 23.10 / Implement better app status

denysbutenko:
**Description**

> On the installed apps page, when you click to start a stopped app, the starting process is shown as a progress in a dialog. Instead use the state tiles in the row to show that the app is starting. Clicking on that tile can bring up this dialog that’s showing now.

**Testing**

- Open the Installed Apps page
- Make a test of the app status


src/app/pages/apps/apps.module.ts
```diff
@@ -121,5 +123,6 @@ import { InstalledAppsComponent } from './components/installed-apps/installed-ap
     MatTooltipModule,
     MatMenuModule,
   ],
+  exports: [],
```
AlexKarpov98, line 126: we don't need empty **exports: []**

AlexKarpov98 (approved):
Looks good.
Good job.

PR #8051, opened 2023-04-10
NAS-121274 / 23.10 / Prevent empty value in Wipe Disk dialog

denysbutenko:
Prevent users from submitting an empty value in the Wipe Disk dialog.

For context, please check the ticket.


AlexKarpov98 (approved):
Works as expected.

PR #8029, opened 2023-04-03
NAS-121308: Resolve issues with old-style timeFormat

rickwargo:
`timeFormat` in local storage may have an old-style format escape `A` that needs to be converted to `aa`. If not converted, the UI does not work reliably, creating hundreds of RangeError messages in the console and potentially freezing the screen.

PR #8052, opened 2023-04-10
NAS-121247 / 23.10 / TrueNAS SCALE - RangeError: Invalid time value on UI

AlexKarpov98:


denysbutenko (approved):
No regressions are found. Looks good 👍🏻

denysbutenko:
It seems to have some regressions. Once you finish the changes, please let me know I will review them again.

AlexKarpov98:
> It seems to have some regressions. Once you finish the changes, please let me know I will review them again.

yeah, @denysbutenko check it now.
the main problem was the unhandled error if some invalid date provided
like in ticket: 
image

denysbutenko (changes requested):
The WebUI does not crash on invalid time format now - that's good.
 
However, it seems like many users need help with this issue when upgrading from Core to Scale, like recent contribution #8029. So, the change in this PR will show the `Invalid date` instead of the actual date if the time format is invalid. The reason for that is the `formatDateTime` pipe is looking for `timeFormat` in local storage, and the `formatDateTime` method in `LocaleService` relies on the user preferences only. And there is no way to overwrite the `timeFormat` value in local storage. So, it makes possible the situation when the user won't see the dates formatted using the pipe and can't fix it through WebUI.

I'm unsure how better to solve this.

For testing, set `timeFormat` value to `undefined` in local storage and check Alerts.

AlexKarpov98:
> 

okay, then we may check for time format & day format in localstorage and if not set or set as invalid - we override it.
@denysbutenko

src/app/services/locale.service.ts
```diff
@@ -69,36 +69,48 @@ export class LocaleService {
   }
 
   formatDateTime(date: Date | number, tz?: string): string {
```
denysbutenko, line 71: No usage of this method is found in the code base. It can be removed.

denysbutenko (approved):
Invalid `timeFormat` or `dateFormat` won't crash the UI, and users can see an actual date.

One small thing. If I change the time format in user settings, it updates a value in local storage, and the `formatDateTime` pipe doesn't reflect the changes (e.g. Alerts) until refreshing the page. 

Other than that, it works well. Nice work.

AlexKarpov98:
@denysbutenko - please check it now, should update the value after the form was submitted with no page reload.

src/app/core/pipes/format-datetime.pipe.ts
```diff
@@ -20,19 +23,44 @@ export class FormatDateTimePipe implements PipeTransform {
...
     private cdr: ChangeDetectorRef,
     @Inject(WINDOW) private window: Window,
   ) {
     this.store$.select(selectTimezone).pipe(untilDestroyed(this)).subscribe((timezone) => {
       this.timezone = timezone;
       this.cdr.markForCheck();
+      this.checkFormatsFromLocalStorage();
+    });
+
+    this.actions$
```
denysbutenko, line 36: I'd like to mention that checking for changes in preferences from the store allows us to avoid new entity injection. However, I don't know if that has a noticeable difference.

denysbutenko (approved):
The system works well and now effectively handles timeFormat/dateFormat preferences. Additionally, the WebUI no longer crashes when presented with invalid values. Any changes made to these preferences by the user are immediately reflected in the UI without requiring a page reload.

PR #8056, opened 2023-04-11
NAS-121344 / 23.10 / Some times signin form won't take any users other than 'root'

AlexKarpov98:


RehanY147 (commented):
@AlexKarpov98 Please explain the problem. This code was added for https://ixsystems.atlassian.net/browse/NAS-119669. Removing it might cause issues.

PR #8062, opened 2023-04-13
NAS-121189 / 23.10 / Minor improvements when installing an app

RehanY147:
Read ticket description for todos


src/app/pages/apps/components/chart-wizard/chart-wizard.component.html
```diff
@@ -27,7 +27,7 @@
           ixTest="save"
           [disabled]="form.invalid || isLoading"
         >
-          {{ 'Save' | translate }}
+          {{ isNew ? 'Install' : 'Update' | translate }}
```
undsoft, line 30: Message extractor won't be picked from this. It should be `isNew ? ('Install' | translate) : ('Update' | translate)`

undsoft (changes requested):
Ticket number is incorrect. I'm assuming this is the ticket that this relates to:
If so, last two items are still not addressed. I still get two dialogs on receiving an error and I can still minimize job dialog when installing an app.

RehanY147:
Wrong branch name.

PR #8064, opened 2023-04-13
NAS-121432 / 23.10 / Various fixes for Available Apps

denysbutenko:
Details in the ticket


RehanY147 (changes requested):
Let's make the installed badge clickable every where. Like on the Available Apps page. Also, the add the tooltip. Also fix the tooltip text to say `"An instance of this app already installed. Click the badge to see installed apps."`

RehanY147 (changes requested):
Everything looks good. This is something extra but falls under the scope of this ticket. We should have a skeleton loading for the app cards for a couple of categories, also category titles.

denysbutenko:
> Everything looks good. This is something extra but falls under the scope of this ticket. We should have a skeleton loading for the app cards for a couple of categories, also category titles.

Sure, good idea. Thanks for bringing it to my attention.

undsoft:
I liked previous loading for app cards more. At least on our m40 loading a list of apps is not a fast operation and animation on cards it too intrusive.
I think we should just have a fake progress bar just below filter. We can add cards, but then maybe let's lose animation to make them less prominent.

undsoft (changes requested):
Also this PR seems to show every app in every category at least on our m40.

denysbutenko:
> Also this PR seems to show every app in every category at least on our m40.

Oops. There were two different `apps` - one coming from the template and the second variable from observable. Fixed 👍🏻 

Also, put back animation for app card placeholders. So, it has a blue line and animated placeholders now.

undsoft (approved):
There's something about loading animation that doesn't sit right with me, but I'm not sure I can pinpoint it.
We can merge and iterate on it later, rest of the stuff is okay.

denysbutenko:
> @denysbutenko I've been playing with animation. I've removed skeleton from header, I think it looks better this way. If you agree, we can merge.

Yep, looks good to me 👍🏻

PR #8066, opened 2023-04-13
NAS-121429 / 23.10 / Improve experience of adding a catalog

AlexKarpov98:


dszidi (approved):
I didn't notice any problems with the code. Unfortunately I could not test a success scenario as truecharts was not installing correctly

image

PR #8067, opened 2023-04-13
NAS-120692 / 23.10 / WIP: Mock Enclosure in dev mode

dszidi:
NOTE: New setup script available.

### WRT setup_env.js
- `setup_env.js` no longer exists. There is a setup script that gets triggered on any non production build that ensures an environment.ts file is present. If remote has not been set, we will be notified via stdout along with the command required to set remote.
- The new way to set remote is via the ui command `yarn ui remote -i <ip address or FQDN>`

### New command `ui`

Command can be triggered via `./ui` or users can make an alias for themselves to avoid constantly typing the path (`./`) in front of the file name every time (recommended)

Too much info to explicitly type out every detail so I will just paste help flag output

### Root Command `ui`
```
% ./ui -h
Usage: ui.ts [options] [command]

WebUI setup utility

Options:
  -V, --version     output the version number
  -h, --help        output usage information

Commands:
  mock [options]    Mocking enclosure ui support
  reset             Reset environment file to initial state
  remote [options]  Set the server WebUI communicates with
  showremote        Show the server WebUI communicates with
  ```

### Subcommand `remote`
```
% ./ui remote -h
Usage: ui.ts remote [options]

Set the server WebUI communicates with

Options:
  -i, --ip <ip_address>  Sets IP address of your server
  -h, --help             output usage information
```

### Subcommand `mock`
```
Usage: ui mock|m [options]

Configure mock settings via config.json files

Options:
  -e, --enable    enable mock config
  -d, --disable   disable mock config
  -g, --generate  generate a mock config file
  -h, --help      display help for command

The mock command sets mock related configurations through the use of json files.
The repo includes some files for common configurations developers might need for testing.
Developers can also create custom config files. See examples below...

EXAMPLES:

  * Load a custom or included configuration file (guided)

  % ui mock | ui m

  * Generating a custom config file (guided)

  The --generate option will provide a series of prompts that results in the creation of a config.json file.
  After finishing the guided install, it will create the file with the provided selections in src/assets/mock/configs

    % ui mock --generate OR ui m -g
```

### Subcommand `mock-opt`

```
ui mo -h
Usage: ui mock-opt|mo [options]

Set specific mock options

Options:
  -e, --enable                        enable mock config
  -d, --disable                       disable mock config
  -M, --showcontrollers               show available controllers for mock
  -S, --showshelves                   show available shelves for mock
  -m, --model <name>                  set controller to mock
  -s, --shelves, [Array|null]         set expansion shelves to mock
  -a, --assign, <existing | default>  set slot assignment strategy
  -r, --reset                         Reset mock configuration to default values
  -l, --list                          show current mock config settings
  -D, --debug                         debug output
  -h, --help                          display help for command

EXAMPLES:

  * Retrieve a list of mockable controller models

    % ui mock-opt -M OR ui mock-opt --showcontrollers

  * Retrieve a list of mockable expansion shelf models

    % ui mock-opt -S OR ui mock-opt --showshelves

  * Set the controller model to mock

    % ui mock-opt -m OR ui mock-opt --model

  * Set the expansion shelf models to mock

    NOTE: It will except no argument or empty string to specify no shelves.
    Specifying shelves must be done via a comma separated string

    % ui mock-opt -s 'es24,es24' OR ui mock-opt --shelves 'es24,es24'

  * Enable mock functionality

    % ui mock-opt -e OR ui mock-opt --enable

  * Disable mock functionality

    % ui mock-opt -d OR ui mock-opt --disable

  * Complete configuration example that creates an M50 with two ES24 shelves

    % ui mock-opt -e -m m50 -s 'es24,es24' -a default
```

BE AWARE:
currently supports the following models...
- M40
- M50
- MINI-3.0-XL+
- MINI-R
- ES24
- ES102

Once this gets merged, I can add other models

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -9,16 +9,17 @@
     "rimraf": "rimraf",
     "changelog": "standard-changelog",
     "commitlint": "commitlint",
-    "start": "ng serve --proxy-config proxy.config.json",
-    "start:dev": "ng serve --proxy-config proxy.config.json",
+    "setup": "cd $(git rev-parse --show-toplevel) && npx ts-node ./src/environments/setup.ts",
```
undsoft, line 12: Also, I'm a bit confused as to what `setup` command does exactly. If i have an `environment.ts` file (even if it's an old one), it doesn't seem to be doing anything.
If I don't have an `environment.ts`, I still have to use `yarn ui` to set the remote.

src/environments/environment.template.ts
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable no-restricted-imports */
```
undsoft, line 1: Is this really necessary?

src/environments/ui.ts
```diff
@@ -0,0 +1,462 @@
+/* eslint-disable */
+import { environment } from './environment';
```
undsoft, line 2: 1. We have scripts folder. It may be more appropriate for this file instead of `environments`.
2. We should remove `setup_env.js`.

src/environments/ui.ts
```diff
@@ -0,0 +1,462 @@
...
+/*
+* Command Setup
+* */
+const program: Command = new Command();
+program
+  .version('0.0.1')
+  .description('WebUI setup utility')
+
+program
+  .command('mock')
```
undsoft, line 64: I feel like it may be easier if there was some sort of guided mode to this.
Right now, as a developer I don't know what models are available, which forces me to open IDE and look through files. At this point, I might as well edit environment.ts manually.
Maybe if it worked as a wizard, explaining every option, it would be easier.

src/environments/ui.ts
```diff
@@ -0,0 +1,462 @@
...
+    mockConfigReport({
+      showHeader: true,
+      showFooter: true,
+    });
+  });
+
+program
+  .command('remote')
+  .description('Set the server WebUI communicates with')
+  .option('-i, --ip <ip_address>', 'Sets IP address of your server')
```
undsoft, line 104: Probably don't need a flag.

src/environments/environment.template.ts
```diff
@@ -0,0 +1,19 @@
...
+  production: false,
+  sentryPublicDsn: 'https://7ac3e76fe2a94f77a58e1c38ea6b42d9@sentry.ixsystems.com/4',
+  mockConfig: {
+    enabled: false,
+    enclosureOptions: {
+      controllerModel: 'M40',
+      expansionModels: [],
+      dispersal: EnclosureDispersalStrategy.Default,
+    },
+    systemProduct: 'TRUENAS-M40',
```
undsoft, line 17: Indentation is weird in resulting `environment.ts`

undsoft (changes requested):
I think I was able to successfully mock it once, but then something happened and my subsequent attempts at doing `yarn ui mock -m ES102` result in
```
/Users/undsoft/projects/work/webui-2/src/environments/ui.ts:277
  environment.mockConfig.systemProduct = models.controllers[value.toUpperCase()].systemProduct;
                                                                                 ^
TypeError: Cannot read properties of undefined (reading 'systemProduct')
```

package.json
dszidi: Just posting this as a matter of record. As per our discussions in slack, setup is now renamed to `check-environment`. The script's sole purpose is to guarantee that an `environment.ts` file exists, otherwise the ui.ts script breaks during imports. The new `yarn check-environment` script is invoked in other commands like `build:prod `and `start:dev` to guarantee this file is not missing.

src/environments/ui.ts
dszidi: Done. You can try `./ui mock-gen` or with the alias `./ui mg`

src/environments/ui.ts
dszidi: Just for future reference. As per our discussions, the design was to have the command without flags get the current state.  Flags are what we use to set values.

src/environments/ui.ts
dszidi: Done and done

src/environments/environment.template.ts
dszidi: Fixed

package.json
```diff
@@ -29,7 +30,8 @@
     "validate-translations": "node scripts/validate_translations.js",
     "generate_licenses": "yarn licenses generate-disclaimer > src/assets/disclaimer.txt && yarn licenses list --json > src/assets/licenses.json",
     "generate_licenses_list": "yarn licenses list > licenses.txt",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "ui": "cd $(git rev-parse --show-toplevel) && yarn run --silent ts-node ./scripts/ui/ui.ts"
```
undsoft, line 34: shouldn't check-environment be used here as well?

undsoft (changes requested):
As discussed before I think we should provide more guidance to a developer on how to use this.
I.e. doing `./ui mock` should list instructions on how to setup a mock and what models are available.

Another issue that we may want to solve is dealing with old env file. Running `./ui` will just fail with random import errors and developer would have to know to go and recreate the env file.

package.json
dszidi: Wanted to give an update. I'm working on this help text issue right now. Commander.js documentation seems to indicate that adding `command.addToHelp('after', 'My help string')` should do what we want but that is not working. Seems our commander.js version is wildly out of date. I've updated it to latest (4.3 -> 10.0.1) and there is a lot of breakage. Next commit should have the help text updates

scripts/ui/ui.ts
```diff
@@ -0,0 +1,630 @@
...
+* Command Setup* */
+const program: Command = new Command()
+  .name('./ui')
+  .version('0.0.1')
+  .description('TrueNAS webui setup utility')
+  .usage('(Call from root directory)')
+  .addHelpText('before', banner())
+
+const mockExamples = `
+EXAMPLES:
```
denysbutenko, line 79: It has examples of how to mock different parts (model, shelves) and how to use a config file. And no examples of how to plug and play, e.g.:

```
yarn ui mock -e -m M50 -s 'es24,es24'
```

or just use existing config
```
yarn ui mock -c m50.config.json
```
That would be useful I believe.

README.md
```diff
@@ -54,7 +54,7 @@ Run the environment configuration script
 (if your ip address changes later you may repeat this step)
 
 ```sh
-$ ./setup_env.js -i <ip address or FQDN of the server where TrueNAS is running>
+$ ./ui remote -i <ip address or FQDN of the server where TrueNAS is running>
```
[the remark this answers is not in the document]
dszidi: There are two reasons I made the ./ui script.

- Adds yarn's --silent flag by default. This eliminates yarn output that only seems to add clutter to the output.
- Less to type

If there was a nice way to help users setup a shell alias for themselves without opening a huge can of worms, I'd do it. Fortunately I think it's safe to assume devs will already know how to do this 

I've set up an alias for myself and all I am using on my own system is `ui remote` or `ui mock` etc I was hoping to get devs as close to that as possible out of the gate

I'll update the README with some better instructions regardless

undsoft:
1. ENOENT: no such file or directory, scandir custom.
2. Can we make `mock` listed above `mock-opt` in `ui` command?
3. Can we rename configs to include model name?
4. And also need to address that issue where developer may have an existing (or missing) environment file that may prevent `ui` command from running.

dszidi:
> 1. ENOENT: no such file or directory, scandir custom.
> 2. Can we make `mock` listed above `mock-opt` in `ui` command?
> 3. Can we rename configs to include model name?
> 4. And also need to address that issue where developer may have an existing (or missing) environment file that may prevent `ui` command from running.

1. I  set .gitignore to only ignore the json files and dropped a readme file in there so the directory would be saved in the repo
2. Done
3. Done
4. I changed the package.json script entry so `check-env` is run in new quiet mode. Environment file problems should be eliminated

undsoft (approved):
I thought `Custom` option would run config generator, but this probably works too. In either case, core functionality is there and rest we can improve on later.

PR #8069, opened 2023-04-13
NAS-121394 / 23.10 / Refactor replication wizard to ix-forms

bvasilenko:
**Testing**

Verify the functionality of **Replication Wizard**


undsoft (changes requested):
Only did superficial testing in UI.
Explorer doesn't seem to work correctly on remote systems.

bvasilenko:
@undsoft ok, all review remarks should be addressed

undsoft (changes requested):
I don't know if you pushed all changes, but here are some of the new things I've discovered in addition to _some of the previous items_ that were not fixed:
* Missing a button to switch to advanced replication creation.
* Missing a line that says `14 snapshots found.` when creating a local replication.
* If you select Encryption and Key, untick Generate and switch to Passphrase, you'll see an extra Encryption Key.
* When I set replication with remote component, old UI would show `Include snapshots with the name` fields. New UI doesn't.
* `SSH Connections` field can be renamed to `SSH Connection`
* Old UI didn't have a confirmation step. I'm okay with not having it too.

bvasilenko:
@undsoft ok, all the remarks should be addressed

src/app/pages/data-protection/replication/replication-wizard/replication-wizard.component.ts
```diff
@@ -1484,33 +283,121 @@ export class ReplicationWizardComponent implements WizardConfiguration {
...
+              }
+              this.existSnapshotTasks.push(...tasks.map((task) => task.id));
+              return of(null);
+            }),
+          ),
+        );
+      }
+      if (requestsTasks.length) {
+        return forkJoin(requestsTasks).pipe(
+          map((createdSnapshotTasks) => this.createdSnapshotTasks = (createdSnapshotTasks || []).filter((tk) => !!tk)),
```
undsoft, line 366: `tk` => `task`

src/app/pages/data-protection/replication/replication-wizard/replication-wizard.component.ts
```diff
@@ -1,1430 +1,229 @@
...
-      Object.keys((this.entityWizard.formArray.get([i]) as UntypedFormGroup).controls).forEach((item) => {
-        const itemConf = _.find(this.wizardConfig[i].fieldConfig, { name: item });
-        if (itemConf.value !== undefined && item !== 'exist_replication') {
-          this.entityWizard.formArray.get([i]).get(item).setValue(itemConf.value);
-        }
+    this.callCreateSnapshots(values).pipe(
+      catchError((err) => { this.handleError(err); return EMPTY; }),
+      untilDestroyed(this),
+    ).subscribe(() => {
+      this.callCreateTasks(values).pipe(
```
undsoft, line 128: Eh, why not just use `switchMap`?

src/app/pages/data-protection/replication/replication-wizard/replication-wizard.component.ts
```diff
@@ -1,1430 +1,229 @@
...
-          );
-        },
-      );
-    }
-
-    this.loader.close();
-    if (!toStop) {
-      this.modalService.closeSlideIn();
-    }
+  getSnapshotsСount(payload: CountManualSnapshotsParams): Observable<EligibleManualSnapshotsCount> {
```
undsoft, line 154: I still get `Name contains both ASCII and non-ASCII symbols: getSnapshotsСount ` in letter `c`.

src/app/pages/data-protection/replication/replication-wizard/steps/replication-what-and-where/replication-what-and-where.component.ts
```diff
@@ -0,0 +1,546 @@
...
+    }
+
+    this.form.controls.source_datasets.setValue(task.source_datasets);
+    this.form.controls.target_dataset.setValue(task.target_dataset);
+    this.form.controls.transport.setValue(task.transport);
+    this.form.controls.name.setValue(task.name);
+  }
+
+  clearReplicationTask(): void {
+    this.form.controls.source_datasets_from.setValue(null);
```
undsoft, line 480: Maybe just do `this.form.patchValue({ ... })`.

undsoft (changes requested):
1. When I meant that we can remove summary, I mean no step 3 at all (as it currently works). We should either have summary or just save on step 2.
2. When I switch to regex and field is empty I get `[EINVAL] count_eligible_manual_snapshots.name_regex: Empty value not allowed` in console. Probably not worth sending request on empty name_regex.
3. When I create replication from Remote to Local, old UI would show `Include snapshots with the name`. New UI does not and replication cannot be created.

undsoft:
I've merged [a PR](https://github.com/truenas/webui/pull/8087) which refactors Replication Form to ix-forms. 
I don't know if there is anything useful for you there.

I've updated this PR.

bvasilenko:
@undsoft my last commit should fix the remaining remarks, so this should be ready for review now

undsoft (changes requested):
1. Auto generated name didn't use to include `mnt` part.
2. When setting up local replication old UI shows `0 snapshots found. Snapshots will be created automatically.`. New UI doesn't show this string.
3. I also believe we are missing colors on these messages.
4. I've selected an existing replication and noticed that Source field didn't get populated correctly.
5. When setting up local-to-remote replication, old UI would only show me two fields in Destination Snapshot Lifetime: Same as source and Never Delete. New UI also shows Custom.

Rest looks okay, tested in multiple scenarios.

bvasilenko:
@undsoft the latest requests should be addressed now

undsoft (approved):
Looks good, but one issue is not resolved.
If you go to our m40 and select an existing replication called `pizza`, you'll see that no source datasets are selected.

PR #8070, opened 2023-04-13
NAS-121291 / 23.10 / Improvements for Apps Details page

bvasilenko:
**Testing**

At the **Apps Details** page

1. If there is no pool selected for the apps, you should be able to still load available apps. If you go to the apps details page for one of the apps in this case, the “Install” button should instead say “Setup Pool First”

2. The permanent dialog to select a pool shouldn’t be showing up. Instead the user should user the select pool button on the available apps dashboard to select pool


src/app/pages/apps/components/app-detail-view/app-detail-view.component.html
```diff
@@ -15,9 +15,8 @@
           mat-button
           color="primary"
           [ixTest]="[app?.name, 'install']"
-          [disabled]="!app"
-          (click)="installButtonPressed()"
-        >{{ 'Install' | translate }}</a>
+          (click)="wasPoolSet ? installButtonPressed() : onChoosePool()"
```
AlexKarpov98, line 18: Let's change  **On**ChoosePool to `showChoosePoolModal()`

and `installButtonPressed()` to `navigateToInstallPage()`

AlexKarpov98 (approved):
Works as described. 
I'd like to change method names only. 🤔 

Screenshot 2023-04-14 at 14 32 17

PR #8071, opened 2023-04-14
NAS-121484 / 23.10 / Similar Apps Navigation is not working

AlexKarpov98:


RehanY147:
@AlexKarpov98 Middleware is creating a new endpoint which will give us a list of similar apps automatically. We don't know need to use `app.available`.

AlexKarpov98:
> @AlexKarpov98 Middleware is creating a new endpoint which will give us a list of similar apps automatically. We don't know need to use `app.available`.

Sure, let's wait for it and then I'll implement it.

RehanY147 (changes requested):
While the middleware ticket is being worked on let's do the following
- Merge `master` to get the latest code fixes
- Get similar apps by comparing categories array. Try to match all categories to other apps and then reduce the number until we have four apps to show as similar apps
Let me know if you have any questions.

AlexKarpov98:
@RehanY147 - now new API is used, please check it out.

undsoft (approved):
Looks good, though backend algorithm can be improved.

PR #8072, opened 2023-04-14
NAS-121445 / 23.10 / Trivial: Waiting to renew Kerberos ticket. Current ticketexpires

AlexKarpov98:
Goal is to filter out `transient: true` jobs.


undsoft (approved):
Should be 
```
expect(jobs[3]).toBeUndefined();
```
otherwise looks good.

PR #8076, opened 2023-04-15
NAS-121504: Remove FibreChannel related components

undsoft:
Removes some unused components and other cleanup.
Code cleanup should be enough.


AlexKarpov98 (approved):
Looks good 👍

PR #8077, opened 2023-04-17
NAS-121440 / 23.10 / Sync cancel buttons in services

denysbutenko:
**Description**
> When identifying the SNMP service edit page cancel button, the data-test tag is defined differently than other service cancel buttons.  This should be consistent.

For testing, the cancel button on the `Services -> SNMP` page should have `data-test="button-cancel"` value.


undsoft (approved):
Looks good, didn't know you could use `routerLink` on `button`s.

PR #8079, opened 2023-04-17
NAS-121187 / 23.10 / Add warnings to General Info in Pool Manager

denysbutenko:


dszidi (approved):
Pretty good so far. Based on the TODO list on the ticket...


- Disks are not loaded immediately when wizard is opened, so these warnings aren’t actually shown until you go to second step and back.
- Prevent user from navigating to second step until disks were loaded to make sure user saw the warning.

This works. No need to leave and return to see warnings

- Show some form of inline loading indicator while we load disks.
Loaders are present

- Make warning more prominent. 

They are definitely obvious

- For exported pools add texts: “The following disks have exported pools on them. Using those disks will make existing pools on them unable to be imported. You will lose any and all data in {pool}.  Select disks you want to use:” 
- For disks with non unique serials take a look at existing messaging in src/app/pages/storage/components/manager/manager.component.ts:164.

These warnings were present as expected.

- Add something similar and ask user to select disks they want to use.
- Make it look better visually with margins and alignment. Consider that there may be a lot of disks (hundreds) with some of those conditions, so find a way to show them.

I would have to see this in action with a lot of disks to really be confident with the layout effectiveness. For now I would say it works fine.


TL;DR - LGTM

image

denysbutenko:
> I would have to see this in action with a lot of disks to really be confident with the layout effectiveness.

@dszidi, you can try to modify the `exportedPoolsOptions$` stream to simulate the needed amount of disks. e.g.

```
exportedPoolsOptions$: Observable<Option[]> = this.exportedPools$.pipe(
    map((pools) => pools.map((pool) => ({ label: pool, value: pool }))),
    map((pools) => {
      for (let i = 0; i < 100; i++) {
        pools.push({ label: `sda${i} (pool ${i})`, value: `pool ${i}` });
      }
      return pools;
    }),
  );
```

or

https://user-images.githubusercontent.com/351613/233527434-5e8aaecf-fff2-48e6-83c2-953b62efb574.mp4

PR #8082, opened 2023-04-18
NAS-121460 / 23.10 / webui uses private `dscache.get_uncached_user` endpoint rather than public `user.get_user_obj` endpoint

AlexKarpov98:


undsoft (approved):
Looks good.

PR #8084, opened 2023-04-19
NAS-121124 / 23.10 / Fixed the resizing on reload shell

RehanY147:

The problem was that when we ran that command or generally on reload, the `ReplaySubject` `lastGeneratedAuthToken$` would remember the last token value, which was invalid at this point. When we went back to the shell, it loaded the wrong token the first couple of times and shell failed to authenticate so it loaded a couple of empty pages before actually loading the right one.

This problem was also partially caused by the missing `take(1)` in the generated auth token repeated action pipe. We only wanted that sequence to execute once every time the timer ticks. But because of that missing `take(1)` pipe, every time the `isAuthenticated$` observable emitted a value, the sequence ran again which caused token to reload without need.

Just test that authentication works as normal and no major authentication errors arise in the UI. Also, that shell loads without the empty blank pages.

### DIFF
--- a/src/app/services/auth/auth.service.spec.ts
+++ b/src/app/services/auth/auth.service.spec.ts
@@ -207,8 +207,8 @@ describe('AuthService', () => {
           { c: false },
         );
         expectObservable(spectator.service.authToken$).toBe(
-          'd',
-          { d: null },
+          '|',
+          {},
         );
       });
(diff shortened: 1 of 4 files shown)


AlexKarpov98 (approved):
Works good.
Nice job.

PR #8087, opened 2023-04-19
NAS-117660 / 23.10 / Refactoring replication form

undsoft:
Testing: test replication form in multiple scenarios.


denysbutenko (commented):
I have found a few things that could be improved. Although they are valid in the master branch also, so they are not related to this PR.

For example:
1. Modifying the value in the `ix-explorer` manually would raise the error dialog with `EAGAIN`.
2. If I go to the 2nd step, switch the `Replication Schedule` value to `Run Once` and then return to the 1st step - I can't move to the 2nd step again.
3. Sometimes the `Start Replication` button is disabled, and it is unclear reasons and how to fix that.

denysbutenko (approved):
Impressive work!

denysbutenko:
I have discovered two types of issues while testing the replication form:

Layout issues: Selecting a long value causes the form layout to break, making certain elements inaccessible.


https://user-images.githubusercontent.com/351613/233325649-876545ce-9329-4943-8e25-efeecef1454e.mp4


https://user-images.githubusercontent.com/351613/233325678-2a0f1bf9-ad42-49f9-ae63-673f60e23b99.mp4


Validation issues: I have identified several cases where submitting invalid values is possible, indicating the need for improvements in this area. Perhaps separately.

- [EINVAL] replication_create.name: Object with this name already exists
- [EINVAL] replication_create.name_regex: Empty value not allowed
- [EINVAL] replication_create.also_include_naming_schema.0.naming_schema: %Y must be present in snapshot naming schema
- [EINVAL] replication_create.also_include_naming_schema.0.naming_schema: %m must be present in snapshot naming schema
- [EINVAL] replication_update.schedule: You can't have schedule for replication that does not run automatically

undsoft:
I've fixed the layout issues, as to validation - there a lot of different scenarios where middleware will return validation errors. I'm okay about improving on this later.

denysbutenko (approved):
Form layout is fixed 👍🏻

PR #8088, opened 2023-04-19
NAS-121502 / 23.10 / Refactor InitiatorFormComponent to ix-forms

bvasilenko:
**Testing**

Verify the functionality of **Shares → iSCSI Configure → Initiator Groups → Add**


src/app/modules/common/dual-list/dual-list.component.ts
```diff
@@ -45,6 +47,12 @@ export class DualListboxComponent<T extends { id: string | number; name?: string
     this.selectedItems = new ListSelectionImpl(this._selectedItems);
   }
 
+  ngOnChanges(changes: IxSimpleChanges<this>): void {
+    if ('items' in changes && !_.isEqual(changes.items.currentValue, changes.items.previousValue)) {
+      this.ngOnInit();
```
undsoft, line 52: Calling one lifecycle hook in another doesn't look right. It's better to extract code in a separate method and call it.

undsoft (approved):
Works well.
Let's rename columns to: left – Connected Initiators, right – Allowed Initiators and `Allowed Initiators (IQN)` to `Add Allowed Initiator (IQN)`

PR #8089, opened 2023-04-20
NAS-121542 / 23.10 / Trailing commas on builtin user full name

AlexKarpov98:


undsoft (changes requested):
Have we checked with middleware about this? If data is coming like this from middleware, it sounds like it should be addressed on middleware side.

AlexKarpov98:
> Have we checked with middleware about this? If data is coming like this from middleware, it sounds like it should be addressed on middleware side.

yes, it comes from middleware.
Gonna talk to @yocalebo

AlexKarpov98:
assigned to @yocalebo

PR #8090, opened 2023-04-20
NAS-121541 / 23.10 / Properly validate username field when creating a local user

AlexKarpov98:
original: https://github.com/truenas/webui/pull/8091


AlexKarpov98:
original: https://github.com/truenas/webui/pull/8091

PR #8093, opened 2023-04-21
NAS-121518 / 23.10 / Fixes for pool creation wizard - data vdevs step

denysbutenko:
See https://ixsystems.atlassian.net/browse/NAS-121518


undsoft (approved):
1. When manual disk selection is used, resulting string is too wide (and also not correct). I.e. I have 2 mixed vdevs , and it just says `Manual layout | 1 × HDD, 1 × STRIPE VDEVs | 240.07 TiB Total`. I think it should just say `Manual layout | 2 VDEVs | 240.07 TiB Total`.
2. I'm not really fond of casting to `unknown` to force SelectOption to work. In the future, somebody can decide to make a change in selects with an assumption that it would not affect other places, and it may break this page. Is there something else we can do?

Otherwise looks good, other points are addressed.

denysbutenko:
> 1. When manual disk selection is used, resulting string is too wide (and also not correct). I.e. I have 2 mixed vdevs , and it just says `Manual layout | 1 × HDD, 1 × STRIPE VDEVs | 240.07 TiB Total`. I think it should just say `Manual layout | 2 VDEVs | 240.07 TiB Total`.

This VDEVs string is uses in two places now:
1. Wizard on 2nd step, when layout is selected manually
2. Configuration Preview component

Should changes affect both places?
`1 × HDD, 1 × STRIPE VDEVs` -> `2 VDEVs`

> 2. I'm not really fond of casting to `unknown` to force SelectOption to work. In the future, somebody can decide to make a change in selects with an assumption that it would not affect other places, and it may break this page. Is there something else we can do?
> 
Perhaps we could:

- Keep it untyped
- Split `sizeAndType` form control to `size` and `type` separately
- Take a look at passing custom types to `SelectOption -> Option` values

undsoft:
1. I would update string in both places.
2. I would go with allowing custom types in SelectOption if it is not too hard, if we can use generics and if it won't break something.

denysbutenko:
> 1. I would update string in both places.
> 2. I would go with allowing custom types in SelectOption if it is not too hard, if we can use generics and if it won't break something.

image

The VDEVs string is updated. It will be `{n} VDEVs` in the Configuration Preview and `VDEVs: {n}` in the Description.

Regarding SelectOption, it involves more changes than I expected initially. So, I created a ticket for this - https://ixsystems.atlassian.net/browse/NAS-121638

PR #8098, opened 2023-04-22
NAS-121636 / None / Remove entity-form 💀🫡

undsoft:
Code review is enough.


denysbutenko (approved):
Hooray! A big chunk of legacy is gone. Test coverage is going up.

PR #8099, opened 2023-04-24
NAS-121447 / 23.10 / Clear subscriptions after websocket close

bvasilenko:
**Summary**

Need to clear the eventSubscriptions object in the `WebsocketService` after closing the WebSocket connection. It was not possible to directly access the `WebsocketService` from the `onClose` handler due to a circular reference error. Therefore, I solved the problem in a slightly different way.

**Testing**

Make sure this bug no longer happens: 
- NAS-121447
- NAS-121488 (duplicate)

Repro steps:

1. Wait for about 15-20 minutes having the WebUI opened
2. Every 60 seconds, the following issue observed:
   - "Connecting..." splash screen appears
   - 2 x CloseErrors appear in the console, see screenshot below
   - when this happens first time, then one additional websocket connection starts, keeping the old one open, without any message from server indicating that it's closed, which makes us think that UI logic is triggering this. see screenshot below
   

   - when this happens second time, then previous connection closes, and new opens, etc
   - life time of websocket connections is always 2 min

   - and, finally, it ends up with the error message `Maximum number of concurrent calls (20) has exceeded`

### DIFF
--- a/src/app/services/ws.service.ts
+++ b/src/app/services/ws.service.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@angular/core';
 import { Router } from '@angular/router';
-import { UntilDestroy } from '@ngneat/until-destroy';
+import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { UUID } from 'angular2-uuid';
 import {
   merge, Observable, of, throwError,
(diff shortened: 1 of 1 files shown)


undsoft (approved):
I've tried reproducing by restarting middlewared service. 
I'm okay with merging this and seeing if it resolves the issue.

PR #8101, opened 2023-04-24
NAS-121582 / 23.10 / Always "Switch to new dataset" message is incorrect on edit

AlexKarpov98:


denysbutenko (changes requested):
It looks like the wrong place for toasts (I know I added it). 
So let's move toasts to the corresponding forms and keep here redirect only.

denysbutenko (approved):
Thanks. It looks good.

PR #8106, opened 2023-04-25
NAS-121531 / 23.10 / Incorrect CPU core/thread display on dashboard when device display <600px

AlexKarpov98:


bvasilenko (approved):
Looks good to me

PR #8110, opened 2023-04-26
NAS-121679 / 23.10 / Update FirstUse state on Installed Apps

denysbutenko:
Add the `Choose Pool` button on the Installed Apps table on first use.

For testing, use a fresh system or `unset pool` to see the needed state.

image

After the pool is selected, you'll see the next state:
image


undsoft (changes requested):
I am not sure this is needed. I'd rather direct users to explore available apps first and only select a pool once they want to install an application.

denysbutenko:
> I am not sure this is needed. I'd rather direct users to explore available apps first and only select a pool once they want to install an application.

Sounds reasonable 👍🏻

denysbutenko:
In both cases, this will appear for user

image

PR #8114, opened 2023-04-26
NAS-121695 / 23.10 / Add tests to pool creation wizard

undsoft:
Also extracts some things into separate components.
Code review is enough.


dszidi (approved):
LGTM

PR #8115, opened 2023-04-27
NAS-121703 / 23.10 / Closing subscriptions when websocket closes

RehanY147:
Currently on master, UI just forgets about the subscriptions when the websocket closes for whatever reason. However, this still leaves a possibility that somewhere we are still subscribed to that event. This is preventive measure to make sure that before forgetting about it, we complete the observable.


AlexKarpov98 (approved):
Nice catch.

PR #8120, opened 2023-04-27
NAS-121177-1 /23.10/  SlideInService refactoring

Raz-Dva:
Created  IxSlideIn2Service.


src/app/modules/ix-forms/components/ix-slide-in/ix-slide-in-ref.ts
```diff
@@ -0,0 +1,29 @@
+import { ComponentRef } from '@angular/core';
+import { Observable, Subject } from 'rxjs';
+import { IxSlideIn2Component } from 'app/modules/ix-forms/components/ix-slide-in/ix-slide-in2.component';
+
+export class IxSlideInRef<T, R = unknown> {
+  private slideInClosed$ = new Subject<R>();
```
undsoft, line 6: You can just make this `readonly afterClosed$` and remove getter.

src/app/services/ix-slide-in2.service.ts
```diff
@@ -0,0 +1,64 @@
...
+  slideInRefList: IxSlideInRef<unknown>[] = [];
+
+  constructor(
+    private location: Location,
+    private router: Router,
+  ) {
+    this.closeOnNavigation();
+  }
+
+  setModal(modal: IxSlideIn2Component): void {
```
undsoft, line 25: Not a modal.

src/app/services/ix-slide-in2.service.ts
```diff
@@ -0,0 +1,64 @@
...
+  open<T, R>(modal: Type<T>, params?: { wide?: boolean; data?: R }): IxSlideInRef<T, R> {
+    this.slideInRefList.forEach((ref) => ref.close());
+    this.slideInRefList = [];
+    const slideInRef = this.slideIn2Component.openSlideIn<T, R>(modal, params);
+    this.slideInRefList.push(slideInRef);
+
+    return slideInRef;
+  }
+
+  close(): void {
```
undsoft, line 38: I don't see why we need this method. Shouldn't `closeAll` be used instead?

undsoft (changes requested):
I'm confused about boundaries between slidein component and slideinservice. In my mind component should equal to one opened slide in and service should manage logic between different slide ins.
The chain of calls to close a slide is sometimes confusing: `slideInRefClose` -> `ixSlideInRef.close()` -> `closeSlideIn`

src/app/modules/ix-forms/components/ix-slide-in/components/ix-modal-header/ix-modal-header.component.ts
```diff
@@ -11,5 +12,13 @@ export class IxModalHeaderComponent {
...
 
-  constructor(public slideInService: IxSlideInService) {}
+  constructor(
+    public slideInService: IxSlideInService,
+    private slideIn2Service: IxSlideIn2Service,
+  ) {}
+
+  close(): void {
+    this.slideInService.close();
+    this.slideIn2Service.closeEvent$.next(true);
```
undsoft, line 22: I don't think you should call something called `closeEvent` directly. You should probably `closeLast` here.

src/app/modules/ix-forms/components/ix-slide-in/ix-slide-in-ref.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  constructor(
+    private slideIn2Service: IxSlideIn2Service,
+    private slideIn2Component: IxSlideIn2Component,
+  ) {}
+
+  close(response?: D): void {
+    this.slideInClosed$.next(response);
+    this.slideInClosed$.complete();
+    this.slideIn2Service.slideInRefMap.delete(this.id);
```
undsoft, line 23: `slideInRefMap` feels like an internal detail of `IxSlideIn2Service` and shouldn't be called from here.

Raz-Dva: Could it look like ?
```
// slideInRef 
close(response?: D): void {
   ...
    this.slideInService.deleteRef(this.id);
  }
  // service 
    deleteRef(id: string): void {
    this.slideInRefMap.delete(id);
  }
```

src/app/services/ix-slide-in2.service.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const lastSlideInRef = Array.from(this.slideInRefMap.values()).pop();
+    lastSlideInRef.close();
+  }
+
+  closeAll(): void {
+    if (!this.isSlideInOpen) { return; }
+
+    this.slideInRefMap.forEach((ref) => ref.close());
+    this.slideInRefMap.clear();
```
undsoft, line 54: Clear is not necessary here as all refs will remove items from `slideInRefMap` when closed.

src/app/modules/ix-forms/components/ix-slide-in/ix-slide-in-ref.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  constructor(
+    private slideInService: IxSlideIn2Service,
+    private slideIn2Component: IxSlideIn2Component,
+  ) {}
+
+  close(response?: D): void {
+    this.slideInClosed$.next(response);
+    this.slideInClosed$.complete();
+    this.slideInService.deleteRef(this.id);
```
undsoft, line 23: What if instead of calling `deleteRef`, we have service subscribe with `take(1)` to `slideInClosed$` for every slide in ref that was opened?

src/app/pages/datasets/components/dataset-form/dataset-form.component.spec.ts
```diff
@@ -66,6 +69,7 @@ describe('DatasetFormComponent', () => {
         EncryptionSectionComponent,
         QuotasSectionComponent,
         OtherOptionsSectionComponent,
+        IxModalHeaderComponent,
```
undsoft, line 72: It's better not to declare `IxModalHeaderComponent`s like this. Instead add `mockProvider(IxSlideInService2)` to providers.

undsoft (approved):
Looks good 👍🏻

src/app/modules/ix-forms/components/ix-slide-in/ix-slide-in-ref.ts
Raz-Dva: In IxSlideInRef, I created a Subject onClose$ with which service will work

src/app/modules/ix-forms/components/ix-slide-in/ix-slide-in-ref.ts
```diff
@@ -1,27 +1,20 @@
...
-
   close(response?: D): void {
+    this.onClose$.next(this.id);
     this.slideInClosed$.next(response);
     this.slideInClosed$.complete();
-    this.slideInService.deleteRef(this.id);
-    this.slideIn2Component.closeSlideIn();
   }
 
   get afterClosed$(): Observable<D> {
```
undsoft, line 20: Getter is unnecessary. You can just rename `slideInClosed$` and make it readonly.

src/app/modules/ix-forms/components/ix-slide-in/ix-slide-in-ref.ts
undsoft: Why do you need a new subject, why not just use `afterClosed`?

src/app/modules/ix-forms/components/ix-slide-in/ix-slide-in-ref.ts
Raz-Dva: Now I understand :+1:

PR #8121, opened 2023-04-27
NAS-121704 / 23.10 / Add R20,R40,R50 to mock utility

dszidi:
To test, enable mock utility and set controller model to R20/R40/R50

```
yarn ui -e -m <r20 | r40 | r50>
```


AlexKarpov98 (approved):
looks good.

PR #8125, opened 2023-04-28
NAS-121596 / 22.12.3 / Fixed creating a new user fails

Raz-Dva:
Go to **Credentials** -> **Users** 
Create a new user. Check the availability of fields that a new user can change.


undsoft (changes requested):
What is `param`? I don't see it being documented anywhere.

Raz-Dva:
This pull request is for bluefin only. The description [NAS-121222](https://ixsystems.atlassian.net/browse/NAS-121222) states that in Bluefin user.shell_choices has 2 arguments. First should be null, second is group IDs list.

undsoft (approved):
I see, looks good then.

PR #8127, opened 2023-04-30
NAS-121735 / 23.10 / Mock Utility disk and pool mocking support

dszidi:
Command structure is now as follows

- `mock`: same as before. Load and generate config.json files. Also global enable/disable
- `mock-enclosure`: Basically just renamed mock-opt command for clarity
- `mock-disks`: Mocks unassigned disks. Options include disk size and repeats
- `mock-pool`: Mocks a single pool with data-topology

The mock command's guided workflows have been updated with all the new options.
If desired, any or all sections can be skipped during the generator questions


src/app/services/ws.service.ts
```diff
@@ -59,19 +59,27 @@ export class WebSocketService {
           return throwError(() => data.error);
         }
 
-        // Mock Data
+        // Mock Data Response
+        const isEnabled = (
```
undsoft, line 63: Perhaps just exclude all of this to `canMockMessage()` function. Comments are unnecessary.

scripts/ui/ui.ts
```diff
@@ -688,6 +803,319 @@ function mockConfig(command: Command, options: CommandOptions): void {
...
+        process.exit(0);
+    }
+  }
+
+  mockConfigReport({
+    showHeader : true,
+    showFooter: true,
+  });
+}
+
```
undsoft, line 1118: This file starts to become swol now.

undsoft (approved):
It looks good, but I also get a typescript import error when coming from master branch.
```
Property 'mockEnclosure' is missing in type '{ enabled: false; enclosureOptions: { controllerModel: string; expansionModels: undefined[]; dispersal: EnclosureDispersalStrategy.Default; }; systemProduct: string; }' but required in type 'MockEnclosureConfig'.
```

I cannot do `ui remote` and have to remove file manually.
Would be good to do something about it if possible.

src/app/services/ws.service.ts
dszidi: Done

PR #8128, opened 2023-05-01
NAS-121551 / 23.10 / Support Ordering at Network > Edit Interface > Link Aggregation Interfaces

bvasilenko:
**Testing**

On the **Network > Edit Interface** form, 

Insert multiple values into the **Link Aggregation Interfaces** input box, and make sure that they are saved in the correct order


dszidi (approved):
When set to FAILOVER, the interfaces can be edited and the order is preserved after save.
Nice job

image

undsoft:
Let's add a handle to list item to indicate that they are draggable.
Снимок экрана 2023-05-02 в 19 37 51

PR #8129, opened 2023-05-01
NAS-120203 / 13.0 / Create/Edit Dataset > Fix Discrepancies between GUI and command line output of dataset checksum

bvasilenko:
**Testing**

On the **Create/Edit Dataset** form,

1. Create a pool `tank1` with the default options and a child dataset `tank1/test1` also with the default options. When creating the `tank1/test1` dataset, make sure that the GUI indicated "Checksum: inherit (on)" 
2. Go back to the “Edit Options” for the `tank1/test1` 

**Expected result:**

The value displayed in the WebUI should match the value returned by the call 
```
zfs get checksum tank1/test1
```

For more details please see ticket.


undsoft (approved):
Looks good.

PR #8134, opened 2023-05-02
NAS-121692 / 23.10 / SSH password login enabled checkbox does not stay set aft…

undsoft:
See ticket for description.


dszidi (approved):
Checkbox reflects saved values.
LGTM

PR #8135, opened 2023-05-02
NAS-121782 / 23.10 / Remove S3 service

undsoft:
Code removed due to middleware changes. Requires fresh middleware.

Code review should be enough.


bvasilenko (approved):
Looks good

PR #8137, opened 2023-05-02
NAS-121712 / 23.10 / Different apps with the same name are incorrectly handled

bvasilenko:
**Summary**

Now when user clicks an "Installed" badge at **Apps > Available**, it should navigate to this app's details page, respecting the app's catalog and train. It should not navigate to the apps from different trains, or catalogs.

**Testing**

1. Have an installed app
2. On the **Apps > Available**, find this app and click "Installed" badge next to an app's name
   **Expected result:** user is navigated to this app's details page. If it's impossible to lookup the app (user has renamed the app during it's installation), then user is navigated to the list of all installed apps instead.


denysbutenko (approved):
Well done!

PR #8140, opened 2023-05-03
NAS-121799 / 23.10 / Minor code cleanup

undsoft:
Code review is enough.


src/app/modules/ix-forms/components/ix-ip-input-with-netmask/ix-ip-input-with-netmask.component.ts
```diff
@@ -71,8 +71,7 @@ export class IxIpInputWithNetmaskComponent implements ControlValueAccessor {
     this.onChange(value);
   }
 
-  private setAddressAndNetmask(ipWithNetmask: string): void {
-    ipWithNetmask = ipWithNetmask || '';
+  private setAddressAndNetmask(ipWithNetmask = ''): void {
```
RehanY147, line 74: Previously if you got a value from somewhere that you didn't know is null, you could call the method with that value safely. Now, it's possible to call the method without a  net mask at all. Maybe a different method is better. `setEmptyAddressAndNetmask(): void { this.setAddressAndNetMask(''); }` and the check can be performed outside the method. As is, you could assume that the method just sets the correct values if you just call it without any values.

src/app/pages/datasets/modules/permissions/containers/dataset-acl-editor/dataset-acl-editor.component.ts
```diff
@@ -102,18 +102,19 @@ export class DatasetAclEditorComponent implements OnInit {
...
-    ).subscribe(() => {
-      this.dialogService.confirm({
-        title: helptext.dataset_acl_recursive_dialog_warning,
-        message: helptext.dataset_acl_recursive_dialog_warning_message,
-      }).pipe(untilDestroyed(this)).subscribe((confirmed) => {
-        if (confirmed) {
-          return;
-        }
+    ).subscribe((confirmed) => {
+      if (confirmed) {
```
RehanY147, line 113: Could have a filter operator

RehanY147 (approved):

src/app/pages/datasets/modules/permissions/containers/dataset-acl-editor/dataset-acl-editor.component.ts
undsoft: True, I may change it in future.

src/app/modules/ix-forms/components/ix-ip-input-with-netmask/ix-ip-input-with-netmask.component.ts
undsoft: It's only called in one place where this is not an issue.

PR #8143, opened 2023-05-03
NAS-121762 / 23.10 / Updates to Sort filter in new Apps

AlexKarpov98:
Check ticket description for more details.
Try different scenarios by changing filtering data :)


src/app/pages/apps/store/available-apps-store.service.ts
```diff
@@ -393,4 +361,120 @@ export class AvailableAppsStore extends ComponentStore<AvailableAppsState> {
       }),
     );
   }
+
+  private sortAppsByName(filteredApps: AvailableApp[]): AppsByCategory[] {
+    const appsByCategory: AppsByCategory[] = [];
+
+    const alphabet = [
```
undsoft, line 368: Not really sustainable. There may be an app starting with a number or cyrillic letter. Need to look at existing letters in catalog apps.

src/app/pages/apps/services/applications.service.ts
```diff
@@ -140,4 +143,47 @@ export class ApplicationsService {
   stopApplication(name: string): Observable<Job<ChartScaleResult>> {
     return this.ws.job('chart.release.scale', [name, { replica_count: 0 }]);
   }
+
+  convertDateToRelativeDate(date: Date): string {
+    const diff = Math.round(((new Date() as unknown as number) - (date as unknown as number)) / 1000);
```
undsoft, line 148: 1. You may be able to use `formatRelative` from `date-fns` here.
2. Seems to be impossible to test this right now, but it looks like granularity is too much. It should be something like `Last 24 hours`, `Last 3 days`, `Last week`, `Last month`, not `5 hours ago`, `5 days ago`.

undsoft (changes requested):

src/app/pages/apps/store/available-apps-store.service.ts
AlexKarpov98: @undsoft - you mean check all the apps names and get unique first letters?

src/app/pages/apps/services/applications.service.ts
AlexKarpov98: @undsoft  - `date-fns` provides a strange solution with time, I was not able to find solution to fix it, 
If we can keep my method, I can remove unnecessary options and leave only few of them like:
`Last 24 hours, Last 3 days, Last week, Last month, Long time ago`

Screenshot 2023-05-03 at 17 45 17

undsoft: Okay, let's do that.

src/app/pages/apps/store/available-apps-store.service.ts
undsoft: Yup.

undsoft (approved):
Looks good.

PR #8147, opened 2023-05-04
NAS-121714 / 23.10 / Placeholder icon is flashing when switching between installed apps

AlexKarpov98:


bvasilenko (approved):
Works good for me

PR #8154, opened 2023-05-05
NAS-121738 / 23.10 / Snapshot Lifetime Unit Would Be Hidden If Custom ScheduleIs Too Long

AlexKarpov98:
Scheduler width & content is fixed like that:
Screenshot 2023-05-05 at 14 11 52


denysbutenko (approved):
👍🏻

PR #8155, opened 2023-05-05
NAS-121839 / 23.10 / Icon fixes for Manual Disk Selection

dszidi:
Testing:

- Check to see that enclosure icon is no longer cut off. It should look like this

image

- Check to see that SSD icon is used when appropriate. Should look like this...

image


RehanY147 (approved):
Nice

PR #8156, opened 2023-05-05
NAS-121844 / 23.10 / Make sure mock utility doesn't break UI

dszidi:
Removed a condition that broke UI if `mock-pool` was disabled

Try enabling/disabling `mock-enclosure` and `mock-disks` with and without `mock-pools` enabled
There should not be any combination that results in browser console error that breaks the UI

### DIFF
--- a/src/app/core/testing/utils/mock-enclosure.utils.ts
+++ b/src/app/core/testing/utils/mock-enclosure.utils.ts
@@ -23,10 +23,7 @@ export class MockEnclosureUtils {
 
   constructor() {
     const diskOptions = this.mockConfig?.diskOptions;
-
-    if (diskOptions?.mockPools) {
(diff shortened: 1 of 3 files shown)


undsoft (changes requested):
I've added this condition, because we had problems with production installations, where we'd get "cannot read `mockPools` of undefined".

dszidi:
> I've added this condition, because we had problems with production installations, where we'd get "cannot read `mockPools` of undefined".

Ah ok. In that case since mockConfig is optional, I'll remove it from the prod environment file and do an earlier check in ws.service as a condition for creating mockUtils. That way mockUtils won't get created at all unless we're in dev mode and mockConfig is set

undsoft (approved):
👍🏻

PR #8157, opened 2023-05-05
NAS-121843 / 23.10 / Add tests to AppDetailComponent

undsoft:
Extracts some code in separate components and adds tests to them.

Testing: check affected components.


RehanY147 (approved):
Should have a bit more margin underneath the "Similar Apps" title. Also, the keywords list is empty. But I guess that might be because the tags field isn't filled in.

PR #8160, opened 2023-05-08
NAS-121377: Make Installed Apps table responsive

bvasilenko:
**Testing**

At **Apps > Installed** check how it works on different resolutions, including mobile screen size


denysbutenko (approved):
Works like on the Datasets page 👍🏻

PR #8162, opened 2023-05-08
Revert "NAS-121686 / 23.10 / Add usableUnusedDisks stream"

RehanY147:
Reverts truenas/webui#8141

The original PR leaves a couple of bugs that need to be addressed. Even if I do nothing and have a couple of disks in the inventory, they don't show up in the drop downs on the second step. Also, if I go back to the first step and mark some checkboxes, the second step options don't update.
Furthermore, the `take(1)` for `usableUnusedDisks$` from the store needs to be removed so when disks update the options can update. But removing that causes infinite emits from the selector because in that subscription we are updating form values which triggers state update which triggers `usableUnusedDisks$` update in an infinite loop.


denysbutenko (approved):
👍🏻

PR #8163, opened 2023-05-08
NAS-121685 / 23.10 / Added support for Mirror and Raidz vdev types

RehanY147:
Ticket description for details


undsoft (approved):
Looks good. 
The only issue I can see is that you can end up with an empty `Width` and Next button will still be active.
Снимок экрана 2023-05-08 в 22 23 23

RehanY147:
With create a new PR to address ticket comments. Merging to avoid conflicts with `master` and other PRs

PR #8166, opened 2023-05-09
NAS-121279 / 23.10 / Restore scrolling logic

bvasilenko:
**Testing**

Go to **Datasets** page, with a lots of datasets to trigger the scrollbar. Alternatively, you can resize your browser's window, or add a temporary DOM element with height of 2000px.

Expected scrolling behavior:

1. Datasets block isn't scrollable. Only the page scrolls.
2. The height of details block is dynamically adjusted as the page scrolls.

https://user-images.githubusercontent.com/20611516/236998605-82b2b9ae-7093-41f8-b174-a734debeaf02.mp4


src/app/pages/datasets/components/dataset-management/dataset-management.component.html
```diff
@@ -65,16 +65,11 @@
         class="dataset-tree-wrapper"
       >
         <div class="dataset-tree-inner">
-          <ix-tree-virtual-scroll-view
```
denysbutenko, line 68: Using `ix-tree-virtual-scroll-view` here is intentional, as removing it will cause performance issues when the system has multiple datasets.

denysbutenko (changes requested):

src/app/pages/datasets/components/dataset-management/dataset-management.component.html
undsoft: @denysbutenko Didn't we manage to implement both virtual scrolling and full page scrolling or am I misremembering things?

denysbutenko: > @denysbutenko Didn't we manage to implement both virtual scrolling and full page scrolling or am I misremembering things?

@undsoft I don't remember this tbh. I think we have skipped backporting `ix-tree-virtual-scroll-view` for Bluefin. It may cause different scrolling behavior between Bluefin and Cobia versions.

denysbutenko (approved):
The dataset tree takes more space and the whole page can be scrolled. Looks good.

PR #8170, opened 2023-05-09
NAS-120602 / 13.0 / Uncommanded reboot from logging into the web interface

undsoft:
To reproduce previous issue.
1. In `reboot.component.ts` replace add:
```
console.log('rebooted');
return;
```
before
```
this.ws.call('system.reboot', {})
```

2. Remove `{ skipLocationChange: true }` in `topbar.component.ts`.
3. Press reboot in UI.
4. Change address in browser to `http://localhost` and press Enter.
Observe that you are at reboot page again.


AlexKarpov98 (approved):
Looks good.

PR #8174, opened 2023-05-10
NAS-121773 / 23.10 / Show capacity for disks only on manage devices

denysbutenko:
Remove the capacity value from VDEV items on the Manage Devices page.

An example
image


bvasilenko (approved):
Looks good to me

PR #8181, opened 2023-05-10
NAS-121716 / 23.10 / Apps - make Details and Installation form pages more responsive

AlexKarpov98:


dszidi (commented):
So far looking quite promising. On the Apps Available page, the cards are getting cut off on mobile screens as they're wider than the viewport. It would be better if the flex layout was column instead of row and their were appropriate margins. Perhaps something like this...


image

dszidi (commented):
On the individual app pages the details cards should come after the screenshots

This is what it looks like now...

image


It needs to look more like this...

image

dszidi (commented):
Installation form looks nice and neat. Well done

image

dszidi (approved):
Code looks good and wIth the app page details now rendered at the bottom of mobile layout, everything looks good.

PR #8183, opened 2023-05-11
NAS-121879 / 23.10 / Implement notes card with markdown

bvasilenko:
**Testing**

At **Applications > Installed** the new "notes" card should appear. See details in the ticket


src/app/pages/apps/components/installed-apps/app-notes-card/app-notes-card.component.html
```diff
@@ -4,19 +4,14 @@ <h3 mat-card-title>{{ 'Notes' | translate }}</h3>
...
-        <div class="notes-item">
-          {{ '$ mychart cli command' | translate }}
-        </div>
+      <div #notes class="notes-list">
+        <markdown [data]="app.info.notes"></markdown>
+      </div>
+      <div *ngIf="isCollapsed && showMoreLess" class="gradient"></div>
+      <div *ngIf="showMoreLess" class="more-less">
+        <span (click)="changeCollapsed()">
+          {{ isCollapsed ? 'View More' : 'Collapse' | translate }}
```
undsoft, line 13: This won't get picked by message extractor. Needs to be `('View More' | translate) : ('Collapse' | translate)`

undsoft (changes requested):
1. Can we alter default styles inside of a markdown block? Currently headers are way too large.
2. View More link implies navigation. Let's use a button.

bvasilenko:
For point 2, the **`<a>`** tag is used, as well as in the components on the screenshots

image

image

undsoft (changes requested):
Let's make headers even smaller, so that they don't stand out. I think h1 in markdown should be less intrusive than h3 on cards.
I'm okay with `19px` with `font-weight: 500`.

As to `a`, in your examples most links are used appropriately and take you to a different place. In notes, View More / Collapse is an action and should be a button.

bvasilenko:
@undsoft ok, header size is adjusted

undsoft (approved):
Looks good overall. There is a small issue when viewing notes on smaller screen. When page is loaded in this mode, `View More` button doesn't appear.

https://github.com/truenas/webui/assets/1481270/27518220-522b-4de0-bd3e-734df81ffea8

bvasilenko:
Was too quick to hit "merge" button. I'll make another PR to fix the last issue.

PR #8186, opened 2023-05-11
NAS-121884 / 23.10 / Sudo Dialog appears 3x after Replication Task Created for remote non root user

AlexKarpov98:


dszidi (approved):
Code looks good. Was able to create replication task as described in ticket without unwanted sudo dialogs

PR #8194, opened 2023-05-14
NAS-121320 / 23.10 / Apps table is updated when applications are updated or removed

Raz-Dva:
Check that the apps table is updated correctly after deleting or updating an apps.


bvasilenko (approved):
Works ok for me

PR #8198, opened 2023-05-15
NAS-121946: App is still marked as installed in available tables after uninstall

AlexKarpov98:


bvasilenko (approved):
Additions to store work ok for me

PR #8202, opened 2023-05-16
zh-Hans: refine translations

KevinMX:
Mainly about ZFS scrub which most were translated as "擦除", literal means: "erase" which is obviously not the real case, and kinda terrifying for zh-Hans users :)

I chose "数据清理" which is the same as what Oracle use for their [Solaris ZFS docs](https://docs.oracle.com/cd/E26926_01/html/E25826/gbbwa.html), literal means "data cleaning". Synology DSM use this translation as well for their BTRFS scrub if I remembered correctly.


denysbutenko (approved):
Appreciate your contribution.

PR #8210, opened 2023-05-18
NAS-121987 / 23.10 / Data-tag missing from Show-Builtin-Users/Groups on Users/Groups page

AlexKarpov98:


undsoft (approved):
Looks good, but let's add `mat-slide-toggle` to `angular-test-ids/require-test-id` in `eslintrc`.

PR #8214, opened 2023-05-19
NAS-121675 / 23.10 / Adding User Auth key doesn't re-evaluate after home directory error - blanks Key on save, and doesn't close right panel

bvasilenko:
**Summary**

Middleware does not support updating `home_create` simultaneously with any other property.
Make a separate `user.update` request for `home_create`.


denysbutenko (approved):
It works well.

PR #8217, opened 2023-05-19
NAS-122057 / 23.10 / Rename official catalog

denysbutenko:
Rename official catalog name from `OFFICIAL` -> `TRUENAS`


bvasilenko (approved):
Looks good to me

PR #8218, opened 2023-05-19
NAS-122048 / 23.10 / Removing tftp and webdav

undsoft:
Testing: see that tftp and webdav related stuff is not there.


dszidi (approved):
LGTM

No tftp.

image

No WebDav

image

PR #8219, opened 2023-05-19
NAS-121998 / 23.10 / Improvements for enclosure options step

AlexKarpov98:
for testing please check ticket description.
Screenshot 2023-05-20 at 00 11 19


denysbutenko (approved):
image

The `ix-select` and `description` has the same `*ngIf` condition. When `isLimitingToSingle` is false there will be rendered parent element which may affect the layout. It would probably be safer to move `ngIf` to a parent element. Otherwise, it looks good 👍🏻

PR #8224, opened 2023-05-22
NAS-122002 / 23.10 / ManualDiskSelectionStore cleanup

RehanY147:
See ticket for details

ManualDiskSelectionDragToggleStore exists because if it is part of the old store, when updating the toggle boolean, as we update the state (because of the immutable patterns, we can't just update one boolean. The whole state has to be a new object), the vdevs gets cloned, which means they are considered updated and get re-rendered on the dom and the dragging can't work because the view we were trying to drag got destroyed in the process. So, drag toggle has to exist in a separate store.

Will add tests in a separate PR.


dszidi (approved):
Code looks good. Drag and drop behavior works as expected. Was successfully able to move disks from disk selection area to vdev and back.

PR #8197, opened 2023-05-15
NAS-121177 / 23.10 / IxSlideInService refactoring

Raz-Dva:
Replacing IxSlideInService with IxSlideIn2Service.

PR #8229, opened 2023-05-23
NAS-122065 / 23.10 / Replace the IxSlideInService with a new IxSlideInService

AlexKarpov98:


AlexKarpov98:
> https://github.com/truenas/webui/pull/8197/files

is it the same PR as this one?

undsoft:
Okay that PR got merged, probably easier to start a new one if there are any old usages that remain. When refactoring avoid using `slideInService.closeLast()`, the idea is that you'll have `slideInRef` like `matDialogRef` that you can use to close component.

Also instead of using methods from a component like this `componentInstance.setupForm()`, we should now pass arguments via `data` and they'll be available via `SLIDE_IN_DATA` token inside.

bvasilenko:
@AlexKarpov98 once PR is ready for review, please remove "WIP" from the description

undsoft (approved):
Looks good.

PR #8234, opened 2023-05-26
NAS-121999 / 23.10 / Added the remaining vdev types steps in the Pool Wizard

RehanY147:


undsoft (changes requested):
1. I don't think we should hide other controls before layout is selected.
2. It's better to reimplement layout limiting logic by using inputs in `LayoutStepComponent`. This will match how rest of the communication to data store was designed. 

- Pass in and implement `canChangeLayout` for steps where layout cannot be changed.
- Implement `limitLayouts` to optionally limit layouts on a specific step.
- I will confirm with middleware, but it's likely that we will want some disks limited by size on some steps, like not allowing smaller Spares than disks used in Data.
4. Then in tests for each steps, you can use `MockComponent(LayoutStepComponent)` and check that correct arguments were passed in. Current tests don't really do anything.
5. It's time to implement `Save And Go To Review` button.
6. Step hint text (and Back/Next buttons) needs to be aligned with controls.
7. Let's adjust `Optional vdev that can be removed` to just `Optional`.

RehanY147:
@undsoft Will create a new PR for style fixes and for tests

src/app/pages/storage/modules/pool-manager/components/pool-manager-wizard/steps/3-data-wizard-step/data-wizard-step.component.html
```diff
@@ -24,4 +28,15 @@
   >
     {{ 'Next' | translate }}
   </button>
+  <!-- TODO: Implement -->
```
undsoft, line 31: Todo comment can be removed.

src/app/pages/storage/modules/pool-manager/components/pool-manager-wizard/steps/3-data-wizard-step/data-wizard-step.component.html
```diff
@@ -1,7 +1,11 @@
+<div class="step-description" [innerHTML]="helptext.data_vdev_description">
```
undsoft, line 1: Needs translation and doesn't need to be innerHTML.

src/app/pages/storage/modules/pool-manager/components/pool-manager-wizard/steps/4-log-wizard-step/log-wizard-step.component.ts
```diff
@@ -8,11 +12,20 @@ import { PoolManagerStore } from 'app/pages/storage/modules/pool-manager/store/p
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class LogWizardStepComponent {
+  @Output() goToLastStep = new EventEmitter<void>();
   protected readonly VdevType = VdevType;
+  readonly helptext = helptext;
 
   protected readonly inventory$ = this.store.getInventoryForStep(VdevType.Log);
-
+  protected readonly hasDataVdevs$ = this.store.topology$.pipe(
```
undsoft, line 20: I don't think we need to have this check on all the steps.

src/app/pages/storage/modules/pool-manager/components/pool-manager-wizard/steps/5-spare-wizard-step/spare-wizard-step.component.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  beforeEach(() => {
+    spectator = createComponent();
+    loader = TestbedHarnessEnvironment.loader(spectator.fixture);
+  });
+
+  it('has back button', async () => {
+    const backBtn = await loader.getHarness(MatButtonHarness.with({ text: 'Back' }));
+    expect(backBtn).toBeTruthy();
+  });
+});
```
undsoft, line 40: Need to mock `LayoutStepComponent` and store and check that everything is being passed between the two.

undsoft (changes requested):
1. Helptext on steps still needs to be horizontally aligned with controls.
2. So, how is not having enough disks supposed to work now? It is currently possible to select Raidz1 - 16GiB SSD - width 3, even when I only 1 matching disk.
3. It is my understanding that all other types are optional, and so we should mention this.

src/app/pages/storage/modules/pool-manager/components/pool-manager-wizard/steps/4-log-wizard-step/log-wizard-step.component.ts
RehanY147: I think we do. Each step has a next button which is not dependent on whether or not there are data vdevs selected. So, user can select next -> next -> next. Or not see the next button enabled when no data vdevs are selected. User shouldn't be able to open other steps when there's an error in a previous step but that's possible right now by just clicking the step title. I'll create a separate ticket for that.

undsoft: It's true that we probably shouldn't allow to go to other steps if first step or data step is not filled in. But there is no use disabling this button on Log step. User either shouldn't be there, or if he's there disabled button won't communicate what the problem is. We should still let him get to review step to see warnings and errors.

undsoft (approved):

PR #8235, opened 2023-05-26
NAS-122082 / 23.10 / Not starting SMB service when adding a share results in broken form

bvasilenko:
**Testing**

Please see repro steps described in the ticket


dszidi (approved):
Code looks good. Hitting no now results in the share being created with the smb service turned off.

LGTM

PR #8242, opened 2023-05-29
NAS-122084 / 23.10 / Apps - update page hierarchy

denysbutenko:
For testing, go through list from the original ticket


undsoft (changes requested):
1. Update All has different font than rest of the buttons. Also would be good to have some space between icon and a button.
2. Let's also make Discover Apps button wider if possible.
3. When I click on Custom App on m40, I get redirected back with `[ENOENT] None: Catalog TRUENAS does not exist` in console. The error is fine, but it's a good opportunity to fix how we handle it. Currently we just silently redirect back, would be better to show an error dialog.
4. I think breadcrumbs are confusing this way. Let's change to no breadcrumbs on Installed, just `Applications` on Discover and `Applications -> Discover` on nested pages like an app installation page. Latter applies to Manage Catalogs too.

undsoft:
BTW let's also rename menu items and call new apps just Apps, while using "Apps (Old)" for old apps. New apps should also be higher on the list.

denysbutenko:
@undsoft breadcrumbs definitions work unpredictably, tbh. So making some arrangements would allow us to remove breadcrumbs from `/apps/installed`, on `/apps/available` there is `Applications -> Discover`, and nested pages have `Applications -> Discover -> Install App` etc. Breadcrumbs need refactoring to be predictable and have more flexibility.

undsoft (approved):
This works too, thanks.

PR #8243, opened 2023-05-29
NAS-122198 / 23.10 / Fix active state for Reporting menu item

denysbutenko:
For more context, follow the original ticket.

For testing, go to Reporting and check if the side nav menu item is active, then change the report type using the dropdown menu and check it again.

### DIFF
--- a/src/app/services/navigation/navigation.service.ts
+++ b/src/app/services/navigation/navigation.service.ts
@@ -117,7 +117,7 @@ export class NavigationService {
       type: MenuItemType.Link,
       tooltip: T('Reports'),
       icon: 'insert_chart',
-      state: 'reportsdashboard/cpu',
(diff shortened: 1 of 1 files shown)


AlexKarpov98 (changes requested):
Seems to be ok.
But now we have another problem -- clicking on the `Reporting` tab (like double click) - right hand switcher disapears.

Screenshot 2023-05-29 at 11 16 10


Please check my video:
https://github.com/truenas/webui/assets/22980553/bccc7ed6-107b-4761-b3bf-3a8f52299bfa

RehanY147 (approved):

denysbutenko:
@AlexKarpov98 I haven't found a way to prevent the toolbar from disappearing. Let's keep it as is.

PR #8245, opened 2023-05-29
NAS-122149 / 23.10 / Remove the disk import feature from webUI for Cobia

AlexKarpov98:


src/app/pages/datasets/components/dataset-management/dataset-management.component.ts
```diff
@@ -55,7 +53,6 @@ import { selectIsSystemHaCapable } from 'app/store/system-info/system-info.selec
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class DatasetsManagementComponent implements OnInit, AfterViewInit, OnDestroy {
```
undsoft, line 55: `layoutService` can be removed from constructor.

undsoft (approved):
The button is not there.
Снимок экрана 2023-05-29 в 12 46 12

PR #8246, opened 2023-05-29
NAS-121898: New replication encryption option (fix for wizard)

AlexKarpov98:


src/app/pages/data-protection/replication/replication-wizard/replication-wizard.component.ts
```diff
@@ -235,6 +235,8 @@ export class ReplicationWizardComponent {
       payload.encryption_key_location = data.encryption_key_location_truenasdb
         ? truenasDbKeyLocation
         : data.encryption_key_location;
+
+      payload.encryption_inherit = data.encryption_inherit;
```
undsoft, line 239: You could just move it up where initial object is built.

undsoft (approved):
Looks good.

PR #8248, opened 2023-05-30
NAS-119953 / 23.10 / Make alert/email settings easier to find

AlexKarpov98:
results:

https://github.com/truenas/webui/assets/22980553/a9de0aee-31b2-475d-af58-afea0bb99e05


undsoft (changes requested):
I think there are too many items in this menu now. Please check with @dszidi on suggestions what to do here.

dszidi:
As per our Slack discussions...

- Create a new page called `Alert Settings`. Show both `Alert Categories` and `Alert Services`  should be on this page as cards one underneath the other (Make sure `Alert Services` is on top as that is almost guaranteed to take up less vertical space)
- Add `Alert Settings` menu item to System Settings submenu.
- Move `Email Settings` to go `System Settings` > `General` as a card.

undsoft (changes requested):
1. When I click Save on alert categories, the loading indicator is not cleared until I click somewhere. Probably a change detection issue.
2. If possible, let's alter headers a little bit. Ideally it would look like this:
```
Alert Settings

<smaller header>Alert Services</>.          [filter][columns][add]
[table]

[everything else]
```

AlexKarpov98:
updated:
Screenshot 2023-06-12 at 13 16 01

undsoft (approved):
Looks good. Some minor adjustments to Email card:
1. I think we can reduce amount of rows. 
For Gmail, we can just show "Send Mail Method: Gmail"
For SMTP: 
- Send Mail Method: SMTP
- From: John <john@john.com> via server.com
Obviously don't show parts that are missing.

src/app/pages/system/general-settings/email/email-card/email-card.component.html
```diff
@@ -0,0 +1,40 @@
...
+      </mat-list-item>
+
+      <ng-container *ixWithLoadingState="emailConfig$ as emailConfig">
+        <ng-container *ngIf="!emailConfig?.oauth?.client_id">
+          <mat-list-item>
+            <span class="label">{{ 'From' | translate }}:</span>
+            <span *ixWithLoadingState="emailConfig$ as emailConfig" class="value">
+              {{ emailConfig.fromname }}
+              {{ emailConfig.fromemail || '-' }}
+              {{ emailConfig.outgoingserver ? (' via ' | translate) + emailConfig.outgoingserver : ''}}
```
undsoft, line 33: Just translating `via` is not a good idea, because in different languages the order of words may not be the same.
It should be `{email} via {server}` when both parts are set.

undsoft (approved):

PR #8249, opened 2023-05-30
NAS-118894 / 23.10 / WebUI changes in Data Protection Section

AlexKarpov98:
Added more tooltips for easier reading of data.
Screenshot 2023-05-30 at 15 54 30

Running button colour updated (instead of orange):
Screenshot 2023-05-30 at 15 55 58


RehanY147 (approved):
Looks good

PR #8250, opened 2023-05-30
NAS-120010: Bring memory size converter functions in one place and reformat

AlexKarpov98:


denysbutenko (approved):
The changes look good. However, `convertBytesToHumanReadable` was marked as deprecated, and it needs to be replaced with a filesize pipe. Anyway, replacing it in one place would be much easier even if it is out of ticket scope.

image

AlexKarpov98:
> The changes look good. However, `convertBytesToHumanReadable` was marked as deprecated, and it needs to be replaced with a filesize pipe. Anyway, replacing it in one place would be much easier even if it is out of ticket scope.
> 
> image

but why are you using `this.storage.convertBytesToHumanReadable` -> it's now only available inside of the IxFormatterService :)

denysbutenko:
> > The changes look good. However, `convertBytesToHumanReadable` was marked as deprecated, and it needs to be replaced with a filesize pipe. Anyway, replacing it in one place would be much easier even if it is out of ticket scope.
> > image
> 
> but why are you using `this.storage.convertBytesToHumanReadable` -> it's now only available inside of the **IxFormatterService** :)

The screenshot was made from another branch.

AlexKarpov98:
> The screenshot was made from another branch.
So, no problem with deprecated method?

denysbutenko:
> @denysbutenko So, no problem with deprecated method?

My point is that `storage.convertBytesToHumanReadable` has a `deprecated` status when `formatter.convertBytesToHumanReadable` don't have it while they are the same and need to be replaced with a filesize pipe in the future. I'd suggest adding the `deprecate` status to `formatter.convertBytesToHumanReadable` just to keep that in mind.

PR #8251, opened 2023-05-31
NAS-122153 / 23.10 / Remove Rsync Service

denysbutenko:
~While testing, the `rsync` service may appear on the services page.~ Added additional filter to exclude removed services. It will be a middleware issue also to adjust the response with available services.


dszidi (approved):
LGTM

PR #8252, opened 2023-05-31
NAS-122214: HA Manual Update UI has "Upload FIle Temporary Location" dropdown that is not used

AlexKarpov98:


undsoft (changes requested):
The task specifically refers to an HA update form, where this select is not used in the request.
We still need it for normal update where it is in fact used.

undsoft (approved):
Looks good.

PR #8254, opened 2023-05-31
NAS-122116 / 23.10 / No Error/notification of duplicate CSR when trying to save

AlexKarpov98:
For current fix it's okay to show modal, but in future we need to fix this TODO:

// TODO: Need to update error handler to open step with an error.
**src/app/pages/credentials/certificates-dash/csr-add/csr-add.component.ts**

Screenshot 2023-05-31 at 18 24 58

Screenshot 2023-05-31 at 18 21 27


RehanY147 (approved):
Looks good

PR #8263, opened 2023-06-02
NAS-122206 / 23.10 / Investigate `dataset does not exist` when selecting source in Replication Wizard

AlexKarpov98:
result:
Screenshot 2023-06-02 at 14 10 58


undsoft:
@AlexKarpov98 Have you looked into why there are non-existing dataset available in ix-explorer here?

AlexKarpov98:
@undsoft - it's the data which comes from the middleware, not sure why it grabs non datasets.
`filesystem.listdir`

with such parameters:
`/mnt/tank [] {order_by: Array(1), limit: 1000}`

I see that middleware sends `file.is_mountpoint` property, if we filter by it - we only see available datasets.

AlexKarpov98:
@undsoft - make sense!

src/app/services/filesystem.service.ts
```diff
@@ -40,11 +40,10 @@ export class FilesystemService {
         map((files) => {
           const children: ExplorerNodeData[] = [];
           files.forEach((file) => {
-            if (file.type === FileType.Symlink || !file.hasOwnProperty('name')) {
-              return;
-            }
-
-            if (!options.showHiddenFiles && file.name.startsWith('.')) {
+            if (
```
undsoft, line 43: Previous version was more readable to me.

undsoft (changes requested):
I get errors when I try to use it.

https://github.com/truenas/webui/assets/1481270/cfa89c43-ac9a-4845-acad-0643e81964f2

AlexKarpov98:
@undsoft - are you sure that you are checking on the correct branch with latest updates? (I tested it and it works fine)

PR #8273, opened 2023-06-06
NAS-121931 / 23.10 / Run Once Replication Task doesn't run automatically after task is created

AlexKarpov98:
Screenshot 2023-06-06 at 14 18 55


denysbutenko (approved):
`.toPromise()` is deprecated. Let's use `lastValueFrom()` instead.

Other than that, the changes look good.

PR #8278, opened 2023-06-07
NAS-122259 / 22.12.4 / Same system Serial number listed for both Controllers

AlexKarpov98:


undsoft (approved):
Looks good. Does this need to go into Cobia?

AlexKarpov98:
> Looks good. Does this need to go into Cobia?

@undsoft - it's already there.

PR #8282, opened 2023-06-08
NAS-121985 / 23.10 / Add Feedback Module

denysbutenko:
Making `extra` as required param. Fixes dialog width on narrow screens.

There is a known issue with CORS - it will be solved at https://github.com/truenas/feedback-service/pull/8. Until that, for testing, you'll need to set up `feedback-service` locally and expose it to some domain, using `ngrok` for example. Update `hostname` in `ix-feedback.service.ts` using the exposed domain, and test form submission in the dialog.

### DIFF
--- a/src/app/modules/ix-feedback/feedback-dialog/feedback-dialog.component.scss
+++ b/src/app/modules/ix-feedback/feedback-dialog/feedback-dialog.component.scss
@@ -2,6 +2,16 @@
   display: block;
   max-width: 100%;
   width: 500px;
+
+  ::ng-deep {
+    ix-textarea {
+      textarea {
+        max-height: 40vh;
+        min-height: 7vh;
+        min-width: 96.5%;
+      }
+    }
+  }
 }
 
 .close-feedback-dialog {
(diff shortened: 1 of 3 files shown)


AlexKarpov98 (changes requested):
On mobile we can make it look better without scrolling.
Screenshot 2023-06-08 at 11 21 40

And I would like to limit the max height of textarea and min-width of it, in order to avoid what you can see on the video:

https://github.com/truenas/webui/assets/22980553/152e59ab-23df-4b26-9a2c-0d85251ba636

denysbutenko:
@AlexKarpov98 
> On mobile we can make it look better without scrolling. Screenshot 2023-06-08 at 11 21 40
> 

It has been fixed in the second commit at this PR. The dialog should have a better look on mobile screens.

> And I would like to limit the max height of textarea and min-width of it, in order to avoid what you can see on the video:
> 
>  Screen.Recording.2023-06-08.at.11.20.44.mov

This is related to the `ix-textarea` component itself; I'm just using it here. If you have any details in mind, please share them. I'll create a ticket for those improvements.

AlexKarpov98:
@denysbutenko - I added fix for textarea in this PR.
https://github.com/truenas/webui/assets/22980553/b64e7f18-8ff1-4fbf-82d8-02fa8d2de03e

AlexKarpov98 (approved):
Looks good to me. ✅

PR #8285, opened 2023-06-08
NAS-122372 / 23.10 / Pool Creation Wizard - disable stepper if current form is invalid (fix)

AlexKarpov98:
If step is invalid - we only can click [Back] or fill form correctly to be able to navigate to other steps.


undsoft (changes requested):
Can we redo this by using store selectors instead? You could have `hasDataVdevs` and remove the need for looking at ids and having to have `hasDataVdevs` input.

AlexKarpov98:
@undsoft - but we only need to make these checks (`hasDataVdevs`) on DATA step.

undsoft:
Then you can use this selector within data step component?

AlexKarpov98:
@undsoft - that component is rendered from the beginning and rendered once. (all the components - they work the same way inside of mat-step)

undsoft (changes requested):
I see what you meant. Perhaps we could use lazy loading with data step.
Another issue is that we shouldn't block Back button on Data step – I may change my mind and want to go back to enable more exported disks.

AlexKarpov98:
@undsoft - it's currently blocked by @RehanY147  fix.
Screenshot 2023-06-13 at 17 12 48

AlexKarpov98:
@undsoft - now it is working, can be reviewed.

undsoft (changes requested):
It's possible to complete step 1 and then just skip all remaining steps by clicking one of the headers.
Also if I go to Data step and the remember that I want to select another disk on first step, I have to click Back twice because previous headers are no available.

I'm open to other suggestions on how to deal with this. Perhaps we could show a warning icon in step header if it's invalid and allow users to freely go between steps.

AlexKarpov98:
@undsoft - I think it's better to allow any step to be opened, but once it was opened and not filled - we can show warning next to the header.
and somehow we also need to keep track of the required steps to disable final submit button.

undsoft:
@AlexKarpov98 Sounds good. Submit button can be disabled if Name or Data vdevs are not added. This could be a store selector.

AlexKarpov98:
gonna open new PR in this ticket:
https://ixsystems.atlassian.net/browse/NAS-122477

PR #8289, opened 2023-06-09
NAS-122385 / 23.10 / Process values ​​in kilobytes

bvasilenko:
**Testing**

At FTP bandwidth limits,

for example, the **Anonymous User Download Bandwidth** field in the WebUI indicates 5 KiB, then the value for `anonuserdlbw` in `ftp.config` should be 5.

- **Expected result:** "5"


AlexKarpov98 (approved):
By default input stores data in bytes, converting KiB to B.
But with this change, this component stores KiB directly - with no default Bytes transformation.
How it works?
Upon save - it converts bytes to kilobytes to be correctly received by the middleware.
Upon loading config - it converts saved kilobytes to bytes to be correctly shown on the UI side.

Works good ✅ .

PR #8290, opened 2023-06-09
NAS-122373 / 23.10 / Create vdev inspector in new pool creation

undsoft:
Testing: see Inspect Vdevs button on Review step in new pool creation.


RehanY147 (changes requested):
Works Ok. But I think the types column on the left can be made to look a bit more clickable. It's not immediately obvious that those items are selectable.

undsoft:
Added some hover effects.

AlexKarpov98 (approved):
Works as expected ✅

PR #8293, opened 2023-06-11
NAS-121381 / 23.10 / Fixing 2FA according to middleware changes

RehanY147:


undsoft (changes requested):
Couldn't test. Both on m40 and on a machine from Jenkins, nothing happens after I click `Log In` button.

RehanY147:
That issue was being caused by a wrong older filter check that I've updated.

undsoft (changes requested):
1. I've enabled 2FA globally and logged out without setting up 2FA for my root user.
When I login, I get "Password is incorrect" and I get prompted for 2FA code (which is incorrect) and then no matter what code I enter, I can still login.
2. I don't think we need to show all that system-wide info in user 2FA form. We just need to explain current status for both system and user.
I.e. "Two-Factor authentication is required on this system, but it's not yet configured for your user. Please configure it now."
or "✅ Two-Factor authentication has been configured"
or "Two-Factor authentication is not enabled on this this system. You can configure your personal settings, but they will have no effect until two-factor authentication is enabled globally by system administrator".
4. I have Save Debug button if I go from Advanced settings to user specific form.
5. Once I've configured 2FA for my user and try to login, I get "Email or password is incorrect" when they are in fact correct. Also password gets cleared up.
6. When I press "Renew", there is no indication of progress or success.
7. If user doesn't scan QR code after setting or renewing a secret, they can lock out themselves from the system. We should show them QR code along with the message explaining the importance of setting it up immediately.
8. If I login with 2FA enabled globally, but my user doesn't have it configured, I should be taken to user form with a message explaining that I'm required to set it up. We can even forbid any other operations before this is done. See if we can use `FullScreenDialogComponent` to render the form.

undsoft (changes requested):
Most things look pretty good.
Some things to fix:
1. Let's make "Two-Factor authentication has been configured. No further actions are required." green (or white with green icon).
2. After I login with the user without 2FA, it seemed that first page after login didn't prompt me to setup 2FA. Only after I tried using menu, I got the warning screen. Let's make sure we take user to 2FA immediately after login in this case.
3. Text is not aligned on this warning screen.
4. If I setup 2FA for my users, and then change settings globally (change otp digits for example), I then cannot login with old users. a) This seems like a middleware bug. b) We should show a warning when global settings are changed that may result in secrets being reset.

RehanY147:
@undsoft I can't reproduce the issue where the first time warning isn't shown. Also, I think the warning text looks OK. I've created a ticket for middleware for the described bug. Added the warning on UI though.

undsoft (changes requested):
1. Warning about user secrets being reset is not necessary when I'm enabling 2FA for the first time.
2. I've enabled 2FA globally without setting it up for my user first. This was on TrueNAS-SCALE-23.10-MASTER-20230616-080848.
3. If I try to login with one user, get OTP field, but then change my mind and want to use another user, this other user may not have 2FA configured, yet the OTP field will still be there.
4. Can we show 'Two-Factor authentication has been configured. No further actions are required.' only after initial QR dialog has been closed?
5. I have issues logging in once 2FA was setup for my user. Will PM you.

undsoft (approved):
Looks good. 
One minor thing that I've noticed is that pressing Enter no longer works correctly in login form.
Previously you could press Enter after entering password to login. It does not work now. Pressing Enter when entering OTP token will instead incorrectly submit original password form.

RehanY147:
@undsoft I've created https://ixsystems.atlassian.net/browse/NAS-122675 to address the bug

PR #8296, opened 2023-06-12
NAS-122402 / 23.10 / Make sure it is possible to edit apps

AlexKarpov98:
Testing: 
Check the new app update process.

Ticket description:
Screenshot 2023-06-13 at 00 40 47

Result:
Screenshot 2023-06-13 at 00 40 00


src/app/pages/apps/components/chart-wizard/chart-wizard.component.ts
```diff
@@ -161,41 +160,43 @@ export class ChartWizardComponent implements OnInit, AfterViewInit, OnDestroy {
         this.cdr.markForCheck();
 
         if (this.activatedRoute.routeConfig.path.includes('install')) {
-          this.makeChartCreate();
+          this.loadApplicationForCreation();
```
RehanY147, line 163: I think checking the route path for keywords could be flaky. Since the route contains app name, an app name can potentially have "install" or "edit" and break the UI. A better approach would be to have a shared component "form" and separate components for install and edit that use the "form" component and set it up to their need with input properties. Then set the routes components correctly based on what we want.

RehanY147 (changes requested):

src/app/pages/apps/components/chart-wizard/chart-wizard.component.ts
AlexKarpov98: @RehanY147 
I think we can update that `includes` part with better/precise check
like here:
Screenshot 2023-06-13 at 12 39 53

AlexKarpov98: In this case it will always work since the route can ends only with these strings.

undsoft (approved):
Looks good to me.

PR #8301, opened 2023-06-14
NAS-122465: Cleaner value changes for PoolManagerWizard controls

RehanY147:
This fixes the bug of infinite value changes loop in the wizard
Known issue: The buttons stay disabled on Spare vdev wizard step. This issue is because of a large design issue that will be solved by https://ixsystems.atlassian.net/browse/NAS-122316


AlexKarpov98 (approved):
Bug is fixed now 👍

PR #8304, opened 2023-06-14
NAS-122282 / 23.10 / "Failed to export RRD data: ERROR" appear when navigate to the furthest position/time, left, in Reporting graphs

AlexKarpov98:
Testing: 
Go to reports page and try to rewind to 2000 year.
You should not be able to rewind further - this fixes the bug.
http://localhost:4200/reportsdashboard/network

Screenshot 2023-06-14 at 13 00 18

Screenshot 2023-06-14 at 13 01 55

### DIFF
--- a/src/app/pages/reports-dashboard/components/report/report.component.html
+++ b/src/app/pages/reports-dashboard/components/report/report.component.html
@@ -24,6 +24,7 @@
           <div *ngIf="localControls && !report?.errorConf" class="line-chart-tools">
             <mat-button-toggle-group aria-label="Steps">
               <mat-button-toggle
+                [disabled]="stepBackDisabled"
                 [matTooltip]="'Step Back' | translate"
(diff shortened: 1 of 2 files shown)


denysbutenko (approved):
I wonder if static `1999` could be replaced with more reliable data, e.g. system build time.
Otherwise, it looks good to me.

AlexKarpov98:
> I wonder if static `1999` could be replaced with more reliable data, e.g. system build time.
> 
> Otherwise, it looks good to me.

Any ideas of how to do it?

denysbutenko:
> > I wonder if static `1999` could be replaced with more reliable data, e.g. system build time.
> > Otherwise, it looks good to me.
> 
> Any ideas of how to do it?

I'd ask somebody from middleware if there is a possibility of extracting this number from existing system information like system build time or birth date. Anyway, it seems to require extra effort and is out of the scope of this bug fix.

PR #8309, opened 2023-06-15
NAS-122353 / 23.10 / Improve loading state for available apps header

denysbutenko:
image

For testing, check available apps header


undsoft (changes requested):
The problem with new solution is that it shows loading animation for numbers every time you change something in filters.

denysbutenko:
> The problem with new solution is that it shows loading animation for numbers every time you change something in filters.

True. With recent changes, there is no needed loader for those numbers - initial values are present from the beginning. I'll code clean up at least.

denysbutenko:
As discussed, I added an internal loader that works the first time only. The loading state is visible when a user is directly loading the available apps page in other cases numbers are ready, and the loading state won't appear.

undsoft (approved):
Looks good.

PR #8315, opened 2023-06-16
NAS-122508 / 23.10 / Updating ng-charts

undsoft:
Test affected charts:
* Charts on dashboard
* Gauge chart on Storage page
* Chart on Datasets page


RehanY147 (approved):
Looks good. All charts work like before.

PR #8316, opened 2023-06-17
NAS-122448 / 23.10 / Redirecting after setting up pool

RehanY147:


dszidi (commented):
So far, testing this on the team's M40 shows that user can select a pool in the dialog and get to the app install form.
However, the pool I set in the dialog doesn’t seem to stick.

Additionally, if I try to set it from Apps>Settings (menu btn)>Choose Pool
my selection won’t stick from there either.
No matter how I try to set the apps pool, I always see the “Setup pool to install” button label text on the App page.

image


To summarize, while the smaller workflow issue of getting the user to the install form is resolved, the “infinite loop” problem is still there in that you can’t ever get to a point where you don’t need to set the pool.

dszidi (approved):
Pool settings are now being retained. LGTM

PR #8318, opened 2023-06-18
NAS-122477 / 23.10 / Warnings in pool creation wizard

AlexKarpov98:
Testing:
Go to new pool creation and check validation errors:
[now we store all errors in store as array]
like:
```
[
              {
                text: this.translate.instant('Name not added'),
                severity: PoolCreationSeverity.Error,
                step: PoolCreationWizardStep.General,
              }
]
```

Screenshot 2023-06-24 at 18 14 12
Screenshot 2023-06-24 at 18 15 01


src/app/pages/storage/modules/pool-manager/components/pool-manager-wizard/pool-manager-wizard.component.ts
```diff
@@ -35,19 +37,21 @@ export class PoolManagerWizardComponent implements OnInit {
...
-  hasEnclosureStep$ = combineLatest([
-    this.store.hasMultipleEnclosuresAfterFirstStep$,
-    this.systemStore$.pipe(waitForSystemFeatures, map((features) => features.enclosure)),
-  ]).pipe(
-    map(([hasMultipleEnclosures, hasEnclosureSupport]) => hasMultipleEnclosures && hasEnclosureSupport),
-  );
+  activeStep: PoolCreationWizardStep;
+  hasEnclosureStep = false;
+
+  wizardRequiredStepsValidityForm = this.fb.group({
```
undsoft, line 46: I don't think we should use form for this. It's more like a Map really.

undsoft (commented):
Approach works so far, but I don't know if it's going to be scalable once more rules as added. For example, such orange warnings as using stripe vdevs. It may be better to just look at store and return an error/warning map grouped by step.

Let's also make Warnings a separate row in review step.

dszidi:
> Approach works so far, but I don't know if it's going to be scalable once more rules as added. For example, such orange warnings as using stripe vdevs. It may be better to just look at store and return an error/warning map grouped by step...

@undsoft  If the potential for validation errors is going to be this numerous, wouldn't it be better to force the user to deal with them in the step in which they occur? Seems unhelpful to let the user to proceed through numerous steps and hoard validation errors as they go. This way the final step would only need to have warnings

AlexKarpov98:
@dszidi - not sure if in real life someone would make as much errors as I did for this demo image :D

undsoft:
@dszidi The idea is to have warnings in both places – first when they occur and then on review step.

src/app/pages/storage/modules/pool-manager/store/pool-manager.store.ts
```diff
@@ -117,6 +137,143 @@ export class PoolManagerStore extends ComponentStore<PoolManagerState> {
     }),
   );
 
+  disknumErrorMessage = helptext.manager_disknumErrorMessage;
+  disknumErrorConfirmMessage = helptext.manager_disknumErrorConfirmMessage;
+  disknumExtendConfirmMessage = helptext.manager_disknumExtendConfirmMessage;
+  vdevtypeErrorMessage = helptext.manager_vdevtypeErrorMessage;
+  exportedPoolsWarning = helptext.manager_exportedDisksWarning;
+
+  readonly poolCreationErrors$ = combineLatest([
```
undsoft, line 146: I think it would be easier to test if warnings related to logic is extracted into a separate service that uses PoolManagerStore.

src/app/pages/storage/modules/pool-manager/store/pool-manager.store.ts
```diff
@@ -91,13 +104,20 @@ export const initialState: PoolManagerState = {
   },
 
   topology: initialTopology,
+
+  wizardRequiredStepsValidity: {
```
undsoft, line 108: I don't think we need to take data from components to be able to validate this. Name and settings used for data vdevs are already in state.

undsoft (changes requested):
> A stripe data vdev is highly discouraged and will result in data loss if it fails

This is not a type of error that should prevent wizard from being submitted.

> The following disks have exported pools on them. Using those disks will make existing pools on them unable to be imported. You will lose any and all data in selected disks.

Doesn't list disks. We should list disks or make a generic warning.

src/app/pages/storage/modules/pool-manager/store/pool-manager.store.ts
```diff
@@ -117,6 +137,143 @@ export class PoolManagerStore extends ComponentStore<PoolManagerState> {
...
+
+        nonEmptyTopologyCategories.forEach(([typologyCategoryType, typologyCategory], i) => {
+          let dataVdevDisknum = 0;
+          let dataVdevType: string;
+          let firstDataVdevType: CreateVdevLayout;
+          let firstDataVdevDisknum = 0;
+
+          if (typologyCategoryType === VdevType.Data) {
+            if (i === 0) {
+              firstDataVdevType = typologyCategory.layout;
```
undsoft, line 191: I think warnings related to editing a pool, such as `First vdev has {n} disks, new vdev has {m}` and `First vdev is a {vdevType}, new vdev is {newVdevType}` can be removed for now. We should prevent user from doing something like when editing.

src/app/pages/storage/modules/pool-manager/store/pool-manager.store.ts
```diff
@@ -117,6 +137,143 @@ export class PoolManagerStore extends ComponentStore<PoolManagerState> {
...
+          }
+
+          if (
+            [VdevType.Dedup, VdevType.Log, VdevType.Special, VdevType.Data].includes(typologyCategoryType)
+            && typologyCategory.vdevs.length >= 1 && typologyCategory.layout === CreateVdevLayout.Stripe
+          ) {
+            if (typologyCategoryType === VdevType.Log) {
+              errors.push({
+                text: this.translate.instant('A stripe log vdev may result in data loss if it fails combined with a power outage.'),
+                severity: PoolCreationSeverity.Error,
```
undsoft, line 233: That was likely a warning before.

src/app/pages/storage/modules/pool-manager/store/pool-manager.store.ts
```diff
@@ -117,6 +137,143 @@ export class PoolManagerStore extends ComponentStore<PoolManagerState> {
     }),
   );
 
+  disknumErrorMessage = helptext.manager_disknumErrorMessage;
+  disknumErrorConfirmMessage = helptext.manager_disknumErrorConfirmMessage;
+  disknumExtendConfirmMessage = helptext.manager_disknumExtendConfirmMessage;
```
undsoft, line 142: This one and above are unused and can be removed.

PR #8319, opened 2023-06-19
NAS-122522 / 23.10 / Fixing infinite loops on inventory changes

RehanY147:


AlexKarpov98 (approved):
Nice, I could not find any infinite loop problem now, good job! 👍

PR #8320, opened 2023-06-19
NAS-122531 / 23.10 / Updates packages

undsoft:
Updates packages that are preventing us from upgrading to Angular 16.

Testing: regression testing.


package.json
```diff
@@ -131,7 +131,7 @@
     "ngx-popperjs": "^15.0.4",
     "ngx-skeleton-loader": "~5.0.0",
     "ngx-translate-messageformat-compiler": "~5.0.1",
-    "ngx-uploader": "~11.0.0",
+    "@angular-ex/uploader": "~15.0.0",
```
AlexKarpov98, line 134: I don't see it's used somewhere
Screenshot 2023-06-22 at 12 11 11

AlexKarpov98 (approved):
Looks good, no bugs found.

package.json
undsoft: You are right.

PR #8323, opened 2023-06-21
NAS-122533 / 23.10 / Fix app sources on the new apps

denysbutenko:
- Render commas using CSS approach
- `getCleanLink` utility function converted to `cleanLink` pipe.
- `Homepage` rendering as a link and uses `cleanLink` pipe.

For testing, check the row with sources on the `AppInfo` and `AppAvailableInfo` cards.

Preview
image
image
image


dszidi (approved):
LGTM

PR #8326, opened 2023-06-22
Allow uploading attachments in the Feedback Dialog

denysbutenko:
- Allow uploading images in the feedback dialog.
- Remove the bearer token
- The page param uses `path` instead of full URL.

For testing, try to upload an image to your review.


AlexKarpov98 (changes requested):
Looks good, but I was not able to submit form successfully.

Screenshot 2023-06-22 at 12 19 56
image

Suggestions:
Maybe we can add small attachment preview?
and add drag & drop feature (currently dragging image to Choose File button - open image in new tab

denysbutenko:
@AlexKarpov98 Certificate issue that prevents the form from being submitted should be resolved now.

AlexKarpov98 (approved):
It worked now.
Screenshot 2023-06-26 at 10 52 04

Is it possible to upload only 1 image, is there API limit?

denysbutenko:
> Is it possible to upload only 1 image, is there API limit?

@AlexKarpov98 yes

PR #8327, opened 2023-06-22
NAS-122537 / 23.10 / Use html2canvas for taking screenshots in Feedback Dialog

denysbutenko:
For testing, try to submit feedback and check the attachment.


src/app/modules/ix-feedback/feedback-dialog/feedback-dialog.component.ts
```diff
@@ -136,4 +132,22 @@ export class FeedbackDialogComponent implements OnInit {
     this.dialogRef.close();
     this.cdr.markForCheck();
   }
+
+  private takeScreenshot(): Promise<File> {
```
undsoft, line 136: Worth extracting into a service.

undsoft (approved):
Tested via mocks, looks good, some image is attached.

1. When "Take screenshot" is unselected, let's change label on file input to "Attach image (optional)".
2. Loading indicator doesn't seem to stop on error.

PR #8342, opened 2023-06-26
NAS-122530 / 23.10 / Use system host id in Feedback Dialog

denysbutenko:
CI passed and code review should be enough for those changes.


undsoft (approved):
Looks good, but I don't know if we need to have host id in global store. Seems like we're going to use it in one component, so we can keep it contained there.

PR #8346, opened 2023-06-26
NAS-122554: API changed in `network.configuration.config`

AlexKarpov98:
Testing: 
Network —> Global Configuration —> Settings 

**Host Name Database** field

Try to setup some hosts items 
Now they shall be saved as an array 
and don’t break the UI 
Use the latest nightly for test because on m40 for example it will break


dszidi (approved):
Code looks good.
Entries seem to be saved just fine LGTM


image



image

PR #8350, opened 2023-06-27
NAS-122609 / 23.10 / NAS_122609: Two dialogs show up on top of each other when clicking "Setup pool to install" without any pools

AlexKarpov98:
Testing:
Screenshot 2023-06-27 at 14 02 04

Now only one modal shall be available

### DIFF
--- a/src/app/pages/apps-old/select-pool-dialog/select-pool-dialog.component.spec.ts
+++ b/src/app/pages/apps-old/select-pool-dialog/select-pool-dialog.component.spec.ts
@@ -115,12 +115,14 @@ describe('SelectPoolDialogComponent', () => {
     jest.spyOn(appService, 'getPoolList').mockReturnValue(of([]));
     spectator.component.ngOnInit();
 
-    expect(spectator.inject(DialogService).confirm).toHaveBeenCalledWith({
-      title: helptext.noPool.title,
-      message: helptext.noPool.message,
-      hideCheckbox: true,
-      buttonText: helptext.noPool.action,
-    });
-    expect(spectator.inject(Router).navigate).toHaveBeenCalledWith(['/storage', 'create']);
+    setTimeout(() => {
+      expect(spectator.inject(DialogService).confirm).toHaveBeenCalledWith({
+        title: helptext.noPool.title,
+        message: helptext.noPool.message,
+        hideCheckbox: true,
+        buttonText: helptext.noPool.action,
+      });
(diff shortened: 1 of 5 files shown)


RehanY147 (changes requested):
I can still see one dialog on top of each other.

AlexKarpov98:
@RehanY147 - only setTimeout helps (needed because without it redirect is not working)

RehanY147 (changes requested):
image

Only moving this method call to the start of the method should be enough

RehanY147 (changes requested):
Dialog disappears too soon when there's at least one pool created.

src/app/pages/apps-old/select-pool-dialog/select-pool-dialog.component.spec.ts
```diff
@@ -115,12 +115,14 @@ describe('SelectPoolDialogComponent', () => {
...
     spectator.component.ngOnInit();
 
-    expect(spectator.inject(DialogService).confirm).toHaveBeenCalledWith({
-      title: helptext.noPool.title,
-      message: helptext.noPool.message,
-      hideCheckbox: true,
-      buttonText: helptext.noPool.action,
-    });
-    expect(spectator.inject(Router).navigate).toHaveBeenCalledWith(['/storage', 'create']);
+    setTimeout(() => {
```
RehanY147, line 118: Tests shouldn't use timeout like that.

src/app/pages/apps-old/select-pool-dialog/select-pool-dialog.component.ts
```diff
@@ -98,8 +98,10 @@ export class SelectPoolDialogComponent implements OnInit {
           }));
           this.pools$ = of(poolOptions);
 
+          this.dialogService.closeAllDialogs();
+
           if (!pools.length) {
-            this.showNoPoolsWarning();
+            setTimeout(() => this.showNoPoolsWarning(), 100);
```
RehanY147, line 104: This settimeout is also not needed

RehanY147 (changes requested):
Now the dialog closes even if you have some pools to select from.

src/app/pages/apps-old/select-pool-dialog/select-pool-dialog.component.ts
AlexKarpov98: @RehanY147 - there is problem with subscription here, because if we `closeAllDialogs` - we loose ref, I think so.

RehanY147 (approved):

PR #8355, opened 2023-06-28
NAS-122638 / 23.10 / Refactoring Catalog table to ix-table2

denysbutenko:
Some components were moved to the `ix-table2` folder to use everything from the `ix-table2.module`

For testing, check the [Catalogs](http://localhost:4200/apps/catalogs) page and compare it with the old table.


undsoft (approved):
Looks good for now.

PR #8369, opened 2023-06-30
Update pools-dashboard.component.html

MDVAurelien:
Add T marker to obtain matche in translation files


MDVAurelien:
> Thanks, but this is not necessary. This is there only temporarily for betas. Once we are happy with new pool creation wizard, there will only be one `Create Pool` button.

Ah ok sorry, I thought it was strange! ;-)

PR #8370, opened 2023-07-02
NAS-121778 / 23.10 / Fixing installed apps updates

RehanY147:
Check that after installing an app, the installed apps table is shown and the app recently installed is shown on the table. Also check that the app has the installed badge on the apps dashboard.

Check the opposite when deleting an app. The table should update and the badge should be removed in apps dashboard.


denysbutenko (approved):
It works great

PR #8373, opened 2023-07-03
NAS-122667: [Scale/Apps]: List of Lists of Strings produce an error

AlexKarpov98:
Testing: 
Go to new apps, find Sonar app.
Screenshot 2023-07-03 at 19 12 35

Try installing it:
Screenshot 2023-07-03 at 19 14 21

Set list of list of strings works (previously it didn't work)

### DIFF
--- a/src/app/pages/apps-old/chart-releases/chart-releases.component.ts
+++ b/src/app/pages/apps-old/chart-releases/chart-releases.component.ts
@@ -185,10 +185,10 @@ export class ChartReleasesComponent implements AfterViewInit, OnInit {
     this.ws.subscribe('chart.release.query').pipe(
       untilDestroyed(this),
     ).subscribe((evt) => {
-      const app = this.chartItems.get(evt.fields.name);
+      const app = this.chartItems.get(evt.fields?.name);
 
       if (app && evt?.fields) {
-        this.chartItems.set(evt.fields.name, { ...app, ...evt.fields });
+        this.chartItems.set(evt.fields?.name, { ...app, ...evt.fields });
       }
       this.filterChartItems();
     });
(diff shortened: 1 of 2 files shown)


src/app/pages/apps-old/chart-releases/chart-releases.component.ts
```diff
@@ -185,10 +185,10 @@ export class ChartReleasesComponent implements AfterViewInit, OnInit {
     this.ws.subscribe('chart.release.query').pipe(
       untilDestroyed(this),
     ).subscribe((evt) => {
-      const app = this.chartItems.get(evt.fields.name);
+      const app = this.chartItems.get(evt.fields?.name);
 
       if (app && evt?.fields) {
-        this.chartItems.set(evt.fields.name, { ...app, ...evt.fields });
+        this.chartItems.set(evt.fields?.name, { ...app, ...evt.fields });
```
bvasilenko, line 191: I've noticed that currently, we have a chance that `set` will be called with `undefined` argument

Shouldn't we skip those which have empty `fields` object?

bvasilenko (changes requested):
Overall looks good, just noticed potential issue, which might need additional fixing

src/app/pages/apps-old/chart-releases/chart-releases.component.ts
AlexKarpov98: Sure, lets skip undefined value

bvasilenko (approved):
Tested on TrueCharts catalog (https://truecharts.org/manual/SCALE/guides/getting-started/)

**Traefik** app proceeds to installation without errors in console

PR #8375, opened 2023-07-03
Extract strings from data-protection.routing for translations

MDVAurelien:
Extract with yarn & yarn extract ;-)


undsoft (approved):
Looks good.

PR #8377, opened 2023-07-04
NAS-121705 / 23.10 / Loading installed app updates automatically

RehanY147:
Check that the installed apps table on the main apps page updates automatically whether it be newly installed apps, apps status updates or apps being removed or edited. Also, check the code to make sure that we are not making periodic calls to make these updates and instead only relying on the `chart.release.query` subscription for updates.


AlexKarpov98 (changes requested):
There is no loading indicator now and on first load I see message that there are no apps.

https://github.com/truenas/webui/assets/22980553/6616c68d-e39f-4a61-9cf2-1915cb75c25e

AlexKarpov98:
@RehanY147 -
During deletion status never changes (always: `Running`)
https://github.com/truenas/webui/assets/22980553/48f94784-224c-4395-9128-4b8fe1485c9a

AlexKarpov98:
Other than that - statuses are updated correctly.

RehanY147:
@AlexKarpov98 I've talked to Evgeny and the status not updating while deleting is a non-issue. I've fixed the loading. Please review again.

AlexKarpov98 (approved):
Works good now.

PR #8378, opened 2023-07-04
NAS-122550 / 23.10 / Deep cloning object before deleting attribute

RehanY147:
Testing instructions are in the ticket description


bvasilenko (approved):
Works good to me!

PR #8382, opened 2023-07-04
NAS-122757 / 23.10 / Maximum number of concurrent calls (20) has exceeded

undsoft:
I was able to reproduce original issue by logging in, closing laptop and waiting for ~10 minutes.
You may be able to put system to sleep or do something similar.


denysbutenko (approved):
LGTM

PR #8383, opened 2023-07-04
NAS-122755 / 23.10 / High token settings values

undsoft:
Testing: see ticket.


AlexKarpov98 (approved):
Works good.
Screenshot 2023-07-05 at 15 22 50

PR #8385, opened 2023-07-05
NAS-122534: Update breadcrumbs

denysbutenko:
For testing, go through pages and check breadcrumbs


bvasilenko (approved):
Works good to me!

PR #8399, opened 2023-07-06
NAS-122781 / UI does not present "Remove" options for L2ARC or SLOG when pool contains RAIDZ data vdevs

chrisperedun:
Added logic in canRemoveDisk and canRemoveVDEV to allow removal of VdevType.Cache and .Log when hasTopLevelRaidz is TRUE - these vdev types are still removable even when a top-level RAIDZ vdev exists so the REMOVE button should be presented for the appropriate disk/vdevs


undsoft (approved):
Looks good, thanks.

PR #8408, opened 2023-07-09
NAS-122641 / 23.10 / Adding vdevs to existing pool

RehanY147:


undsoft (changes requested):
* Design calls for Current Configuration and New Devices cards.
* When trying to add more data vdevs, I got an error saying that " This pool already has some DISK Data Vdevs. You can only add vdevs of DISK type.". This doesn't look right, I don't know what DISK type vdevs are.
* Ideally, if I can only add vdevs of certain layout during a step, I should not even be allowed to select other layouts.
* There is some logic that prevents Next/Save button on Data vdevs to be active when nothing is selected. This needs to be removed, as I may be adding vdevs of a different type. I think we could conceivably remove it from creation wizard as well, since there are errors in place.

src/app/interfaces/storage.interface.ts
```diff
@@ -21,6 +21,7 @@ export interface VDev {
   stats: TopologyItemStats;
   status: TopologyItemStatus;
   unavail_disk: unknown;
+  disk?: string;
```
undsoft, line 24: Does this field come from middleware? If we are adding this field on frontend, it's better to extend original interface and add new field there.

undsoft (changes requested):
1. `Current Configuration Preview` can be renamed to just `Current Configuration`.
2. I've got Next button disabled on the first step for some reason. 
Снимок экрана 2023-07-12 в 13 13 42

3. Data step says "A stripe vdev is highly discouraged..." yet there is no mention about Stripe. We could remove this warning when adding vdevs, or maybe show disabled layout dropdown.
4. On my system Current configuration says "Data: No Redundancy". It's better if it tried to describe vdevs or at least show "Custom Layout" or something like this.
5. Current Configuration says "Spare: VDEVs not assigned", while New Devices says "Spare: None". Let's standartize to just "None" in both cases.
6. I was no able to select Width on Log step for some reason, even though I had available disks.

src/app/interfaces/storage.interface.ts
RehanY147: This comes from middleware. For stripe vdevs, each vdev is only assigned one disk even if you add multiple disks in one stripe vdev, a new vdev is created for each disk. And that vdev has an empty `children` array and has `disk` as disk name.

undsoft (approved):
Next button on the first step is still disabled for me for some reason.
Other than that looks good.

PR #8411, opened 2023-07-10
NAS-122654: Use new endpoints for Reports

denysbutenko:
For testing, go through reports and make sure that every report is rendered correctly

### DIFF
--- a/src/app/interfaces/api-directory.interface.ts
+++ b/src/app/interfaces/api-directory.interface.ts
@@ -767,6 +767,8 @@ export type ApiDirectory = {
   'reporting.config': { params: void; response: ReportingConfig };
   'reporting.graphs': { params: QueryParams<ReportingGraph>; response: ReportingGraph[] };
   'reporting.clear': { params: void; response: void };
+  'reporting.netdata_get_data': { params: ReportingQueryParams; response: ReportingData[] };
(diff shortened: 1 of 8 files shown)


undsoft (changes requested):
1. CPU Usage has weird scale on the left now. It says 500m, 1000m, 1500m. I don't know what those are.
2. It also looks differently compared to old chart. Idle used to have background and when I compare two charts, they seem to show different values. I understand that it's different reporting API, but I would imagine that CPU Usage should be roughly the same between them.
3. Multiple disk charts look different. Hard to say if this is because of backend or not.
4. Memory usage says "KiloMebibytes/s" on the axis, legend on right uses KiB instead of GiB and GiB is missing from `buffers`.
5. Incorrect units on Network.
6. NFS doesn't show anything for me. If this chart is not available, should menu item be removed?
7. Same for Partition.
8. System uptime is in seconds now instead of days and `Processes` chart is absent.
9. Arc: incorrect units, missing charts, console errors ("Cannot read properties of undefined (reading 'replace' at LineChartComponent.renderGraph (line-chart.component.ts:181:20))

undsoft (approved):
Appears stable enough to merge. Let's merge and then correct if needed.

PR #8414, opened 2023-07-10
NAS-122759 / 23.10 / Apps Container Card Icon Links to Shell Doesn't Open the Shell for the Pod and Logs Doesn't Open Logs

AlexKarpov98:
Testing:
Check Pod Logs & Pod Shell from installed app.
You should see some results, not infinite loaders or so.
Notice - wait for app to be in ACTIVE status, deploying or stopped - hides the buttons.


denysbutenko (approved):
Looks good 👍🏻

PR #8417, opened 2023-07-11
NAS-122870 / 23.10 / List of strings is converted to list of objects when the field key has same name as a key in the same level of the parent.

AlexKarpov98:
Testing:
Check test case -> that's how it should work, try it in real life.

Problem was that we took incorrect schemaNode as schema for building result, like if we have 2 items keys, then it took first one as reference even tho we need to take second one.

Now we are looking into the schema node path to be 100% sure that we take correct schema node as a referenct.

You can use [m40](http://m40g3-137.dc1.ixsystems.net/ui/dashboard) device and **radarr** app:

http://localhost:4200/apps/available/TRUENAS/community/radarr/install

these form elements shall be tested:
Screenshot 2023-07-11 at 15 30 32


bvasilenko (approved):
Worked good for me, tested with radarr

PR #8419, opened 2023-07-11
NAS-122896 / 23.10 / Add table toggle column component

bvasilenko:
**Testing**

Code review should be enough


src/app/modules/ix-table2/components/ix-table-body/ix-table-body.component.spec.ts
```diff
@@ -1,22 +1,36 @@
...
+  yesNoField: boolean;
+  enabledField: boolean;
 }
 
 const testTableData: TestTableData[] = [
-  { numberField: 1, stringField: 'a', booleanField: true },
-  { numberField: 2, stringField: 'c', booleanField: false },
-  { numberField: 4, stringField: 'b', booleanField: false },
-  { numberField: 3, stringField: 'd', booleanField: true },
+  {
```
undsoft, line 22: While having this kind of integration test is good, I think it would be better to (also) have more isolated tests in column specific folders.

dszidi (approved):
Looks good

undsoft (approved):
Thanks.

PR #8421, opened 2023-07-11
NAS-122595 / UI does not allow REMOVE operations on single-device data vdevs

chrisperedun:
Updated UI logic, see line references below

421: show REMOVE for single-disk vdevs in mirror pools
423+426: hide ONLINE, OFFLINE, and REPLACE for spare vdev (fixed case sensitive matching)
430: hide REPLACE for cache vdev
563: hide EXTEND for cache and spare vdev


AlexKarpov98 (approved):
Looks good.

kmoore134:
@undsoft Since this is going into 13.0 STABLE branch for 13.1 eventually, can I also have your quick review? 

Thanks!

undsoft (approved):
Looks good to me too.

PR #8422, opened 2023-07-12
NAS-122872 / 23.10 / Add iSCSI ALUA setting to SCALE HA WebUI

AlexKarpov98:
Testing: check ticket description.


undsoft (approved):
Looks good, but please check with Brian if we can write something more descriptive than just "ALUA". Ideally we'd have a more descriptive label and a tooltip.

AlexKarpov98:
updated:
Screenshot 2023-07-13 at 11 50 03

PR #8424, opened 2023-07-12
NAS-122854 / 23.10 / Reduce cognitive complexity to 40

denysbutenko:
Check `app-schema.service.ts` changes


src/app/services/schema/app-schema.service.ts
```diff
@@ -637,65 +639,74 @@ export class AppSchemaService {
...
-            path,
-            chartSchemaNode: subquestion,
-            formGroup,
-            config,
-            isParentImmutable: !!schema.immutable || isParentImmutable,
-          }),
-        );
+    for (const subquestion of schema.subquestions) {
+      subscription.add(
+        this.addFormControls({
```
RehanY147, line 657: This method name is not very intuitive. It should be named something that indicates what it returns. It's not just adding a form control. It's also returning a subscription to form control's value changes. Suggesstion: `getNewFormControlChangesSubscription` or maybe it can be divided between two methods. One to create a form control and one to get the subscription. Or, we can pass the subscription collection to the method so adding the subscription can be done inside the method. Something should be done.

src/app/services/schema/app-schema.service.ts
```diff
@@ -637,65 +639,74 @@ export class AppSchemaService {
...
-          formField.hidden$ = new BehaviorSubject<boolean>(false);
-        }
-        if (newFormControl.value === schema.show_subquestions_if) {
-          formField.hidden$.next(false);
-          formField.enable();
-        } else {
-          formField.hidden$.next(true);
-          formField.disable();
+      const formField = formGroup.controls[subquestion.variable] as CustomUntypedFormField;
+      this.initializeFormField(formField, newFormControl.value, schema, subquestion, isNew, isParentImmutable);
```
RehanY147, line 668: Same goes for this method name. The main purpose is to hide or unhide disable or enable the form control. Not initialize it. Maybe name it `toggleFieldHiddenOrDisabled` or `toggleFieldVisibility` or something similar.

src/app/services/schema/app-schema.service.ts
```diff
@@ -637,65 +639,74 @@ export class AppSchemaService {
...
+    }
+
+    subscription.add(newFormControl.valueChanges.subscribe((value) => {
```
RehanY147, line 671: In continuation of my earlier comment, maybe adding all the subscriptions can be a separate method call. This might not be possible depending on the context so take it with a grain of salt.

src/app/services/schema/app-schema.service.ts
```diff
@@ -637,65 +639,74 @@ export class AppSchemaService {
...
+          parentControl.hidden$ = new BehaviorSubject<boolean>(false);
         }
-      });
 
-      subscription.add(newFormControl.valueChanges.subscribe((value) => {
-        schema.subquestions.forEach((subquestion) => {
-          const parentControl = (formGroup.controls[subquestion.variable].parent as CustomUntypedFormField);
-          if (!parentControl.hidden$) {
-            parentControl.hidden$ = new BehaviorSubject<boolean>(false);
+        parentControl.hidden$.pipe(take(1)).subscribe((isParentHidden) => {
```
RehanY147, line 678: See if possible to have a variable `hidden` on the `parentControl` as well which stays update to date with the `hidden$` observable. That way we don't have to rely on a subscription here to make subsequent recursion calls.

RehanY147 (changes requested):

RehanY147 (approved):

PR #8426, opened 2023-07-12
NAS-122866 / 23.10 / Add `inherit_encryption` field

bvasilenko:
Please see ticket


denysbutenko (approved):

src/app/modules/ix-table2/directives/ix-body-cell.directive.ts
```diff
@@ -29,7 +29,7 @@ export class IxTableBodyCellDirective<T> implements AfterViewInit {
       this.column.type,
     );
 
-    componentRef.instance.setRow(this.row);
+    componentRef.instance.row = this.row;
```
undsoft, line 32: This isn't a correct fix and will break functionality in other places.

bvasilenko:
@undsoft the fix for the test is ready for review

PR #8432, opened 2023-07-13
NAS-122895 / 23.10 / Updates to pool wizard

AlexKarpov98:
Testing: see ticket.


dszidi (approved):
Code looks good. Strings in components and tests have been updated as per ticket instructions.

PR #8436, opened 2023-07-13
NAS-116080: Fix network dashboard widget issues

bvasilenko:
**Summary**

1. Updated the axis labels in the Network dashboard widget. Changed "M" into "Mb/s"

2. Fixed the inconsistency between the axis labels provided in megabits/s and the In/Out values of the widget provided in mebibytes/s

3. Made sure that there's no negative Y axis label ("-476.8M") 

**Testing**

Go to **Dashboard** and check axes of the charts on the **Network** widget

### DIFF
--- a/src/app/pages/dashboard/components/widget-network/widget-network.component.ts
+++ b/src/app/pages/dashboard/components/widget-network/widget-network.component.ts
@@ -99,7 +99,7 @@ export class WidgetNetworkComponent extends WidgetComponent implements OnInit, A
       },
       tooltip: {
         callbacks: {
-          label(tooltipItem) {
+          label: (tooltipItem) => {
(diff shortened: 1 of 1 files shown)


undsoft (commented):
I think it would be better to keep showing outgoing below axis, but if possible to have labels that don't have negative signs.
There was also a feedback item that we are picking random values for axis. I.e. we should show have ticks at 5MiB/s, 10MiB/s and 15MiB/s instead of 4.232MiB/s, 9.521MiB/s and so on.

undsoft (changes requested):
👆🏻

bvasilenko:
@undsoft ok, now the requested remark should be addressed

PR #8437, opened 2023-07-13
NAS-122959 / 23.10 / Create Pool (New) : Minimum disk size selected pool cannot be created

AlexKarpov98:
Testing: see ticket and you can use this device.


undsoft (changes requested):
Something is wrong code according to tests, because you have the same disks reused in multiple vdevs. This is incorrect.

src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/generate-vdevs-treat-as-minimum.spec.ts
```diff
@@ -50,17 +50,14 @@ describe('GenerateVdevsService - treat disk size as minimum', () => {
       expect(vdevs).toEqual({
         [VdevType.Data]: expectDisks([
           ['enclosure1-disk1', 'enclosure1-disk2'],
-          ['enclosure1-disk3', 'enclosure2-disk1'],
-          ['enclosure2-disk2', 'enclosure3-disk1'],
+          ['enclosure1-disk1', 'enclosure1-disk2'],
+          ['enclosure1-disk1', 'enclosure1-disk2'],
```
undsoft, line 54: This looks incorrect.

AlexKarpov98:
@undsoft 
Finally I found out that we don't check disk type while returning available disks count..
Please check it now.

undsoft (changes requested):
We can discuss this during team meeting, but I think previous approach was better.
There is nothing in the UI that tells user that "Treat disk size as minimum" takes into account disk types.
Instead as far as I can tell the problem is in the fact that `findSuitableDisks` takes disk type into account AND after this is addressed, there is an issue that the same disks are reused multiple types (see `removeUsedDisks`).

AlexKarpov98:
@undsoft  - you mean once we selected `Treat as minimum` - we should count any disk --> HDD or SSD?

src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/generate-vdevs-treat-as-minimum.spec.ts
```diff
@@ -97,15 +94,9 @@ describe('GenerateVdevsService - treat disk size as minimum', () => {
...
-          ['enclosure1-disk1', 'enclosure1-disk2', 'enclosure1-disk3'],
-        ]),
-        [VdevType.Log]: expectDisks([
-          ['small-ssd1'],
-        ]),
-        [VdevType.Spare]: expectDisks([
-          ['small-ssd2'],
-        ]),
+        [VdevType.Data]: expectDisks([['enclosure1-disk1', 'enclosure1-disk2', 'enclosure1-disk3']]),
+        [VdevType.Log]: expectDisks([['enclosure2-never-used-too-small']]),
```
undsoft, line 98: Let's do something with `enclosure2-never-used-too-small`. It's name doesn't make sense anymore . It can probably be removed.

src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/generate-vdevs-maximize-dispersal.spec.ts
```diff
@@ -130,11 +130,11 @@ describe('GenerateVdevsService - maximize dispersal', () => {
       ]),
       [VdevType.Log]: expectDisks([
         ['small-ssd1'],
-        ['small-ssd2'],
+        ['enclosure2-never-used-too-small'],
```
undsoft, line 133: 1. `enclosure2-never-used-too-small` was designed not to be used in tests. Let's change it to 1GB, so that it doesn't appear. Not sure why it worked before.
2. Even after I change it to 1Gb, it tests still pass, which doesn't appear to be correct.

src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/grouped-disks.class.ts
```diff
@@ -10,30 +10,34 @@ import { getDiskTypeSizeMap } from 'app/pages/storage/modules/pool-manager/utils
...
+      }
+      for (const diskSize of Object.keys(this.diskMap.SSD)) {
         if (category.diskSize > Number(diskSize)) continue;
-        matchingTypeAndSize = matchingTypeAndSize.concat(matchingType[diskSize]);
+        matchingTypeAndSize = matchingTypeAndSize.concat(this.diskMap.SSD[diskSize]);
       }
     } else {
-      matchingTypeAndSize = matchingType[category.diskSize];
+      matchingTypeAndSize = Array.from(
+        new Map([...hddDisks, ...ssdDisks].map((value) => [value.devname, value])).values(),
```
undsoft, line 33: I don't think it's a good idea to return one property when method declares to return the whole interface.

src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/grouped-disks.class.ts
```diff
@@ -10,30 +10,34 @@ import { getDiskTypeSizeMap } from 'app/pages/storage/modules/pool-manager/utils
...
-    const matchingType = this.diskMap[category.diskType];
     let matchingTypeAndSize: UnusedDisk[] = [];
+    const ssdDisks = this.diskMap.SSD[category.diskSize] || [];
+    const hddDisks = this.diskMap.HDD[category.diskSize] || [];
+
     if (category.treatDiskSizeAsMinimum) {
-      for (const diskSize of Object.keys(matchingType)) {
+      for (const diskSize of Object.keys(this.diskMap.HDD)) {
+        if (category.diskSize > Number(diskSize)) continue;
+        matchingTypeAndSize = matchingTypeAndSize.concat(this.diskMap.HDD[diskSize]);
```
undsoft, line 25: Variable name lost its meaning.

src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/grouped-disks.class.ts
```diff
@@ -10,30 +10,34 @@ import { getDiskTypeSizeMap } from 'app/pages/storage/modules/pool-manager/utils
...
+      }
+      for (const diskSize of Object.keys(this.diskMap.SSD)) {
```
undsoft, line 27: Can we make this prettier (i.e. unify two cases)?

undsoft (changes requested):
Approach appears to be correct.

src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/grouped-disks.class.ts
AlexKarpov98: in the end it's an array of data with same interface, we call `.values()` to receive values from Map by key.

src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/grouped-disks.class.ts
AlexKarpov98: yeah, I gonna refactor, I just tried the general idea, if it appears to be correct, I'll refactor it all.

src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/grouped-disks.class.ts
AlexKarpov98: yes, it will be made as a part of refactoring, first of all I wanted to know that I'm moving in the right direction.

src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/generate-vdevs-maximize-dispersal.spec.ts
AlexKarpov98: if I try with 0.99GB it's not appearing

undsoft (approved):
Looks good.

PR #8441, opened 2023-07-14
NAS-122921: Auto select values in pool creation if only one option is available

AlexKarpov98:
Testing:
Please check ticket.
You should be able to see select's to be autofilled if there is only one option available.

Screenshot 2023-07-14 at 10 40 36


undsoft (changes requested):
I think this new functionality work across all steps for some reason. I've been using Data step, but some of other steps got populated.
Снимок экрана 2023-07-14 в 15 08 15

src/app/modules/ix-table2/interfaces/table-column.interface.ts
```diff
@@ -6,7 +6,7 @@ export abstract class ColumnComponent<T> {
   sortBy?: (row: T) => string | number;
   sortable?: boolean;
 
-  protected row: T;
+  row: T;
```
undsoft, line 9: Keep in mind that fix in master is better and should be used instead of this one.

undsoft (approved):
Looks good, thanks 👍🏻

src/app/modules/ix-table2/interfaces/table-column.interface.ts
AlexKarpov98: > 

yeah, sorry, didn't update it - because I started my master fix here and continued as separate branch

PR #8443, opened 2023-07-14
Missing translate tags

MDVAurelien:
Missing translate tags in syslog-card.component.html


undsoft:
This is not really necessary. `syslog_server` is just a user-entered string and `syslog_transport` has values like UDP and TCP, which I assume are usually not translated.

MDVAurelien:
it's strange because the related tools tips for these two fields doesn't work.

undsoft:
Right. This needs to be fixed in `syslog-form.component.html` in `[tooltip]="tooltips.syslog_transport"` and others.

MDVAurelien:
OK I'll take a look in these files

PR #8445, opened 2023-07-14
NAS-xxxxxx: Fix tests

AlexKarpov98:
Current master is broken:
This fixes tests pass.
Screenshot 2023-07-14 at 20 26 23


src/app/modules/ix-table2/interfaces/table-column.interface.ts
```diff
@@ -6,7 +6,7 @@ export abstract class ColumnComponent<T> {
   sortBy?: (row: T) => string | number;
   sortable?: boolean;
 
-  protected row: T;
+  row: T;
```
undsoft, line 9: I don't think it should be public. It's not meant to be accessible from outside now. It's better to figure out a way to call `setRow` in tests.

undsoft (approved):
This still accesses protected field from outside when setting `props`, but I'm going to merge it to unblock master.

PR #8446, opened 2023-07-14
Missing translate tags

MDVAurelien:


undsoft:
Unfortunately this will not work because these values are coming from API. There are no strings for string-extractor to extract from files and they won't be available for translation in JSON files.
This is currently a limitation of our translation system.

PR #8449, opened 2023-07-17
NAS-122953 / 23.10 / Fix `home_mode` issue

bvasilenko:
**Testing**

See ticket for repro steps


AlexKarpov98 (approved):
It worked for me with no errors compared to master.

PR #8451, opened 2023-07-17
NAS-122967 / 23.10 / Maximum number of concurrent calls (20) has exceeded when exporting a pool

AlexKarpov98:
Testing: see ticket


dszidi (approved):
Code looks good and no error message after export

image

PR #8455, opened 2023-07-18
NAS-122870 / 23.10 / [ported] List of strings is converted to list of objects when the field key has same name as a key in the same level of the parent.

AlexKarpov98:
Ported change, we need to get it back to the master branch 😺


bvasilenko (approved):
LGTM

PR #8458, opened 2023-07-18
NAS-122940 / 23.10 / WebUI does not update by itself after setting up a pool for apps

AlexKarpov98:
Testing: see ticket.


dszidi (approved):
Was able to successfully reset pool and get to installation form
LGTM

PR #8459, opened 2023-07-18
NAS-116866 / 23.10 / Allowing scrub to be paused

undsoft:
Testing: start a scrub, pause a scrub, unpause a scrub, have fun.


AlexKarpov98:
@undsoft 
It makes me available to CLICK many times before it actually changed the scrub "Status" of PAUSE/RESUME.

https://github.com/truenas/webui/assets/22980553/82e9b188-32da-4bd6-be92-790e17baa808

undsoft:
Unfortunately the delay is because there is a delay between `ws.call` completing and job status updating. If I add a simple loading indicator, it looks worse because indicator stops, but then UI is still not updated.
There is probably a way to solve this, but it's too much effort for now.

PR #8461, opened 2023-07-19
NAS-123094 / 23.10 / Replace pageHeader with a directive

undsoft:
And updates some tooltips.

Testing: check page headers.


bvasilenko (approved):
Works good to me

PR #8462, opened 2023-07-19
NAS-122964 / 23.10 / Font for errors on New Pool Manual selection is too small to be readable

AlexKarpov98:
Testing: see ticket


undsoft (changes requested):
Having `error` class inside label component is outside of its scope.
It's better to just replace usages of ix-label for showing errors with custom div with class.

undsoft (approved):
👍🏻

PR #8463, opened 2023-07-19
NAS-121927 / 23.10 / Adding support for dRAID

undsoft:
Adds support for dRAID and NAS-122961: Fixes vdev size estimation.

Testing:
* Open old and new pool manager and compare size estimations for different vdev layouts.
* Using fresh middleware try to add vdevs of dRAID type.


RehanY147 (approved):
Looks good. Size estimates match on old and new pool wizard. Can add pools of Draid type.

PR #8465, opened 2023-07-20
NAS-122909: Update system info on dashboard init

bvasilenko:
**Summary**

Fix for Uptime is not updating.

**Testing**

On **Dashboard** page, at **System Info** widget the **Uptime** should update, when user navigates away to another page, and then returns back to dashboard.


denysbutenko (approved):
Looks good.

Just wondering if real-time updates are possible there. For example, get the initial uptime value from the system info and then start the counter increment locally. Take a look at the `serverTime` variable in `reports.service.ts`.

PR #8470, opened 2023-07-21
NAS-122998 / 23.10 / Setting that hidden flag correctly so form control is hidden as well as disabled

RehanY147:
Check ticket description for testing instructions


dszidi (approved):
LGTM

image

PR #8475, opened 2023-07-22
NAS-122954 / 23.10 / Reconfiguring layout options when inventory changes

RehanY147:


dszidi (approved):
LGTM
Following the steps in the ticket no longer results in disabled "Width" and "Number of VDEVs" fields

PR #8476, opened 2023-07-24
Add reset button to pool manager step actions

dszidi:
Adds a reset button for each topology step in pool manager.
This PR also replaces current reset button when manual selection exists

Testing:

- create Data vdevs
- go to Log vdev step and configure one or more vdevs
- hit reset button in step actions
- check that all selections in step are reset to default values
- check that selections in other steps are not affected


RehanY147 (approved):
Looks good. Can reset form or steps individually.

PR #8477, opened 2023-07-24
NAS-123178 / 23.10 / Added default config for loading

RehanY147:
Previously, if the requests to fetch data for the dashboard took long, you would see nothing on the dashboard for a while. This regression was caused by a recent effort to refactor data loading on the dashboard. This PR fixes the regression.


denysbutenko (approved):
Looks good.

PR #8483, opened 2023-07-25
NAS-122757 / 23.10 / Maximum number of concurrent calls (20) has exceeded

undsoft:
Testing: test that pings work, open Storage page and switch to another tab for 5 minutes or close laptop lid. Observe no error upon return.


denysbutenko (approved):
Works well. No errors were seen after the session timeout.

PR #8484, opened 2023-07-25
NAS-122911 / 23.10 / Disk icon should leave more space for labels

dszidi:
Testing:

- Go to manual disk selection dialog in pool wizard
- Check that disk details are in a single column layout
- Check the disk icons on the left. The labels should have enough space to accommodate values like `428 GiB` and `nvme12n12`
- Also make sure that hdd and ssd icons are centered


bvasilenko (approved):
Looks good to me!

PR #8485, opened 2023-07-26
NAS-123110 / 23.10 / NFS 'Add SPN' option should be presented after 'Require Kerberos' is in DB

bvasilenko:
**Testing**

On the **System Settings \> Services \> NFS** page,

To reproduce the issue, follow these steps:

1. Open the NFS configuration page.
2. Select the **Require Kerberos for NFSv4** option.
3. Click **Add SPN** button (adjust one of the conditions on which it's displayed, which checks the system-wide kerberos support)
4. Fill **Add SPN dialog** and submit
5. Save NFS configuration

* Expected Result: First the **Require Kerberos for NFSv4** setting is saved, then the request `nfs.add_principal` is made, but not earlier.


undsoft (commented):
I would use a different approach. I feel like enabling setting is outside of the scope for AddSpnDialog. It may also mislead the user. 
I may for example, tick "Require Kerberos for NFSv4", click Add SPN and then decide against saving the form, not knowing that some settings were already updated.

I would try something like this: when Require Kerberos for NFSv4 is enabled, show Add SPN button (with other conditions), but keep it disabled until the form is saved. Add tooltip over button to communicate to the user that form needs to be saved.

undsoft (changes requested):
.

bvasilenko:
@undsoft requested changes should be ready for review now

undsoft (approved):
Thanks, I think this works better.

PR #8488, opened 2023-07-26
NAS-123226 / 23.10 / Extract translations on every commit from staged files only

denysbutenko:
Note: This PR is just for testing an idea if this approach will work for us.

```
git commit -m "Extract translations on every commit from staged files only"  
21.25s user 
3.17s system 
114% CPU 
21.289 total
```

```
git commit -m "Extract translations on every commit from all files" 
47.47s user 
5.15s system 
106% CPU 
49.624 total
```


undsoft (commented):
Would we need to run `yarn extract` every now and then to remove old translations?

denysbutenko:
> Would we need to run `yarn extract` every now and then to remove old translations?

Good question. Let me check it.

denysbutenko:
@undsoft there is one way only to remove old translations with `--clean` param. So, this idea won't work to keep translations in sync.

PR #8495, opened 2023-07-27
NAS-123023 / 23.10 / Showing virtual IPs on network card

undsoft:
On a HA system, see that we are now showing VIP addresses on Network -> Interfaces.


dszidi (approved):
LGTM

image

PR #8496, opened 2023-07-27
NAS-123149 / 23.10 / Using the endpoint with the latest token always

RehanY147:


RehanY147:
Reproduction steps:
1. Open the file ticket form and fill in all the fields including attaching screenshots but don't submit the form.
2. Let the view sit for 5 minutes and don't touch the mouse/mouse pad even. Let the UI idle completely
3. The dialog to extend session will appear after 5 minutes
4. Choose to extend the session
5. Wait for a few more minutes
6. Now submit the form that you filled in in the first step
7. The upload will fail but the form will submit successfully.

Admittedly, the reproduction is a hit and miss. But this PR ensures that the issue is addressed

undsoft (approved):
Fix looks reasonable.

PR #8497, opened 2023-07-28
NAS-122283 / 23.10 / Fix reporting zoom level for 6 months

denysbutenko:
For testing, go to reporting and zoom out to 6 months, ensuring the time interval is accurate and should match ~180+- days.


undsoft (approved):
👍🏻

PR #8498, opened 2023-07-28
NAS-123185 / 23.10 / Manually checking for forbiddenValues error for name

RehanY147:


bvasilenko (approved):
Works good to me!

image

PR #8499, opened 2023-07-28
Fix Edit SMB ACL Share tracebacks

dszidi:
Testing:

- Go to https://nas-url/sharing/smb and trigger "Edit Share ACL" row action from table row
- In slide in form Who field, select User.
- If there is a current value for the User field, ensure it shows name and not id
- Combobox options for User field should show `builtin` users being shown last
- Saving the form should now work without traceback
- Repeat this process with groups

### DIFF
--- a/src/app/modules/ix-forms/classes/group-combobox-provider.ts
+++ b/src/app/modules/ix-forms/classes/group-combobox-provider.ts
@@ -32,5 +32,5 @@ export class GroupComboboxProvider implements IxComboboxProvider {
       );
   }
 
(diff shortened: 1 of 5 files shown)


denysbutenko (changes requested):
It works well for changing users. And something went weird with groups. Eventually, values are not saved or displayed as a number.

https://github.com/truenas/webui/assets/351613/ac5cbcb6-871e-489f-a461-aab702057f61

denysbutenko (approved):
Looks good.

PR #8500, opened 2023-07-28
NAS-123155 / 23.10 / Check for pending upgrade on ha systems before showing 2fa modal

RehanY147:
This PR also makes it so that HA related data is loaded right after a successful login call. However, it also will reload that store data every 5 minutes when the token is refreshed. Arguable if that should be restricted to only once per login session.


dszidi (commented):
For code review, this looks fine to me.

RehanY147:
I'm testing for this PR but running into issues. Talking to middleware. Will update when possible

PR #8503, opened 2023-07-31
NAS-123274 / 23.10 / Remove usages of CoreService for getting disk temperatures

bvasilenko:
**Testing**

Emulating "Enclosure":
1. run `yarn ui mock -e`
2. run `yarn ui mock-enclosure -e`
3. navigate to **System Settings > Enclosure**

**Expected result:** Disk temperatures should still be fetched every N seconds, according to the new approach. So please test the closure page to make sure that temperatures are still updated properly. `

### DIFF
--- a/src/app/pages/system/view-enclosure/components/enclosure-disks/enclosure-disks.component.ts
+++ b/src/app/pages/system/view-enclosure/components/enclosure-disks/enclosure-disks.component.ts
@@ -18,11 +18,7 @@ import {
   Enclosure, EnclosureElement, EnclosureSlot, EnclosureView,
 } from 'app/interfaces/enclosure.interface';
 import { CoreEvent } from 'app/interfaces/events';
-import {
-  CanvasExtractEvent,
-  DiskTemperaturesEvent,
-  DriveSelectedEvent,
-} from 'app/interfaces/events/disk-events.interface';
+import { CanvasExtractEvent, DriveSelectedEvent } from 'app/interfaces/events/disk-events.interface';
 import { LabelDrivesEvent } from 'app/interfaces/events/label-drives-event.interface';
 import { Pool } from 'app/interfaces/pool.interface';
 import { Disk, TopologyDisk } from 'app/interfaces/storage.interface';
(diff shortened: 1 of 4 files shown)


src/app/pages/system/view-enclosure/components/enclosure-disks/enclosure-disks.component.ts
```diff
@@ -292,11 +286,11 @@ export class EnclosureDisksComponent implements AfterContentInit, OnDestroy {
     this.themeUtils = new ThemeUtils();
     this.diskTemperatureService.listenForTemperatureUpdates();
 
-    core.register({ observerClass: this, eventName: 'DiskTemperatures' }).pipe(untilDestroyed(this)).subscribe((evt: DiskTemperaturesEvent) => {
+    this.diskTemperatureService.temperature$.pipe(untilDestroyed(this)).subscribe((data) => {
```
RehanY147, line 289: We shouldn't make the subject public. Instead, let's make the subject private and make an observable of the subject public.

RehanY147 (changes requested):
This logic is very complicated to understand. I think since we are in the area, we can do a refactor of the `DiskUpdateService` and make it a store. Also, make the logic in `DiskTemperaturesService` a bit more easier to understand.

RehanY147:
@bvasilenko You can disregard my last review. I will create a new ticket for that.
Also, can you give me a more specific instructions on what to test for? I have a mock system, so I don't know how I am going to check for changing temperature updates etc.

bvasilenko:
> You can disregard my last review. I will create a new ticket for that.

Ok!

> Also, can you give me a more specific instructions on what to test for? I have a mock system, so I don't know how I am going to check for changing temperature updates etc.

3. navigate to **System Settings > Enclosure**

image

4. switch to **Temperature** view
image

RehanY147 (approved):

RehanY147 (changes requested):
Same request is being made twice every N seconds. Plase review.

bvasilenko:
@RehanY147 now requested changes should be done

RehanY147 (approved):

PR #8505, opened 2023-07-31
NAS-123320 / 24.04 / Separate calls and jobs directories

undsoft:


denysbutenko (approved):
Looks good.

PR #8506, opened 2023-08-01
NAS-122401 / 24.04 / Correct capitalization in TrueNAS on apps views

denysbutenko:
Use `TrueNAS` as a catalog name on the apps view, other catalog names still capitalize the first letter only.


dszidi (approved):
Was not able to test but code looks good

PR #8509, opened 2023-08-01
NAS-123101 / 24.04 / MinIO Enterprise Configured for SNMD Stuck in Deploying

AlexKarpov98:
Testing: 
Check enterprise MiNio App on m40 device.

Try creating new one -> setting up storage configuration:
Screenshot 2023-08-01 at 18 03 22

You should be able to create/edit with no errors related to it.


src/app/services/schema/app-schema.service.ts
```diff
@@ -544,10 +547,14 @@ export class AppSchemaService {
       }
 
       if (relation.operatorName === '=') {
-        this.handleEqualOperatorNameSubscription(payload, relation);
+        subscription.add(
+          timer(0).pipe(take(1)).subscribe(() => this.handleEqualOperatorNameSubscription(payload, relation)),
```
bvasilenko, line 551: I've noticed these `timer` calls, and I think there should be a better way to implement this. My suggestion is to try `detectChanges` or one of the `NgZone` methods

bvasilenko (approved):
Works overall, just a minor remark

src/app/services/schema/app-schema.service.ts
AlexKarpov98: tried your recommendations, didn't work for me.

PR #8511, opened 2023-08-01
NAS-123319 / 24.04 / The Layout Selection Field is Missing When Creating a New Pool Using The New Pool Creation Wizard After Adding a Disk to an Existing Pool Using The New Pool Creation Wizard

AlexKarpov98:
Testing: 
Go to Add to Pool screen:
Screenshot 2023-08-01 at 18 50 53

Go to DATA step, you should not see Layout selection.
Screenshot 2023-08-01 at 18 51 15

After that -> go to the default Create Pool screen -> Data STEP 
Screenshot 2023-08-01 at 18 51 56

There you should see Layout select. 

Previously it was not there after the steps I described.


denysbutenko (approved):
Looks good

PR #8512, opened 2023-08-02
NAS-122296 / 24.04 / Update time interval when manual zoom on Reports

denysbutenko:
For testing, go to reporting and select range using a cursor on any chart, then check the `Start/End` dates below the chart. Also, try to mouse over the x-axis to see the exact date and time above the legend.


dszidi (changes requested):
For my testing I selected a time range from 7:20 to 7:22. It seems the chart is not consistent with showing the time zones. The graphs labels look like they're based on client time and the stamps below are based on system localization settings

image


image

dszidi (approved):
Looks good

image

PR #8518, opened 2023-08-02
NAS-123273 / 24.04 / Extract drag handle into separate component/directive

AlexKarpov98:


undsoft:
I can live with the error. Component looks simple, I don't see how it can introduce an error like this, could be interaction with something else.

dszidi (approved):
LGTM

PR #8529, opened 2023-08-03
NAS-123278 / 24.04 / `zfs.snapshot.query` `REMOVED` event will not include child snapshots

AlexKarpov98:
I checked that this ticket will not break anything on the UI since we don't use `{"recursive": true}` option.
But I found out that subscribing to updates, will not update the list
// event.fields is not sent for some reason and that's why list is not updated, so this gonna help

So I came up with an idea. (check code changes)

### DIFF
--- a/src/app/pages/datasets/modules/snapshots/store/snapshot.effects.ts
+++ b/src/app/pages/datasets/modules/snapshots/store/snapshot.effects.ts
@@ -47,7 +47,7 @@ export class SnapshotEffects {
     switchMap(() => {
       return this.ws.subscribe('zfs.snapshot.query').pipe(
         filter((event) => event.msg !== IncomingApiMessageType.Removed),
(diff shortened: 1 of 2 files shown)


denysbutenko (approved):
The ticket requirements and PR content are confusing. So, I checked if the snapshot list is updating in different scenarios:

The snapshot list is updated when:

- a new snapshot is added
- an existing snapshot is deleted

The snapshot list is NOT updated when:

- snapshot is cloned

Also, there is [a middleware PR#11797](https://github.com/truenas/middleware/pull/11797) that addresses missing fields for `zfs.snapshot.query` event.

undsoft (changes requested):
Appears to be not needed after recent middleware changes.

AlexKarpov98:
It's tested with latest MiddleWare updates --> and is working fine now.
`event.fields` are included.

Screenshot 2023-08-11 at 13 37 07

PR #8531, opened 2023-08-04
NAS-123271 / 24.04 / Remove web worker from reporting

denysbutenko:
Note: It is MVP without web worker and other optimizations for data flow need to do.
For testing make sure no regressions on reporting.


undsoft (approved):
Everything appears to be in order.

PR #8535, opened 2023-08-07
NAS-123402 / 23.10 / Hiding vdevs count field for spare vdev step

RehanY147:
Test that the Spare vdev step doesn't have the Vdevs Count field. It should only be allowed to have 1 vdev. Also, test that if there is already a spare vdev in the pool, you are not allowed to add a new spare vdev.


AlexKarpov98 (approved):
Works as described.

PR #8536, opened 2023-08-07
NAS-123295 / 23.10 / Added error handling on Shares Dashboard 

RehanY147:


undsoft (commented):
I feel like the video is suggestive of there being a problem when saving smb acl, not necessarily when loading it.

dszidi (approved):
Looks good

RehanY147:
@undsoft Andrew commented on the ticket that the issue was the when saving, websocket was throwing an error and UI wasn't showing that error. So it seemed like the save was successfull.

undsoft:
@RehanY147 I just don't see how adding error handling on loading solves the issue in the ticket. Were you able to reproduce it by rebooting the system?

undsoft (changes requested):
.

RehanY147:
This seems like a middleware issue upon investigation. Closing.

PR #8537, opened 2023-08-07
NAS-123351 / 24.04 / Takeout `checkForUpdate`

bvasilenko:
**Summary**

After refactoring, the `checkForUpdate` was moved to a place in code, which was out of reach of `untilDestroy` pipe.

Moveover, the call `checkForUpdate` was waiting for the sys info to be fetched, without need to do so.

**Testing**

1. Tweak the code to simulate delayed error in `checkForUpdate`
2. Drop update-related keys from your `sessionStorage`
3. Go to Dashboard, not stay too long
4. Quickly navigate from Dashboard to any other page (/system/services, /storage/, ...)

**Expected result:**
Once you left Dashboard page, you should never see an pop-up with an error from Pt "1"

### DIFF
--- a/src/app/pages/dashboard/components/widget-sys-info/widget-sys-info.component.ts
+++ b/src/app/pages/dashboard/components/widget-sys-info/widget-sys-info.component.ts
@@ -136,10 +136,8 @@ export class WidgetSysInfoComponent extends WidgetComponent implements OnInit {
         error: (error) => {
           console.error('System Info not available', error);
         },
-        complete: () => {
-          this.checkForUpdate();
-        },
       });
+      this.checkForUpdate();
     }
     if (this.sysGenService.getProductType() === ProductType.ScaleEnterprise) {
       this.store$.select(selectIsHaLicensed).pipe(untilDestroyed(this)).subscribe((isHaLicensed) => {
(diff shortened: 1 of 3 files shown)


undsoft (changes requested):
I think there is another problem. This error doesn't have error handling, yet error is shown.
We do this in `handleError` in `ErrorHandlerService`, I think it is too much.

bvasilenko:
@undsoft now requested changes should be addressed

undsoft (changes requested):
It is better to remove showing dialog in `ErrorHandlerService.handleError()`. I think it's better not to show error dialog on every JS error.

src/app/pages/dashboard/components/widget-sys-info/widget-sys-info.component.ts
```diff
@@ -284,15 +284,20 @@ export class WidgetSysInfoComponent extends WidgetComponent implements OnInit {
...
+          sessionStorage.updateAvailable = 'false';
+          return;
+        }
 
-      this.updateAvailable = true;
-      sessionStorage.updateAvailable = 'true';
+        this.updateAvailable = true;
+        sessionStorage.updateAvailable = 'true';
+      },
+      error: (err) => {
```
undsoft, line 298: Error handler can be removed, uncaught errors will be logged to sentry automatically.

undsoft (approved):
Looks good.

PR #8555, opened 2023-08-08
NAS-123457 / 24.04 / Rsync: Create New SSH Connection opens dialog

denysbutenko:
For testing, go to data protection dashboard and try to create a new rsync task. choose the `Create New` option for the SSH connection dialog should appear. make sure that form in dialog works well.


AlexKarpov98 (approved):
Screenshot 2023-08-09 at 16 15 49

Form is now visible in a modal, but we have problem when saving - it adds horizontal scrollbar.

denysbutenko:
> Form is now visible in a modal, but we have problem when saving - it adds horizontal scrollbar.

Good catch. There was horizontal scrolling even when no saving. Fixed.

PR #8560, opened 2023-08-08
NAS-123415 / 24.04 / "Mixing vdev layouts not allowed" error appears when creating a new pool

AlexKarpov98:
Testing:
How it was reproducing before:
- Create new Pool with RaidZ1 `Data` Layout step.
- Add Vdev for newly created Pool
- Create new pool with any Layout in `Data` step.
--- Here error showed up 
image

You can try on latest nightly to reproduce the bug; (you can as well check ticket description video)

On localhost:
Try to go ahead with above steps -> you should no longer see that error.


undsoft (approved):
Looks good.

PR #8574, opened 2023-08-10
NAS-123480 / 24.04 / WebDAV Install missing "Add Share" button. produces error unable to resolve

AlexKarpov98:
Testing: see ticket.
Make sure that Shares are working on `Install WebDAV` page.

Screenshot 2023-08-10 at 16 33 05


undsoft (approved):
It works, but if we expect something like `false` it may be more straightforward to just check for it.

PR #8582, opened 2023-08-11
NAS-123151 / 24.04 / HA: Alert is duplicated

AlexKarpov98:
Testing:
Check that alerts are working and new one comes on the list.

Screenshot 2023-08-11 at 13 28 47

The end solution came up after talking to V. Vinogradenko.

### DIFF
--- a/src/app/modules/alerts/store/alert.effects.ts
+++ b/src/app/modules/alerts/store/alert.effects.ts
@@ -29,7 +29,7 @@ import { alertIndicatorPressed } from 'app/store/topbar/topbar.actions';
 @Injectable()
 export class AlertEffects {
   loadAlerts$ = createEffect(() => this.actions$.pipe(
-    ofType(adminUiInitialized, alertIndicatorPressed),
+    ofType(adminUiInitialized, alertIndicatorPressed, alertReceivedWhenPanelIsOpen),
     switchMap(() => {
       return this.ws.call('alert.list').pipe(
         map((alerts) => alertsLoaded({ alerts })),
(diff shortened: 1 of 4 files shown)


src/app/modules/alerts/store/alert.effects.ts
```diff
@@ -4,27 +4,27 @@ import { Store } from '@ngrx/store';
...
 import { AlertSlice, selectDismissedAlerts, selectUnreadAlerts } from 'app/modules/alerts/store/alert.selectors';
 import { WebSocketService } from 'app/services/ws.service';
 import { adminUiInitialized } from 'app/store/admin-panel/admin.actions';
+import { alertIndicatorPressed } from 'app/store/topbar/topbar.actions';
 
 @Injectable()
 export class AlertEffects {
   loadAlerts$ = createEffect(() => this.actions$.pipe(
-    ofType(adminUiInitialized),
+    ofType(adminUiInitialized, alertIndicatorPressed, alertReceived),
```
undsoft, line 27: This is different from loading alerts when alerts panel open. This will query `alert.list` every time there is a new alert or an update to an existing one.
If we want to still do this, I feel like the larger refactoring is needed, e.g. it may be necessary to trigger this update behaviour only when panel is open.

AlexKarpov98: @undsoft 
I agree, but on the other hand - we want to know a `fresh counter` and for this we always need to grab a list of alerts I assume.
Screenshot 2023-08-11 at 14 22 51

AlexKarpov98: or we may handle counter with subscription and events we receive somehow.
And only fetch fresh alerts if the `Alerts panel` is opened and new event received.

AlexKarpov98:
@undsoft - give it a try now.

src/app/modules/alerts/store/alert.effects.ts
```diff
@@ -41,34 +44,26 @@ export class AlertEffects {
...
         switchMap((event) => {
-          switch (event.msg) {
-            case IncomingApiMessageType.Added:
-              return of(alertAdded({ alert: event.fields }));
-            case IncomingApiMessageType.Changed:
-              return of(alertChanged({ alert: event.fields }));
-            default:
-              return EMPTY;
+          if ([IncomingApiMessageType.Added, IncomingApiMessageType.Changed].includes(event.msg) && isAlertsPanelOpen) {
+            return of(alertReceivedWhenPanelIsOpen());
```
undsoft, line 52: `alertReceivedWhenPanelIsOpen` doesn't seem to go anywhere?

undsoft (changes requested):
1. We are not loading alerts via `alert.list` when panel is open like discussed?
2. If I open two tabs and dismiss/reopen alerts in second tab, indicator won't update in first tab.

AlexKarpov98:
@undsoft 

> 1. We are not loading alerts via `alert.list` when panel is open like discussed?
> 2. If I open two tabs and dismiss/reopen alerts in second tab, indicator won't update in first tab.

Now it's working as expected

undsoft (approved):
Looks good, thanks.

PR #8593, opened 2023-08-14
NAS-123526 / 24.04 / Extend session may appear when there is no active connection

AlexKarpov98:
Testing: 
Check that you are logged out even tho there is no ws connection and user is logged in.
Extend session modal will appear as usual and closes the session, before once there was no connection - it stuck.

image

example: 
11:55:19 on screenshot
but now 12:03:59 and session still is not closed


denysbutenko (approved):
Looks good

PR #8597, opened 2023-08-14
NAS-122599 / 24.04 / Apps filter should reset when the user navigates away from the apps dashboard page

AlexKarpov98:
Testing:
See ticket.

Final result:

https://github.com/truenas/webui/assets/22980553/b30770f6-4a85-489b-b90c-b705bd9c5926


AlexKarpov98:
@undsoft - it only resets filters once we completely navigate away from the `/apps` section
When we view details, installed apps or something related to `/apps` route -> filters will not be cleared.

bvasilenko (approved):
Looks good to me

PR #8604, opened 2023-08-16
NAS-122287 / 24.04 / Reporting:  Graph resets from a past date/time to current date/time when change the zoom increment

AlexKarpov98:
Testing: see ticket


undsoft (changes requested):
I am not sure how it currently works, but it feels kinda strange. I think my natural expectation is that when I zoom in or out, I zoom in and out to the center of the chart. Start of the period like suggested in the ticket could do too. 
Right now it seems to zoom in to end of the period?

AlexKarpov98:
@undsoft 
yes - currently it uses the `end of the period`
We can use `start of the period` but it's annoying if I want to get to the end date (like today)

Like:
I want to see some data for 2021 year
I make it 6 months -> roll back (<<)

And now I want to get fresh data 
I click >> Forward

And now I want to see the current date only -> viewing data by HOURS 
But what I see now -> I need to click forward per hour (starting from the 6 months old)

undsoft:
I see your point.
Perhaps we could add a special condition that when you are rightmost position, it zooms in/out to the end date, but if you have scrolled back it zooms in to the center.
Because in an alternative scenario I want to see specific day months ago. I zoom out, go back and then zoom, but it jumps around.

AlexKarpov98:
@undsoft - with middle date - now it's easier to navigate and get to the TODAY's date.

https://github.com/truenas/webui/assets/22980553/2f25e032-8024-4445-9328-a53d45a6e36c

undsoft (changes requested):
1. When I zoom out, graphs jump to a different date.
2. I still think we should have a special condition when chart is scrolled to the right. In this case, it should zoom in to latest date.

undsoft (changes requested):
I think it works correctly at zooming in and out using center of the chart.
However, it does not seem to correctly stick to rightmost position when the chart's end date is current date.

AlexKarpov98:
@undsoft - added check for isToday logic.
what should we count as well as rightmost position?

src/app/pages/reports-dashboard/components/report/report.component.ts
```diff
@@ -255,41 +262,62 @@ export class ReportComponent extends WidgetComponent implements OnInit, OnChange
...
     }
     this.zoomLevelIndex -= 1;
-    const rrdOptions = this.convertTimespan(this.currentZoomLevel);
+
+    const halfPeriodMilliseconds = this.getHalfPeriodMilliseconds();
+
+    const rrdOptions = this.convertTimeSpan(
+      this.currentZoomLevel,
+      ReportStepDirection.Backward,
+      isToday(this.currentEndDate) ? this.currentEndDate
```
undsoft, line 312: Indentation is wrong, plus it's hard to follow. Best to change into `if` statements.

undsoft (changes requested):
Zooming to center appears to be working okay.
Zooming in the rightmost position doesn't appear to work correctly, because when I zoom out and in the End Date changes.

https://github.com/truenas/webui/assets/1481270/66770a0b-abc0-4c0c-aac8-16a7d4694cec

undsoft (approved):
Mission accomplished 🎈🥳

PR #8608, opened 2023-08-16
NAS-123525 / 24.04 / Removed CoreService

RehanY147:


src/app/pages/reports-dashboard/reports.service.ts
```diff
@@ -35,47 +33,13 @@ export class ReportsService implements OnDestroy {
   private hasNfs = false;
   private hasPartitions = false;
 
+  private legendEventEmitter$ = new Subject<LegendDataWithStackedTotalHtml>();
+  readonly legendEventEmitterObs$ = this.legendEventEmitter$.asObservable();
```
undsoft, line 37: This is unnecessary as subjects are already observables.

undsoft (approved):
Looks good, thanks 👍🏻

src/app/pages/reports-dashboard/reports.service.ts
RehanY147: I only do this because subjects allow emitting new values if you expose them directly. This subject is not supposed to emit values outside of this service. But if we expose it publicly, that becomes possible.

PR #8613, opened 2023-08-17
NAS-123558 / 24.04 / Applications do not report CPU, RAM, or disk usage

AlexKarpov98:
Testing: see ticket
`TODO: Hide for Cobia, this is not going to be implemented.`


denysbutenko (approved):
Looks good.

PR #8618, opened 2023-08-17
NAS-123270 / 24.04 / Disable global mock when remote is changed

denysbutenko:
For testing, make sure the global mock is disabled after `yarn ui remote -i ...` execution.


AlexKarpov98 (approved):
👍

PR #8621, opened 2023-08-18
NAS-123426 / 24.04 / Fix default home path

bvasilenko:
**Testing**

1. **Datasets > Add Dataset** Add a dataset named e.g. `home`

image

2. **Shares > Windows (SMB) Shares > Add**
   - Path - this dataset (`home`)
   - Purpose - **No presets**
   - Use as Home Share - `true`

image

3. **Credentials \> Local Users \> Add**
   - Full name - for example, `newUser`
   - Password - for example, `qwerty`
   - Create Home Directoty - `true`

image

**Expected Behavior:** A user should be added, with a Home Directory similar to `mnt/my pool/home/newUser`

image

### DIFF
--- a/src/app/pages/account/users/user-form/user-form.component.spec.ts
+++ b/src/app/pages/account/users/user-form/user-form.component.spec.ts
@@ -125,7 +125,7 @@ describe('UserFormComponent', () => {
 
       const usernameInput = await loader.getHarness(IxInputHarness.with({ label: 'Username' }));
       await usernameInput.setValue('test');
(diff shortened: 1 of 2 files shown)


RehanY147 (changes requested):
The form can be submitted and the home directly is filled in but I'm confused about what you said

> Home directory field automatically updated with the path to your share and the username

What about the username? I don't see anything about a username. Please elaborate.

bvasilenko:
@RehanY147 I've rewritten the testing instructions, please check the updated steps

RehanY147 (approved):

PR #8632, opened 2023-08-19
NAS-123603 / 24.04 / Remove leftovers ngx-datatable

denysbutenko:


dszidi (approved):
Looks good. A quick `git grep ngx-datatable` shows that the only mentions left are in python and selenium test files. I'm assuming those will be taken care of by their department

PR #8645, opened 2023-08-22
NAS-122321: Text field unreadable

AlexKarpov98:
Testing:
Check that Replication Task - STATE error modal message is readable.

**Before:**
Screenshot 2023-08-22 at 14 07 50

**After:**
Screenshot 2023-08-22 at 14 07 36


undsoft (approved):
I'm not 100% sure it resolves original problem, but changes look okay.

PR #8646, opened 2023-08-22
NAS-123647 / 24.04 / Add Replication Task Advanced Form [SAVE] button - is always active

AlexKarpov98:
Testing:
Check Replication Task Advanced form SAVE button.
It should be disabled if form is invalid. (previously it was always active)


bvasilenko (approved):
LGTM

PR #8647, opened 2023-08-22
NAS-121466 / 24.04 / Topology overview hides VDEV information if there is a warning for redundancy mismatch

AlexKarpov98:
Testing: see ticket:

**Before:**
Screenshot 2023-08-22 at 17 51 51

**After:**
Screenshot 2023-08-22 at 17 51 57


denysbutenko (approved):
Preview looks good.

PR #8652, opened 2023-08-23
NAS-115255 / 24.04 / Persist theme choice in localstorage

AlexKarpov98:
Testing: 
See ticket & Check theme behaviour.
Reload the page after theme was set, try changing themes.

https://github.com/truenas/webui/assets/22980553/43404e9a-cb3a-4493-83ee-f49128588590


src/app/modules/layout/components/auth-layout/auth-layout.component.ts
```diff
@@ -1,22 +1,33 @@
...
+    private themeService: ThemeService,
+    @Inject(WINDOW) private window: Window,
+  ) {
     // Translator init
     language.setLanguageFromBrowser();
   }
 
   ngOnInit(): void {
-    this.setCssVars(this.theme);
+    let theme: Theme = defaultTheme;
```
undsoft, line 23: I think it would be better to extract code here and below to methods in `ThemeService`.

undsoft (approved):
Looks good, thanks.

PR #8662, opened 2023-08-23
german translation fix

jonnytest1:


AlexKarpov98 (approved):
Thanks.

AlexKarpov98:
> btw are there any solutions for adding tooltips via the language file ?
> 
> for example i'm thinking adding a short eplainer for what a vdev is above every occurence of VDEV might be neat

Currently we have tooltips via the language file, maybe it's not available for all tooltips?

jonnytest1:
ok thx i'll take a look 🫡

PR #8664, opened 2023-08-24
NAS-123631: Move services forms to slide-in

AlexKarpov98:
Testing: 
Check that services are now opening in SlideIn instead of router navigation.

_i.e:_
Screenshot 2023-08-24 at 11 47 01


src/app/pages/services/services.component.html
```diff
@@ -29,14 +29,7 @@
...
           [diameter]="30"
-          [matTooltip]="
-            (service.state === ServiceStatus.Running
-              ? 'Stopping'
-              : 'Starting'
-            )
-              | uppercase
-              | translate
-          "
+          [matTooltip]="(service.state === ServiceStatus.Running ? 'Stopping' : 'Starting') | uppercase | translate"
```
undsoft, line 32: Both old and new string won't actually be correctly parsed by translate message extractor.
It should be something like '? (string | translate) : (another string | translate)'

src/app/pages/services/services.component.html
```diff
@@ -75,6 +68,7 @@
           matTooltipPosition="left"
           [ixTest]="[service.service, 'edit']"
           [attr.aria-label]="'Edit' | translate"
+          [attr.aria-service]="service.name"
```
undsoft, line 71: `aria-service` is not a valid attribute.

undsoft (approved):
Cancel buttons in forms do not work anymore. These can be removed.
Otherwise looks good.

undsoft:
Actually for tests you can use `getHarnessInCell` for now and I'll create a ticket to do something better.

PR #8665, opened 2023-08-24
NAS-123683: fixed node version

jonnytest1:


bvasilenko:
@undsoft would you please share your opinion on if it'll work for CI?

undsoft (approved):
Looks good, thanks.

PR #8673, opened 2023-08-25
NAS-123469 / 24.04 / Refactor CronCardComponent to ix-table2

AlexKarpov98:
Testing: 
test updated CronCardComponent

### DIFF
--- a/src/app/pages/system/advanced/cron/cron-card/cron-card.component.html
+++ b/src/app/pages/system/advanced/cron/cron-card/cron-card.component.html
@@ -1,3 +1,58 @@
-<ix-conf-table
-  [conf]="tableConfig"
-></ix-conf-table>
+<mat-card>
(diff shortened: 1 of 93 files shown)


dszidi (changes requested):
The row width shouldn't shrink when details are expanded

image

AlexKarpov98:
> The row width shouldn't shrink when details are expanded
> 
> 
> 
> image
> 
> 

That's the other bug I opened.
Btw - you tested wrong place, you need to test Card, not List page.

dszidi:
> ... That's the other bug I opened. Btw - you tested wrong place, you need to test Card, not List page.
 
 Apologies. I misread one of the file paths from the diff so I also checked the cron page.

dszidi (approved):
The card in System Advanced works as expected.
 LGTM
 
image

PR #8676, opened 2023-08-27
NAS-123521 / 24.04 / Show readonly layout field for Data VDEVs when adding disks to existing pool

dszidi:
This PR exposes preselected layout field for data vdevs when adding vdevs to existing pool

To test:
- go to add vdevs page
- On Data step check to see that layout field is visible as a read only input field
- Check that the read only field has tooltip text

### DIFF
--- a/src/app/pages/storage/modules/pool-manager/components/pool-manager-wizard/components/layout-step/automated-disk-selection/automated-disk-selection.component.html
+++ b/src/app/pages/storage/modules/pool-manager/components/pool-manager-wizard/components/layout-step/automated-disk-selection/automated-disk-selection.component.html
@@ -7,7 +7,7 @@
     [required]="true"
   ></ix-select>
 </div>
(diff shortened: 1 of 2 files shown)


bvasilenko (approved):
Works good to me

src/app/pages/storage/modules/pool-manager/components/pool-manager-wizard/components/layout-step/automated-disk-selection/automated-disk-selection.component.html
```diff
@@ -7,6 +7,16 @@
     [required]="true"
   ></ix-select>
 </div>
+<ng-template #showLayoutInfo class="layout-container">
```
undsoft, line 10: `class` won't be applied on an ng-template.

undsoft (approved):
Nice.

PR #8679, opened 2023-08-28
NAS-123699 / 24.04 / Cobia BETA.1: Sudo Commands fields on Edit User are not saving

AlexKarpov98:
Testing: see ticket


undsoft (approved):
Looks good. I will create a ticket to add explicit `canAddNew` input in the future.

PR #8680, opened 2023-08-28
NAS-123698 / 24.04 / Clarify Manual Selection for Spares

AlexKarpov98:
Testing: see ticket


src/app/pages/storage/modules/pool-manager/components/manual-disk-selection/manual-disk-selection.component.ts
```diff
@@ -58,10 +59,13 @@ export class ManualDiskSelectionComponent implements OnInit {
         }
       }
 
+      this.hideAddVdevButton = this.data.vdevsLimit && vdevs.length >= this.data.vdevsLimit;
```
undsoft, line 62: Setting `hideAddVdevButton` doesn't belong to something called `isSaveDisabled$`.
You need to create a new observable for it.

undsoft (approved):
Would be good to add a test case.

PR #8681, opened 2023-08-28
NAS-123675 / 24.04 / Fixes for app info card

denysbutenko:
App Info Card should display:

- application name specified by the user and not the chart name.
- docker image tag as the app version for custom apps

BETA-1:
image

Changes:
image


AlexKarpov98 (approved):
Works good.
Screenshot 2023-08-29 at 17 15 12

PR #8684, opened 2023-08-28
NAS-120309 / 24.04 / cdk-overlay-pane class HTML divs build up in the DOM and cause the UI to become sluggish

bvasilenko:
See ticket please


undsoft (approved):
Looks good, thanks.

PR #8690, opened 2023-08-29
NAS-123695 / 24.04 / Bug – fast clicking user can click on Data, but end up on Enclosure Options

AlexKarpov98:
Testing: see ticket


denysbutenko (approved):
I stay on Data step when disks are loaded. Looks good.

PR #8691, opened 2023-08-29
NAS-123563 / 24.04 / duplicate "ZFS ARC Result" graphs in reporting section of webUI

AlexKarpov98:
Testing: see ticket


undsoft:
@AlexKarpov98 Do we understand why the original issue was happening? It appears that those two charts were showing different things.

AlexKarpov98:
@undsoft - seems like data is different, it's the object:

Screenshot 2023-08-29 at 17 32 11

AlexKarpov98:
Screenshot 2023-08-29 at 17 34 57
@undsoft  - we can render it like that, what do you think?

RehanY147 (changes requested):
I agree with what Evgeny said. I'm not sure what to look out for when testing.

undsoft (changes requested):
This approach is correct, thought I think it would be better to have an exception to correct `demand_data` to `Demand Data`.
But titles for some other charts don't look good now, e.g. `Disk I/O sdm - sdm` and `Interface Traffic eno1 - eno1`.

AlexKarpov98:
> This approach is correct, thought I think it would be better to have an exception to correct `demand_data` to `Demand Data`. But titles for some other charts don't look good now, e.g. `Disk I/O sdm - sdm` and `Interface Traffic eno1 - eno1`.

ok, in this case I need to check if there are more than one item with same name - then I'll add an extra title.

AlexKarpov98:
@undsoft @RehanY147
 
Screenshot 2023-09-01 at 15 43 35

Final result

undsoft (approved):
This works, thanks.

AlexKarpov98:
@RehanY147 - I need your approve for merge :)

RehanY147 (approved):

PR #8696, opened 2023-08-29
Dataset storage use could be finer graded

bvasilenko:
**Testing**

On **Datasets** page, the accurate volume should be displayed for a dataset. 

For example, for a dataset with 1.43 terabyte,  the "1\.43TiB" should be reported in the dataset level. You should compare the value displayed to the volume displayed on **Data Protection** page, when adding a replication task


undsoft (approved):
Thanks, looks good.

undsoft:
backport

PR #8697, opened 2023-08-30
NAS-123810 / 24.04 / Fix initialization on ixDetailsHeight directive

denysbutenko:
For testing, go to Apps and check that the extra scrollbar does not appear. Ensure other places that use `ixDetailsHeight` are still functional after switching between pages (e.g., from Datasets to Apps pages).

image

### DIFF
--- a/src/app/core/components/directives/details-height/details-height.directive.ts
+++ b/src/app/core/components/directives/details-height/details-height.directive.ts
@@ -48,6 +48,7 @@ export class IxDetailsHeightDirective implements OnInit, OnDestroy, OnChanges {
 
     this.element.nativeElement.style.height = this.heightCssValue;
     this.window.addEventListener('scroll', this.onScrollHandler, true);
+    setTimeout(() => this.onScroll());
   }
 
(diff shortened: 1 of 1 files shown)


src/app/core/components/directives/details-height/details-height.directive.ts
```diff
@@ -48,6 +48,7 @@ export class IxDetailsHeightDirective implements OnInit, OnDestroy, OnChanges {
 
     this.element.nativeElement.style.height = this.heightCssValue;
     this.window.addEventListener('scroll', this.onScrollHandler, true);
+    setTimeout(() => this.onScroll());
```
RehanY147, line 51: Why is `setTimeout` needed?

RehanY147 (changes requested):

src/app/core/components/directives/details-height/details-height.directive.ts
denysbutenko: There is an issue with height detection if you switches pages between apps and dataset.

RehanY147 (approved):

PR #8703, opened 2023-08-30
NAS-123811 / 24.04 / Layout fixes for Datasets and Apps

denysbutenko:
image

image

The height for both columns should be equal.


dszidi (approved):
Looks good

PR #8704, opened 2023-08-30
NAS-123815 / 24.04 / Remove unused component

undsoft:
Code review is enough.


AlexKarpov98 (approved):
We won't miss these component. 👍

PR #8705, opened 2023-08-30
NAS-122963 / 24.04 / Pool creation (New) : Pool Name field does not validate at time of entry

AlexKarpov98:
Testing: check pool name validator.
As well test that Wizard is blocked for creation once there are errors.
(the error comes from the middleware)

Screenshot 2023-08-30 at 16 32 56

### DIFF
--- a/src/app/enums/default-validation-error.enum.ts
+++ b/src/app/enums/default-validation-error.enum.ts
@@ -11,4 +11,5 @@ export enum DefaultValidationError {
   Number = 'number',
   Cron = 'cron',
   Ip2 = 'ip2',
+  InvalidPoolName = 'invalidPoolName',
 }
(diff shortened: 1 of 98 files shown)


src/app/enums/default-validation-error.enum.ts
```diff
@@ -11,4 +11,5 @@ export enum DefaultValidationError {
   Number = 'number',
   Cron = 'cron',
   Ip2 = 'ip2',
+  InvalidPoolName = 'invalidPoolName',
```
undsoft, line 14: This is a very custom error and doesn't belong here.
It may be better to use `IxValidatorsService.customValidator`.

undsoft (approved):
Looks good.

PR #8707, opened 2023-08-30
NAS-123749 / 24.04 /  "Spares" category from pool creation "Configuration Preview" inaccurate/unneeded

AlexKarpov98:
Testing: see ticket


src/app/pages/storage/modules/pool-manager/components/configuration-preview/configuration-preview.component.ts
```diff
@@ -20,7 +21,14 @@ export class ConfigurationPreviewComponent {
 
   protected name$ = this.store.name$;
   protected encryption$ = this.store.encryption$;
-  protected topology$ = this.store.topology$;
+  protected topology$ = this.store.usesDraidLayout$.pipe(switchMap((usesDraidLayout) => {
```
RehanY147, line 24: This will limit `topology$` to only emit when `usesDraidLayout$` emits first. At least we should be using the `map` operator on `topoology$`

RehanY147 (changes requested):
There are too many new null checks with `?` are introduced here. A better approach would be to ensure that it's not possible for that value to be `null/undefined` for all these instances. This has potential to cause issues in the future. Better to cover it now.

RehanY147 (approved):

PR #8718, opened 2023-08-31
NAS-123762 / 24.04 / Fix JIRA token in File Ticket form

denysbutenko:
For testing see details in the ticket


AlexKarpov98 (approved):
I was able to login to Jira.
Screenshot 2023-09-01 at 14 06 53

But I can't create ticket:

Screenshot 2023-09-01 at 14 07 51

PR #8721, opened 2023-08-31
NAS-123802 / 24.04 / Fix position of children in dataset tree

bvasilenko:
**Testing**

On **Datasets** page,

1. Create 2 datasets, with identical names, but the second one containing `-` character and some arbitrary postfix in it's name. For example:
   1. **pool/data**
   2. **pool/data-system**
2. Select one of these datasets, and create a child dataset. For example, **media**
   * **Expected result:** media is displayed under "pool/data"


dszidi (approved):
Looks good. Child dataset is now showing under the correct parent for datasets with hyphenated names

PR #8723, opened 2023-09-01
NAS-123818 / 24.04 / Add explicit flag to allow/disallow new entries in ix-chips

AlexKarpov98:
Testing: see ticket


bvasilenko (approved):
Tested on Apps - works good!

PR #8725, opened 2023-09-01
NAS-123744 / 23.10 / Removing password from VM display device

undsoft:
Testing: VM -> Display button.


AlexKarpov98 (approved):
Code looks good to me.

PR #8727, opened 2023-09-03
NAS-123437 / 24.04 / Disable delete button when snapshot has dependent clones

bvasilenko:
**Summary**
When setting the `recursive` parameter to `true`, the deletion also failed. Therefore, the button was disabled and a tooltip was added.

**Testing**

On Datasets page, after choosing a dataset (for example ix-applications) , and following Manage snapshots hyperlink

Then see ticket please


denysbutenko (changes requested):
image

There is an issue with UI

bvasilenko:
@denysbutenko the requested changes should be addressed now

denysbutenko (changes requested):
A new docker image is not available. It's needed to validate the fix.

denysbutenko (changes requested):
There is another error appears in the console.

https://github.com/truenas/webui/assets/351613/2063f5d6-5e9c-4468-b45c-22ff3c9ea2e3

bvasilenko:
@denysbutenko requested changes should be addressed now

denysbutenko (approved):
No errors were observed. Looks good.

PR #8732, opened 2023-09-04
NAS-123802 / 23.10 / Fix position of children in dataset tree

bvasilenko:
Backported

Original PR: https://github.com/truenas/webui/pull/8721


undsoft (approved):
Backported

PR #8753, opened 2023-09-05
NAS-123920 / 24.04 / Snackbar message not shown when deleting a snapshot

undsoft:
When snapshot is deleted via Delete button (as opposed to a checkbox), snackbar message is going to be shown now.


dszidi (approved):
LGTM

PR #8764, opened 2023-09-05
NAS-123723 / 23.10 / View Enclosure Elements contain empty Descriptor column

AlexKarpov98:
Testing: see that Descriptor field on Enclosure page is no longer there for Cobia & Bluefin:
Screenshot 2023-09-06 at 00 45 46


denysbutenko (approved):
Looks good.

PR #8765, opened 2023-09-05
NAS-121104 / 24.04 / Adding close button to manual validation error

RehanY147:
To test this PR, submit the user creation form while copying over the user id from an older user (that already exists). Middleware will report an error about the conflicting user id. You should see a 'x' button next to the error. Clicking that should remove the error and allow you to submit the form again without having to edit the user id field.


AlexKarpov98 (commented):
Interesting solution, I like it 👍 .
Button is there and is working fine 
Screenshot 2023-09-06 at 15 15 09

But I think it's not fixing the original issue.

Screenshot 2023-09-06 at 15 18 18

If I see this error and I update password - error is still there, well I still can close it but it's not handled automatically.

My temporary fix https://ixsystems.atlassian.net/browse/NAS-121010 is removing the error right now, but if you remove my fix - error will still be there until I close it.

@undsoft - what do you think?

src/app/modules/ix-forms/components/ix-errors/ix-errors.component.ts
```diff
@@ -71,6 +71,9 @@ export class IxErrorsComponent implements OnChanges {
         untilDestroyed(this),
       ).subscribe(() => {
         const newErrors: string[] = Object.keys(this.control.errors || []).map((error) => {
+          if (error === 'ixManualValidateError') {
```
undsoft, line 74: Is there a constant we can use?

undsoft (approved):
This works as an escape hatch in our current system.

PR #8788, opened 2023-09-07
NAS-123931 / 24.04 / Time axis on the reporting charts values incorrect on zooms

AlexKarpov98:
Testing: See ticket.

Go to reports and try to zoom chart by hands, you will be stopped at some point (max 5 minutes)
If you try to zoom in greater (shorter distance) than 5 minutes (like 2 minutes)
You should be not able to do it and reset to previous valid time range.

Screenshot 2023-09-07 at 22 38 47


AlexKarpov98:
> Looks good. Is there any way to disable the animations when the user is already at the maximum zoom level?
> 
> Currently when I get to max zoom level, the start and end date stay static as expected, but the graph does a bunch of animated movement which will likely confuse the user.

@dszidi  I'll investigate if it's possible to disable animation in such scenario!

PR #8801, opened 2023-09-08
NAS-123962 / 24.04 / Showing device/disk on Extent table

undsoft:
Testing: see new column in iSCSI -> Extents table. Also add memory size formatting to Filesize field in Extent form (when type is File).


undsoft:
@dszidi I'm open to suggestions on a better column name.

dszidi (approved):
Looks good

PR #8807, opened 2023-09-08
NAS-124025 / 24.04 / Refactoring init shutdown scripts table

undsoft:
Check table in Advanced settings. Cron table also gained Add button.


denysbutenko (approved):
Looks good.

Minor things are noticed:

- The `When` column displays raw values, e.g., `PREINIT`. Dropdown options look differently.
- Table is reloading when the form is just closed.

undsoft:
Will leave `When` column as is for now. Need better support for value transformations in ix-table2.

PR #8811, opened 2023-09-11
NAS-111637 / 24.04 / Refactor admin-layout.component.ts

denysbutenko:
Extracted sidenav logic into service.

For testing, look for sidenav regressions.


src/app/modules/layout/components/admin-layout/admin-layout.component.ts
```diff
@@ -41,150 +38,95 @@ import { waitForSystemInfo } from 'app/store/system-info/system-info.selectors';
...
       this.menuName = state;
       this.subs = subItems;
-      this.isOpen = true;
+      this.isOpenSecondaryMenu = true;
     }
   }
 
   onMenuClosed(): void {
-    this.isOpen = false;
+    this.isOpenSecondaryMenu = false;
```
RehanY147, line 129: This `isOpenSecondaryMenu` should also be part of the nav service instead of completely being managed by the component

src/app/modules/layout/components/admin-layout/admin-layout.component.ts
```diff
@@ -209,4 +151,13 @@ export class AdminLayoutComponent implements OnInit, AfterViewInit, OnDestroy {
         }
       });
   }
+
+  private listenToRouteChange(): void {
```
RehanY147, line 155: This can also be setup inside the service instead of leaving it up to the component. Maybe service can emit an observable on when to close the navbar at most but the decision making should happen inside the service itself.

RehanY147 (changes requested):

RehanY147 (changes requested):
UI Crashing after signin

RehanY147 (approved):
Navar seems to be working ok

PR #8812, opened 2023-09-11
NAS-123859 / 24.04 / Fix hostname for HA System on widget-sys-info

denysbutenko:
For testing, check ticket details and ensure the hostname is correctly displayed on the sys-info widgets. Check against the HA system and make sure there are no regressions for non-HA systems.


undsoft (changes requested):
1. This looks somewhat complicated.
2. And I'm confused as to why this is needed, if we have a `failover.call_remote system.info` in `ngOnInit`. Shouldn't it return correct hostname for the other machine?

undsoft (approved):
Thanks, looks good.

PR #8818, opened 2023-09-11
NAS-124013 / 24.04 / Refactor table in AllowedAddressesCardComponent

AlexKarpov98:
Refactoring to ix-tables


undsoft (approved):
Would be good to have a better empty state, otherwise looks good.

PR #8835, opened 2023-09-13
NAS-124047 / 24.04 / Refactor tables in SharesDashboardComponent

AlexKarpov98:
Testing: 
Check Shares route and ticket.

Result:
Screenshot 2023-09-13 at 14 08 57


src/app/modules/ix-table2/components/ix-table-body/cells/ix-cell-text/ix-cell-text.component.html
```diff
@@ -1 +1 @@
-{{ value }}
+{{ value || ' - ' }}
```
bvasilenko, line 1: Noticed that this change will affect all the falsy values. Examples of affected values are `false`, `""`, `0`, `null`, `undefined`. And difference between `""` and `undefined` (or between `0` and `null`) could play significant role in some configurations. I suggest to use a more strict check to display only for empty values (`null` or `undefined`)

src/app/modules/ix-table2/components/ix-table-body/ix-table-body.component.html
```diff
@@ -6,7 +6,10 @@
     [hidden]="isLoading"
     (click)="onToggle(row)"
   >
-    <td *ngFor="let column of displayedColumns; index as idx">
+    <td
+      *ngFor="let column of displayedColumns; index as idx"
+      [ngClass]="column.cssClass ? column.cssClass : ''"
```
bvasilenko, line 11: Maybe it would be easier to do `column.cssClass || ''`

bvasilenko (changes requested):
In UI width of the heading needs a bit more work:

image

src/app/modules/ix-table2/components/ix-table-body/cells/ix-cell-text/ix-cell-text.component.html
AlexKarpov98: > Noticed that this change will affect all the falsy values. Examples of affected values are `false`, `""`, `0`, `null`, `undefined`. And difference between `""` and `undefined` (or between `0` and `null`) could play significant role in some configurations. I suggest to use a more strict check to display only for empty values (`null` or `undefined`)

okay, make sense

AlexKarpov98:
> In UI width of the heading needs a bit more work:
> 
> image

@bvasilenko - can you help me reproducing this UI bug?
Which device size should I use?

Screenshot 2023-09-17 at 17 01 47

bvasilenko (approved):
LGTM

PR #8841, opened 2023-09-13
NAS-124045 / 24.04 / Add support for value transformations in ix-table2

bvasilenko:
**Testing**
On pages where `ix-table2` is used, test the table displaying and sorting and execution of row actions


AlexKarpov98 (approved):
Works good. Can we cover it with tests as well?

PR #8844, opened 2023-09-13
NAS-123780 / 24.04 / Mock Utility: support more expansion shelves

dszidi:
Testing:
Make sure the following shelves render properly. Check all element views not just disks view.

- ES12
- E16
- E24
- ES60


undsoft (approved):
As far as I can tell this looks okay.

PR #8846, opened 2023-09-13
NAS-124100 / 24.04 / Extract components from TopbarComponent

undsoft:
Check two new components: HA icon in topbar and ix logo.
You can test different HA states by disabling it in Settings -> Failover.


denysbutenko (approved):
image
image

Looks good.

PR #8859, opened 2023-09-15
NAS-123861 / 24.04 / Added last_run columns

RehanY147:
Check for `'Last Run'` columns on the tables on Data Protection dashboard as well as individual pages.


denysbutenko (changes requested):
`Cannot read properties of undefined (reading '$date')`

RehanY147:
> `Cannot read properties of undefined (reading '$date')`

Please provide more info. What version of the system are you on? For what table did you get this info? Also please share the task object returned from middleware for that row. @denysbutenko

PR #8861, opened 2023-09-16
NAS-123651 / 23.10 / Fixing units conversion

RehanY147:
For testing, check that the values for 'In' and 'Out' values on the widgets for interfaces on the dashboard are accurate in units and values.


AlexKarpov98 (approved):
👍

PR #8866, opened 2023-09-17
NAS-124147 / 24.04 / Add support for cssClass for cells and tooltip for text-cells

AlexKarpov98:


src/app/modules/ix-table2/components/ix-table-body/cells/ix-cell-text/ix-cell-text.component.html
```diff
@@ -1 +1,3 @@
-{{ value }}
+<span [matTooltip]="value?.toString()">
```
bvasilenko, line 1: I noticed that you've assuming that `value` will always have `toString()` method. I'd suggested to allow a broader range of values, and include even those without `toString()` method, and use `[matTooltip]="value"` until there's a reason doing otherwise

bvasilenko (approved):
Looks good overall, just one minor note

src/app/modules/ix-table2/components/ix-table-body/cells/ix-cell-text/ix-cell-text.component.html
AlexKarpov98: image

PR #8867, opened 2023-09-17
NAS-124068 / 24.04 / Refactor SMART tests card in Data Protection

AlexKarpov98:
Testing: check Data protection -> S.M.A.R.T tasks
Screenshot 2023-09-19 at 18 08 54


bvasilenko (approved):
Looks good overall.

One thing I noticed that the order of the blocks on Data Protection page has been modified. At the end of the refactoring, we should remember to ensure to maintain the order of the blocks on Data Protection dashboard.

AlexKarpov98:
> Looks good overall.
> 
> One thing I noticed that the order of the blocks on Data Protection page has been modified. At the end of the refactoring, we should remember to ensure to maintain the order of the blocks on Data Protection dashboard.

yes, I will handle it after merging all cards.
https://ixsystems.atlassian.net/browse/NAS-124185

PR #8871, opened 2023-09-18
NAS-124137 / 23.10 / Start/Restart iSCSI service when iSCSI wizard closes

RehanY147:
Test two things,
1. When share of any type (iSCSI, Smb) is created on the Shares Dashboard page, the tables would start showing two headers when the slide in form was closed successfully. Test that this doesn't happen anymore when you submit the share creation form/wizard.
2. Test that at the end of the iSCSI wizard when it's submitted successfully, UI prompts the user to either start the iSCSI service when it's disabled or restart it if it's already enabled. If restarted, UI must also check for any active iSCSI connections and warn the user that restarting the service might close those sessions and ask for another confirmation. If confirmed, UI should restart the service and close the form.


bvasilenko (approved):
1. Works well
2. Works well, with mock for iSCSI connection

denysbutenko:
backport

PR #8875, opened 2023-09-18
NAS-124150 / 24.04 / Extracting more components from topbar

undsoft:
Testing: the easiest way to test new code is by altering `if` conditions locally.


denysbutenko (approved):
image
image


Looks good.

PR #8880, opened 2023-09-19
NAS-124066 / 24.04 / Refactor Cloud Sync Tasks in Data Protection

AlexKarpov98:
Testing: check Data protection -> Cloud Sync tasks

Screenshot 2023-09-19 at 18 26 00


denysbutenko (changes requested):
Observed unexpected glitches on the state button.

https://github.com/truenas/webui/assets/351613/56b5e84a-3c5c-4270-961c-ab1caf0970d4

When clicked on the `Running` state it shows Logs dialog instead of job progress.

image

denysbutenko:
There also color mismatch between `Running` state on Dashboard and full page table:

image
image

AlexKarpov98:
> There also color mismatch between `Running` state on Dashboard and full page table:
> 
> 
> 
> image
> 
> image
> 
> 

I think running looks better with the primary color rather than with orange, i will check

AlexKarpov98:
`When clicked on the Running state it shows Logs dialog instead of job progress.`
@denysbutenko - I believe that this should be handled as separate ticket.
This button was developed separately. [https://ixsystems.atlassian.net/browse/NAS-124146]

AlexKarpov98:
@denysbutenko - please have a look at it now

src/app/modules/ix-table2/interfaces/table-column.interface.ts
```diff
@@ -26,6 +28,6 @@ export abstract class ColumnComponent<T> {
 }
 
 export type Column<T, C extends ColumnComponent<T>> = {
-  type?: new (matDialog?: MatDialog) => C;
+  type?: new (matDialog?: MatDialog, translate?: TranslateService, dialogService?: DialogService) => C;
```
denysbutenko, line 31: Following this approach does not looks good. For every specific need it is extending common component. Maybe [inject](https://angular.io/api/core/inject) function can help to get necessary services in the state button component and keep this clean?

denysbutenko (approved):
Other looks good.

PR #8881, opened 2023-09-19
NAS-124133 / 24.04 / Ensure iSCSI Wizard creates Initiator Group

denysbutenko:
For testing, go to Shares dashboard and create iSCSI share with empty initiators field. Ensure Initiator Group is created.


denysbutenko:
`Enable Service` dialog was removed, it is implemented in another PR#8871

PR #8885, opened 2023-09-19
NAS-124139 / 24.04 / Refactor `ShellService`

bvasilenko:
**Testing**

Terminal UI component should have no regressions

### DIFF
--- a/src/app/modules/terminal/components/terminal/terminal.component.ts
+++ b/src/app/modules/terminal/components/terminal/terminal.component.ts
@@ -106,7 +106,7 @@ export class TerminalComponent implements OnInit, OnDestroy {
       take(1),
       tap((token) => {
         this.initializeWebShell(token);
-        this.shellService.shellOutput.pipe(untilDestroyed(this)).subscribe(() => {});
+        this.shellService.shellOutput$.pipe(untilDestroyed(this)).subscribe(() => {});
         this.initializeTerminal();
       }),
       untilDestroyed(this),
@@ -114,7 +114,7 @@ export class TerminalComponent implements OnInit, OnDestroy {
   }
 
   ngOnDestroy(): void {
-    this.shellService.closeWebsocketConnection();
+    this.shellService.disconnect();
   }
 
   onResize(): void {
@@ -175,14 +175,9 @@ export class TerminalComponent implements OnInit, OnDestroy {
   }
 
   initializeWebShell(token: string): void {
-    this.shellService.token = token;
+    this.shellService.connect(this.conf.connectionData, token);
 
-    if (this.conf.setShellConnectionData) {
-      this.conf.setShellConnectionData(this.shellService);
-    }
-    this.shellService.connect();
-
-    this.shellService.shellConnected.pipe(untilDestroyed(this)).subscribe((event: ShellConnectedEvent) => {
+    this.shellService.shellConnected$.pipe(untilDestroyed(this)).subscribe((event: ShellConnectedEvent) => {
       this.shellConnected = event.connected;
       this.connectionId = event.id;
       this.updateTerminal();
@@ -196,10 +191,7 @@ export class TerminalComponent implements OnInit, OnDestroy {
   }
 
   reconnect(): void {
-    if (this.conf.setShellConnectionData) {
-      this.conf.setShellConnectionData(this.shellService);
-    }
-    this.shellService.connect();
+    this.shellService.connect(this.conf.connectionData);
   }
 
   onFontSizeChanged(newSize: number): void {
--- a/src/app/modules/terminal/components/terminal/terminal.component.ts
+++ b/src/app/modules/terminal/components/terminal/terminal.component.ts
@@ -52,6 +52,7 @@ export class TerminalComponent implements OnInit, OnDestroy {
 
   private fitAddon: FitAddon;
   private attachAddon: XtermAttachAddon;
+  private token: string;
 
   readonly toolbarTooltip = this.translate.instant(`<b>Copy & Paste</b> <br/>
                   Context menu copy and paste operations are disabled in the Shell. Copy and paste shortcuts for Mac are <i>Command+C</i> and <i>Command+V</i>. For most operating systems, use <i>Ctrl+Insert</i> to copy and <i>Shift+Insert</i> to paste.<br/><br/>
@@ -105,8 +106,8 @@ export class TerminalComponent implements OnInit, OnDestroy {
     this.authService.authToken$.pipe(
       take(1),
       tap((token) => {
-        this.initializeWebShell(token);
-        this.shellService.shellOutput$.pipe(untilDestroyed(this)).subscribe(() => {});
+        this.token = token;
+        this.initializeWebShell();
         this.initializeTerminal();
       }),
       untilDestroyed(this),
@@ -174,8 +175,8 @@ export class TerminalComponent implements OnInit, OnDestroy {
     return true;
   }
 
-  initializeWebShell(token: string): void {
-    this.shellService.connect(this.conf.connectionData, token);
+  initializeWebShell(): void {
+    this.shellService.connect(this.conf.connectionData, this.token);
 
     this.shellService.shellConnected$.pipe(untilDestroyed(this)).subscribe((event: ShellConnectedEvent) => {
       this.shellConnected = event.connected;
@@ -191,7 +192,7 @@ export class TerminalComponent implements OnInit, OnDestroy {
   }
 
   reconnect(): void {
-    this.shellService.connect(this.conf.connectionData);
+    this.shellService.connect(this.conf.connectionData, this.token);
   }
 
   onFontSizeChanged(newSize: number): void {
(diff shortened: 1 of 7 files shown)


src/app/modules/terminal/xterm-attach-addon.ts
```diff
@@ -1,32 +1,34 @@
+import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { Terminal, IDisposable, ITerminalAddon } from 'xterm';
+import { ShellService } from 'app/services/shell.service';
 
 /**
  * This is basically the same as xterm-addon-attach
  * https://github.com/xtermjs/xterm.js/tree/master/addons/xterm-addon-attach
  * but it always sends messages as binary.
  */
+@UntilDestroy()
```
undsoft, line 10: UntilDestroy relies on `ngDestroy`, which probably wouldn't work for this class.
You may need to do something else here.

undsoft (commented):
I've altered template to always show reconnect button. When I press it, I can see `[EINVAL] id: null not allowed\n` in console, but the button appears to be working.

Overall refactoring looks good, thought it's probably not what @RehanY147 envisioned.

bvasilenko:
@RehanY147 would you please clarify is that matches expected outcome of the task. If not, please provide an actual-vs-expected comparison, or anything else to clarify the expected outcome. Thanks!

src/app/modules/terminal/xterm-attach-addon.ts
```diff
@@ -36,27 +38,7 @@ export class XtermAttachAddon implements ITerminalAddon {
   }
 
   private sendBinary(data: string): void {
-    if (this.socket.readyState !== 1) {
-      return;
-    }
     const buffer = this.encoder.encode(data);
```
RehanY147, line 41: I think we can move this line to the service as well. In the interest of having as few pre-reqs to using the service method as possible. Keeping the interface simple.

src/app/services/shell.service.ts
```diff
@@ -19,29 +27,52 @@ export class ShellService {
     command: string;
   };
 
-  private shellCmdOutput: ArrayBuffer;
   shellOutput = new EventEmitter<ArrayBuffer>();
   shellConnected = new EventEmitter<ShellConnectedEvent>();
```
RehanY147, line 31: `shellConnected` and `shellOutput` should be private and exposed with getters that only return the underlying observable with `asObservable()` instead of the subject itself as the `next` method can be called on these `EventEmitter` objects which exposes parts of the API we don't want exposed.

RehanY147 (changes requested):

RehanY147:
> @RehanY147 would you please clarify is that matches expected outcome of the task. If not, please provide an actual-vs-expected comparison, or anything else to clarify the expected outcome. Thanks!

I'm lost. What are we discussing? Please clarify.

RehanY147 (changes requested):
I was hoping to redo all things that were being done in the old shell service. A lot of the logic would still be the same but done better if possible. In a manner that makes it difficult for the developer using this service to do the wrong thing. Make the API/interface safer. 

Remove the "pendingMessages" array. A message is either sent right away or not sent at all. We shouldn't allow the user to type commands in the shell until the shell is connected instead of making the UI too smart and keeping track of what the user is doing while the shell is trying to connect.

Another way of thinking about it is, write a service that only exposes a method to send messages, an observable to consume messages, and a boolean observable for connection status (anything else that might be required but we should prefer not exposing anything publicly unless there's no work around). Then rework the components using this service. Move message pre-processing into the service and any general message post processing as well since this is a tailor made service for this component. Remove things that are not used.

Ideally, the component should have to do nothing more than calling the `send` method on the service. All the things that need to be done before sending a message or before a message is consumed should be done inside the method.

bvasilenko:
@RehanY147 ok, now the requested changes should be addressed

src/app/services/shell.service.ts
```diff
@@ -1,78 +1,104 @@
...
-  connect(): void {
-    this.socket = new WebSocket(
-      (this.window.location.protocol === 'https:' ? 'wss://' : 'ws://')
-        + environment.remote + '/websocket/shell/',
-    );
-    this.socket.onmessage = this.onmessage.bind(this);
-    this.socket.onopen = this.onopen.bind(this);
-    this.socket.onclose = this.onclose.bind(this);
+  connect(connectionData: TerminalConnectionData, token?: string): void {
+    this.connectionData = connectionData;
```
RehanY147, line 42: `token` and `connectionData` don't need to be class members here. We can use the `bind` method to pass the params to `onOpen` and remove the class variables. The calling component can hold the token value of token in a variable to pass it on when reconnecting.

src/app/services/shell.service.ts
```diff
@@ -1,78 +1,104 @@
...
+  private connectionUrl = (this.window.location.protocol === 'https:' ? 'wss://' : 'ws://') + environment.remote + '/websocket/shell/';
+  private isConnected = false;
+
+  private token: string;
+  private connectionData: TerminalConnectionData;
+
+  private shellOutput = new EventEmitter<ArrayBuffer>();
+  private shellConnected = new EventEmitter<ShellConnectedEvent>();
+
+  get shellOutput$(): Observable<ArrayBuffer> {
```
RehanY147, line 28: Even though it was my suggestion initially to make these getters, since we're only going to initialize the `shellOutput` and `shellConnected` EventEmitter instances once, we can use a readonly variable for the observable value instead. As in getters, every time the value is re-evaluated, the `asObservable` method is called again fetching a new observable. Something like 
`readonly shellOutput$ = this.shellOutput.asObservable();`

src/app/services/shell.service.ts
```diff
@@ -81,29 +107,31 @@ export class ShellService {
...
-      this.pendingMessages.push(payload);
+  send(data: string): void {
+    if (this.isConnected) {
+      const buffer = this.encoder.encode(data);
+      this.ws$.next(buffer);
+    }
+  }
+
+  disconnect(): void {
+    if (this.ws$) {
```
RehanY147, line 133: This check should also include `if (this.ws$ && !this.ws$.closed) { ... }`

src/app/modules/terminal/components/terminal/terminal.component.ts
```diff
@@ -106,17 +106,15 @@ export class TerminalComponent implements OnInit, OnDestroy {
       take(1),
       tap((token) => {
         this.initializeWebShell(token);
-        this.shellService.shellOutput.pipe(untilDestroyed(this)).subscribe(() => {});
+        this.shellService.shellOutput$.pipe(untilDestroyed(this)).subscribe(() => {});
```
RehanY147, line 109: Can you elaborate why we need this empty subscription please? If it is for the websocket to stay connected, there is a permanent subscription inside the service to `ws$` which would suffice.

src/app/services/shell.service.ts
```diff
@@ -1,78 +1,104 @@
...
 
-  connect(): void {
-    this.socket = new WebSocket(
-      (this.window.location.protocol === 'https:' ? 'wss://' : 'ws://')
-        + environment.remote + '/websocket/shell/',
-    );
-    this.socket.onmessage = this.onmessage.bind(this);
-    this.socket.onopen = this.onopen.bind(this);
-    this.socket.onclose = this.onclose.bind(this);
+  connect(connectionData: TerminalConnectionData, token?: string): void {
```
RehanY147, line 41: It doesn't seem like the `token` is optional

RehanY147 (changes requested):
This issue might not be caused by this refactor but since we are working on this, I think we should look at it. If I run 'exit' command in the shell and then click reconnect, the shell UI breaks. With each line being shown multiple times when I use it.

bvasilenko:
@RehanY147 the requested changes should be applied

src/app/services/shell.service.ts
```diff
@@ -81,29 +93,31 @@ export class ShellService {
...
-    } else {
-      this.pendingMessages.push(payload);
+  send(data: string): void {
+    if (this.isConnected) {
+      const buffer = this.encoder.encode(data);
+      this.ws$.next(buffer);
+    }
+  }
+
+  disconnect(): void {
```
RehanY147, line 118: `disconnectIfSessionActive` would be a better name for this method.

RehanY147 (approved):
Looks good. Works Ok. Doesn't seem to have any regressions.

src/app/services/shell.service.ts
```diff
@@ -1,78 +1,90 @@
...
-    this.socket.onmessage = this.onmessage.bind(this);
-    this.socket.onopen = this.onopen.bind(this);
-    this.socket.onclose = this.onclose.bind(this);
+  connect(connectionData: TerminalConnectionData, token: string): void {
+    this.disconnect();
+
+    this.ws$ = this.webSocket({
+      url: this.connectionUrl,
+      openObserver: {
+        next: () => this.onOpen(connectionData, token),
```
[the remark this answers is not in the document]
bvasilenko: arrow-function preserves the execution scope as well as `.bind`, so it should be safe to keep it

PR #8886, opened 2023-09-19
NAS-123945 / 24.04 / NAS-124077: Removing of(undefined)

RehanY147:
There is no concrete way to test this. I've removed instances where undefined is thrown. Also, in the websocket service job method, we were throwing the job object. Instead, we are now throwing the job error.

### DIFF
--- a/src/app/pages/apps/store/apps-filter-store.service.ts
+++ b/src/app/pages/apps/store/apps-filter-store.service.ts
@@ -4,7 +4,8 @@ import { ComponentStore } from '@ngrx/component-store';
 import { TranslateService } from '@ngx-translate/core';
 import _ from 'lodash';
 import {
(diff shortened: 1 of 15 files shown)


src/app/pages/apps/store/apps-filter-store.service.ts
```diff
@@ -4,7 +4,8 @@ import { ComponentStore } from '@ngrx/component-store';
 import { TranslateService } from '@ngx-translate/core';
 import _ from 'lodash';
 import {
-  Observable, combineLatest, map, of,
+  EMPTY,
+  Observable, combineLatest, map,
```
dszidi, line 8: These should also be broken out into separate lines, shouldn't it?

dszidi (approved):
Code review only:
Other than the little code style comment it looks good to me.

PR #8892, opened 2023-09-20
NAS-124152 / 24.04 / Show fallback image when product image loading has failure

denysbutenko:
Changes:

- When product image returns 404 not found error - show fallback image instead of broken one
- Ensure there is single slash in the path: `/assets/images//servers/F100.png`

For testing, use system from the ticket.


bvasilenko (approved):
LGTM

image

PR #8899, opened 2023-09-20
NAS-123287 / 23.10 / ES60G2 

dszidi:
Putting this up so progress can be tracked. Treat as a draft until assets are received

To test:

Since this is just a shelf, the only place this will show up is in enclosure management.
You can test against a real unit or alternatively use mock utility via the following command...

`yarn ui me -e -m m40 -s es60g2`


undsoft (approved):
Looks good via mocks, let's see that 13.0 PR.

PR #8903, opened 2023-09-21
NAS-124136: Fix Sudo Dialog appears after Replication Task Created

denysbutenko:
For testing check the [ticket](https://ixsystems.atlassian.net/browse/NAS-124136)

### DIFF
--- a/src/app/pages/data-protection/replication/replication-wizard/steps/replication-what-and-where/replication-what-and-where.component.ts
+++ b/src/app/pages/data-protection/replication/replication-wizard/steps/replication-what-and-where/replication-what-and-where.component.ts
@@ -8,7 +8,7 @@ import { FormBuilder } from '@ngneat/reactive-forms';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { TranslateService } from '@ngx-translate/core';
 import {
-  debounceTime, map, merge, Observable, of,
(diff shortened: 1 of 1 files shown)


RehanY147 (commented):
Waiting for testing instructions

denysbutenko:
Good morning @RehanY147, the steps for testing described in the ticket.

RehanY147 (changes requested):
Ideally, we should be only getting this dialog for one task only once. I'm getting it twice. Once when I make changes to the input fields on the form. Second after the form has been closed.

denysbutenko:
> Ideally, we should be only getting this dialog for one task only once. I'm getting it twice. Once when I make changes to the input fields on the form. Second after the form has been closed.

Good catch. Close/Cancel also considered as an action now.

RehanY147 (changes requested):
We're not getting the "Task Started" info dialog. Because the slidein form component closes, all subscriptions are dumped.

denysbutenko:
> We're not getting the "Task Started" info dialog. Because the slidein form component closes, all subscriptions are dumped.

That's true although it is not related to this ticket. The other thing I noticed is that the wizard lack of a success notification after task is created when the replication form has a success notification. It was synced between wizard and form. There is a ticket created for the issue: https://ixsystems.atlassian.net/browse/NAS-124413

RehanY147 (approved):

PR #8908, opened 2023-09-21
NAS-123257 / 24.04 / Consider allowing typing into enum fields (dropdowns)

AlexKarpov98:
Testing:
I asked @stavros-k to setup an app for this one, it was set on m40 machine.
`Radarr App from community train`

We should be able to select value from the autocomplete, if it's custom - we reset it.
There may be MANY autocomplete suggestions (100000+)
So I show only 50 per autocomplete session, in order to make app work smoothly. 
(if change input, it updates autocomplete values limited to 50 items)

Result: 

https://github.com/truenas/webui/assets/22980553/ed3313e6-ed25-4740-8281-c31ec4f7b2ad


bvasilenko (approved):
Enum works good to me

PR #8917, opened 2023-09-22
NAS-123861 / 24.04 / Added last_run column

RehanY147:
Rework of https://github.com/truenas/webui/pull/8859 because of middleware and UI changes


undsoft (approved):
Looks good.

PR #8925, opened 2023-09-24
NAS-124161 / 24.04 / During pool creation on Cobia nightlies, if dRAID is selected as a layout other options are not available

AlexKarpov98:
Testing:
see ticket, there should be no errors in console as well.

### DIFF
--- a/src/app/pages/storage/modules/pool-manager/components/pool-manager-wizard/pool-manager-wizard.component.html
+++ b/src/app/pages/storage/modules/pool-manager/components/pool-manager-wizard/pool-manager-wizard.component.html
@@ -99,7 +99,7 @@
       </mat-step>
 
       <mat-step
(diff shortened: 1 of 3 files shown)


src/app/pages/storage/modules/pool-manager/components/pool-manager-wizard/pool-manager-wizard.component.html
```diff
@@ -99,7 +99,7 @@
       </mat-step>
 
       <mat-step
-        *ngIf="!(usesDraidLayout$ | async) && !alreadyHasSpare"
+        *ngIf="!(usesDraidLayout$ | async) && !alreadyHasSpare && state.topology[PoolCreationWizardStep.Spare]"
```
undsoft, line 102: I think it's time to extract all of this into `hasSpareStep$` or something similar.

undsoft (approved):
Looks good, thanks.

PR #8927, opened 2023-09-24
NAS-124123 / 24.04 / Removing vdevs from the manual disk selection dialog doesn't update the automated dropdowns

AlexKarpov98:
Testing: see ticket


dszidi (approved):
LGTM

PR #8928, opened 2023-09-24
NAS-124183 / 23.10 / Unsubscribe from 'reporting.realtime' after leaving the dashboard

bvasilenko:
**Testing**

Please see ticket


RehanY147 (changes requested):
We can make the service module scoped and add the `untilDestroyed` pipe in the subscriptions on the service.

bvasilenko:
@RehanY147 I suggest to keep the code as it is, for the sake of simplicity

src/app/pages/dashboard/components/dashboard/dashboard.component.ts
```diff
@@ -269,6 +272,10 @@ export class DashboardComponent implements AfterViewInit, OnDestroy {
     this.availableWidgets = conf;
   }
 
+  private subscribeToResourceUsageUpdates(): void {
+    this.resourcesUsageStore$.subscribeToResourceUsageUpdates().pipe(untilDestroyed(this)).subscribe();
```
RehanY147, line 276: I think the previous method name still works and is better than this one.

RehanY147 (approved):

PR #8935, opened 2023-09-25
NAS-124308 / 24.04 / Extracting network checking code from topbar

undsoft:
Testing: use machine from CI. Go to Networking and edit one of the interfaces (for example by changing description). Test how topbar icon works in when testing/saving changes/reloading page not on Network page.


RehanY147 (changes requested):
On `master` as soon as you click "Test Changes" in the table prompt, the icon is hidden in the top bar.

undsoft:
Looked into how master work. I like new way better. For example, master hides icon on other pages too, but shows it if you reload the page. New behaviour just consistently shows an icon when there are network changes.

PR #8940, opened 2023-09-26
NAS-124015 / 24.04 / Suppress an error dialog

bvasilenko:
**Summary**

You may use replication steps from the ticket. However, an easier way to replicate the error in your WebUI is described below
- at the **pool.dataset.get_quota** API call temporary replace the `dataset.id` with a arbitrary name of non-existing dataset.
- Expand any dataset. Then hit F5 and you'll be able to see the error:

image

From UX point of view after a navigation to a **Datasets** page should not automatically trigger an error pop-up, because user is just viewing, not executing an edit or delete operation, thus user is not expecting any pop-ups.

**Testing**

To verify, at the **pool.dataset.get_quota** API call temporary replace the `dataset.id` with a arbitrary name of non-existing dataset.

1. Go to Datasets
2. Expand any dataset
3. Hit F5

**Expected result:** you should see no error pop-up.

### DIFF
--- a/src/app/pages/datasets/components/dataset-capacity-management-card/dataset-capacity-management-card.component.ts
+++ b/src/app/pages/datasets/components/dataset-capacity-management-card/dataset-capacity-management-card.component.ts
@@ -102,7 +102,7 @@ export class DatasetCapacityManagementCardComponent implements OnChanges, OnInit
       },
       error: (error: WebsocketError) => {
         this.isLoadingQuotas = false;
-        this.dialogService.error(this.errorHandler.parseWsError(error));
+        this.errorHandler.handleError(error);
         this.cdr.markForCheck();
       },
     });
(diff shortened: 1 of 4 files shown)


undsoft (changes requested):
So, this silences a lot of errors in a lot of scenarios, not just when user has incorrect dataset id in url.
Can we have a narrower solution that would check for datasets presence in `listenForRouteChanges`? Ideally, we would select first dataset in the tree like we normally do.

bvasilenko:
ok, @undsoft now the requested changes should be addressed

undsoft (approved):
It's fine, but I'd prefer to do a proper redirect, so that url doesn't say `/datasets/non-existent` while pointing to dataset `A`. I'd rather we just redirect to `/datasets/A`.

bvasilenko:
backport

PR #8943, opened 2023-09-26
NAS-124300: Refactor CloudsyncListComponent to ix-tables2

AlexKarpov98:
Testing: check CloudsyncListComponent work


RehanY147 (changes requested):
Left and right margins are different for buttons in the details row. Buttons should be wider.
The attribute names should be highlighted i.e., bold
For tests, we should also test that the actual call is made. Just testing until the confirmation dialog is shown is not enough.

AlexKarpov98:
@RehanY147 
`Left and right margins are different for buttons in the details row. Buttons should be wider.`
- That's the new style for ix-table2, it should be fine.
Regarding tests, I can't test API calls for some reason, investigating, give it a look as well

this test for example:
```
  it('shows confirmation dialog when Run Now button is pressed', async () => {
    jest.spyOn(spectator.inject(DialogService), 'confirm');
    await table.clickToggle(0);

    const runNowButton = await loader.getHarness(MatButtonHarness.with({ text: 'Run Now' }));
    await runNowButton.click();

    expect(spectator.inject(DialogService).confirm).toHaveBeenCalledWith({
      title: 'Run Now',
      message: 'Run this Cloud Sync now?',
      hideCheckbox: true,
    });
  });
  ```

RehanY147 (changes requested):
Adding `expect(spectator.inject(WebSocketService).job).toHaveBeenCalledWith('cloudsync.sync', [1]);` at the end of that test returns true. The test passes.

RehanY147 (approved):

PR #8944, opened 2023-09-26
NAS-123801 / 23.10 / Better error tracking for 'ix-list'

RehanY147:
The issue was that for an `ix-list` control which is of type `string[]`, when middleware complained about a validation error, it didn't have a way of specifying which specific control had the erroneous value. Furthermore, there was no way to track validity of the `FormArray` instance attached to the `ix-list` component. This PR adds `ix-errors` `ix-list` component and requires that the `FormArray` instance is passed as a property input for the `ix-list` component. That way, when the validity of the `FormArray` changes, we can show the error under the field name.

Follow ticket steps in order to reproduce errors that showcase the new behavior.


RehanY147:
These changes will be made in 24.04. `NAS-123801-alt` will handle changes for cobia via https://github.com/truenas/webui/pull/8962

PR #8950, opened 2023-09-27
NAS-124044 / 23.10 / Adapt to 0 values

RehanY147:
If temp values are 0, that means temperature values are not available for the CPU and we should show the user `'N/A'`


denysbutenko (changes requested):
Getting `Hottest: -Infinity°C (All Cores)`

denysbutenko (approved):
LGTM

PR #8951, opened 2023-09-27
NAS-124314 / 24.04 / Adjusting to middleware changes for UI temps

RehanY147:
In the new middleware changes, the `temperature` field has been removed. Only temperatures in Celcius are provided now. Furthermore, if `temperature_celcius` field is null, that means temperatures are not available on the system and we have to show `'N/A'` to the user. Also, fixes the bug where if the temps were 0, the user would just be shown the skeleton loading indicator. Now, if temps are 0, user will be shown `'N/A'`. Otherwise, they will be shown the temperature values. If the temperature values are still loading, only then they should be shown the skeleton loader.


AlexKarpov98 (changes requested):
There is now an issue with - Infinity.
Previously we had loader instead.

Screenshot 2023-10-03 at 17 50 13

AlexKarpov98 (approved):
Now it works fine, thanks.

PR #8952, opened 2023-09-27
NAS-124077 / 23.10 / Encoding text before sending

RehanY147:


bvasilenko (approved):
I've tried various hotkeys in bash, vim, etc
Works well

PR #8954, opened 2023-09-27
NAS-124335 / 23.10 / Improvements for mobile layout for Installed Apps page

RehanY147:
See ticket descriptions for things to check out


dszidi (approved):
Works as per ticket description. Would be better if the table didn't introduce a scroll bar just for a single row

image

PR #8956, opened 2023-09-27
NAS-124232 / 23.10 / Added query param to fetch `display_available` attribute

RehanY147:
Wait for https://github.com/truenas/middleware/pull/12200 to be ported to `cobia` before testing. This removes the need for UI to make decisions about whether or not to show the display button for running vms. Now, middleware will tell us if a display is available or not.

### DIFF
--- a/src/app/pages/vm/vm-list/vm-list.component.ts
+++ b/src/app/pages/vm/vm-list/vm-list.component.ts
@@ -13,6 +13,7 @@ import helptext from 'app/helptext/vm/vm-list';
 import wizardHelptext from 'app/helptext/vm/vm-wizard/vm-wizard';
 import { ApiParams } from 'app/interfaces/api-directory.interface';
 import { EmptyConfig } from 'app/interfaces/empty-config.interface';
+import { QueryParams } from 'app/interfaces/query-api.interface';
 import {
   VirtualizationDetails,
   VirtualMachine, VirtualMachineUpdate,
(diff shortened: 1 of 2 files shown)


src/app/pages/vm/vm-list/vm-list.component.ts
```diff
@@ -49,6 +50,11 @@ const noMemoryError = 'ENOMEM';
 export class VmListComponent implements EntityTableConfig<VirtualMachineRow> {
   title = this.translate.instant('Virtual Machines');
   queryCall = 'vm.query' as const;
+  queryCallOption: QueryParams<VirtualMachine, { extra: { retrieve_display_available_info: boolean } }> = [[], {
+    extra: {
+      retrieve_display_available_info: true,
```
yocalebo, line 55: Hold off on this change. I've asked the middleware PR to remove the extra query params as I don't believe it's helpful. We should just show the `display_available` key by default.

yocalebo (changes requested):

src/app/pages/vm/vm-list/vm-list.component.ts
yocalebo: Okay, you don't need to specify these extra params in the query. The `display_available` key will be there by default now.

denysbutenko:
@RehanY147 do we have Cobia nightly builds for testing or perhaps there is a system available with the changes?

RehanY147:
@denysbutenko I'm not sure. You'd have to ask Evgeny

denysbutenko (approved):
No luck to find a system with a build includes the changes and VM support for testing. Let's move on.

undsoft (changes requested):
So what about other usage of `checkDisplay` where it is used to decide whether port number is shown?

undsoft (approved):

PR #8958, opened 2023-09-27
NAS-124339 / 24.04 / Sorting on installed applications screen

AlexKarpov98:
Testing:
Check installed apps page, you should be able to sort by APPLICATION, STATUS, UPDATE AVAILABLE (if there are some apps with update available)

Screenshot 2023-09-27 at 18 16 22


bvasilenko (approved):
Works good to me

PR #8962, opened 2023-09-28
NAS-123801 / 23.10 / Showing error in dialog if form not found

RehanY147:
Showing error in a dialog. Check out the ticket description for details.

### DIFF
--- a/src/app/modules/ix-forms/services/form-error-handler.service.ts
+++ b/src/app/modules/ix-forms/services/form-error-handler.service.ts
@@ -66,13 +66,16 @@ export class FormErrorHandlerService {
           .find((controlOfArray) => isExactMatch(errorMessage, controlOfArray.value));
       }
 
-      control.setErrors({
-        manualValidateError: true,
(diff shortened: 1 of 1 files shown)


src/app/modules/ix-forms/services/form-error-handler.service.ts
```diff
@@ -66,13 +66,16 @@ export class FormErrorHandlerService {
...
       }
 
-      control.setErrors({
-        manualValidateError: true,
-        manualValidateErrorMsg: errorMessage,
-        ixManualValidateError: { message: errorMessage },
-      });
-
-      control.markAsTouched();
+      if (!control) {
```
undsoft, line 69: Similar to line 55?

undsoft (approved):
Looks good.

src/app/modules/ix-forms/services/form-error-handler.service.ts
RehanY147: This follows another check for specific array item where we compare value instead of looking for the field by name in the form group.

PR #8969, opened 2023-09-29
NAS-123907 / 24.04 / Do not show New & Updated in search results

AlexKarpov98:
Testing:
Go to Available Apps section --> Search for any app and see that `New & Updated Apps`  section is not there while searching.

Screenshot 2023-09-29 at 11 07 56


dszidi (approved):
LGTM

image

PR #8970, opened 2023-09-29
NAS-124222 / 24.04 / Login page error mesage Username is required

AlexKarpov98:
Testing: 
See that on login page you don't see `Validators.required` error if **username** or **password** field is empty.
Form shall stay disabled if value is empty tho.
Screenshot 2023-09-29 at 11 18 58

### DIFF
--- a/src/app/views/sessions/signin/signin-form/signin-form.component.html
+++ b/src/app/views/sessions/signin/signin-form/signin-form.component.html
@@ -25,7 +25,7 @@
       type="button"
       color="primary"
       ixTest="log-in"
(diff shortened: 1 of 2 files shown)


undsoft (changes requested):
The problem with this approach is that if you have autocomplete for form values you may end up with form being filled in, but login button disabled.

I'm okay with not having those checks on Login button. I think we can also keep the required validation on OTP field.
Снимок экрана 2023-10-01 в 11 56 34

undsoft (approved):
👍🏻

PR #8971, opened 2023-09-29
NAS-124250 / 24.04 / Details panel does not always properly populate when clicking apps

AlexKarpov98:
I was not able to reproduce the bug, but I did fix based on an error provided.


undsoft (approved):
👍🏻

PR #8973, opened 2023-09-29
NAS-123925 / 24.04 / Added JobProgressDialogComponent

RehanY147:
This creates a new way of showing job progress and subcribing to main job events i.e., Success, Abort, Failure and Progress updates.


src/app/modules/common/dialog/job-progress/job-progress-dialog.component.ts
```diff
@@ -0,0 +1,167 @@
...
+    onFailure?: (job: Job) => void;
+    onAbort?: (job: Job) => void;
+    onProgress: (progress: JobProgress) => void;
+  };
+  config: {
+    title: string;
+    description: string;
+  };
+  flags: {
+    showRealtimeLogs: boolean;
```
undsoft, line 23: realtime logs are only used in one form, I'd consider creating a separate component for this.

src/app/modules/common/dialog/job-progress/job-progress-dialog.component.ts
```diff
@@ -0,0 +1,167 @@
...
+    autoCloseOnSuccess: boolean;
```
undsoft, line 24: I think it's a good idea to decide this and have it work one way.

src/app/modules/common/dialog/job-progress/job-progress-dialog.component.ts
```diff
@@ -0,0 +1,167 @@
...
+  };
+}
+
+@UntilDestroy()
+@Component({
+  templateUrl: './job-progress-dialog.component.html',
+  styleUrls: ['./job-progress-dialog.component.scss'],
```
undsoft, line 31: onpush

src/app/services/dialog.service.ts
```diff
@@ -1,20 +1,85 @@
...
+import { Job, JobProgress } from 'app/interfaces/job.interface';
 import { ConfirmDialogComponent } from 'app/modules/common/dialog/confirm-dialog/confirm-dialog.component';
 import { ErrorDialogComponent } from 'app/modules/common/dialog/error-dialog/error-dialog.component';
 import { FullScreenDialogComponent } from 'app/modules/common/dialog/full-screen-dialog/full-screen-dialog.component';
 import { GeneralDialogComponent, GeneralDialogConfig } from 'app/modules/common/dialog/general-dialog/general-dialog.component';
 import { InfoDialogComponent } from 'app/modules/common/dialog/info-dialog/info-dialog.component';
+import { JobProgressDialogComponent, JobProgressDialogConfig } from 'app/modules/common/dialog/job-progress/job-progress-dialog.component';
 import { MultiErrorDialogComponent } from 'app/modules/common/dialog/multi-error-dialog/multi-error-dialog.component';
 
+export class JobProgressDialog {
```
undsoft, line 21: Needs to go in some other file.

src/app/services/dialog.service.ts
```diff
@@ -1,20 +1,85 @@
...
+  private readonly _onProgress$ = new Subject<JobProgress>();
+
+  readonly afterSuccess$ = this.onSuccess$.asObservable();
+  readonly afterAbort$ = this.onAbort$.asObservable();
+  readonly afterFailure$ = this.onFailure$.asObservable();
+  readonly onProgress$ = this._onProgress$.asObservable();
+
+  private _matDialogRef: MatDialogRef<JobProgressDialogComponent>;
+
+  set matDialogRef(ref: MatDialogRef<JobProgressDialogComponent>) {
```
undsoft, line 34: Getter/setter doesn't seem to provide value.

src/app/services/dialog.service.ts
```diff
@@ -1,20 +1,85 @@
...
 import { GeneralDialogComponent, GeneralDialogConfig } from 'app/modules/common/dialog/general-dialog/general-dialog.component';
 import { InfoDialogComponent } from 'app/modules/common/dialog/info-dialog/info-dialog.component';
+import { JobProgressDialogComponent, JobProgressDialogConfig } from 'app/modules/common/dialog/job-progress/job-progress-dialog.component';
 import { MultiErrorDialogComponent } from 'app/modules/common/dialog/multi-error-dialog/multi-error-dialog.component';
 
+export class JobProgressDialog {
+  private readonly onSuccess$ = new Subject<Job>();
+  private readonly onAbort$ = new Subject<Job>();
+  private readonly onFailure$ = new Subject<Job>();
+  private readonly _onProgress$ = new Subject<JobProgress>();
```
undsoft, line 25: leading underscore doesn't match our styleguide

src/app/services/dialog.service.ts
```diff
@@ -1,20 +1,85 @@
...
+
+  readonly afterSuccess$ = this.onSuccess$.asObservable();
+  readonly afterAbort$ = this.onAbort$.asObservable();
+  readonly afterFailure$ = this.onFailure$.asObservable();
+  readonly onProgress$ = this._onProgress$.asObservable();
```
undsoft, line 30: What's the difference between using `asObservable` vs returning a subject, which is also an observable?

undsoft (changes requested):
I think we need to decide whether we want to handle errors directly in jobProgress dialog.
Though it may be less customizable, I think it may be a good idea – most of the time you want to delegate everything about showing job status to this dialog.

src/app/services/dialog.service.ts
RehanY147: Subjects are allowed to emit values. The intention is to only expose buttons that are needed to the class's user. They don't need to do emits. So why expose that.

RehanY147:
@undsoft For real time logs, I think we can get away with one flag instead of a whole component. Also, in regards to handling errors, sometimes the user might not want to do anything with an error. Such as when running data protection tasks on the Data Protection dashboard.

src/app/classes/job-progress-dialog-ref.class.ts
```diff
@@ -0,0 +1,29 @@
...
+import { JobProgressDialogComponent } from 'app/modules/common/dialog/job-progress/job-progress-dialog.component';
+
+export class JobProgressDialogRef {
+  constructor(private readonly jobProgressDialog: JobProgressDialog) { }
+
+  getDialogRef(): MatDialogRef<JobProgressDialogComponent> {
+    return this.jobProgressDialog.matDialogRef;
+  }
+
+  onSuccess(): Observable<Job> {
```
undsoft, line 14: I think we may also need something called `onClose`.

src/app/services/dialog.service.ts
```diff
@@ -120,4 +125,32 @@ export class DialogService {
...
     }
   }
+
+  jobProgress<M extends ApiJobMethod>(
+    { job$, title, description }: {
+      job$: Observable<Job<ApiJobDirectory[M]['response']>>;
+      title: string;
+      description: string;
+    },
+    showRealtimeLogs: boolean,
```
undsoft, line 135: `showRealtimeLogs` then needs to be part of config object.

src/app/services/dialog.service.ts
```diff
@@ -120,4 +125,32 @@ export class DialogService {
       openDialog.close();
     }
   }
+
+  jobProgress<M extends ApiJobMethod>(
+    { job$, title, description }: {
+      job$: Observable<Job<ApiJobDirectory[M]['response']>>;
+      title: string;
+      description: string;
```
undsoft, line 133: Title and description should be optional.

undsoft (changes requested):
I still think we should handle errors in job dialog. You as a developer delegate showing job status to a dialog when you open it. Error to me is also a type of a job status.
If you don't care about job failing, then to me you should use something like `startJob` without the dialog.
Currently if you mess with a job by altering call params, job dialog will keep showing progress indicator.

RehanY147:
@undsoft Test on the team m40 and open the `DataProtectionDashboard` and a request is made the fails so far so you can see what happens when job fails.

src/app/classes/job-progress-dialog-ref.class.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  onFailure(): Observable<Job> {
+    return this.jobProgressDialog.afterFailure$;
+  }
+
+  onProgressUpdate(): Observable<JobProgress> {
+    return this.jobProgressDialog.onProgress$;
+  }
+
+  onClose(): Observable<unknown> {
```
undsoft, line 30: `Observable<void>`

undsoft (approved):
I think it looks pretty good now.

PR #8975, opened 2023-09-29
NAS-124255 / 24.04 / Miss-match between step name and list item on the Pool Creation Wizard

AlexKarpov98:
Check label:
Screenshot 2023-09-29 at 12 59 07


dszidi (approved):
LGTM

PR #8976, opened 2023-09-29
NAS-124397 / 24.04 / Notes card in apps UI does not show full text

undsoft:
Testing: use inspector to put a long unbreakable string in App -> Notes.


denysbutenko (approved):
image

👍🏻

PR #8977, opened 2023-09-30
NAS-124335 / 23.10 / Improved styles for movile view for installed apps

RehanY147:
Check out the ticket description for details


undsoft (approved):
Looks good, mobile view is easier to use now.

PR #8978, opened 2023-09-30
NAS-124360 / 24.04 / Fixed dataset form

RehanY147:
Fixed method


src/app/pages/datasets/components/dataset-form/dataset-form.component.ts
```diff
@@ -111,22 +111,22 @@ export class DatasetFormComponent implements OnInit {
     this.isLoading = true;
     this.cdr.markForCheck();
 
-    this.datasetFormService.ensurePathLimits(this.slideInData.datasetId).pipe(
+    this.datasetFormService.isPathLengthAndDepthSafe(this.slideInData.datasetId).pipe(
```
undsoft, line 114: Method name now implies that there will be no side effects like closing a slide-in. You either need to use a different name or move side-effect outside of the method.

bvasilenko (approved):
Works good for me

Oct-03-2023 08-51-00

PR #8979, opened 2023-09-30
NAS-123925 / 24.04 / Parsing job events with operators

RehanY147:
This PR takes away the job of parsing events from the job progress dialog. Instead, we will have operators to get those checks for you. Another argument for this way of doing things is that the job progress dialog can be closed and we can still continue to process those callbacks.


undsoft (approved):
I like the operators themselves and there is a place for them in the code, but I don't think we should subscribe to them instead of the dialog.
My main reason is transferring control to a job dialog feels more natural, it's in line with what's happening from the user perspective and matches behaviour of other dialogs.
Just opening dialog without callbacks feels weird.

PR #8985, opened 2023-10-02
NAS-124362 / 24.04 / Simplify data loading boilerplate for ix-table2

bvasilenko:
**Testing**

Check the described functionality in the table2 on user's page, 
Also please check other pages, where the table2 is used.


undsoft (changes requested):
I think there are conceptual issues with this approach.
1. I don't think `setEmptyType` or `isLoading$` belongs to `ArrayDataProvider`. `ArrayDataProvider` doesn't really know about loading rows.
2. You've integrated empty directive inside of `ix-table2`. I'd prefer to keep it out and to simplify the directive (or create a new one). Not every table will need a loading indication or a message for empty rows, so it's not really part of the core package.
3. It may be a good idea to use some other page as an example. User list takes loading states and errors from the store, which is unusual when compared to most other pages. Most other pages just load something from `ws.call`.

I would try to implement a new async data provider to have loading statuses in it and I would try to create new/simplify the ix-empty directive.

undsoft (changes requested):
Requesting changes as per our discussion. Let's experiment with creating a new `AsyncDataProvider` alongside with `ArrayDataProvider`. They both will derive from common interface, but AsyncDataProvider will know about loading data and loading status. It will take in an observable and spit out information necessary to render empty state.

Empty state should be a separate directive or component and should not be a part of ix-table core, because not every table will need it.

bvasilenko:
@undsoft the requested changes should be addressed now

src/app/modules/ix-table2/async-data-provider.ts
```diff
@@ -0,0 +1,41 @@
+import { BehaviorSubject, Observable, Subscription } from 'rxjs';
+import { EmptyType } from 'app/enums/empty-type.enum';
+import { ArrayDataProvider } from 'app/modules/ix-table2/array-data-provider';
+
+export class AsyncDataProvider<T> extends ArrayDataProvider<T> {
+  private _subscription = new Subscription();
```
undsoft, line 6: underscore in front of the property name is not recommended.

src/app/modules/ix-table2/async-data-provider.ts
```diff
@@ -0,0 +1,41 @@
...
+  private _emptyType$ = new BehaviorSubject<EmptyType>(EmptyType.Loading);
+
+  get emptyType$(): Observable<EmptyType> {
```
undsoft, line 9: BehaviourSubject is already an observable. Can you use it directly and just mark property as `readonly`?

src/app/modules/ix-table2/async-data-provider.ts
```diff
@@ -0,0 +1,41 @@
+import { BehaviorSubject, Observable, Subscription } from 'rxjs';
+import { EmptyType } from 'app/enums/empty-type.enum';
+import { ArrayDataProvider } from 'app/modules/ix-table2/array-data-provider';
+
+export class AsyncDataProvider<T> extends ArrayDataProvider<T> {
+  private _subscription = new Subscription();
+  private _emptyType$ = new BehaviorSubject<EmptyType>(EmptyType.Loading);
```
undsoft, line 7: Should we also expose `isLoading$` and `error$`? What do you think.

undsoft (changes requested):
Direction is correct overall.

undsoft (changes requested):
I think I forgot to write all my comments last time. I think we shouldn't do:
```
this.dataProvider.emptyType$.pipe(untilDestroyed(this)).subscribe(() => this.cdr.markForCheck());
```
This feels like a workaround. I'd rather table would handle change detection on its own. If we need to update `TableProvider` interface to use observables for rows, we can do that.
Array data provider can still have `setRows(rows)` as it currently has.

bvasilenko:
@undsoft the last remark should be fixed now, and this PR now includes other pages

undsoft (approved):
The only issue that I see is that with new approach calling `refresh()` will have a side effect of executing everything in the original observable chain.
The way its currently written it may not be immediately obvious to the developer that calling `refresh` in say `getCronJobs` will execute `tap((cronjobs) => this.cronjobs = cronjobs)` in a different method.

But it's still an improvement over what we previously had and we can fix this in other tickets. Thanks.

bvasilenko:
> The way its currently written it may not be immediately obvious to the developer that calling refresh in say getCronJobs will execute tap((cronjobs) => this.cronjobs = cronjobs) in a different method.

@undsoft Ok, I will address this request in my next table2 ticket

PR #8991, opened 2023-10-02
NAS-124305 / 24.04 / Added IxChainedSlideInComponent for chaining multiple slide-in forms

RehanY147:
Check out the code and the Edit/Add form on Data Protection -> Cloud Sync Tasks -> Add/Edit
On the slide-in form, under the Credentials dropdown, click the "Add New" button and another slide-in form should open. Check for any bugs. Try submitting forms and closing them without form submission. Try breaking the UI.


dszidi (changes requested):
I tested adding a new credential and it fails to submit and exit the subform.

Steps to reproduce:
- Go to Data Protection Dashboard
- Add new Cloud Sync Task
- Trigger subform by selecting "Add New" in the Credential field in the Remote section
- Fill out the sub form with the values listed below and hit save.

```
Provider: FTP
Name: FTP
Host: put in any IP
Port: 21
Username: jdoe
Password: abcd1234
```

Expected result:
Form is saved and subform should close getting user back to the original cloud credential form.

Actual result:
The loading animation goes on infinite loop and the sub form fails to close.
If I hit the close button ("x" on top right) the form will close but selecting the "Add New" option in the Credential form no longer triggers the subform. I have to reload the page to get subform functionality to work again.

dszidi (approved):
Seems to be working fine now. I can create a new credential and it seems to save it just fine. 
Code LGTM and I didn't notice any regressions

image

undsoft (changes requested):
I'd like to see a version where we have visual indication underneath that there is a second (third, etc) slide in there.
Also, I'm assuming we are not going to have two slide-in services in the future and we'll replace original service at some point.

RehanY147:
@undsoft Here's with added border and some shadow

https://github.com/truenas/webui/assets/24204846/9b6cbfcc-3260-4485-963a-f7253e849436

undsoft:
@dszidi @RehanY147 What about something like this?
Снимок экрана 2023-10-23 в 13 20 25

css for this is twice less on `right` and `padding-top` with `filter: brightness(0.75)` and `filter: brightness(0.5)`. Borders are only applied to the left side with `--lines` as color.

dszidi:
I can appreciate the effect trying to be attained but IMO visually this looks odd. The entire material paradigm is based around the paper metaphor and the way the slide in forms are being stacked here adds more depth than the context suggests. With material it's normally one sheet of paper slightly raised over another. The proposed design suggests an endless stack fading into the horizon. 

With material design, the main approaches that seem to be used in these types of situations are to either 

A) extend the subform out like we do with our side nav menu + submenus
B) have the new form completely hide the original form and provide a back button. The back button is a well understood action that exists on multiple platforms and is enough to signal to the user that there is a previous context that they can go to.

I would tend to lean on option B since A will become messy with more than one sub form.
My suggestion is to provide a back button next to the form title and perhaps even rename the save button to something more descriptive like "Save and return to <foo> form"

undsoft (changes requested):
Making some of the changes will be problematic. I think we may be able to get back button in the header to work, altering Save button is going to be messy.
@RehanY147 Let's create a quick hack to see how back button may look like. If we like it, then we can code a proper solution.

undsoft (changes requested):
As discussed, let's see an option for:
`<- Header`
on secondary slide-ins.
You can just hard code this for now.

undsoft:
Yeah, I suppose this could work. The icon needs to be something different and we need to remove border from the slide-in, but otherwise looks okay.
@dszidi Thoughts?

PR #8998, opened 2023-10-03
NAS-124340 / 24.04 / NAS-124340 Stopped cron jobs shown as tasks

bvasilenko:
**Testing**

Add mocks (see ticket) and ensure that the tooltip is displayed correctly


undsoft (approved):
Looks good.

PR #8999, opened 2023-10-03
NAS-124298 / 24.04 / Add sudo enabled dialog to replication form and unit tests

denysbutenko:
For testing, check `Sudo Enabled` dialog on `Replication Form` and `Replication Wizard`


AlexKarpov98 (approved):
👍

PR #9013, opened 2023-10-04
NAS-124413 / 24.04 / Waiting for dialog before destroying the component

RehanY147:


undsoft (approved):
Looks good, but please extract code into a method.

PR #9020, opened 2023-10-05
NAS-124447 / 24.04 / Refactor Uptime on Dashboard

denysbutenko:
For testing, check uptime on dashboard. ensure uptime and datetime updating every minute on the sys info widget.


undsoft (approved):
Well, it works, but the way it updates sometimes results in a minor discrepancy between system time and webui time.
E.g. I may have 10:36 on my computer, but it's going to say 10:35 in webui.

denysbutenko:
> Well, it works, but the way it updates sometimes results in a minor discrepancy between system time and webui time. E.g. I may have 10:36 on my computer, but it's going to say 10:35 in webui.

Yeah, it may happen due to internal timer that emits value each 60 seconds. Looks like it can be avoided when updating interval being reduced.

PR #9021, opened 2023-10-06
NAS-113941 / 24.04 / Sync email validation rules WebUI/Middleware

bvasilenko:


denysbutenko (changes requested):
Let's add unit tests for new validator.

denysbutenko (commented):
However, it works well while basic checking. Consider of replacing builtin validator with new one for other components.

denysbutenko:
As final point would be good to prevent using built-in validator in the future. Take a look on eslint config, specifically at `no-restricted-syntax` that allows add linter rule for this case.

bvasilenko:
Ok, requested changes should be addressed now

src/app/modules/ix-forms/validators/email-validation/email-validation.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+
+describe('emailValidator', () => {
+  const validate = emailValidator();
+
+  it('should not return an error when value is valid email', () => {
+    expect(validate(new FormControl('test@mail.com'))).toBeNull();
+    expect(validate(new FormControl('a@a.a'))).toBeNull();
+  });
+
+  it('should return an error when value is valid email', () => {
```
undsoft, line 12: when value is invalid email.

undsoft (approved):
Looks good.

denysbutenko (approved):
Looks good. Thanks for adding tests and linter rule.

PR #9022, opened 2023-10-09
NAS-124480 / 24.04 / "Switched to new dataset" does not work anymore

AlexKarpov98:
Testing: Create new dataset -> you should be navigated to the newly created dataset now.
`We need to wait for datasets to be load before redirecting to newly created one.`


bvasilenko (approved):
Works good

PR #9024, opened 2023-10-09
NAS-123891 / 24.04 / NAS-123773: syslog option for system dataset is deprecated and should be removed

AlexKarpov98:
Testing: see that there are no `syslog` option anymore. See ticket.


undsoft (approved):
Looks good, the option is not there.

PR #9032, opened 2023-10-10
Revert "NAS-124392: Remove web shell (#8967)"

undsoft:
I removed web shell in 13.0-u6, when it should be removed in 13.1.


AlexKarpov98 (approved):
👍

PR #9036, opened 2023-10-11
NAS-124363 / 24.04 / Reducing test output

undsoft:
After tests complete, check to see that there are less than 52k lines of warnings (some warnings may still be there).


AlexKarpov98 (approved):
Test output is reduced:
Screenshot 2023-10-11 at 23 22 54

PR #9039, opened 2023-10-12
NAS-124606 / 24.04 / Duplicate name error is not shown in SMB shares

AlexKarpov98:
Testing: see ticket


bvasilenko (approved):
Works fine!

PR #9042, opened 2023-10-13
NAS-124445: Fix wrong state on VM page

bvasilenko:
**Summary**

This bug was occurring due to a race condition between `vm.query` and `vm.virtualization_details`. If the latter response is too slow, then `supported: false` flag gets ignored, which causes an issue described in the ticket.

**Testing**

Use mocking on front-end (hard-code `supported: false` , and add some delay to the `vm.virtualization_details` response)

**Expected result:**

On **Virtualization** page you will see **Validation is not supported** error block.

### DIFF
--- a/src/app/pages/vm/vm-list/vm-list.component.ts
+++ b/src/app/pages/vm/vm-list/vm-list.component.ts
@@ -3,7 +3,8 @@ import { MatDialog, MatDialogRef } from '@angular/material/dialog';
 import { Router } from '@angular/router';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { TranslateService } from '@ngx-translate/core';
-import { filter, switchMap } from 'rxjs/operators';
+import { lastValueFrom, of } from 'rxjs';
+import { catchError, filter, map, switchMap } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


src/app/pages/vm/vm-list/vm-list.component.ts
```diff
@@ -112,22 +113,28 @@ export class VmListComponent implements EntityTableConfig<VirtualMachineRow> {
...
+  prerequisite(): Promise<boolean> {
+    return lastValueFrom(
+      this.vmService.getVirtualizationDetails().pipe(
+        map((virtualization) => {
+          this.virtualizationDetails = virtualization;
+          this.hasVirtualizationSupport = virtualization.supported;
+          this.canAdd = virtualization.supported;
+          return true;
+        }),
+        catchError(() => {
```
RehanY147, line 125: Why are we enabling settings on error instead of showing an error state?

RehanY147 (changes requested):
image
Getting this error in the console if supported is `false`

bvasilenko:
@RehanY147 the requested changes should be addressed now

RehanY147 (approved):
The paginator shows up on top of the error message on the vm list page. Let's fix that as well.

PR #9045, opened 2023-10-13
NAS-124608 / 24.04 / Add a note explaining how to configure session timeout

AlexKarpov98:
Testing: 
Check Session expiring modal and see an updated like below:

Screenshot 2023-10-16 at 15 22 44

Click `Advanced Settings → Sessions`, you shall be redirected to the card (highlighted)

Screenshot 2023-10-13 at 14 29 40


src/app/modules/common/dialog/session-expiring-dialog/session-expiring-dialog.component.html
```diff
@@ -0,0 +1,25 @@
+<h1 mat-dialog-title>{{ options.title | translate }}</h1>
+<div mat-dialog-content id="confirm-msg">
+  <div class="message-content">
+    <span [innerHTML]="options.message | translate"></span>
+
+    <p>
+      {{ 'Change Token Lifetime in' | translate }}
+      <b (click)="viewSessionsCard()">
```
undsoft, line 8: Needs to look like a link so that people know to click it.

RehanY147 (changes requested):
The link takes out straight to the settings page. Was the session extended in that case? Or not? It's vague. We should also mention what will happen if they just click the link right away.

AlexKarpov98:
@RehanY147 

yes, session is extended in this case.
We may add Text about it but I don't think it's necessary (my opinion). 
Can you provide a text in this case if you think it's needed?

AlexKarpov98:
@RehanY147 
Screenshot 2023-10-16 at 15 22 44

undsoft (approved):
Looks good.

RehanY147 (approved):

PR #9046, opened 2023-10-13
NAS-124609 / 24.04 / Do not open Task Manager when using Upgrade All

AlexKarpov98:
Testing: see ticket


bvasilenko (approved):
Works good

PR #9056, opened 2023-10-16
NAS-124430 / 23.10.0 / Duplicate zvols and datasets are displayed upon initial creation and disappear when refresh browser

AlexKarpov98:
ATM - is not reproducing and reported can't reproduce either.

### DIFF
--- a/src/app/pages/datasets/store/dataset-store.service.ts
+++ b/src/app/pages/datasets/store/dataset-store.service.ts
@@ -1,5 +1,6 @@
 import { Injectable } from '@angular/core';
 import { ComponentStore } from '@ngrx/component-store';
+import { uniqBy } from 'lodash';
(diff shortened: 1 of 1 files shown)


undsoft:
So, there is some underlying problem that we need to find. This just puts a bandaid on it.

AlexKarpov98:
> So, there is some underlying problem that we need to find. This just puts a bandaid on it.

yeah, that's why it was a draft PR :) 
I'm waiting for response from @BonnieF

undsoft:
@AlexKarpov98 Ah, my mistake then. I've rescheduled it for 23.10.1.

denysbutenko:
Ping me when it's ready for review.

denysbutenko (changes requested):
Review is paused until further notification.

AlexKarpov98:
moved back to the queue.

PR #9059, opened 2023-10-17
NAS-124641 / 24.04 / Fix uptime string for HA systems

denysbutenko:
Changes

- Removed `elapsedTime`
- Do update `system.info` with 30 sec interval.
- Added `uptime` pipe to get correct string from raw date values.
- Skip values from active node on Standby widget - apply from `failover.call_remote` only

For testing, check on m40 if two widgets display correct information and update it every 30 secs.


bvasilenko (approved):
Works good to me

PR #9060, opened 2023-10-17
NAS-124667 / 24.04 / Use ix-cell-actions for actions for existing refactored tables to keep one style for all refactored tables & update getRow usage where custom indexes used

AlexKarpov98:
Testing: check refactored tables for regressions.

In this PR:
- Unified actions column using `actionsColumn`.
- Removed unused .scss files and style classes.
- Fixed lot's of SlideIns for getting data only after something was updated but not only if SlideIn was closed.
- Extended `ix-cell-actions` logic with `dynamicTooltip` & `disabled` props.
- Fixed styles for ix-table2 buttons to be justified correctly.
- Fixed some places where we can use `getValue()` instead of `template`


dszidi (approved):
LGTM

PR #9065, opened 2023-10-18
NAS-124681 / 24.04 / Race condition in interface form

AlexKarpov98:
Testing: check ticket I think code review is enough.


bvasilenko (approved):
Works good!

PR #9067, opened 2023-10-18
NAS-123988 / 24.04 / Skip processing state updates when the document is hidden

denysbutenko:
To avoid memory leaks and save memory usage.

https://developer.mozilla.org/en-US/docs/Web/API/Document/hidden


AlexKarpov98 (approved):
Nice improvement 👍

PR #9072, opened 2023-10-19
NAS-124607 / 24.04 / NAS-124607 ix-table2 backend pagination and sorting

bvasilenko:
New `ApiDataProvider` implements server-side pagination strategy server-side and sorting strategy for ix-table2

**Testing**

1. Test server-side pagination and server-side sorting in the table on a new **Audit** page (route [/audit](http://localhost:4200/audit)) 
   - generate test data by cloning https://github.com/iXsystems/ixlab and running `testing/python/audit_test_script.py` on your box with SMB service running

3. Check for regressions on other the pages, where ix-table2 is used


src/app/modules/ix-table2/api-data-provider.ts
```diff
@@ -0,0 +1,166 @@
...
+  getParams(): ApiCallParams {
+    return {};
+  }
+
+  paginate<T>(rows: T[], pagination: TablePagination): T[] {
+    return paginateFn(rows, pagination);
+  }
+
+  handleCurrentPage(loadFn: () => void, skipLoadFn: () => void): void {
+    skipLoadFn();
```
undsoft, line 34: Can you explain what `skipLoadFn` is and why we need it?

src/app/modules/ix-table2/async-data-provider.ts
```diff
@@ -1,31 +1,20 @@
...
   constructor(
-    private request$: Observable<T[]>,
+    private request$: Observable<ApiCallDirectory[ApiCallMethod]['response']>,
   ) {
     super();
-    this.refresh();
   }
 
-  refresh(): void {
+  load<K>(transformFunction?: () => OperatorFunction<K, T[]>): void {
```
undsoft, line 13: Since you can pass an observable in when creating a provider, I don't see the need to `transformFunction`, you can transform the original observable.

src/app/modules/ix-table2/interfaces/data-provider.interface.ts
```diff
@@ -2,12 +2,15 @@ import { Observable } from 'rxjs';
 import { TablePagination } from 'app/modules/ix-table2/interfaces/table-pagination.interface';
 import { TableSort } from 'app/modules/ix-table2/interfaces/table-sort.interface';
 
-export interface TableProvider<T> {
+export interface DataProvider<T> {
   rows: T[];
```
undsoft, line 6: `rows` should be removed as it now means different things depending on the child class used.

src/app/pages/system/advanced/init-shutdown/init-shutdown-card/init-shutdown-card.component.ts
```diff
@@ -84,18 +84,19 @@ export class InitShutdownCardComponent implements OnInit {
...
+    this.loadScripts();
   }
 
   onAdd(): void {
     this.openForm();
   }
 
   loadScripts(): void {
-    this.dataProvider.refresh();
+    this.dataProvider.load<InitShutdownScript[]>(() => pipe(
```
undsoft, line 97: This looks interesting, but I think a more native approach would be to just return an observable and subscribe to it.

undsoft (changes requested):

src/app/modules/ix-table2/api-data-provider.ts
bvasilenko: The purpose of arguments of `handleCurrentPage`:
-  to give the sorting strategy a chance to decide either to do full reload on sorting (for server-side), or just sort existing rows (for client-side), based on polymorphism  to avoid `if`-branching
- same for pagination strategy

I've now improved the naming of these 2 arguments.

src/app/modules/ix-table2/async-data-provider.ts
bvasilenko: This was an attempt to reach something we discussed earlier. However, I agree that it's too complex. `transformFunction` is removed now

src/app/modules/ix-table2/interfaces/data-provider.interface.ts
bvasilenko: done!

src/app/pages/system/advanced/init-shutdown/init-shutdown-card/init-shutdown-card.component.ts
bvasilenko: this is addressed as well

src/app/modules/ix-table2/components/ix-table-head/head-cells/ix-header-cell-checkbox/ix-header-cell-checkbox.component.html
```diff
@@ -1,6 +1,6 @@
 <mat-checkbox
   ixTest="header-checkbox"
-  [checked]="allChecked"
+  [checked]="allChecked | async"
```
undsoft, line 3: `allChecked$`

undsoft (approved):
I like it. New variable names make it clear what's going on. Smart solution for handling subscriptions inside of a table.
Nice work!

PR #9076, opened 2023-10-19
NAS-124670 / 24.04 / Reporting disks view improvements for large disk systems

AlexKarpov98:
Testing: See updated select component with "Select All" option.
Screenshot 2023-10-27 at 00 56 36

Screenshot 2023-10-27 at 00 57 17


undsoft (changes requested):
It works, but 1) icon doesn't match the result of the select options 2) I don't think we need it by default

AlexKarpov98:
> It works, but 1) icon doesn't match the result of the select options 2) I don't think we need it by default

I agree, PR is updated now 😺 (and images descriptions)

undsoft (approved):
Looks good, thanks.

PR #9080, opened 2023-10-20
NAS-124718 / 24.04 / Various code improvements

undsoft:
Testing: test for regressions in some of the affected places.


bvasilenko (approved):
Works good!

PR #9092, opened 2023-10-21
NAS-124740 / 24.04 / Refactoring InitShutdownListComponent

undsoft:
Testing: test InitShutdownListComponent.
Timeout column removed deliberately.


bvasilenko (changes requested):
Works good overall.

Noticed an issue with **Delete script > Cancel** : It still deletes , despite that Cancel was clicked.

undsoft:
@bvasilenko Good catch, but I think you tested the wrong table. This one is the one you get to by clicking on table header.

bvasilenko (approved):
Works good!

PR #9093, opened 2023-10-22
NAS-124716 / 24.04 / Fixing test attr for `actionsColumn`

RehanY147:


undsoft (changes requested):
We need to come up with something more specific. If you have 4 tables with multiple rows, it will still be hard to press at specific rows.
We've tried to solve something similar elsewhere – see `getTestAttr`.

You can check with RE to see what works for them.

undsoft (changes requested):
That's cool, but can we do something even better, for example by having test prefix on a `createTable` or something, so that other controls (like toggles in some columns) also get proper ids?

undsoft:
Cannot say that I like the idea of assuming that `id` property is going to be present in the object. Not sure, what prettier solution can we have.
@bvasilenko ideas?

RehanY147:
@undsoft I'm out of ideas other than to go back an iteration where we left it up to the components to add unique ids to everything instead of receiving a prefix. I think it is a safe assumption that table rows will have an id attribute to their objects.

bvasilenko:
@undsoft @RehanY147 

- How about instead of relying on the object, to supply zero-based indexes to distinguish between same types of controls
- I also like the current version with giving the option to have id or key column, but we should give an ability to specify the name of that column on per-object basis

undsoft (changes requested):
Okay, seems they would have preferred to have unique ids on controls inside the table.
What do you think about something like this:
```
columns = createTable<Catalog>([
   ...columns
],
{
  rowTestId: (row: Catalog) => ['catalogs', row.name],
});
```

This would provide a way to generate ids for rows and nested controls.
Then we could put an id on `tr` and buttons/links inside of them.

PR #9094, opened 2023-10-23
NAS-124695 / 24.04 / Create Cloudsync Wizard

denysbutenko:
Acceptance criteria listed in [the ticket](https://ixsystems.atlassian.net/browse/NAS-124695).

For testing, open Data Protection page and press on the `Add` button on the `Cloud Sync Tasks` card.


RehanY147 (changes requested):
If I choose PULL direction, I don't see the second directory explorer input. For Google Drive provider.

denysbutenko:
> If I choose PULL direction, I don't see the second directory explorer input. For Google Drive provider.

OK, it has been addressed. If existing credentials are chosen, `Folder` will appear instantly on the second step. It may take a few seconds to appear if new credentials are created. Also, some other adjustments are made.

src/app/pages/data-protection/cloudsync/cloudsync-wizard/cloudsync-wizard.component.ts
```diff
@@ -0,0 +1,131 @@
...
+        this.dialogService.error(this.errorHandler.parseWsError(err));
+        this.rollBack();
+      },
+    });
+  }
+
+  rollBack(): void {
+    const requests: Observable<unknown>[] = [];
+
+    this.createdProviders.forEach((credential) => {
```
RehanY147, line 98: Why are we making the assumption that if the task creation fails, the added credential must be deleted too. If the user wants to fix the bug and come back to add the task again, they would prefer to not go through the hassle of creating a provider again. Some providers are quite a hassle to create. You have to open another window, login into another service and provide permissions etc. So, we should be careful about rolling back credentials created.

src/app/pages/data-protection/cloudsync/cloudsync-wizard/steps/cloudsync-provider/cloudsync-provider.component.ts
```diff
@@ -0,0 +1,277 @@
...
+        },
+        error: (error) => {
+          this.isLoading$.next(false);
+          this.formErrorHandler.handleWsFormError(error, this.form);
+          this.cdr.markForCheck();
+        },
+      });
+  }
+
+  openAdvanced(): void {
```
RehanY147, line 193: The Advanced Options button should be on the 2nd step since it takes user to the `CloudSyncFormComponent` and not the `CloudCredentialsFormCompondnt`

src/app/pages/data-protection/cloudsync/cloudsync-wizard/steps/cloudsync-what-and-when/cloudsync-what-and-when.component.ts
```diff
@@ -0,0 +1,515 @@
...
+          this.form.controls.bucket.disable();
+          this.form.controls.bucket_input.enable();
+          this.dialog.closeAllDialogs();
+          this.dialog.confirm({
+            title: error.extra ? (error.extra as { excerpt: string }).excerpt : `${this.translate.instant('Error: ')}${error.error}`,
+            message: error.reason,
+            hideCheckbox: true,
+            buttonText: this.translate.instant('Fix Credential'),
+          }).pipe(filter(Boolean), untilDestroyed(this)).subscribe(() => {
+            // const extra: NavigationExtras = { state: { editCredential: 'cloudcredentials', id: credential.id } };
```
RehanY147, line 420: Comments

RehanY147 (changes requested):
We should not be showing the rest of the form if the user chooses an existing credential. And we shouldn't be showing any input fields until the user has chosen if they want to use an existing provider or create a new one. We should add an option "Create New" in the provider drop down and only shown the form fields if the user clicks on that option.

src/app/pages/data-protection/cloudsync/cloudsync-wizard/cloudsync-wizard.component.ts
denysbutenko: Good point. I agree. Let's keep providers for reusing.

src/app/pages/data-protection/cloudsync/cloudsync-wizard/steps/cloudsync-provider/cloudsync-provider.component.ts
denysbutenko: Makes sense

denysbutenko:
> We should not be showing the rest of the form if the user chooses an existing credential. And we shouldn't be showing any input fields until the user has chosen if they want to use an existing provider or create a new one. We should add an option "Create New" in the provider drop down and only shown the form fields if the user clicks on that option.

OK, requested changes are addressed.

RehanY147 (approved):

PR #9097, opened 2023-10-23
NAS-124690 / 23.10.1 / Allow layouts to change for metadata step

RehanY147:
Check ticket description for testing


AlexKarpov98 (approved):
I can change layout on metadata step.
Please fix tests.

PR #9102, opened 2023-10-24
NAS-124247 / 24.04 / Removed builtin flags from presets requests

RehanY147:
Save an ACL list as a preset. When you reuse the preset, you shouldn't get more entries than the ones you originally added to the preset. Save the ACL without changing the preset and it shouldn't error out and save successfully.


AlexKarpov98 (approved):
I was able to do it  without errors out and saved successfully.

PR #9104, opened 2023-10-24
NAS-124763 / 24.04 / Incorrect indentation in dataset tree

AlexKarpov98:
Testing: 
check that rows are aligned correctly now:
Screenshot 2023-10-24 at 17 58 57

Before it was like:
image


undsoft (approved):
Looks good, please check if this needs to go into Cobia.

PR #9110, opened 2023-10-25
NAS-124792 / 24.04 / Refactor ScrubListComponent

undsoft:
Testing: test scrub list full page table (by clicking on header in Data Protection -> Scrub Tasks header).


bvasilenko (approved):
Works good!
Noticed that **Schedule** column does not have a tooltip (might be useful to see the cropped part of the text). Probably should create a ticket for that

PR #9111, opened 2023-10-25
NAS-124732 / 24.04 / Request:  Improve sorting for Replace Disk Search for DRAIDs 

AlexKarpov98:
Testing: Go to Devices and click Replace
Disks input shall appear and now they are sorted alphabetically and filtering is available.
As well selected item is highlighted (fixed this issue - before it was not)

Screenshot 2023-10-25 at 16 49 39


RehanY147 (changes requested):
Since we're doing this for replace operation, we might as well do it for Extend disk option as well.

RehanY147 (approved):

PR #9112, opened 2023-10-25
NAS-123820 / 24.04 / Remove `selected` property from ChartRelease interface

AlexKarpov98:
Testing: check that app selection works as before

### DIFF
--- a/src/app/interfaces/chart-release.interface.ts
+++ b/src/app/interfaces/chart-release.interface.ts
@@ -74,9 +74,6 @@ export interface ChartRelease {
   history: { [key: string]: ChartReleaseVersion };
   resources?: ChartResources;
   stats?: ChartReleaseStats;
-
-  // TODO: Frontend field, move to another interface.
-  selected?: boolean;
 }
 
 export interface ChartReleaseStats {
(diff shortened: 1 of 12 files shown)


denysbutenko (changes requested):
Some regressions are found. `Check/Uncheck All` does not work properly. Checkboxes at rows are still unchecked. Checkboxes randomly being unchecked when checking others. See recording: 

https://github.com/truenas/webui/assets/351613/d215f00a-ae16-4ed5-bc70-9c5a94a42b3a

denysbutenko (changes requested):
Those regressions are still reproducible.
> Checkboxes randomly being unchecked when checking others. The selected count exceeds the installed apps.

https://github.com/truenas/webui/assets/351613/9f41ff43-fb88-4da3-a99f-075ea46c7ad2

AlexKarpov98:
@denysbutenko - now it is fixed, the problem was that I stored objects and the cursor for the object changed after updates came.
As well I fixed sorting issue - since when data got updated - it returned to the default sorting.

denysbutenko (approved):
Looks good 👍🏻

PR #9113, opened 2023-10-25
NAS-103365 / 24.04 / add if a user has ssh key installed in user drop down

AlexKarpov98:
Testing: go to Users page and see `new` SSH column.
Screenshot 2023-10-25 at 18 33 38


undsoft (changes requested):
I don't now if this information is so important as to put it as one of the columns.
I'd prefer to see it in details.

undsoft (approved):
Looks good 👍🏻

PR #9114, opened 2023-10-26
NAS-124731 / 24.04 / Remove the "Remove" and "Extend" buttons for the DRAID pool

bvasilenko:
**Testing**

1. Connect to `remote: '10.220.16.10', // m60-100.dc1.ixsystems.net`
2. On Storage > (Pool block) > Manage Devices page,

   - **Expected Result:** When actions “Extend” and “Remove” are not applicable (see ticket), the respective buttons should not be displayed in WebUI in the aforementioned conditions.


AlexKarpov98 (approved):
works as expected.

PR #9115, opened 2023-10-26
NAS-124818 / 24.04 / Missing back button on dashboard widget

undsoft:
Help widget on main dashboard will now show a back button when in mobile mode.


dszidi (approved):
Back button is there now. LGTM

image

PR #9123, opened 2023-10-27
NAS-124762 / 24.04 / Extract the same code used for table/details layout

AlexKarpov98:
Testing: check Datasets, Devices, Installed apps screens for regressions.
Added new scss mixin


dszidi (approved):
Looks good. Checked Datasets, Installed Apps and data protection dashboard. Did not see any regressions.

PR #9128, opened 2023-10-30
NAS-124846: Standardize dRAID and RAIDZ pool creation width

AlexKarpov98:
Testing: 
Check that for any Draid configuration the minimum amount of `Data Devices` is **2**
Screenshot 2023-10-30 at 10 38 33


dszidi (approved):
LGTM. Draid configurations no longer offer option of single disk in Data Devices field

image

PR #9129, opened 2023-10-30
NAS-124841 / 24.04 / Expose job creator credentials

AlexKarpov98:
Testing see ticket:
and new credentials prop
Screenshot 2023-10-31 at 13 36 14
Screenshot 2023-10-31 at 13 36 26

`credentials` can be `UNIX_SOCKET`, `ROOT_TCP_SOCKET`, `LOGIN_PASSWORD`, `API_KEY` or `TOKEN`,
        depending on what authentication method was used.
        
        For `UNIX_SOCKET` and `LOGIN_PASSWORD` logged-in `username` field will be provided in `data`.
        For `API_KEY` corresponding `api_key` will be provided in `data`.
        For `TOKEN` its `parent` credential will be provided in `data`.


undsoft (commented):
If `type` is a enum, I'd rather we show our strings instead of altering backend strings. They are not translatable.
I'd also prefer to see:
```
Created by: root (password login)
```

PR #9131, opened 2023-10-30
NAS-124231 / 24.04 / Add `Create New` option for Cloud Credentials

AlexKarpov98:
Testing: see ticket

Result:

https://github.com/truenas/webui/assets/22980553/f7d009dd-1845-48ac-be23-5ca2739e5685


RehanY147 (changes requested):
This PR seems to be in conflict with https://github.com/truenas/webui/pull/9094. The other PR is replacing the form with a wizard and adding ways to add new providers. Discuss with @denysbutenko please.

AlexKarpov98:
@RehanY147 - @denysbutenko told me:
`
Cloud Credentials form is still in use despite adding Cloudsync Wizard. Wizard also have Create New option to add new credential although it renders the cloud credentials form below the credentials select. I am not sure if this way suitable for the form. so I think there is no conflict between PRs.
`

RehanY147 (changes requested):
This is too impactful of a change to accommodate one small form. I think we should just open another slide in form instead of changing the ix-modal-header, meant for ix-slide-ins, to handle dialogs. Plus, as a team preference, we don't want to show forms in dialogs.

AlexKarpov98:
> This is too impactful of a change to accommodate one small form. I think we should just open another slide in form instead of changing the ix-modal-header, meant for ix-slide-ins, to handle dialogs. Plus, as a team preference, we don't want to show forms in dialogs.

@RehanY147 - then I can close this PR?

RehanY147:
@AlexKarpov98 I think so. Instead of opening the form in a dialog, let's use a slide-in to open this form. Or use the new `IxChainedSlideInService`

RehanY147 (commented):
^

AlexKarpov98:
Instead of opening the form in a dialog, let's use a slide-in to open this form. Or use the new IxChainedSlideInService

PR #9138, opened 2023-10-31
NAS-124895 / 24.04 / Selecting Custom retention policy once, keeps Save button disabled even after choosing other retention option

AlexKarpov98:
Testing: check ticket


bvasilenko (approved):
Works as expected

PR #9140, opened 2023-10-31
NAS-124843 / 24.04 / Inconsistent landing page, when clicking "navigation" paths

AlexKarpov98:
Testing: see ticket and try to find any regression.


RehanY147 (changes requested):
If you go to `Datasets` and on the `Data Protection` card click on `'Manage Snapshots'` you get to the snapshots page. If you click the `Snapshots` breadcrumb there, you get redirected to the same page (navigation shouldn't happen since the user is already on this page) and the breadcrumb changes to just `Datasets` (should be `Datasets -> Snapshots`)

undsoft:
I don't think it should. I don't think we need to show current page as part of the breadcrumb, only parent pages.

AlexKarpov98:
@RehanY147 @undsoft - so, what is the final solution?
I as well think that we should show only parent pages, I'll recheck to match this on all pages and will update you.

AlexKarpov98:
@undsoft @RehanY147 - okay, so now it is updated and all routes don't show a breadcrumb for the current page but only for parent if present.

undsoft (approved):
When I click on Credentials breadcrumb on Groups page, I get to Users page.
I think the breadcrumb should be white and non-clickable.

Other than that breadcrumbs look reasonable.

AlexKarpov98:
@RehanY147 - I can't merge my PR, waiting for your feedback/approve 😃

RehanY147 (approved):

PR #9143, opened 2023-10-31
NAS-124908 / 24.04 / Figure out why tests sometimes fail

AlexKarpov98:
Testing: see that pipeline tests succeeded.
Problem in some test cases with getting text "`in 1 day`" I assume because when we try to get straight 24 hours difference it depends on the day time and returned `in 23h` sometimes instead of `in 1 day` when we use `cronParser.parseExpression`.

I fixed it previously to generate 25h instead of 24h, but let's keep this PR update as well for the history.
It is fixed by expanding to 26 hours for example, in this case we'll always get `in 1 day` and not `in 23h` neither`in 26h`


denysbutenko (changes requested):
The title sounds like we need to understand why it failed. The changes have increased day hours to 26 hours to pass the tests. The question is still open. It wasn't an issue for a while. When did it start happening, and why?

AlexKarpov98:
> The title sounds like we need to understand why it failed. The changes have increased day hours to 26 hours to pass the tests. The question is still open. It wasn't an issue for a while. When did it start happening, and why?

It fails because of setting 24h - I think, because sometimes library converts it to 23h instead of 1 day.
I din't see any fails after this update.

denysbutenko:
> > The title sounds like we need to understand why it failed. The changes have increased day hours to 26 hours to pass the tests. The question is still open. It wasn't an issue for a while. When did it start happening, and why?
> 
> It fails because of setting 24h - I think, because sometimes library converts it to 23h instead of 1 day. I din't see any fails after this update.

Then, we could have a bug in UI while unit tests have a hacky workaround. We need to figure out why it is happening with facts to ensure that it's a bug in external dependency or CI or somewhere else.

AlexKarpov98:
@denysbutenko - so, the final verdict - we pass data to `formatDistanceToNow` function which should return string with something like `in 1 day` or `in 22h`.
The value which is sent there is `new Date(new Date().getTime() + (24 * 60 * 60 * 1000))`

I think it's up to `formatDistanceToNow` from `'date-fns'` to fix the bug, it's weird

AlexKarpov98:
@denysbutenko - it's not a bug on our side, because we don't generate that text manually, and it depends on the input it receives

denysbutenko:
Well. If there is a bug in `date-fns`. When did it start happening? Can we downgrade to the previous version until we wait for a fix upstream?

AlexKarpov98:
@denysbutenko @undsoft - let's see if it will fail once again, but I think it's a think:
```
but totally depends on the 3rd party library
in test I only provide data - and that data is managed by the Library
and transformed data is returned (edited) 
```

AlexKarpov98:
I see we have `"date-fns": "~2.28.0",` - for a long time, not sure if someone opened a ticket regarding such behaviour.

denysbutenko:
> I see we have `"date-fns": "~2.28.0",` - for a long time, not sure if someone opened a ticket regarding such behaviour.

If [date-fns](https://github.com/date-fns/date-fns/issues?q=is%3Aissue+is%3Aopen+formatDistanceToNow) wasn't upgraded for a long time. When did the first known failure occur and what is changed?

AlexKarpov98:
> > I see we have `"date-fns": "~2.28.0",` - for a long time, not sure if someone opened a ticket regarding such behaviour.
> 
> If [date-fns](https://github.com/date-fns/date-fns/issues?q=is%3Aissue+is%3Aopen+formatDistanceToNow) wasn't upgraded for a long time. When did the first known failure occur and what is changed?

It may be an existing version bug, there were added tests which were not there before, I think we can count this ticket as closed at this point and get back to it if an error will occur again, but it should not.

AlexKarpov98:
Closed to see if issue can be reproduced and test fail (should not be the case)

PR #9145, opened 2023-11-01
Update fr.json

MDVAurelien:
Add some French translations. Only a few translations remain. It's almost over ;-)


AlexKarpov98 (approved):
Thanks. It's not over yet 😄

PR #9146, opened 2023-11-01
NAS-124940 / 24.04 / Main layout on audit log page

AlexKarpov98:
Testing: 
see ticket and check table, I added data via html replacing ix-table2-->table-->tbody value (copy/paste)
as well I removed border-radius from header  card&search, on screenshot it's rounded still..

Selected state:
Screenshot 2023-11-01 at 12 32 27

Unselected state:
Screenshot 2023-11-01 at 12 38 05

Empty state:
Screenshot 2023-11-01 at 12 38 15

Smaller devices as regular, hide details tab
Screenshot 2023-11-01 at 12 32 51


AlexKarpov98:
Final update:
Screenshot 2023-11-01 at 18 00 09

PR #9148, opened 2023-11-01
NAS-124948 / 24.04 / Add warning to Sync setting

AlexKarpov98:
Testing: 
See ticket
Screenshot 2023-11-01 at 14 47 23


src/app/modules/common/dialog/confirm-dialog/confirm-dialog.component.html
```diff
@@ -38,6 +38,7 @@ <h1 mat-dialog-title>{{ options.title | translate }}</h1>
     {{ options.cancelText | translate }}
   </button>
   <button
+    *ngIf="options.buttonText"
```
undsoft, line 41: I don't think it's necessary. We already have `hideCancel`. The button in the dialog should say something like `Okay` instead of `Cancel`.

undsoft (changes requested):
If you create a dataset with Sync = Disabled and then try to edit it, you'll get a warning again.

undsoft (approved):
👍🏻

AlexKarpov98:
backport-23.10.1

PR #9149, opened 2023-11-01
NAS-124729: Request:  Add name of Pool to Storage > Devices UI in title bar and scroll bar to aid with larger quantity disked systems

AlexKarpov98:
Testing:
see ticket, and Devices page details

Screenshot 2023-11-01 at 15 46 28

### DIFF
--- a/src/app/pages/storage/modules/devices/components/disk-details-panel/disk-details-panel.component.html
+++ b/src/app/pages/storage/modules/devices/components/disk-details-panel/disk-details-panel.component.html
@@ -15,13 +15,18 @@
 
     <span class="prefix">{{ 'Details for' | translate }}</span>
 
-    <div class="full-path">
+    <div class="full-path" [matTooltip]="detailsTitleTooltip">
       <ix-topology-item-icon
         class="title-icon"
         [topologyItem]="topologyItem"
         [disk]="disk"
       ></ix-topology-item-icon>
-      <span class="device-name">{{ title }}</span>
+      <span class="device-name">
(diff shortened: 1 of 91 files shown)


dszidi (changes requested):
I agree with the ticket reporter that having the pool name on the page would be helpful. However the header of the details section is not the place we should put it.  IMO it would make sense to put it in the page title instead. There are a couple of reasons for this. 

To start with, there is redundancy between the breadcrumb and the page title. If we look at the screenshot below, it would be better to have the title be `PoolName` or `PoolName Devices` rather than the current scenario where `Devices` is used both in the breadcrumb and the title

image
 
...
...
...

More importantly, the details section gets hidden on smaller screens.

image

dszidi:
@undsoft What do you think about the above comments?

AlexKarpov98:
> @undsoft What do you think about the above comments?

that makes sense to me as well, I missed that part that it's hidden on smaller screen and user won't be able to get that info from the start page where nothing is selected.

dszidi:
> @dszidi Screenshot 2023-11-07 at 16 31 33

That shows the VDEV name. We need the pool name

dszidi (changes requested):
image

@AlexKarpov98  The title is showing the vdev name. We need the pool name

AlexKarpov98:
@dszidi - now it's updated.
Screenshot 2023-11-08 at 00 04 36

dszidi (approved):
LGTM

PR #9152, opened 2023-11-02
NAS-124785 / 24.04 / Fix network interface bandwidth

bvasilenko:
**Summary**

Middleware already applies **kilobytes -> bytes** conversion to values returned by netdata daemon.  So properties `received_bytes_rate` , `sent_bytes_rate` carry values in bytes as their name suggests.

**Testing**

1. Throttle browser's bandwith at Developer tools > Network 
2. Force refresh the page, while reading metrics from Network widget on Dashboard page in another browser


AlexKarpov98 (approved):
Works good.

PR #9153, opened 2023-11-02
NAS-124904 / 24.04 / Refactoring iSCSI tables - part 1

denysbutenko:
Part 1 includes refactoring for Targets, Extents and TargetExtents tables.

For testing, open the Shares dashboard and click on the `iSCSI` header. Check refactored tables for regressions.


denysbutenko:
> Make sure there are loading states for table actions like delete and toggle buttons. There's a delay after clicking the delete button on the associated targets button. Please also make sure that after the delete operation is confirmed, there's a loader shown so the user doesn't navigate away before deletion is complete. Tables look good otherwise.

Okay, I added a loader while getting ISCSI sessions before confirmation appears.

RehanY147 (approved):

PR #9156, opened 2023-11-02
NAS-124247: Removed builtin flags from presets requests

RehanY147:
Check ticket for testing


dszidi (approved):
Custom ACLs are now saved correctly without extra access controls

image

PR #9158, opened 2023-11-02
NAS-124629 / 24.04 / Adapting to new 2fa changes

RehanY147:
Read ticket for testing details


RehanY147:
Tests will be added by https://ixsystems.atlassian.net/browse/NAS-125007

bvasilenko (changes requested):
The implemented functionality works as expected,

However I've noticed that there's an unused interface `UserTwoFactorConfigUpdate`.
I've also noticed that ticket's description mentions: "... now these both attrs have been moved to being per user configuration instead of it being applied to every one. "  So, I would bring this into the scope of adaptation to 2-factor changes, and suggested to implement input fields for "OTP Digits" and "Interval" at the individual 2-factory settings page. If there's another ticket for this, please let me know

RehanY147:
@bvasilenko The part you're talking about, when the user goes to renew their secret, UI asks them for both of those fields in a dialog before renewing their secret. These fields are not needed on the settings page.

bvasilenko (approved):
Works as expected

PR #9159, opened 2023-11-02
NAS-124962 / 24.04 / Fix scrolling on pane-details layout

AlexKarpov98:
Testing:
Check ticket and all related pages with similar layout:
- Datasets
- Devices
- Apps
- Audit

Upon scroll you should see search & table header sticked to the top while scrolling.
There should be less scrollbars or only one (to control the page scroll)

### DIFF
--- a/src/app/modules/ix-table2/components/ix-table-body/ix-table-body.component.html
+++ b/src/app/modules/ix-table2/components/ix-table-body/ix-table-body.component.html
@@ -2,7 +2,9 @@
   <tr
     class="row"
     [ixTest]="getTestAttr(row)"
-    [ngStyle]="{ cursor: detailsTemplate ? 'pointer' : 'auto', display: isLoading ? 'none' : 'grid' }"
+    [ngStyle]="{ cursor: detailsTemplate ? 'pointer' : 'auto' }"
+    [ngClass]="{ hidden: isLoading }"
+    [hidden]="isLoading"
     (click)="onToggle(row)"
   >
     <td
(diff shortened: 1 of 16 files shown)


src/app/modules/ix-table2/components/ix-table-body/ix-table-body.component.html
```diff
@@ -3,6 +3,7 @@
     class="row"
     [ixTest]="getTestAttr(row)"
     [ngStyle]="{ cursor: detailsTemplate ? 'pointer' : 'auto' }"
+    [ngClass]="{ hidden: isLoading }"
     [hidden]="isLoading"
```
undsoft, line 7: Perhaps somebody was trying to write `[class.hidden]` the line below?

undsoft (changes requested):
Details side panel also need to stick to the top and not just scroll through:
Снимок экрана 2023-11-03 в 10 24 12

This may be outside of the scope of this ticket because of virtual scrolling, but I think scrolling on Datasets page can be improved. Try with `APPS/a-lot-of-datasets` on our m40.

AlexKarpov98:
@undsoft - separate ticket is created to handle virtual scroll behaviour.
https://ixsystems.atlassian.net/browse/NAS-125044

undsoft (changes requested):
Seems like I shouldn't have scrolling in details panel here, since there is enough space to render everything when scrolled up.

https://github.com/truenas/webui/assets/1481270/d2c8e697-01f9-41a3-addb-6407e884b345

undsoft (approved):
Let's add a little more space here, so that it's not so up there. 
Снимок экрана 2023-11-07 в 10 59 50

Otherwise looks good.

PR #9162, opened 2023-11-03
NAS-125003 / 24.04 / Dashboard is missing Loading State (only blank screen)

AlexKarpov98:
Testing: see ticket and result.
I think we can just show loaders with static 6 items, we don't need more or less. After some updates we got empty page and then widgets.

Result:
Screenshot 2023-11-03 at 12 13 35


undsoft (approved):
Nice improvement, thanks.

PR #9168, opened 2023-11-06
NAS-124969 / 24.04 / Add Smb Status page

denysbutenko:
Add Smb status overview page that includes tabs to navigate between different types of SmbStatus information. Other tables will be implemented separately.


RehanY147 (commented):
Need instructions to test this. I can't get a record to show up on that page.

denysbutenko:
> Need instructions to test this. I can't get a record to show up on that page.

The table was implemented before and already tested.
For testing, ensure that
- SMB Status page has Sessions tab
- SMB Session table is rendered on SMB Status page
- URL is match `/sharing/smb/status/sessions`
- if go to `/sharing/smb/status` check redirect to `/sharing/smb/status/sessions`

RehanY147 (changes requested):
The page looks good, however, I only see one tab on the page. Are there supposed to be more tabs there? Also, there should be a way to redirect to the new sessions page on the Shares list page that comes up when you click the Table header on the shares dashboard.

denysbutenko:
> The page looks good, however, I only see one tab on the page. Are there supposed to be more tabs there? 

Yes, it is. There are a couple of tickets for each new view.

> Also, there should be a way to redirect to the new sessions page on the Shares list page that comes up when you click the Table header on the shares dashboard.

Make sense. Unfortunately, the full-page table uses a legacy entity table component that makes extending the page header a bit tricky. Let's add it when the table is being refactored to the ix-table2 component.

RehanY147 (approved):
Let's add a ticket for the refactor. Lest this detail is missed when someone else creates it./

PR #9169, opened 2023-11-06
NAS-124907 / 24.04 / Update change detection to on push in some components, part 2

bvasilenko:
Testing

Do regression resting of the affected blocks


undsoft (approved):
Looks good. No regressions found.

PR #9172, opened 2023-11-06
NAS-124947 / 24.04 / [Apps/UI]: Immutable fields in an item, leaks in other items too

AlexKarpov98:
Testing: see ticket, on m40 I tested with `komga` app:
http://localhost:4200/apps/installed/TRUENAS/community/komga-krpv/edit

Newly added items in edit mode shall be editable:
Screenshot 2023-11-06 at 14 06 25

Correct logic: 
IMMUTABLE field - is only for EDIT mode, but it should not spread on the ADD NEW ITEM ON THE LIST in EDIT mode.
We made a bug where we spread IMMUTABLE field for new item on the list in edit mode.


stavros-k (approved):
Works fine!

src/app/services/schema/app-schema.transformer.ts
```diff
@@ -33,15 +33,13 @@ export function isCommonSchemaType(type: ChartSchemaType): boolean {
...
-  const {
-    schema, chartSchemaNode, isNew, isParentImmutable,
-  } = payload;
+  const { schema, chartSchemaNode } = payload;
 
   return {
     controlName: chartSchemaNode.variable,
     title: chartSchemaNode.label,
     required: schema.required,
-    editable: (!isNew && (!!schema.immutable || isParentImmutable)) ? false : schema.editable,
```
AlexKarpov98, line 44: this logic should only be cared by middleware, like `schema.ediable`

AlexKarpov98: disabled field state is managed in `app-schema.service.ts`

denysbutenko (approved):
👍🏻

PR #9175, opened 2023-11-07
NAS-124972 / 24.04 / Add smb status locks table

denysbutenko:
Depends on #9168.

For testing, create an SMB share and connect to it. Then go to shares dashboard - SMB status - locks tab.


src/app/modules/ix-table2/base-data-provider.ts
```diff
@@ -61,6 +62,11 @@ export class BaseDataProvider<T> implements DataProvider<T> {
     this.updateCurrentPage(this.allRows);
   }
 
+  setExpandedRow(row: T): void {
+    this.expandedRow = row;
+    this.expandedRow$.next(row);
```
undsoft, line 67: Does not appear to be used.

src/app/pages/sharing/smb/smb-lock-list/smb-lock-list.component.html
```diff
@@ -0,0 +1,72 @@
...
+        [dataProvider]="dataProvider"
+      ></thead>
+      <tbody
+        ix-table-body
+        [columns]="columns"
+        [dataProvider]="dataProvider"
+        [isLoading]="dataProvider.isLoading$ | async"
+      >
+        <ng-template let-lock ix-table-details-row [dataProvider]="dataProvider">
+          <h3>{{ 'Open Files' | translate }}</h3>
```
undsoft, line 33: Separate component and test case?

denysbutenko:
> It looks good.
> 
> 1. I like that table controls are below in the table on Locks. We should do in the future the same for Sessions. We could potentially then move tabs over to the right.

The table controls for sessions were moved from the page header in #9221.

> 2. I think there can be a lot of open files. I don't think UI currently will be able to handle this.

My understanding is that each open file has its table row. If there are multiple files I added collapsible sections and the first is expanded by default. I am open to suggestions on how to do it better.

> 3. We also seem to show a lot of technical information. I would try to simplify and remove/hide everything that is not immediately useful. May be worth checking with Andrew as to what he thinks is useful to show here.

Okay, I am waiting for Andrew's input on this.

src/app/modules/ix-table2/base-data-provider.ts
denysbutenko: Ah, there is another implementation based on EventEmitter that was added recently.

denysbutenko:
Since open files require pagination, MatExpansionPanels were replaced with IxTable2 inside a collapsible row. Columns adjusted to show useful info (server, uid, opened at).

undsoft (approved):
Yeah, looks good. Can't say that I'm particularly excited about nested tables, but I don't have any good ideas and this is something we improve in the future.

PR #9176, opened 2023-11-07
NAS-125000 / 24.04 / Extract common css for button groups

denysbutenko:
For testing, check button groups on the Jobs List and Reporting pages.

Preview:
image
image

image
image


undsoft (approved):
Reporting page currently shows selected state for the last button you clicked. This is not necessary, perhaps button group is not the best choice for that page.
Looks okay on jobs.

denysbutenko:
> Reporting page currently shows selected state for the last button you clicked. This is not necessary, perhaps button group is not the best choice for that page. Looks okay on jobs.

Okay, I changed button groups on reporting to use simple mat-buttons.

PR #9177, opened 2023-11-07
NAS-121316: Backups Widget Markup

bvasilenko:
image

image


undsoft:
Hi, looks good overall to me.
1. I think we can update card with tasks to say "Other tasks" somewhere above counts.
2. And all Manage links can be replaced with one going to Data Protection dashboard.
3. Would also be good to see (or get a description of what happens) if there are more tasks then presented.

dszidi:
I liked it at first glance. However I’m not sure about showing the last 5 tasks. 
I’m thinking of scenarios like for example a user has 10 tasks and the last 5 were successful.
If the user is just glancing at it, they may very well miss the little status message next to the card title because all they see are all the green success buttons

Also, the task rows are presented identical to how it is in the data protection dashboard. On the DP dash those buttons show the log message. That behavior is probably not appropriate to do here, but it is what users will expect since they look the same.

Could we try an alternative design that focuses on the summary without the individual task rows. 
Maybe have 3 sections for Replication, Cloud and Rsync tasks
For presentation ideas, take a look at the storage widget and see how it shows a pool
Each task type could use a similar multi column approach. 

1. Title and "Manage" Link
2. Send Tasks
3. Receive Tasks (Optional)

Some datapoints for the send and receive columns
      
- `<n>` `<send | receive>` tasks configured
- Total Failed `<send | receive>` tasks
- Last successful task `<timestamp>`
- `<n>` successful `<send | receive>` tasks this week

For presentation ideas, we can do something similar to how the storage widget shows a pool

dszidi (changes requested):
Check above comment ^^

bvasilenko:
I've updated backup widget UX

Visibility of "Documentation" section should depend on number of types of existing tasks:

- When one or 2 types of tasks exist, the  "documentation" section is displayed, highlighting the need of backup
- When all 3 types of tasks exist, the "documentation" section should be hidden

image

image

image

image

dszidi (changes requested):
After getting some feedback we’d like to request another revision.

- Rename the card title to “Backups” (remove the word “Tasks”)
- Split the card into three tiles.
- First tile is called “Tasks” where we show one row per task types Rsync | Replication | Cloud Sync (non Storj providers)
- Second tile is called Storj where we similarly list Storj cloud sync tasks
- Third tile can be for “Documentation” where we can have links like what you already have listed.


Tasks and Storj entry rows should have the following columns

1. Tile Title (see example)
2. Number of send tasks
3. Manage link

image

undsoft:
@bvasilenko Wait on this. It seems there is some confusion.

dszidi (changes requested):
After some discussion, let's try three different screens/scenarios

1. Fully populated widget
2. One task type with no tasks at all
3. Empty state (No tasks at all for any task type)
4. Three task types with pull tasks only

### 1: Fully populated widget

Show a scenario where all three task types (Cloud Sync, Rsync, Replication) has tasks to list.
I've made an example for what a populated tile should look like that you can use as reference

image


### 2. One task type with no tasks at all

This changes the tile layout by hiding the task type that has no tasks at all. There should only be tiles for the two remaining task types


### 3. Empty state (No push tasks at all for any task type)

For this scenario we should iterate on your previous design.

image

The empty state message on top should read `Backup <a>to cloud</a> / <a>to another TrueNAS</a>` 

Then underneath list backup targets

- Cloud Sync to Storj or similar provider
- ZFS Replication to another TrueNAS/ZFS server
- Rsync to another server

### 4. Three task types with pull tasks only

Scenario has three task types and none of them have push tasks. In this scenario the message next to the card title should be changed to show the empty state message:
`Backup <a>to cloud</a> / <a>to another TrueNAS</a>`

bvasilenko:
@dszidi 

3. If there are three states with a difference in task type:

* Two types have both pull and push. The third also has both pull and push. - Display three cards (1st state)

* Two types have both pull and push. The third has only pull. - Display rows (3rd state)

* Two types have both pull and push. The third is empty. - Display two cards (2nd state)

Did I understand correctly? In other words, should we use the card view only if each type either necessarily has push (pull is optional), or has neither push nor pull? In all other cases - the widget should display `ix-table2` with rows?

4. I don't quite understand what to do in the 4th point. Should we show rows in ix-table2, and where should I place the row with the message? Instead of Backup Tasks?

undsoft:
This does not use ix-table2. The layout is similar to what you already had. We show this:
285949244-551e06df-555d-4d28-bf91-46169129c5ae

Basically:
1. If you have any tasks of a certain type, we show a section for that type. If you don't have tasks of a certain type, section is now shown.
2. If don't have any tasks at all, we show empty state.
3. If you have some tasks, but none of them are backup push tasks, we show banner in the bottom prompting you to create a backup task (Point 4).

If you feel like this is enough to start with implementation, we can close this PR and start making tickets for implementation.

bvasilenko:
I have final designs, so I'll proceed with creating tickets for implementation

For the 1st case:

image
image
image

For the 2nd case:

image

For the 3rd case:

image
image
image

PR #9178, opened 2023-11-07
NAS-125044 / 24.04 / Add Window Virtual Scroll on the datasets page instead of scrolling separate container with virtual scroll 

AlexKarpov98:
Testing: 
Scrolling on the datasets page should be via window scroll and not via limited container height viewport as before.

https://github.com/truenas/webui/assets/22980553/764e4288-7dfb-451b-9eec-996f4452fc7a


src/app/pages/datasets/components/dataset-management/dataset-management.component.ts
```diff
@@ -304,9 +305,10 @@ export class DatasetsManagementComponent implements OnInit, AfterViewInit, OnDes
   private listenForDatasetScrolling(): void {
     this.subscription.add(
       this.scrollSubject
-        .pipe(distinctUntilChanged(), untilDestroyed(this))
+        .pipe(debounceTime(5), untilDestroyed(this))
```
bvasilenko, line 308: Noticed a hard-coded delay. I'd suggested to move it to the named constant. And perhaps the delay length could be changed to 16ms (60 FPS) without degradation of smoothness of scrolling, but not sure.

bvasilenko (approved):
Works as expected

PR #9182, opened 2023-11-08
NAS-124941 / 24.04 / Basic search on Audit

undsoft:
Testing:
Mostly code review, but you can search by Event column.
In basic mode, you can type.
In advanced mode for now, you need to press any key in the field and then search (query is fixed for now).


AlexKarpov98 (approved):
Looks good.
Some improvements:
- Would be better if we can submit search via ENTER button as well.
- And would be good to have `Switch to basic` and `Switch to advanced` focusable, so I can reach it via tab button.
Screenshot 2023-11-09 at 12 25 21

undsoft:
Will fix enter button in future PRs.

PR #9183, opened 2023-11-08
Add Checkbox to enable integrated metrics server

shadofall:
This will add Checkbox to give users a safe way to enable the integrated metrics server that apps like homepage in the IX community train can take advantage of. Currently the only way to enable the metrics server would be via CLI which if improperly done could result in the loss of user applications.


undsoft:
@sonicaj Is this something we want to do?

undsoft (commented):
Looks good to me, but let's wait for confirmation from @sonicaj.

sonicaj (approved):
LGTM!

PR #9186, opened 2023-11-09
NAS-125002 / 24.04 / Break table headers to new line in ix-table2

AlexKarpov98:
Result:
Screenshot 2023-11-09 at 14 59 17


dszidi (approved):
LGTM

image

PR #9188, opened 2023-11-09
NAS-125007 / 24.04 / Added tests for `RenewTwoFactorDialogComponent`

RehanY147:


src/app/interfaces/two-factor-config.interface.ts
```diff
@@ -1,20 +1,28 @@
...
+  provisioning_uri: string;
+  secret_configured: boolean;
+  interval: number;
+  otp_digits: number;
+}
+
+export interface UserTwoFactorConfigUpdate {
+  interval: number;
+  otp_digits: number;
+}
```
AlexKarpov98, line 28: empty line missing

AlexKarpov98 (changes requested):
Tests are failing

AlexKarpov98 (approved):
👍

PR #9192, opened 2023-11-12
NAS-125124 / 24.04 / Reports page is not showing preloader on initial load if I reload the page, I see empty page

AlexKarpov98:
Similar to https://github.com/truenas/webui/pull/9162
We are getting blank screen on reports page for a fraction of a second. 
Test that you see loaders and no empty screen.


dszidi (approved):
Looks good. Preloader shows on reload

PR #9193, opened 2023-11-12
NAS-124968 / 24.04 / Display problem with the French translation

AlexKarpov98:
Testing: see ticket 
Result:
Button are wrapped if text is not fitting for 1 row, as well added UI improvement with margin top for the buttons wrapper 

Screenshot 2023-11-13 at 11 22 17

before margin:
Screenshot 2023-11-13 at 00 39 59

### DIFF
--- a/src/app/pages/system/general-settings/save-config-dialog/save-config-dialog.component.scss
+++ b/src/app/pages/system/general-settings/save-config-dialog/save-config-dialog.component.scss
@@ -1,5 +1,16 @@
 :host {
   display: block;
   width: 360px;
-}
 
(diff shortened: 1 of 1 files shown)


undsoft:
Let's make the dialog wider as well to try to fit them on one row.

AlexKarpov98:
@undsoft - make sense, I updated it. (screenshot on the top of the page)

undsoft (approved):
::ng-deep is probably unnecessary, otherwise okay.

RehanY147 (approved):

PR #9194, opened 2023-11-13
NAS-125090 / 24.04 / Privileges management UI

bvasilenko:
**Testing**

See ticket


src/app/pages/account/groups/privilege/privilege-form/privilege-form.component.html
```diff
@@ -0,0 +1,53 @@
...
+        <ix-select
+          formControlName="roles"
+          [label]="'Roles' | translate"
+          [options]="rolesOptions$"
+          [multiple]="true"
+        ></ix-select>
+
+        <ix-checkbox
+          formControlName="web_shell"
+          [label]="'Web Shell' | translate"
```
undsoft, line 36: Let's call it 'Web Shell Access'

src/app/pages/account/groups/privilege/privilege-list/privilege-list.component.scss
```diff
@@ -0,0 +1,7 @@
+.table-container {
+  ::ng-deep ix-empty-row {
+    background: var(--bg2);
```
undsoft, line 3: Seems very specific.

src/app/pages/account/groups/group-list/group-list.component.html
```diff
@@ -13,6 +13,14 @@
     >
       <label>{{ 'Show Built-In Groups' | translate }}</label>
     </mat-slide-toggle>
+    <a
+      mat-button
+      color="primary"
```
undsoft, line 18: Let's make it a non-primary button and move it after Add button.

undsoft (changes requested):
Looks good overall and works okay, but I think we should make improvements for some of the fields in the form.

bvasilenko:
@undsoft requested changes should be addressed now

undsoft (approved):
Overall looks good. Two small items before we can merge:
1. Some built-in groups cannot be removed. Trying to remove them will result in an API error. We can hide delete button for those items based on whether `builtin_name` is set.
2. Privileges button needs to be in gray and go after `Show Built-in Groups` on groups page.

src/app/pages/account/groups/privilege/privilege-form/privilege-form.component.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  get title(): string {
+    return this.isNew
+      ? this.translate.instant('New Privilege')
+      : this.translate.instant('Edit Privilege');
+  }
+
+  readonly rolesOptions$ = this.ws.call('privilege.roles').pipe(
+    map((roles) => roles.map((role) =>({ label: role.title, value: role.name }))),
```
AlexKarpov98, line 47: **=>({**
Space is missing **=> ({**

PR #9195, opened 2023-11-13
NAS-124826 / 23.10.1 / Using rawvalue instead of value

RehanY147:
When editing a dataset, set value for the `Metadata (Special) Small Block Size` field in the `Advanced Fields`. Save the form successfully and then edit it again. The field should be populated with the last saved value.

### DIFF
--- a/src/app/pages/datasets/components/dataset-form/sections/other-options-section/other-options-section.component.ts
+++ b/src/app/pages/datasets/components/dataset-form/sections/other-options-section/other-options-section.component.ts
@@ -133,6 +133,7 @@ export class OtherOptionsSectionComponent implements OnInit, OnChanges {
     private formatter: IxFormatterService,
     private ws: WebSocketService,
     private datasetFormService: DatasetFormService,
+    private ixFormatterService: IxFormatterService,
   ) {}
 
   ngOnInit(): void {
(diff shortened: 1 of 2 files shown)


undsoft (changes requested):
It looks good, but while testing I've noticed another bug not related to your changes.
If I set field to 512B and then switch back to Inherit (0), it will not be correctly shown as inheriting.

RehanY147:
@undsoft If I set that value to Inherit and send that with the payload to middleware, later on, when the dataset is fetched again, it just reports the value to be 0. No mention that it was inherited.

undsoft (changes requested):
Middleware correctly reports value source as Default when you do this.
See `getFieldValue`.

src/app/pages/datasets/components/dataset-form/sections/other-options-section/other-options-section.component.ts
```diff
@@ -207,6 +208,11 @@ export class OtherOptionsSectionComponent implements OnInit, OnChanges {
       return;
     }
 
+    let specialSmallBlockSize = getFieldValue(this.existing.special_small_block_size, this.parent) as (number | 'INHERIT');
+    if (specialSmallBlockSize !== 'INHERIT') {
```
undsoft, line 212: It seems you are only interested in whether value was inherited. It's worth extracting relevant code from `valueOrInherit` into a new method.

undsoft (approved):
Works correctly now.

RehanY147:
The value changes based on if it is inherited or not. So, that code needs to stay I think.

PR #9196, opened 2023-11-13
NAS-125091 / 24.04 / Implementing a directive for hiding UI elements based on role

AlexKarpov98:
Testing: see that tests are passing or mock user in `auth.service`

```
  get user$(): Observable<LoggedInUser> {
    return this.loggedInUser$.asObservable();
  }
```
and test other scenarios.

Directive checks user role and allows/disallows rendering some of the components where we use it.


bvasilenko (approved):
Works as expected,

However I noticed discrepancy between `SharingWrite` and `SharingSmbWrite` in the role mapping: 

User roles taken from `roleMapping` not recursively
and
- from one side, `SharingWrite` does not include `SharingSmbRead`
- but from another side, `SharingSmbWrite` does include `SharingSmbRead`

**Expected:** unified approach to inclusion of "read" permissions into "write" permissions

AlexKarpov98:
@bvasilenko - nice catch.
I've added:
- auto Read permission if write provided
- recursiveness for roles

PR #9198, opened 2023-11-13
NAS-125008 / 24.04 / Apps: When a `show_if` tries to evaluate a field that is defined after it, results in both fields being hidden in the UI

AlexKarpov98:
Testing: 

See tests updated, the order of the schema items changed but still it should work as expected.
Currently it was not working - because if there were no control yet, we created new one (hidden by default) [what is wrong]

Please use m40 device and go here:
Go to Apps -> Discover -> Minio 👇 
http://localhost:4200/apps/available/TRUENAS/charts/minio/install

or Ask @stavros-k for an app if the one I provided is not there anymore.

Before: (we didn't see `Minio Certificate` even tho it has no **show_if** logic)
image

Now: (we see `Minio Certificate` and don't see `Minio Domain Name` because it depends on the Minio Certificate value and by default it's null, so we don't want to show Minio Domain Name in this case)
Screenshot 2023-11-13 at 19 06 50

### DIFF
--- a/src/app/modules/ix-forms/components/ix-input/ix-input.component.ts
+++ b/src/app/modules/ix-forms/components/ix-input/ix-input.component.ts
@@ -244,8 +244,8 @@ export class IxInputComponent implements ControlValueAccessor, OnInit, OnChanges
     });
 
     // handling initial value formatting from value to label
-    if (this.value !== undefined && this.value !== null) {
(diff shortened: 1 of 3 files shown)


dszidi (approved):
Tested this as per example in the ticket description.
Domain field on the Minio Install form now only shows when a certificate is actually selected. LGTM

PR #9200, opened 2023-11-14
NAS-125135 / 24.04 / Network gateway config popup offers skip, but is blocked by required

AlexKarpov98:
Testing: see ticket and latest comment, the idea is that when blur event is triggered - errors showup and the SKIP button moves down, so it's hart to click it correctly because of that.
NOW - Errors are only shown if value is not empty


undsoft (changes requested):
1. `inputBlur` is marked as deprecated and should not be used. 
2. Removing errors manually on blur is the same thing as simply removing required validator.
3. I wasn't even able to reproduce the original problem.

AlexKarpov98:
> 3. I wasn't even able to reproduce the original problem.
@undsoft - to reproduce - try to be focused on input and click SKIP 
Errors show up and then button changes it's position on UI, so click is misaligned and not counted.

AlexKarpov98:
> 1. `inputBlur` is marked as deprecated and should not be used.
> 2. Removing errors manually on blur is the same thing as simply removing required validator.
@undsoft  - what do you think about other possible solutions?

undsoft (changes requested):
If the issue only appears if user manually puts cursor inside of the field and does not appear if user presses Skip directly, I would just close the ticket.

AlexKarpov98:
@undsoft - well, the issue is that input is auto focused and once I click SKIP input's onBlur triggered and errors show up, SKIP button changes position and it's not clicked correctly.
Clicking it for the second time with errors - will be fine and correctly handled

AlexKarpov98:
@undsoft - check it out now
on initial blur - it will not raise an error 
on second blur - error appear


https://github.com/truenas/webui/assets/22980553/a23f2d05-05dd-49e6-97ba-34ef197ceac0

undsoft (changes requested):
It works, but I don't understand why. And unless you can explain why it works, I would prefer not to add code that we don't understand :)
I think the actual bug is in the fact that `IxInputComponent` calls `onChange` in `blurred` regardless of whether value was changed or not. Whether fixing this will break something else – this needs testing, but this is a more sustainable solution.

AlexKarpov98:
@undsoft - I thought about the `change` - it’s default behaviour (every input event is counter, so focus, blur are counted in this case, and error can be shown)

So we'll get a REQUIRED error anyway after loosing focus, it works with any input.
So if it's required you focused it and unfocused - you see REQUIRED error.

image

AlexKarpov98:
@undsoft  in this case, I think, we can close the ticket (it's default input, form validation logic) or do something I proposed before, show error only when field is not empty.

AlexKarpov98:
@undsoft - now it works as expected

undsoft (changes requested):
I am not sure this is correct. I am fairly certain that it's Angular's job to decide whether control is dirty and how to utilize the update strategy. Our job is to correctly implement   `ControlValueAccessor` and call `onChange`/`onTouch`.

Maybe try to compare whether new value is different on change?

We will also need test cases for this.

AlexKarpov98:
@undsoft - now it should work as expected.

undsoft (approved):
Thanks, looks good.
Nice job on taking time to figure out and fix the root cause.

You may want to update `lastKnownValue` in other cases in `ix-input` where `onChange` is called.

PR #9201, opened 2023-11-14
NAS-125213 / 24.04 / From wizard to Advanced options in card/list are not updating on new items creation (switching from wizard to form)

AlexKarpov98:
Testing: see ticket


bvasilenko (approved):
Works as expected

PR #9202, opened 2023-11-14
NAS-124941 / 24.04 / Advanced search query parser

undsoft:
Testing: go to /audit and type in advanced search.
You can type using sql-like syntax: `Username = "Bob" OR Age > 29`
As you type, you can see how the query is being parsed in console.


AlexKarpov98 (approved):
Works good! 👍 
Screenshot 2023-11-15 at 17 07 34
Screenshot 2023-11-15 at 17 07 37

PR #9205, opened 2023-11-15
NAS-124666 / 24.04 / Create new tickets with FeedbackDialog

denysbutenko:
- Add new component IxButtonGroup to ix-forms
- Merge FileTicketForm into Feedback Dialog
- Merge FileTicketLicensedForm into Feedback Dialog

For testing, ensure no regressions for feedback and ticket creation.


RehanY147 (changes requested):
The two forms are large enough that they should be their own smaller components to be used inside of the dialog component.

denysbutenko:
> The two forms are large enough that they should be their own smaller components to be used inside of the dialog component.

I wonder if that suggestion is a good idea. However, there is another PR with implementation using multiple components approach #9256. Please take a look at it.

denysbutenko:
Closing this PR since is outdated. See https://github.com/truenas/webui/pull/9256

PR #9207, opened 2023-11-15
NAS-125209 / 24.04 / Pools missing in GUI after manually initiating scrub tasks on some pools

AlexKarpov98:
Testing: see ticket. Code review is enough.


bvasilenko (approved):
Looks good

PR #9209, opened 2023-11-16
NAS-125192 / 24.04 / Hide "Mixed Capacity Vdevs" alert for vdevs with ≤2GiB difference eg: vdevs created with different SWAP settings

bvasilenko:
**Testing**

1. Go to **System > Advanced > Storage**, reset **Swap Size** to 2Gb (default value)

2. Create pool with 1 VDEV

2. Adjust **Swap Size** to a different value, for example 1Gb

4. Add another VDEV to your pool

**Expected:** No warnings next to "mixed capacity" at **Storage > Topology**


denysbutenko (approved):
Consider adding a test case for this. Looks good.

PR #9213, opened 2023-11-16
NAS-125246 / 24.04 / Use Translated Names for Privilleges

AlexKarpov98:
Testing: 
You should now see it updated http://localhost:4200/credentials/groups/privileges:

Screenshot 2023-11-16 at 15 23 27
Screenshot 2023-11-16 at 15 23 44


undsoft (changes requested):
That's cool, but strings are not actually translated, only marked for translation (translation pipe is not called).
See `translateOptions` and other methods.

AlexKarpov98:
> That's cool, but strings are not actually translated, only marked for translation (translation pipe is not called). See `translateOptions` and other methods.

ohhh, really, I didn't notice there are no translations added actually, my bad.

AlexKarpov98:
> That's cool, but strings are not actually translated, only marked for translation (translation pipe is not called). See `translateOptions` and other methods.

@undsoft  - I checked, we already have these translations in `.json` files.
I used them previously for a [directive for hiding UI elements based on role](https://ixsystems.atlassian.net/browse/NAS-125091)

undsoft:
@AlexKarpov98 They are in .json files because you have marked them for translation with T. They will not be in fact translated if you switch truenas language to something else.

AlexKarpov98:
@undsoft - now it's using translation strings 😃

undsoft (approved):
Thanks, looks good.

PR #9216, opened 2023-11-17
Connect audit table state to page URL

bvasilenko:
**Testing**

See ticket

### DIFF
--- a/src/app/modules/ix-table2/api-data-provider.ts
+++ b/src/app/modules/ix-table2/api-data-provider.ts
@@ -154,6 +154,7 @@ export class ApiDataProvider<T, M extends ApiCallMethod> extends BaseDataProvide
       () => this.load(),
       () => this.updateCurrentPage(this.rows),
     );
+    this.controlsStateUpdated.emit();
   }
 
   setPagination(pagination: TablePagination): void {
@@ -162,6 +163,7 @@ export class ApiDataProvider<T, M extends ApiCallMethod> extends BaseDataProvide
       () => this.load(),
       () => this.updateCurrentPage(this.rows),
     );
+    this.controlsStateUpdated.emit();
   }
 
   private prepareParams(params: ApiCallDirectory[M]['params']): ApiCallDirectory[M]['params'] {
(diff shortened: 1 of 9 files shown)


AlexKarpov98 (commented):
Screenshot 2023-11-18 at 23 37 14
when I visit Audit page directly - I get errors.
@bvasilenko

denysbutenko (changes requested):
The audit page layout is broken. Getting same error as Alex mentioned.

```error-handler.service.ts:37 TypeError: Cannot read properties of undefined (reading 'pageSize')
    at template (audit.component.html:69:36)```

bvasilenko:
@denysbutenko the issue should be fixed now, would you please take a look again?

denysbutenko (approved):
The audit page has been successfully loaded, and the corresponding URL is as follows:

`http://localhost:8080/ui/audit/%7B%22sorting%22%3A%7B%22propertyName%22%3Anull%2C%22direction%22%3Anull%2C%22active%22%3Anull%7D%2C%22pagination%22%3A%7B%22pageNumber%22%3A1%2C%22pageSize%22%3A50%7D%7D`

Is it possible to unquote query parameters? Maybe [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) can be applied here.

Also, the ticket mentioned: `When user updates filters (support both basic and advanced mode), column sorting or pagination update URL.` 
It makes me think that query params should be empty unless user applies filters. Please clarify if it's okay to have them by default.

Otherwise, it meets the requirements. 👍🏻

denysbutenko (commented):
The search input's clear button seems non-functional in this PR. Initially, I thought it might not have been implemented yet. However, during testing #9222, I observed that the clear button is functional. Could this be a regression? Kindly verify this behavior.

bvasilenko:
@denysbutenko this is not a regression, clear button only clears the text from the input if user has clicked the Search button.


https://github.com/truenas/webui/assets/20611516/e5ebe342-6399-4283-8b19-c069636480c5

PR #9219, opened 2023-11-18
NAS-124454 / 24.04 / Reinitializing stores after sign-in

RehanY147:


undsoft (changes requested):
This ties multiple stores to an AuthService, which is not something these store really should care about it.
I would try:
1. Not calling `initialize` in constructor and instead calling it directly in components that utilize the store.
2. Check if [providing Store in component providers](https://ngrx.io/guide/component-store/lifecycle#ondestroy) can solve our issues. Keep in mind that there may be a conflict betwen `UntilDestroy()` decorator and `ngOnDestroy` in the store itself.
3. If this doesn't work well, `initialize()` returns a subscription that you could use to unsubscribe manually in the parent component.

PR #9220, opened 2023-11-19
NAS-125229 / 24.04 / Query builder

undsoft:
Testing: go to /audit and use advanced search.
You can inspect console to make sure that parsed query is correct, but also use it in real API requests.

Note:
* Human readable names are not supported. You need to use properties from `BaseAuditEntry`, e.g. `username` not `Username`.
* Syntax errors won't be shown.


bvasilenko (approved):
Works well for the most common cases. 

While testing, I've noticed a few issues with non-standard operators. These may be addressed in either this PR or another one.

1. list-accepting operators `in`, `nin`, `rin`, `rnin` convert only 1st item from the list, for example `event in CREATE,CONNECT` expected to be converted into `["event", "in", ["CREATE", "CONNECT"]]` filter query. 
2. params builder does not yet support some operators listed at `/api/docs/` :
   - case insensitivity `C` prefix for an operator , for example  `C=`
3. pagination is broken, since `{count: true}` request does not include the `query-filter`.

undsoft:
> list-accepting operators in, nin, rin, rnin convert only 1st item from the list
The correct syntax is `event in ("CONNECT", "WRITE")`. The reason why `event in "CONNECT"` works is because there is currently no validation as to which operators can be used with what operands.

> params builder does not yet support some operators listed at /api/docs/ 
Will create a separate ticket for this.

> pagination is broken, since {count: true} request does not include the query-filter.
Fixed.

PR #9221, opened 2023-11-19
NAS-124970 / 24.04 / Add smb status shares table

denysbutenko:
Changes:

- Added SMB status shares table
- Moved table-related controls from the page header on SMB Sessions. Like on iSCSI pages.
- Added unit tests

For testing, create an SMB share and connect to it. Then go to shares dashboard - SMB status - shares tab.


src/app/pages/sharing/smb/smb-status/components/smb-share-list/smb-share-list.component.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  loadData(): void {
+    this.dataProvider.load();
+  }
+
+  onListFiltered(query: string): void {
+    this.filterString = query?.toString()?.toLowerCase();
+    this.dataProvider.setRows(this.shares.filter((session) => {
+      return [
+        session.session_id,
```
AlexKarpov98, line 71: maybe we can add more fields to be searched, is that useful?

AlexKarpov98 (commented):
Looks good.
I was able to test it with mock data.

According to our existing breadcrums - please remove that last part showing current page. `Smb Status`. It should be `Sessions -> SMB` 

Screenshot 2023-11-24 at 13 22 38

____

Another question is about these TABS:
Screenshot 2023-11-24 at 13 33 11

On NFS sessions page we merged that toggle group button:
Screenshot 2023-11-24 at 13 34 49

@dszidi @undsoft - what do you think about tabs, should we use them here?

undsoft:
I would not use button group here. Button group works on other pages, because it switches between different kinds of the same thing (e.g. sessions of different type). 
Tabs are more appropriate for switching between different things.

AlexKarpov98 (approved):
Looks good, thanks.

PR #9223, opened 2023-11-20
NAS-125259 / 13.0 / ES60G2 fails to be displayed in UI

AlexKarpov98:
Testing: 
See ticket, use `r50bm` machine from the ticket.
Visit http://localhost:4200/system/viewenclosure 
And see that it's loading. (Disks on TRUENAS-R50BM (0))


dszidi (approved):
LGTM

PR #9225, opened 2023-11-20
NAS-124951 / 23.10.1 / Remove required for periodic_snapshot_tasks, Disable ssh_credentials when transport is local

bvasilenko:
**Summary**

1. Ticket's video contains a mandatory field `periodic_snapshot_tasks` which was not filled. This field is not mandatory in the middleware, so I made `periodic_snapshot_tasks` not mandatory in WebUI as well
2. Ticket's screenshots depict expectation of the required ssh credentials field, for LOCAL, where obviously this should not be there. This is fixed in a 2nd commit

**Testing**
Data Protection > Replication Tasks > Add > Advanced Replication Creation


undsoft (approved):
Looks good, thanks.

PR #9229, opened 2023-11-21
NAS-124932 / 24.04 / Support for empty mointpoint

undsoft:
In dataset-store.service.ts loadDatasets set mountpoint to null for some datasets, for example via:
```
datasets: datasets.map(dataset => ({ ...dataset, mountpoint: null })),
```
(this affects root dataset only).
Test that UI doesn't break.


bvasilenko (approved):
Works as expected 

image

PR #9236, opened 2023-11-23
NAS-124602 / 23.10.1 / Reload UI during updates

undsoft:
Testing: update a system and see that UI reloads after update is complete.
Background tabs and tabs in other browser should also reload.


denysbutenko (approved):

undsoft:
@denysbutenko Thanks!

PR #9248, opened 2023-11-27
Update fr.json

MDVAurelien:
Update some French translation and fix error.  In fact the word dataset is not really translatable so it is better to leave it as is.


undsoft (approved):
Thanks!

PR #9249, opened 2023-11-27
NAS-125368: App Details Card is updating Workloads card many times - it is not possible to click "shell" or "view logs" sometimes because of it

AlexKarpov98:
**Testing:** see ticket and result:

https://github.com/truenas/webui/assets/22980553/72dfe5b4-1e68-4af0-898c-19624d86a2cd

Now there are no more containers disappearing each time we reload data - what is much better for the user.

Before: Each time on reload card became smaller and skeleton appeared - which is really annoying for the end user I suppose.
Screenshot 2023-11-27 at 12 02 58


dszidi (approved):
Tested this by looking at the n8n app on the team's M40. The containers remain visible and  by extension the shell and logs buttons can be clicked.

PR #9251, opened 2023-11-27
NAS-125369: Replace all `[key: string]: TYPE` with Record<string, TYPE>

AlexKarpov98:
Code review is enough.


undsoft (approved):
It's okay (though I'm not sure about what the differences are).
If we are going to replace these, it's a good idea to enable https://typescript-eslint.io/rules/consistent-indexed-object-style/

PR #9257, opened 2023-11-28
NAS-125399 / 24.04 / Add Button Group Component to IxForms

denysbutenko:
We agreed to split PRs into smaller parts for easier reviews. 
Extracted from PRs #9205 #9256 
There is no actual usage of this component yet. Tests should pass and code review is enough.

### DIFF
--- a/src/app/modules/ix-forms/components/ix-button-group/ix-button-group.component.spec.ts
+++ b/src/app/modules/ix-forms/components/ix-button-group/ix-button-group.component.spec.ts
@@ -1,0 +1,58 @@
+import { NgControl } from '@angular/forms';
+import { MatButtonToggleChange, MatButtonToggleModule } from '@angular/material/button-toggle';
+import { Spectator, createComponentFactory } from '@ngneat/spectator/jest';
+import { MockComponent } from 'ng-mocks';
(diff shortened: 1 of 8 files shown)


src/app/modules/ix-forms/components/ix-button-group/ix-button-group.component.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    expect(spectator.component.isDisabled).toBe(true);
+  });
+
+  it('should emit value change', () => {
+    const onChangeSpy = jest.spyOn(spectator.component, 'onChange');
+    spectator.component.onValueChanged({ value: 'test' } as MatButtonToggleChange);
+    expect(onChangeSpy).toHaveBeenCalledWith('test');
+  });
+
+  describe('setDisabledState()', () => {
```
undsoft, line 34: I think ideally you wouldn't be calling these methods. These could be set to `protected`.

A better approach is to create a `FormControl` to interact with (set value, read value, set disabled) and check how it communicates with mat-button-toggle.
For example, see `IxInputComponent`.

undsoft (approved):
Button group looks and works okay, but I would update the style of tests.

PR #9259, opened 2023-11-28
NAS-125122 / 23.10.2 / Only accept events with truenas.com origin

RehanY147:
Go to add a new Cloud Backup Credential. Select Yandex as the provider and click the "Log in to Provider" button. Go through the process that follows and log in to Yandex. Once done successfully, you should be redirected to UI and the token and client id and secret fields should be filled in. You should be able to submit the form.


undsoft:
backport

PR #9261, opened 2023-11-29
NAS-125373 / 24.04 / Apply column and value transformations to audit advanced search

AlexKarpov98:
Testing: 
see tests & ticket & browser console.

Screenshot 2023-12-01 at 00 57 31


src/app/modules/search-input/services/advanced-search-autocomplete.service.ts
```diff
@@ -81,17 +82,22 @@ export class AdvancedSearchAutocompleteService<T> {
     if (
       /^["'].*["']?$/.test(currentQuery[from - 1]) && /^["'].*["']?$/.test(currentQuery[to])
     ) {
-      updatedValue = updatedValue.replace(/['"]/g, '');
+      updatedValue = updatedValue.replace(/^["'](.*)["']$/, '$1');
```
bvasilenko, line 85: 1. I've noticed that the regular expression `^["'](.*)["']$` is used to detect a string surrounded by double quotes `"` and single quotes `'` at the same time, giving a false success for mixed qoutes `"test'`. I'd suggested to have 2 separate regular expressions for single quotes `'` and double quote `"`.
2. Also, looks like same regular expression `^["'](.*)["']$` is used to detect a string surrounded by quotes giving a false success for `"test"test"`. I've suggested to replace `.*` with `[^"]` for double quotes and with `[^']` for single quotes, so that the string could not contain a closing quote. 
3. It might be reasonable to extract the regular expression into a reusable constant
4. Not completely sure about this, but we could update tests to include a couple of invalid user's input similar to the examples in "1" and "2"

bvasilenko (approved):
Works as expected. I've just suggested a few code improvements

src/app/modules/search-input/services/advanced-search-autocomplete.service.ts
AlexKarpov98: I've updated code, in this case autocompletion will not work if you try to wrap it in unmatched quotes like:
'hello"
And query will be invalid without transformations:

double quotes -- works:
Screenshot 2023-12-04 at 12 11 03

single quotes -- works:
Screenshot 2023-12-04 at 12 11 09

mixed quotes -- error and no suggestions:
Screenshot 2023-12-04 at 12 11 38 1

"test"test" -- will also fail and blocks transformations

So the transformation method is only called once the query counts as VALID.
Autocomplete suggests when quotes are matching, I don't think that user will do something like **"hello'** or **"he"llo"** -- we have auto-closing quotes which will enforce this should not happen.

PR #9263, opened 2023-11-29
NAS-124964 / 24.04 / Only server side pagination/sorting in ApiDataProvider

RehanY147:
Separate `ApiDataProvider` from `AsyncDataProvider`

This PR separates responsibilities of `ApiDataProvider` from `AsyncDataProvider`. `AsyncDataProvider` will be responsible for handling the pagination, sorting and filtering on the front-end.

`ApiDataProvider` will be responsible for handling the pagination, sorting and filtering on the back-end. `AsyncDataProvider` already has default sorting and pagination from `BaseDataProvider`


bvasilenko (approved):
responsibilities `AsyncDataProvider` and `ApiDataProvider` are split as expected

PR #9264, opened 2023-11-30
NAS-125305 / 24.04 / Create Dataset button in SMB form

bvasilenko:
**Testing**
See ticket

### DIFF
--- a/src/app/pages/sharing/smb/smb-form/smb-form.component.ts
+++ b/src/app/pages/sharing/smb/smb-form/smb-form.component.ts
@@ -22,7 +22,7 @@ import {
   take,
   tap,
 } from 'rxjs/operators';
-import { DatasetShareType } from 'app/enums/dataset.enum';
+import { DatasetPreset } from 'app/enums/dataset.enum';
 import { ServiceName } from 'app/enums/service-name.enum';
 import { ServiceStatus } from 'app/enums/service-status.enum';
 import { helptextSharingSmb } from 'app/helptext/sharing';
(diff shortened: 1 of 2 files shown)


src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -70,6 +71,10 @@ export class SmbFormComponent implements OnInit {
 
   title: string = helptextSharingSmb.formTitleAdd;
 
+  createDatasetProps = {
+    acltype: DatasetAclType.Nfsv4,
```
undsoft, line 75: There is a ShareType property to be used here instead. It's like a preset.
For SMB it should be `DatasetShareType.Smb` and for NFS - `DatasetShareType.Multiprotocol`.

src/app/modules/ix-forms/components/ix-explorer/ix-explorer.component.ts
```diff
@@ -36,6 +39,8 @@ export class IxExplorerComponent implements OnInit, OnChanges, ControlValueAcces
   @Input() required: boolean;
   @Input() root = mntPath;
   @Input() nodeProvider: TreeNodeProvider;
+  @Input() canCreateDataset = false;
```
undsoft, line 42: Also needs a test case for new properties.

undsoft (changes requested):
It works, but we can make some improvements.

1. I should only be able to create a dataset when what I have currently selected is a dataset and not a folder (created via mkdir for example) or a file (with getFilesystemNodeProvider directoriesOnly = false).
2. Let's add background to the area underneath Create Dataset like in screenshot in the ticket.
3. Let's make the dialog a bit wider.
4. If I select /mnt/parent and not expand it, after dataset is created via dialog, it would be good to expand, so that I can see my selection.

bvasilenko:
@undsoft the look and feel of the button is closer to the design now:

image

undsoft (approved):
Looks good, let's maybe do something about progressbar styles so that content doesn't jump when it appears.

PR #9270, opened 2023-12-02
NAS-125458: Expanding rows on the Groups page fails

AlexKarpov98:
Testing: 
see ticket, there was problem with Same Component Class in different modules imported into another module. 
In this case -- `IxTableExpandableRowComponent`

So now for ix-table2 I renamed `IxTableExpandableRowComponent` --> `IxTable2ExpandableRowComponent`


undsoft (approved):
It works, but doesn't selector in group-details-row html need to be changed to table2?

AlexKarpov98:
> While testing on m40 and on the Groups page, I'm getting the following error:
> 
> ```
> 
> Error: Parameter "key" required
> 
>     at gt.instant (ngx-translate-core.mjs:578:19)
> 
>     at group-list.component.ts:154:47
> 
> ```
> 
> If error not appears try to toggle 'Show Built-In Group`

@denysbutenko this error is fixed in another PR, I'll pull changes and let you know

denysbutenko (approved):

AlexKarpov98:
> It works, but doesn't selector in group-details-row html need to be changed to table2?

@undsoft - sure, I replaced it, it works.

PR #9272, opened 2023-12-02
NAS-125510: Privileges page JS error

AlexKarpov98:
Testing: see ticket and check that there are no errors in console.

JS error was being raised because a key that was not present in a map was provided.


undsoft (approved):
👍🏻

PR #9273, opened 2023-12-02
NAS-125438: Fix step checking

RehanY147:
Check the correct step for existing pools

Create a new pool with a metadata vdev.
Go to edit the pool devices and see if the metadata step has the correct vdev type that was assigned when creating the pool originally.


undsoft (approved):
I was able to create and edit a pool with metadata vdevs.

PR #9274, opened 2023-12-03
NAS-125508 / 24.04 / Checking for undefined values

RehanY147:


src/app/services/network.service.ts
```diff
@@ -70,7 +70,8 @@ export class NetworkService {
 
   subscribeToInOutUpdates(): Observable<AllNetworkInterfacesUpdate> {
     return this.ws.subscribe('reporting.realtime').pipe(
-      map((event) => event.fields.interfaces),
+      map((event) => event.fields?.interfaces || null),
```
undsoft, line 73: || null is unecessary

undsoft (approved):
Well, the fix works, but do we know why the issue is happening? Is there a middleware error or something?

RehanY147:
@undsoft It could be just an event from `reporting.realtime` that doesn't have any updates for `interfaces`. UI tries to read what isn't there?

PR #9278, opened 2023-12-03
NAS-125507 / 24.04 / Show username in top bar

AlexKarpov98:
Testing: see ticket result 👇 :
Screenshot 2023-12-04 at 00 23 13


bvasilenko (approved):
Looks good, just one adjustment needed:

Let's apply `cursor: pointer` to the username, similar to cursor applied to all other buttons.

PR #9283, opened 2023-12-04
NAS-125502 / 24.04 / Don't check for mixed capacity

RehanY147:
On the manual disk selection modal, try adding disks of different sizes to a vdev. You should see a warning that this kind of configuration where you have mixed capacity is not recommended. But you should still be able to save the selection.


william-gr:
backport

PR #9284, opened 2023-12-04
NAS-125509 / 24.04 / Refactoring hasRoles directive

undsoft:
You won't be able to do any meaningful tests because other things are broken, so code review only.


bvasilenko (approved):
LGTM

PR #9286, opened 2023-12-04
Fix typo dataset-form.ts form translation

MDVAurelien:
Hi guys, the translation of dataset_form_name_tooltip tooltip does not work. (French translate OK in json). I think it's because there are the characters \\' several times in the sentence. I hope this will fix the bug.

translate1

### DIFF
--- a/src/app/helptext/storage/volumes/datasets/dataset-form.ts
+++ b/src/app/helptext/storage/volumes/datasets/dataset-form.ts
@@ -2,7 +2,7 @@ import { marker as T } from '@biesbjerg/ngx-translate-extract-marker';
 
 export default {
   dataset_parent_name_tooltip: T('Parent dataset path (read-only).'),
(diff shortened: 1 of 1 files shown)


undsoft (changes requested):
I don't know. I think it's unlikely to be because of slashes. 
And tooltip is translated for me in `master` already.

Снимок экрана 2023-12-05 в 12 30 56

MDVAurelien:
It's strange I tested it on MASTER 20231130 and it didn't work but I didn't have a more recent MASTER. The MASTER can be retrieved here: https://download.truenas.com/truenas-scale-cobia-nightly/?
It's strange I tested it on MASTER 20231130 and it didn't work but I didn't have a more recent MASTER. The MASTER can be retrieved here: https://download.truenas.com/truenas-scale-cobia-nightly/ ?

PR #9289, opened 2023-12-07
NAS-125376 / 24.04 / Fix changes and emit in AdvancedSearchComponent

AlexKarpov98:
Testing: 
see ticket.
- check that once query is valid - you can change from ADVANCED to BASIC mode and vice versa, data is not lost.
- check that once we have query params in place, page is reloaded for example - input should be pre-filled from query params.

example:

https://github.com/truenas/webui/assets/22980553/9ee7c31b-af78-4d97-8acd-7f79ccd7b6ca


dszidi (approved):
Looks good. Submitted a valid query and was able to toggle between basic and advanced modes without losing the results in the table.

image

image

PR #9291, opened 2023-12-07
NAS-125497 / 24.04 / Fix what happens when trying to use user with reduced privileges

undsoft:
Testing:
1. Create a user and a group.
2. Assign READONLY or SHARING MANAGER privilege to the group.
3. Login as user.

Known issues
- You may get Access denied errors.
- 2FA may be broken.


bvasilenko (approved):
Directive works on "Shares" page as expected for readonly-privileged user.

("Sharing Manager" role seems to be not supported yet on the latest nightly.)

PR #9293, opened 2023-12-07
NAS-125610 / None / Allow full admins to access system settings without meeting 2FA requirements

undsoft:
Testing: 
1. Enable 2FA in global settings as an admin.
2. Observe that you can still access System settings pages, but other normal pages will require you to setup your 2FA device.


AlexKarpov98:
@undsoft  I am not able to get to the System Settings pages, I get warning
 
Screenshot 2023-12-08 at 12 09 18

https://github.com/truenas/webui/assets/22980553/62629456-b449-4df7-9d5c-6d79488b8e01

As well I used `Authy` app for scanning QR code, and now I can't login to the app, I get `Two-Factor Authentication Code is required` error

undsoft:
@AlexKarpov98 Weirdly, I had the same issue, but after reloading without cache, it worked.

PR #9294, opened 2023-12-07
NAS-125484 / 24.04 / Incorrect When time displayed on periodic snapshot tasks screen

AlexKarpov98:
Testing: 
See ticket and result and test case: 👇 
Screenshot 2023-12-07 at 17 38 22


RehanY147 (approved):
Looks good! This would be a good opportunity to add a couple of example inputs in the hours and minutes field when setting a custom period for the snapshots. As is, it is not easy to understand what you can put in those fields.

PR #9295, opened 2023-12-07
NAS-125612 / None / Simplify mocking of user roles

undsoft:
Also fixes smaller bugs.

Testing: mostly code review.


denysbutenko (approved):
Looks good

src/app/services/auth/auth.service.ts
```diff
@@ -172,18 +172,24 @@ export class AuthService {
...
 
-    if (currentRoles.includes(Role.FullAdmin)) {
-      return true;
-    }
+        if (currentRoles.includes(Role.FullAdmin)) {
+          return true;
+        }
 
-    return roles.some((role) => currentRoles.includes(role));
+        return roles.some((role) => currentRoles.includes(role));
```
RehanY147, line 190: Is there a scenario where all the roles being checked must exist? Not at least one?

RehanY147 (approved):

RehanY147 (changes requested):
Shouldn't `hasRole` wait for the user to load? The user loads as soon as the user signs in. So, at most, the wait would be a couple of seconds. As is, the user of the service might have to add a check themselves for the user object emitted by `AuthService` to be truthy.

src/app/services/auth/auth.service.ts
undsoft: Not that I'm aware of.

undsoft:
It's a good point, but I don't think `hasRole` has to wait for the user necessarily. The way code is currently structured, user won't get redirected inside before user object is loaded (I think).
I will also refactor `refreshUser` to wait when it is called inside the admin panel.

PR #9296, opened 2023-12-07
NAS-125536 / 24.04 / Replace Auxiliary Groups with chips input

denysbutenko:
- Add support for Option interface for ix-chips
- Replace `Auxiliary Groups` with `ix-chips` input in the user form

For testing, try to edit user groups add/edit/remove using the chips component.


dszidi (approved):
Looks good!

PR #9299, opened 2023-12-09
NAS-124716 / 24.04 / Corrected test attribute for ix-table2 columns

RehanY147:


src/app/pages/account/users/user-list/user-list.component.ts
```diff
@@ -63,7 +63,9 @@ export class UserListComponent implements OnInit {
         .join(', ') || this.translate.instant('N/A'),
       sortable: true,
     }),
-  ]);
+  ], {
+    rowTestId: (row: User) => row.uid.toString(),
```
undsoft, line 67: We should also include what it is in test id, e.g. `user-1`, not just `1`

undsoft (changes requested):
The direction is good. There are some things to improve to make various tables differentiate between themselves.

denysbutenko (commented):
CI checks are stopped - I need the latest docker image for testing.

undsoft (approved):
Looks good, thanks. I'm sure automation engineers will appreciate it.

PR #9307, opened 2023-12-11
NAS-125608 / 24.04 / Do not let users with insufficient roles to login to UI

undsoft:
Testing: login in various scenarios (normal user, user with reduced privileges, etc).


dszidi (approved):
Testing on a CI machine, everything seems to work just fine.
Users belonging to groups with readonly role are met with the following error dialog upon form submission

image

dszidi:
@undsoft I tested it again after our discussion in slack. It works as expected. This is a nitpick that won't affect approval but I think the error message should say something something along the lines of "Insufficient privilieges to log in. Check with your administrator".
Currently it states the username or password is incorrect which is misleading

image

PR #9310, opened 2023-12-12
NAS-125570 / 24.04 / Allow privilege to be attached in group form

AlexKarpov98:
Testing:
Go to Groups page:
http://localhost:50131/credentials/groups

You should see `Privileges` input:
Screenshot 2023-12-12 at 18 03 54

Result:

https://github.com/truenas/webui/assets/22980553/97114292-a9ab-49ed-8bd0-e34c061afc22


denysbutenko:
Hey @AlexKarpov98,

Here are some updates to ix-chips. You can see what we've done in this pull request: https://github.com/truenas/webui/pull/9296.

In short, ix-chips can now resolve number values from passed options objects. This means we can pass form control with group ids `number[]` values and then display them as group names `string[]` and then save data as group ids `number[]`.

Please take a look when you can. This new 'resolveValue' feature might be helpful for your work. Feel free to let me know if you have any questions.

denysbutenko (approved):
It works great for me, actually.

src/app/pages/account/groups/group-form/group-form.component.ts
```diff
@@ -68,8 +77,19 @@ export class GroupFormComponent implements OnInit {
 
   ngOnInit(): void {
     this.setupForm();
+    this.getInitialPrivileges();
   }
 
+  readonly privilegesProvider: ChipsProvider = (query: string) => {
+    return this.ws.call('privilege.query', []).pipe(
```
RehanY147, line 84: This query should be sent to middleware. Instead of being applied locally once the response has been fetched. Otherwise, we don't need to make the same request and fetch the same results every time query changes.

src/app/pages/account/groups/group-form/group-form.component.ts
```diff
@@ -158,6 +169,81 @@ export class GroupFormComponent implements OnInit {
     });
   }
 
+  private togglePrivilegesForGroup(groupId: number): Observable<Privilege[]> {
+    const requests$: Observable<Privilege>[] = [];
+
+    if (this.selectedPrivilegeIds) {
+      const privileges = this.privilegesList
```
RehanY147, line 176: Is this filter required? User should only be allowed to select valid privilege values when they are editing the form.

src/app/pages/account/groups/group-form/group-form.component.ts
```diff
@@ -158,6 +169,81 @@ export class GroupFormComponent implements OnInit {
...
+              [...privilege.local_groups.map((group) => group.gid), groupId],
+            ),
+          ]),
+        );
+      });
+    }
+
+    if (this.existingPrivilegesRemoved) {
+      const privileges = this.privilegesList
+        .filter((privilege) => this.existingPrivilegesRemoved.some((privilegeId) => privilege.id === privilegeId));
```
RehanY147, line 194: Same here

RehanY147 (changes requested):

src/app/pages/account/groups/group-form/group-form.component.ts
AlexKarpov98: @RehanY147  - what do you mean?
this is the provider for chips component?

src/app/pages/account/groups/group-form/group-form.component.ts
AlexKarpov98: @RehanY147  - yes, how without filter we'll know what was added and what was removed?
How to save data eventually?

src/app/pages/account/groups/group-form/group-form.component.ts
AlexKarpov98: @RehanY147  - how do you expect to know which privilege was removed?
We need to also update the PRIVILEGE where the group was removed from the privilege.

src/app/pages/account/groups/group-form/group-form.component.ts
AlexKarpov98: > This query should be sent to middleware. Instead of being applied locally once the response has been fetched. Otherwise, we don't need to make the same request and fetch the same results every time query changes.

we have the same for privileges form, we need to grab all the items once and filter locally I suppose

RehanY147 (approved):

PR #9312, opened 2023-12-13
NAS-125654 / 24.04 / Make `Rate this page` to be conditional

denysbutenko:
For testing, check the feedback dialog on the topbar or the support card.


src/app/modules/ix-feedback/ix-feedback.service.ts
```diff
@@ -54,4 +56,21 @@ export class IxFeedbackService {
       });
     });
   }
+
+  checkIfFeedbackAllowed(): Observable<boolean> {
+    return combineLatest([
+      this.ws.call('system.info'),
```
undsoft, line 62: Some of this info should be already in the store.

undsoft (changes requested):
It works, but there are some things to improve:
1. UI jumps when you open it when feedback is disabled. We need to solve this somehow, for example by not showing feedback options by default and showing a loading indicator, or by loading blacklist status beforehand.
2. When feedback is allowed, form still starts on Report a bug. It should start with feedback screen.

denysbutenko:
> It works, but there are some things to improve:
> 
> 1. UI jumps when you open it when feedback is disabled. We need to solve this somehow, for example by not showing feedback options by default and showing a loading indicator, or by loading blacklist status beforehand.
> 2. When feedback is allowed, form still starts on Report a bug. It should start with feedback screen.

Okay, I changed its behavior to check blacklist status beforehand (on feedback-service initialization). The second issue is fixed.

undsoft (approved):
Looks good, tested with mocks.

PR #9313, opened 2023-12-13
NAS-125670 / 24.04 / Use `user.set_password` API for user password reset

undsoft:
Testing: test password reset form.


denysbutenko (approved):
Looks good

PR #9314, opened 2023-12-13
NAS-125301 / 24.04 / Additional validation when creating a share

AlexKarpov98:
Testing: go to Shares
http://localhost:4200/sharing

Try to add SMB share: 
- with existing name
- if there are no SMB users

You should see:
image


src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -152,7 +153,10 @@ export class SmbFormComponent implements OnInit {
 
   form = this.formBuilder.group({
     path: ['', Validators.required],
-    name: ['', [Validators.required]],
+    name: ['', {
+      validators: [Validators.required],
+      asyncValidators: [this.smbNameValidator.bind(this)],
```
bvasilenko, line 158: Noticed that you've used `.bind` , however it's not widely used across the project (only 23 usages of `.bind` in all code files).

I'd suggested to use arrow function `() => ` instead.

src/app/pages/sharing/smb/smb-form/smb-name-validator.service.ts
```diff
@@ -0,0 +1,70 @@
...
+    private ws: WebSocketService,
+    private translate: TranslateService,
+  ) { }
+
+  private noSmbUsersError = this.translate.instant('TrueNAS server must be joined to Active Directory or have at least one local SMB user before creating an SMB share');
+  private nameExistsError = this.translate.instant('Share with this name already exists');
+
+  private errorText: string;
+
+  validateSmbName = (control: AbstractControl, existingName: string): Observable<ValidationErrors | null> => {
```
bvasilenko, line 25: I've noticed that naming of this function is about validation of the name, while the function is responsible for more than that - it's prepared to be responsible for handling any other error, including the "smbUsersError".

I'd suggested a concise self-documenting naming for this function, for example "validate"

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -537,4 +527,8 @@ export class SmbFormComponent implements OnInit {
       }),
     );
   }
+
+  private smbNameValidator(control: AbstractControl): Observable<ValidationErrors | null> {
+    return this.smbNameValidationService.validateSmbName(control, this.existingSmbShare?.name);
```
bvasilenko, line 532: There's a noticeable delay before the validation is not yet complete, and the user is able to hit "Save". 

Suggestion: we could disable "Save" button to prevent the user from running into server-side validation error, and ending up with no outcome and no prefilled form to re-submit.

bvasilenko (approved):
Works as expected overall, just a few nice-to-have improvements

PR #9316, opened 2023-12-13
NAS-125701 / 24.04 / Audit Page QA improvements

AlexKarpov98:
Testing of improvements: (see ticket or below)
_p.s I've added a couple of Ukrainian 🇺🇦  translations, so you can check using this beautiful language._

- Add autofocus on the Basic search & Advanced Search when switching between them.
- Submit Basic & Advanced Search by clicking Enter
- Allow switch to basic & switch to advanced to be clicked by Enter
- Fix first column SERVICE (to use translated value) from auditServiceLabels.has()
- Localise Advanced autocomplete placeholder, currently it’s static 'Service = "SMB" AND Event = "CLOSE"'
- Localise Event data column, currently values are not marked for translations
- Fix calendar Week Days color, now it’s DARK on the DARK background
- Hide Export as CSV button if no query results
- Add autocomplete suggestions for “Middleware” service
- Fix Event Data search, it's not working
- Fix TODO --> // TODO: Issue: reset icon will not trigger table update
- Hide calendar in advanced autocomplete input upon SEARCH request (currently it stays active until I move focus in the input)
- Remove Unused `ViewConfigurationState` from `/audit/store/view-configuration.store.ts`


src/app/modules/search-input/components/advanced-search/advanced-search.component.ts
```diff
@@ -35,6 +39,7 @@ export class AdvancedSearchComponent<T> implements OnInit {
 
   @Output() paramsChange = new EventEmitter<QueryFilters<T>>();
   @Output() switchToBasic = new EventEmitter<void>();
+  @Output() runSearch = new EventEmitter<SearchQuery<T> | void>();
```
undsoft, line 42: There seem to be two conflicts with current architecture:
1. By emitting a query in `runSearch` some of the functionality of `paramsChange` is duplicated.
2. Advanced search was designed not to know that it is advanced, i.e. to emit QueryFilters instead of `SearchQuery`. The idea here was that maybe in the future we could use advanced and basic search components separately.

I think it would be a good idea to change this to always emit void and to rely on `paramsChange` for new params.

src/app/pages/audit/components/audit/audit.component.ts
```diff
@@ -237,6 +266,8 @@ export class AuditComponent implements OnInit, AfterViewInit, OnDestroy {
       this.dataProvider.setParams([(query as AdvancedSearchQuery<AuditEntry>).filters]);
     }
 
+    this.advancedSearchAutocomplete.showDatePicker$.next(false);
```
undsoft, line 269: This breaks encapsulation of advanced search component. Audit component shouldn't care about internal details of advanced search autocomplete. There should be some other way to implement this, perhaps by closing date picker when focus is lost or something like this.

src/app/modules/search-input/components/basic-search/basic-search.component.ts
```diff
@@ -15,15 +16,21 @@ export class BasicSearchComponent implements AfterViewInit {
...
 
   protected resetInput(): void {
     this.query = '';
-    this.queryChange.emit(this.query);
+    this.runSearch.emit({ query: '', isBasicQuery: true });
+    this.focusInput();
+  }
+
+  private focusInput(): void {
+    (this.searchControl?.nativeElement as HTMLInputElement)?.focus();
```
undsoft, line 34: You can also add `HTMLInputElement` as a generic in `ElementRef` on line 21.

undsoft (changes requested):
Nice job on testing 👍🏻

undsoft (approved):
It would be better not to introduce a reset argument in paramsChange.
Instead just emit `paramsChange` first and `runSearch` second when user resets the input.

PR #9319, opened 2023-12-14
NAS-125696 / 24.04 / ix-chips options are transparent at some point

AlexKarpov98:
Testing: 👇 

Before: _(notice Readonly Administrators background)_
Screenshot 2023-12-14 at 18 16 16

After:
Screenshot 2023-12-14 at 18 17 10


bvasilenko (approved):
LGTM!

PR #9321, opened 2023-12-14
NAS-125720 / 24.04 / NAS-125720 - Add License Text to WebUI Repo

kmoore134:


undsoft:
Shouldn't it be called `LICENSE`. Seems like this is what github understands https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository

kmoore134:
Good point. I had seen the preference for GPL was to use COPYING, so I've just sym-linked it to make sure both bases are covered.

undsoft (approved):
👍🏻

undsoft:
Looks like github doesn't understand symlinks. Front page now says  `Unknown, GPL-3.0 licenses found`

kmoore134:
GitHub is lame :) But it answers the question as to which files to use, the answer is "both" work. I'll just standardize on LICENSE then

PR #9324, opened 2023-12-15
NAS-125656 / 24.04 / Remove categories and move Jira login button

denysbutenko:
This PR depends on https://github.com/truenas/middleware/pull/12699

For testing, check the feedback dialog for regressions.


bvasilenko (changes requested):
Please re-request review once the blocking PR is merged

denysbutenko:
@bvasilenko middleware [PR](https://github.com/truenas/middleware/pull/12699) has been merged, and this PR can be tested with the latest nightly build

undsoft (changes requested):
The Login to Jira button should just submit the form once oAuth successfully completes.
We can also rename the button `Login To Jira To Submit`.

denysbutenko:
> The Login to Jira button should just submit the form once oAuth successfully completes. 

Good point - I'll check it. Just to clarify, do you mean cases when the form is filled and valid to be submitted.

> We can also rename the button `Login To Jira To Submit`.

Okay let's rename it

undsoft:
Yes, it should just act a submit button. You may need to create a new component, not sure.

src/app/modules/ix-feedback/feedback-dialog/feedback-dialog.component.ts
```diff
@@ -372,9 +369,14 @@ export class FeedbackDialogComponent implements OnInit {
 
     this.form.controls.token.valueChanges.pipe(
       filter((token) => !!token),
+      distinctUntilChanged(),
+      delay(10),
       untilDestroyed(this),
     ).subscribe((token) => {
       this.feedbackService.setOauthToken(token);
+      if (this.form.valid && !this.isLoading) {
+        this.onSubmit();
```
undsoft, line 378: So, user presses login to Jira button -> it updates token in a form -> this triggers submit in `addFormListeners`.
This won't be obvious to any future developer.
We should do it the right way by extracting common code from `OauthButtonComponent` into a service and calling it when normal submit button in this form is used.

undsoft (changes requested):

src/app/modules/ix-feedback/feedback-dialog/feedback-dialog.component.ts
denysbutenko: > We should do it the right way by extracting common code from OauthButtonComponent into a service and calling it when normal submit button in this form is used.

I am not sure if I get you. Can you explain the idea a bit more?

undsoft: Original Jira auth button doesn't work well in this form because it was not designed to be used in this scenario. Instead of a form control like in other forms, we need a generic button that shouldn't be responsible for Jira auth.
We need to put a normal button and start oAuth flow when it is pressed. To avoid copy-pasting existing code, we should reuse some of the code related to this from the OauthButtonComponent.

denysbutenko: If we use `ix-oauth-button` directly it allows us to remove `JiraOauthComponent` and keep it clear. Is it something we are looking for?

undsoft: Yeah, we can probably do that, it has that (loggedIn) output which should be more obvious to everyone.

src/app/modules/ix-feedback/feedback-dialog/feedback-dialog.component.html
```diff
@@ -91,13 +91,13 @@ <h1 matDialogTitle>{{ 'How would you rate this page?' | translate }}</h1>
...
+    <ix-oauth-button
       *ngIf="showJiraButton"
-      [formControl]="form.controls.token"
-      [label]="'Token' | translate"
-      [tooltip]="tooltips.token | translate"
-      [required]="true"
-    ></ix-jira-oauth>
+      testId="login-to-jira"
+      [oauthType]="oauthType.Jira"
+      [oauthUrl]="'https://support-proxy.ixsystems.com/oauth/initiate?origin='"
```
undsoft, line 98: Let's move it into class property.

src/app/interfaces/api/api-call-directory.interface.ts
```diff
@@ -788,29 +787,7 @@ export interface ApiCallDirectory {
...
-  'smart.update': { params: [SmartConfigUpdate]; response: SmartConfig };
-  'smart.config': { params: void; response: SmartConfig };
-  'smart.test.manual_test': { params: [SmartManualTestParams[]]; response: ManualSmartTest[] };
-  'smart.test.query': { params: QueryParams<SmartTestTask>; response: SmartTestTask[] };
-  'smart.test.query_for_disk': { params: [disk: string]; response: SmartTestTask[] };
-  'smart.test.create': { params: [SmartTestTaskUpdate]; response: SmartTestTask };
-  'smart.test.results': { params: QueryParams<SmartTestResults>; response: SmartTestResults[] };
-  'smart.test.update': { params: [id: number, update: SmartTestTaskUpdate]; response: SmartTestTask };
-  'smart.test.delete': { params: [id: number]; response: boolean };
+  'system.set_time': { params: [number]; response: void };
```
undsoft, line 790: Don't see usages for set_time.

undsoft (approved):
I think it may be worth splitting this component into different forms, but we can do it later.
It works okay.

src/app/interfaces/api/api-call-directory.interface.ts
denysbutenko: seems like merge issues let me check

src/app/modules/ix-feedback/feedback-dialog/feedback-dialog.component.html
denysbutenko: oauthUrl is used in other scenarios also

undsoft: I mean we should extract 'https://support-proxy' into a property of the feedback-dialog.component.

bvasilenko (approved):
Works as expected

PR #9328, opened 2023-12-15
NAS-125616 / 24.04 / An Internally Signed Certificate Requires a Page Reload To Display After Creation On The Certificates Page

AlexKarpov98:
Testing: see ticket (there is video)

Additional:

I tried to update tests to include checks of API calls and value `emit`, but was not able to find `DELETE` button after I open modal, and can't find any working test with same conditions.

```    
    expect(spectator.inject(MatDialog).open).toHaveBeenCalledWith(ConfirmForceDeleteCertificateComponent, {
      data: certificates[0],
    });
```

```
ailed to find element matching one of the following queries:
    (_MatButtonHarness with host element matching selector: "[mat-button], [mat-raised-button], [mat-flat-button],
                             [mat-icon-button], [mat-stroked-button], [mat-fab], [mat-mini-fab]" satisfying the constraints: text = "Delete")
```

### DIFF
--- a/src/app/pages/credentials/certificates-dash/certificate-authority-list/certificate-authority-list.component.ts
+++ b/src/app/pages/credentials/certificates-dash/certificate-authority-list/certificate-authority-list.component.ts
@@ -260,7 +260,9 @@ export class CertificateAuthorityListComponent implements OnInit {
         }),
         untilDestroyed(this),
       )
-      .subscribe();
+      .subscribe(() => {
+        this.getCertificates();
+      });
   }
 
   doSignCsr(certificate: CertificateAuthority): void {
--- a/src/app/pages/credentials/certificates-dash/certificate-authority-list/certificate-authority-list.component.ts
+++ b/src/app/pages/credentials/certificates-dash/certificate-authority-list/certificate-authority-list.component.ts
@@ -1,6 +1,6 @@
 import { HttpErrorResponse } from '@angular/common/http';
 import {
-  ChangeDetectionStrategy, ChangeDetectorRef, Component, OnInit,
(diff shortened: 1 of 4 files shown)


denysbutenko (changes requested):
When `Sign CSR` or `Delete` is completed, the dashboard reloads as expected.
Nothing happens after `Revoke` is completed (needs data reloading).

It's not related, but it was hard to find the `Sign CSR` button due to content overlap. See if a quick fix is possible or create a ticket for that.
image

AlexKarpov98:
@denysbutenko - in ticket details I didn't find anything about the need to reload the second table. We only update `Certificates` table in this case. Or once you REVOKED you didn't see this action button?
Screenshot 2023-12-18 at 11 01 36

2. we can add that button as `Text Button` below the row content, what do you think?

denysbutenko:
> @denysbutenko - in ticket details I didn't find anything about the need to reload the second table. We only update `Certificates` table in this case. Or once you REVOKED you didn't see this action button? Screenshot 2023-12-18 at 11 01 36
> 

Yes, no `Revoked` icon/button exists until the page is refreshed manually. 

https://github.com/truenas/webui/assets/351613/fb988b60-e3c0-4fdf-8bc0-679439a0fda1


> 2. we can add that button as `Text Button` below the row content, what do you think?

At first, I thought of tweaking the style to give our buttons more room or maybe bumping up the z-index to stay on top of the text. Perhaps we need a more one-size-fits-all fix for our tables. But for now, just zooming out the page seems to do the trick for this and other similar issues.

AlexKarpov98:
@denysbutenko ah, you mean CA, yeah, we definitely should refetch data here.

denysbutenko (approved):
👍🏻

PR #9329, opened 2023-12-15
NAS-125726 / 24.04 / Remove unused code

undsoft:
- Removes unused code and features that were removed on middleware.
- Replaces default empty path when creating user with a better one according to Andrew

Testing: code review.


AlexKarpov98 (approved):
Looks good 👍

PR #9330, opened 2023-12-16
Added directive to require roles to enable buttons

RehanY147:


src/app/directives/common/common-directives.module.ts
```diff
@@ -14,12 +15,14 @@ import { LetDirective } from './app-let.directive';
     LetDirective,
     IfNightlyDirective,
     HasRoleDirective,
+    MatButtonRolesDirective,
```
undsoft, line 18: We shouldn't use Mat prefix.

src/app/directives/common/mat-button-roles/mat-button-roles.directive.ts
```diff
@@ -0,0 +1,70 @@
...
+import { Role } from 'app/enums/role.enum';
+import { AuthService } from 'app/services/auth/auth.service';
+
+@UntilDestroy()
+@Directive({
+  selector: '[ixMatButtonRoles]',
+  providers: [MatTooltip],
+})
+export class MatButtonRolesDirective {
+  protected _elementClass: string[] = [];
```
undsoft, line 17: Prefer not to use leading underscore.

src/app/pages/sharing/nfs/nfs-form/nfs-form.component.ts
```diff
@@ -101,6 +105,7 @@ export class NfsFormComponent implements OnInit {
     private slideInRef: IxSlideInRef<NfsFormComponent>,
     private store$: Store<ServicesState>,
     @Inject(SLIDE_IN_DATA) private existingNfsShare: NfsShare,
+    private authService: AuthService,
```
undsoft, line 108: unused.

src/app/modules/ix-forms/validators/has-role-validation/has-role-validation.ts
```diff
@@ -0,0 +1,31 @@
+import {
+  AsyncValidatorFn, FormGroup, ValidationErrors,
+} from '@angular/forms';
+import {
+  Observable, first, map, of,
+} from 'rxjs';
+import { Role } from 'app/enums/role.enum';
+import { AuthService } from 'app/services/auth/auth.service';
+
+export function hasRoleAsyncFgValidator(
```
undsoft, line 10: Unused.

src/app/pages/sharing/nfs/nfs-form/nfs-form.component.html
```diff
@@ -122,6 +122,7 @@
           type="submit"
           color="primary"
           ixTest="save"
+          [ixMatButtonRoles]="[Role.SharingNfsWrite]"
```
undsoft, line 125: `ixRequireRoles`?

undsoft (changes requested):
It looks good, but if we are to introduce a strike-out as a design language element, it should be also applied to icons.

src/app/pages/sharing/nfs/nfs-form/nfs-form.component.html
RehanY147: This won't work on non mat-button elements. And we exclusively use mat-buttons on our forms so the naming should be reflective I think. Plus, it won't work on other non-button elements to don't want it to be mistakenly used on anything other than a mat-button.

bvasilenko (approved):
Works as expected

PR #9331, opened 2023-12-17
NAS-125754 / 24.04 / Fixed injection of `KubernetesStore` to `SetupPoolDialogComponent`

RehanY147:
Go to the App details page and click on the Setup Pool button to make the dialog appear.

Expected: The dialog opens
Actual: nothing shows up. The screen goes a bit dark. Errors in the browser console.


undsoft (changes requested):
Not sure what adding `viewContainerRef` does, but I still get an error about missing KubernetesStore when I click Settings -> Choose Pool.

RehanY147:
@undsoft That was a different instance of the dialog where the update was also needed. Adding the viewContainerRef gives the same injector context as the component that opened the dialog.

undsoft (approved):
Dialogs open now.

PR #9332, opened 2023-12-17
NAS-125756 / 24.04 / Hide Restart and other options from limited users

undsoft:
Testing: create a limited role and observe that Restart, Power Off and Reset to defaults option from General settings are not there.

### DIFF
--- a/src/app/modules/layout/components/topbar/power-menu/power-menu.component.html
+++ b/src/app/modules/layout/components/topbar/power-menu/power-menu.component.html
@@ -14,12 +14,24 @@
     {{ 'Log Out' | translate }}
   </button>
 
-  <button name="power-restart" mat-menu-item ixTest="restart" (click)="onReboot()">
+  <button
+    *ngIf="isSysAdmin$ | async"
+    name="power-restart"
+    mat-menu-item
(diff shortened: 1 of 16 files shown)


RehanY147 (changes requested):
Soon as I login with root, UI crashes with these errors in the console
image

undsoft:
Are you on fresh middleware?

src/app/modules/layout/components/topbar/power-menu/power-menu.component.html
```diff
@@ -14,12 +14,24 @@
...
+    ixTest="restart"
+    (click)="onReboot()"
+  >
     <ix-icon name="replay"></ix-icon>
     {{ 'Restart' | translate }}
   </button>
 
-  <button name="power-shut-down" mat-menu-item ixTest="shut-down" (click)="onShutdown()">
+  <button
+    *ngIf="isSysAdmin$ | async"
```
RehanY147, line 29: We could put both buttons with the same `ngIf` into one `ng-container`

RehanY147 (approved):
image
The styles are a little messed up but looks good otherwise

PR #9335, opened 2023-12-18
NAS-125760 / 24.04 / Offer to start SMB or NFS services from dataset form

denysbutenko:
For testing, try to create a new dataset with SMB and NFS shares and ensure the Start Service dialog appears (only when the selected service is stopped). 

Note: The Start Service dialog does not support multiple services simultaneously, so two dialogs for NFS and SMB services will appear.


bvasilenko (approved):
Works as expected

PR #9336, opened 2023-12-18
NAS-125706 / 24.04 / Fix mixed values in ix-chips

denysbutenko:
For testing:

> Open add user form.
> Fill in details and select builtin_users from the list in Auxiliary Groups.
> Try to save.
> Get error: Item#0 is not valid per list types: [group] Not an integer


undsoft (approved):
Auxiliary input in ix-chips works okay now.

PR #9340, opened 2023-12-18
NAS-125354 / 24.04 / Added ixRequiresRoles directive

RehanY147:


denysbutenko (changes requested):
It works great for me. Just minor cosmetic things:

- Disabled elements use `cursor: not-allowed`
- Slightly different font color between disabled menu item and new directive elements (2nd img)
- Directive content is focused by default in the menu (1st img) 


image 


image

src/app/pages/sharing/components/shares-dashboard/nfs-card/nfs-card.component.ts
```diff
@@ -62,6 +63,7 @@ export class NfsCardComponent implements OnInit {
           iconName: 'delete',
           tooltip: this.translate.instant('Delete'),
           onClick: (row) => this.doDelete(row),
+          roles: [Role.SharingNfsWrite],
```
undsoft, line 66: Let's also call it `requiresRoles` for consistency.

src/setup-jest.ts
```diff
@@ -92,6 +93,7 @@ defineGlobalsInjections({
     AppLoaderModule,
   ],
   providers: [
+    mockAuth(),
```
undsoft, line 96: I don't know if mocking authentication globally is a good approach, there are places where this may not be desirable. Let's let developers mockAuth individually in their tests.

src/assets/styles/other/_material-reduction.scss
```diff
@@ -863,3 +863,21 @@ body .mat-calendar-previous-button::after,
 body .mat-calendar-next-button::after {
   margin: 10px;
 }
+
+.role-missing button {
```
undsoft, line 867: Let's move this somewhere, either in a separate file or even in the component with ng-deep.

undsoft (changes requested):
Looks good.
It would be good to make line move visible – it's hard to see on the Save button. Maybe we could draw it with `::before` css pseudo-element.

RehanY147:
@denysbutenko The focus issue will have to be managed by another ticket. It has the same issue as just disabling the buttons with the disable attribute. Too complicated.

src/app/directives/common/requires-roles/requires-roles-wrapper.component.ts
```diff
@@ -0,0 +1,14 @@
+import { Component, Input, TemplateRef } from '@angular/core';
+
+@Component({
+  selector: 'ix-requires-roles-wrapper',
+  template: `
+<span class="{{class + ' role-missing'}}" [matTooltip]="'Missing permissions for this action' | translate" [matTooltipPosition]="'above'">
```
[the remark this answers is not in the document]
RehanY147: The `ngClass` gives conditional classes. These aren't conditional classes. These are just classes passed on from the host element.

undsoft (approved):
Okay, that's probably good enough.
Please incorporate Denyses suggestions about `cursor: not-allowed`. For `ngClass` you may be able to do something like `[class]="['role-missing', class]"`

denysbutenko (approved):
It looks good

PR #9344, opened 2023-12-19
NAS-125717 / 24.04 / Add processes charts to Reporting > CPU

denysbutenko:
For testing, go to Reporting and check the `System Active Processes` chart on the CPU tab.

- Add a new chart - System Active Processes
- Fix condition for UPS menu item. For charts another ticket - https://ixsystems.atlassian.net/browse/NAS-125159


bvasilenko (approved):
Works as expected

PR #131, opened 2017-11-28
Fix 25825

Nesma1988:
Fixes: https://redmine.ixsystems.com/issues/25825
Showing re-silvering process in topbar

PR #9347, opened 2023-12-19
NAS-125354 / 24.04 / Added ixAddNewOption directive

RehanY147:
There are two examples.
The Replication Add Wizard -> Advanced Options -> SSH Connection (under Transport Options)
The Add Cloud Sync Task -> Credential (under Remote)

On the slide-in form for Cloud Sync tasks, the credentials select should have the option 'Add New'. When you click it, the form should open to add `Cloud Credentials`. Once successfully added a cloud credential, the cloud credential form should close and UI should go back to the Cloud Sync Task card. The provider select should have the newly created value as the selected value.


undsoft (changes requested):
Looks promising, just curious if we could do something in the directive itself to listen to what new item was added and add it to the list of options, so that there is less code outside of the component to handle this.
We could also require forms to implement a certain interface to make sure they return something standard when they are closed.

RehanY147:
@undsoft We don't know what the value and label is every time. It's not the same attribute every time. That's the main problem with trying to add the option in the directive.

src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
```diff
@@ -123,7 +127,8 @@ export class CloudsyncFormComponent implements OnInit {
   readonly transferModeOptions$ = of(mapToOptions(transferModeNames, this.translate));
 
   credentialsList: CloudsyncCredential[] = [];
-  readonly credentialsOptions$ = this.cloudCredentialService.getCloudsyncCredentials().pipe(
+  readonly fetchCredentialsOptions$ = new Subject<void>();
```
denysbutenko, line 130: Unused

src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
```diff
@@ -67,6 +69,8 @@ export class CloudsyncFormComponent implements OnInit {
       : this.translate.instant('Edit Cloud Sync Task');
   }
 
+  CloudCredentialsFormComponent = CloudCredentialsFormComponent;
```
denysbutenko, line 72: That's weird. Maybe we can follow other approaches. For example, what if we can do a map of possible options and components and then use an enum to avoid importing the whole component like that?

src/app/pages/data-protection/cloudsync/cloudsync-form/cloudsync-form.component.ts
```diff
@@ -180,8 +184,24 @@ export class CloudsyncFormComponent implements OnInit {
     private store$: Store<AppState>,
     protected cloudCredentialService: CloudCredentialService,
     @Inject(SLIDE_IN_DATA) private editingTask: CloudSyncTaskUi,
+    @Inject(SLIDE_IN_CLOSER) protected closer$: Subject<unknown>,
   ) { }
 
+  newCloudCredAdded(value: unknown): void {
+    this.credentialsOptions$ = this.cloudCredentialService.getCloudsyncCredentials().pipe(
```
denysbutenko, line 191: duplicate at line [#131](https://github.com/truenas/webui/pull/9347/files#diff-07f66c4dfca3295b93b45129b462776611624e53d3e6494943c35260eee67dbaR131). lets merge them

src/app/pages/data-protection/cloudsync/cloudsync-task-card/cloudsync-task-card.component.ts
```diff
@@ -51,6 +51,8 @@ export class CloudSyncTaskCardComponent implements OnInit {
   jobStates = new Map<number, string>();
   readonly jobState = JobState;
 
+  private ixChainedSlideInService = inject(IxChainedSlideInService);
```
denysbutenko, line 54: I like this approach, but it breaks component consistency. Could we move it to the constructor?

src/app/modules/ix-forms/components/ix-select/ix-add-new-option.directive.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  ngOnInit(): void {
+    this.ixSelect.controlDirective.control.valueChanges.pipe(untilDestroyed(this)).subscribe({
+      next: (value) => {
+        if (value === addNewValue) {
+          const close$ = new Subject();
+          this.chainedSlideIn.pushComponent({ component: this.ixAddNewOption, close$ });
+          close$.pipe(take(1), untilDestroyed(this)).subscribe({
+            next: (response) => {
+              this.newOptionAdded.emit(response);
```
denysbutenko, line 51: Do you think we can update select value and options from here? It may allow us to eliminate callbacks.

denysbutenko (changes requested):
It works okay and needs some improvements.

src/app/modules/ix-forms/components/ix-select/ix-select-with-new-option.directive.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  abstract setValueFromSlideInResult(
+    result: ChainedSlideInCloseResponse,
+    valueSetterCallback: (value: string | number) => void
+  ): void;
+  abstract getFormComponentType(): ComponentType<unknown>;
+  abstract fetchUpdatedOptions(): Observable<Option[]>;
+
+  ngAfterViewInit(): void {
+    if (this.ixSelect) {
```
undsoft, line 91: return early

src/app/modules/ix-forms/components/ix-select/ix-select-with-new-option.directive.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  setDisabledState(isDisabled: boolean): void {
+    this.disabled = isDisabled;
+  }
+
+  private chainedSlideIn = inject(IxChainedSlideInService);
+  private translateService = inject(TranslateService);
+
+  ngOnInit(): void {
+    this.fetchUpdatedOptions().pipe(
```
undsoft, line 67: Maybe just call it `fetchOptions`.

undsoft (approved):
An interesting approach with abstract class. Looks nice.

Don't forget to remove Create New option from Replication form.

RehanY147:
@undsoft @denysbutenko This PR has also reworked the `cloudsync-wizard` to use the ChaindSlideInService and also added a swap slide-in functionality. Test
- The component that opened the original wizard/form should still get the response when the form is submitted even if you switch to form from wizard or vise-versa while using the slide-in.
- The CloudsyncWizard component should work as before without any regressions.
- Test the add new option and the abstract class that is used to create new ix-selects with add-new option

undsoft (changes requested):
Cloudsync wizard - when I add new item, I cannot advance to next step. Next button doesn't do anything.
Rsync task form - maybe unrelated to this PR, but I couldn't get Save button to be enabled.

RehanY147:
@undsoft It's working for me.

https://github.com/truenas/webui/assets/24204846/1517bf22-0e2e-4faf-8a7c-5bdccb575c84

undsoft (approved):
Yeah, weird. Appears to be working okay now.

denysbutenko (changes requested):
I have noticed a few regressions:
- Editing cloud sync task shows Credentials select empty.
- Credential options are not sorted anymore.
- I see an empty slide-in panel when I go to `/credentials/backup-credentials` and click on the edit button cloud credential. Error in console `No provider for InjectionToken CHAINED_SLIDE_IN_REF!`

RehanY147:
@denysbutenko I don't think sorting the options was an intentional behavior. Now they are just sorted the other way around
image

But good catch on the CloudCredentialsForm! Thanks!

src/app/modules/ix-forms/components/ix-slide-in/components/ix-chained-slide-in/ix-chained-slide-in.component.ts
```diff
@@ -14,12 +16,10 @@ import { ChainedComponentSeralized as ChainedComponentInfoSeralized, IxChainedSl
 export class IxChainedSlideInComponent implements OnInit {
   @Input() id: string;
   @ViewChild('componentWrapper') container: HTMLElement;
-  protected components: ChainedComponentInfoSeralized[];
+  protected components: ChainedComponentSeralized[];
```
denysbutenko, line 19: Typo in `seralized`

denysbutenko (approved):
Looks good, great work 👍🏻

PR #9350, opened 2023-12-20
NAS-125773 / 24.04 / Audit improvements 3

AlexKarpov98:
Testing: see ticket and try on the UI side (as well maybe there are some improvements)

Screenshot 2023-12-20 at 17 41 35
Screenshot 2023-12-20 at 17 41 54


src/app/pages/audit/utils/get-log-important-data.utils.ts
```diff
@@ -28,6 +28,13 @@ function getMiddlewareLogImportantData(log: MiddlewareAuditEntry, translateServi
     case AuditEvent.Authentication: {
       const credentialType = log.event_data?.credentials.credentials;
       const credentialTypeKey = credentialTypeLabels.get(credentialType);
+
+      if (log.event_data?.error) {
+        return translateService.instant(T('Credentials Issue: {credentials}'), {
```
undsoft, line 33: Let's call it `Failed Authentication`

undsoft (changes requested):
I don't think issue 4 in the ticket is fixed:
1. Switch to last page in results.
2. Search for `Service != "Middleware"`
See "No results" for a second or two and then results appear.

undsoft (approved):
I think it may be better to update current page somewhere after we do `this.countRows().pipe(`.
Otherwise works well.

PR #9359, opened 2023-12-21
NAS-125808 / 24.04 / Disable restricted controls for Credentials pages

undsoft:
Testing: test credential pages as Readonly and Sharing Manager user.


AlexKarpov98 (commented):
Generally it works as expected, buttons are disabled, like:
Screenshot 2023-12-22 at 10 54 28

But one problem I noticed is that form can still be submitted via ENTER, for example, and then I see error modal:
Screenshot 2023-12-22 at 10 55 21

For readonly user - we still see ADD button, and for some cases I immediately see error modal (for example on Certificates page after I click ADD):

Screenshot 2023-12-22 at 11 05 43

AlexKarpov98:
@undsoft - is that OK case?
https://github.com/truenas/webui/assets/22980553/559cbeef-0f4b-4f89-897e-4fffd955abd6

undsoft:
@AlexKarpov98 The sidebar should not be closing – fixed. As for some endpoints being unavailable, this is expected. There is a middleware ticket about it.

AlexKarpov98 (approved):
Looks good 👍

PR #9360, opened 2023-12-22
NAS-125809 / 24.04 / Disable restricted controls for System Settings pages

denysbutenko:
Check ticket for details


undsoft (changes requested):
Other items under System Settings menu, such as Boot, Services, etc should also be reviewed.

For Update, General and Advanced settings:
- Update - Check for Updates Daily and Download if Available
Needs to be disabled and we might as well add error handling here.

- CRON job, init shutdown scripts and others
We can allow edit icon, just don't allow Save button.
There may be additional fields in the form.

- Self-Encrypting Drive
Can be fully hidden for non admins

undsoft (changes requested):
Looks pretty good.

Here are the things I found:
- I can see `update.set_auto_download` being called when I visit Update page
- We may need to do something for Running toggle on services page. Currently it's not obvious that it is disabled.
- Same for Start Automatically checkboxes.
I would suggest adding support for these elements in restricted roles directive.
- Alert services - delete is not disabled.

denysbutenko:
> Looks pretty good.
> 
> Here are the things I found:
> 
> * I can see `update.set_auto_download` being called when I visit Update page
> * We may need to do something for Running toggle on services page. Currently it's not obvious that it is disabled.
> * Same for Start Automatically checkboxes.
>   I would suggest adding support for these elements in restricted roles directive.

Okay, I added an extra check for the first case and extended styles for new elements.

> * Alert services - delete is not disabled.

There is a lack of support in the entity table actions. I saw some work done at https://github.com/truenas/webui/pull/9368. Let's update it once PR merged.

undsoft (approved):
Looks good, but there are additional roles for Sharing Manager that allow it to edit SMB/iSCSI services.
See https://ixsystems.atlassian.net/browse/NAS-125769

PR #9364, opened 2023-12-22
NAS-125805 / 24.04 / Disable restricted controls for Storage pages

undsoft:
Testing: test Storage pages as readonly and sharing manager users.


AlexKarpov98 (commented):
1. We have problem with submitting via Enter, example:

https://github.com/truenas/webui/assets/22980553/39876233-b4c2-4635-98fe-3d563a759615


2. As well we can hide non needed buttons, like START OVER or RESET STEP
Screenshot 2023-12-26 at 15 26 08

3. Once I open Import Pool, I immediately get error and will be able to submit form via ENTER:
Screenshot 2023-12-26 at 15 27 56

4. Maybe we can disable Add disks button?
Screenshot 2023-12-26 at 15 29 38

5. I can submit forms via Enter
Screenshot 2023-12-26 at 15 30 36

undsoft:
> Enter button
This is a good point, I will make a separate ticket for it.

> Hiding Start over and add buttons.
They were deliberately left available. The idea is that even though the user won't be able to make changes, he may be interested in seeing how those changes can be made, hence all the buttons are there.

AlexKarpov98 (approved):
Works good 👍 
 
_We should just not forget for `Enter form submitting` ticket 😄_

PR #9365, opened 2023-12-22
Prevent text overflow in entity-table-row-details

nitn3lav:
Currently, if the text in the right column of `entity-table-row-details` is too long and doesn't contain enouch characters that usually allow line breaks (like spaces and hyphens) the text will overflow. Because the text is in a table this also messes with the layout of the columns of the `entity-table`:

This PR fixes this by adding the `word-break: break-word` CSS property to the `p`-element:


undsoft (approved):
Thank you for your contribution!

PR #9366, opened 2023-12-23
Clarify what the VM password field is for

seanthegeek:
Ideally, the New VM wizard UI should also be updated to show that the Bind and Password fields are related to the VNC display device. I'm not sure what the best way to do that is.

- Group the fields visually
- Rename the fields to "Display bind" and "Display password" and update all of the translations


RehanY147:
jira

PR #9368, opened 2023-12-26
NAS-125806 / 24.04 / Disable restricted controls for Shares and Data Protection pages

bvasilenko:
**Testing**

1. Create a new local group on our server, with `READONLY` role and another group with `SHARING_MANAGER` role at **Credentials > Local Groups > Privileges**.

2. Log in with this user, and visit the **Shares** page, click through every button and clickable icon. 
   - **Expected result:** if action is illogical, or corresponding API call results in an error, then this action should not be available in WebUI for the current user

5. Do the same for **Data Protection** page.


AlexKarpov98 (changes requested):
Let's fix all of the 🔴 RED checks first: 👇 

Screenshot 2023-12-26 at 15 11 22

src/app/pages/services/components/service-smb/service-smb.component.ts
```diff
@@ -60,6 +61,7 @@ export class ServiceSmbComponent implements OnInit {
     multichannel: [false, []],
   });
 
+  readonly requiresRoles = [Role.SharingNfsWrite, Role.SharingManager, Role.SharingWrite];
```
denysbutenko, line 64: `Role.SharingNfsWrite` seems like a wrong role here

AlexKarpov98 (approved):
It works as expected. 👍

denysbutenko (approved):
👍🏻

PR #9371, opened 2023-12-26
NAS-125807 / 24.04 / Disable restricted controls for Datasets and Network pages

undsoft:
Testing: test corresponding pages as readonly and sharing manager roles.


RehanY147 (approved):
I think we should also disable the following buttons
- Add Dataset (Primary button on Datasets page since that form save button is disabled anyway and it doesn't show any information in that case)
- Add Zvol for the same reasons

PR #9373, opened 2023-12-26
NAS-125903 / 24.04 / Add edit icons to Quotas lists

undsoft:
Testing: see new edit buttons on User and Group lists.


denysbutenko (approved):
There is no tooltip on the edit buttons. Other than that, it looks good.

PR #9374, opened 2023-12-27
NAS-125915 / 24.04 / Simplify error handlers

undsoft:
Rationale:
Current individual handlers may catch a JS error and fail, or a job error handler may catch a WS error in some scenarios.

Testing:
Test some of the affected places.


AlexKarpov98 (approved):
LGTM 👍

PR #9377, opened 2023-12-28
NAS-125125 / 24.04 / It is a requirement to specify source bootenv when creating a new bootenv in UI

AlexKarpov98:
Code review is enough.


undsoft (approved):
👍🏻

PR #9378, opened 2023-12-28
NAS-125676 / 24.04 / NFS servers setting now supports null entry

AlexKarpov98:
Testing: see ticket and comments.
You should now see:
Screenshot 2023-12-28 at 14 10 56


denysbutenko (approved):
Looks good 👍🏻

PR #9381, opened 2023-12-29
NAS-125931 / 24.04 / Prevent user from pressing Enter on restricted controls

AlexKarpov98:
Testing: see video and ticket

You should not be able to submit the form via ENTER if you don't have WRITE access.
On the video:
- first is `root` user - has privilege
- second is `non admin` user - no privilege to submit form

https://github.com/truenas/webui/assets/22980553/18f8ee8a-2d5b-462e-b149-f7bb76907428

### DIFF
--- a/src/app/directives/common/requires-roles/requires-roles.directive.ts
+++ b/src/app/directives/common/requires-roles/requires-roles.directive.ts
@@ -1,5 +1,5 @@
 import {
-  ComponentRef, Directive, HostBinding, Input, TemplateRef, ViewContainerRef,
+  ComponentRef, Directive, HostBinding, Input, OnDestroy, TemplateRef, ViewContainerRef,
 } from '@angular/core';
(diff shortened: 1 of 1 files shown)


undsoft (changes requested):
I don't think it's a good solution.
1. This breaks scope of responsibilities for this directive. It's not obvious for any future developer that putting `ixRequiresRoles` on a button will disable the whole form submission.
2. This behaviour may or may not be needed. I may theoretically want to disable one button in a form, but allow another button to submit.
3. I believe angular also frowns upon direct dom manipulation.

Let's try disabling the button from getting focus state instead.

AlexKarpov98:
proposed treatments are worse then the problem.
User pressing submitting form by being clever will simply result in an error.
Us trying to fix it will either require another directive or a more complicated current directive.

PR #9383, opened 2023-12-31
NAS-125624 / 24.04 / Removed has role guard

RehanY147:
Removes the `ixHasRoles` guard to block routes if they don't have certain roles. It has now been established that this guard is not needed. Bare minimum role required to access UI automatically allows access to all routes by default. Which makes it useless.


AlexKarpov98 (approved):
👍

PR #9386, opened 2024-01-01
NAS-125728 / 23.10.2 / Stop reporting errors when system is rebooting

RehanY147:
This PR aims to stop listening to errors in the `WebSocketService` when the `system.reboot` method is called. It should start listening to errors again as soon as the websocket is connected anew.


undsoft (changes requested):
Pending comment.

PR #9387, opened 2024-01-01
NAS-125421 / 23.10.2 / Add check for undefined before concatenating values

RehanY147:
Add check for undefined value before using array values


AlexKarpov98 (approved):
👍

PR #9392, opened 2024-01-03
NAS-125611 / 24.04 / Truncated display of IPv6 address in Network screen

AlexKarpov98:
Testing: see ticket


RehanY147 (changes requested):
In the after screenshots, the IPv4 address is truncated still. Also, for the iPhone 11 Pro screen (available in Firefox browser by default), the addresses are also truncated. See below
image

AlexKarpov98:
@RehanY147 - nice catch, please check it once again please.

RehanY147 (approved):

PR #9396, opened 2024-01-04
NAS-124814 / 23.10.2 / Handling case for no disks for a specific pool

RehanY147:
Handling case where missing disks for a specific pool make the topology card fail on the Storage page for pools.


AlexKarpov98 (approved):
👌

PR #9399, opened 2024-01-04
NAS-125495 / 24.04 / Enclosure management - list of drives off the screen and not scrollable

AlexKarpov98:
Testing: see ticket
Screenshot 2024-01-04 at 15 03 27


undsoft (approved):
I can see the scrollbar, thanks.

PR #9401, opened 2024-01-04
NAS-125942 / 24.04 / Add unique id to sentry tags and new tickets payload

denysbutenko:
Add host id and session id to sentry events and new tickets payload.


src/app/services/system-general.service.ts
```diff
@@ -48,12 +52,18 @@ export class SystemGeneralService {
...
-    ]).subscribe(([isStable, sysInfo]) => {
+      this.store$.select(selectSystemHostId).pipe(filter(Boolean)),
+      this.sessionId$,
+    ]).subscribe(([isStable, sysInfo, hostId, sessionId]) => {
       if (!isStable) {
         Sentry.init({
           dsn: environment.sentryPublicDsn,
           release: sysInfo.version,
         });
+        Sentry.configureScope((scope) => {
```
undsoft, line 63: I think it's time to create a new service.

denysbutenko: Good idea 👍🏻

AlexKarpov98 (approved):
Looks good

undsoft (approved):
Looks good, I was able to find my errors in Sentry after reporting a bug.
Don't forget to change `if (!isStable) {` to correct condition in the service before merging.

undsoft:
Also it's worth looking into what conditions we have for enabling error reporting. I think we should still report errors on nightlies to Sentry, but we shouldn't report them if it's a dev machine (localhost).

denysbutenko:
> Looks good, I was able to find my errors in Sentry after reporting a bug. Don't forget to change `if (!isStable) {` to correct condition in the service before merging.

What is the correct condition? It wasn't changed in this PR.

> Also it's worth looking into what conditions we have for enabling error reporting. I think we should still report errors on nightlies to Sentry, but we shouldn't report them if it's a dev machine (localhost).

Before there was an additional flag to enable reporting, but it was removed from the middleware side. I agree it needs to be revisited. @undsoft would you mind creating a new ticket for that and assign to me?

undsoft:
Well, I would imagine that it's when the system is stable that we mostly want to submit issues. I'll describe this in the new ticket.

PR #9402, opened 2024-01-04
NAS-125984 / 23.10.2 / Removing current name from validated smb share names list when editing

RehanY147:
See ticket for testing details


denysbutenko (approved):
👍🏻

PR #9403, opened 2024-01-04
NAS-125777 / 23.10.2 / Added null check

RehanY147:


undsoft (approved):
Looks good.

undsoft:
Does this need to go into Dragonfish?

RehanY147:
Yes. Should be ported.

PR #9406, opened 2024-01-05
NAS-125925 / 24.04 / Incorrect field type for "ACME Server Directory URI" field

AlexKarpov98:
Testing:
see ticket,
You should be able to set 
`Custom ACME Server Directory URI`
or 
Pre select one

CUSTOM:
Screenshot 2024-01-05 at 12 52 45

PRE SELECT:
Screenshot 2024-01-05 at 12 53 01


denysbutenko (approved):
Looks good

PR #9409, opened 2024-01-07
Fix typo in German translation: "Algorythmus"

NicoWeio:
See for example https://de.wikipedia.org/wiki/Algorithmus

I can remove the trailing newline if it bothers you – just tell me.


undsoft (approved):
Thanks, new line will probably get removed automatically by our software in the future.

PR #9412, opened 2024-01-08
NAS-125351 / 23.10.2 / In pCloud only directories in root dir can be selected, not lower levels

AlexKarpov98:
Testing: 
See ticket & use COBIA.
Now you should be able to get child nodes, as well fixed the path which is stored.

https://github.com/truenas/webui/assets/22980553/ed8f1321-61a6-491a-aada-09dcca8e461e


bvasilenko (approved):
Works as expected

PR #9413, opened 2024-01-08
NAS-126688 / 24.04 / Using more specific roles for some actions

undsoft:
A bunch of smaller fixes:
- Will no longer attempt to login with empty token.
- Uses specific roles instead of compound roles like ReplicationManager and SharingManager.
- Fixes regression in shell.
- Uses more specific roles instead of Full admin.

Testing: test some of the affected places.


AlexKarpov98 (approved):
Tested, it all works good. 👍

PR #9414, opened 2024-01-08
NAS-125235 / 24.10 / Updates for new Enclosure API

dszidi:
WIP PR

NOTES:

The new enclosure endpoint affects mock utility, storage generator and the enclosure UI page itself. The changes in this PR mostly deal with user facing updates to enclosure page. Storage generator and mock utility has been updated just enough to keep UI compiling.

Issues that exist but are out of the scope of this PR include...

- ix-test directives need to be updated
- mock utility templates other than M50, ES24, ES102, MINI-XL+ need to be updated with real data
- mock utility needs to be able to detect when query params are used

Waiting on further changes from middleware (Currently hard coding values to prevent errors)
- Add zfs disk errors to new endpoint 
- Pool health

What needs to be tested. 

- UI compiles
- Enclosure page works on rackmount products
- Enclosure page works on MINI products
-  tests (.spec.ts) are passing


undsoft (approved):
It works when tested superficially on m60 provided and m40 that we have.

PR #9417, opened 2024-01-08
NAS-126701 / UI errors on Manage Devices page - Add EXTEND for single-device LOG and remove EXTEND on RAIDZ vdev member disks

chrisperedun:
Adds EXTEND button for single-device LOG vdevs and removes misleading EXTEND on RAIDZ vdev member disks

isRaidzParent function updated (was checking the passed item itself, intention is to check if the parent item is RAIDZ type) - ref #7883 

canExtendDisk function updated to add match on VdevType.Log as a single-disk LOG should still be extended


RehanY147 (approved):
Thank you!

PR #9418, opened 2024-01-09
NAS-125837 / 24.04 / Only replace interfaces which are part of the update instead of replacing object

RehanY147:


src/app/pages/network/components/interfaces-card/interfaces-card.component.ts
```diff
@@ -188,7 +199,17 @@ export class InterfacesCardComponent implements OnInit, OnChanges {
...
   private subscribeToUpdates(): void {
     this.networkService.subscribeToInOutUpdates().pipe(untilDestroyed(this)).subscribe((updates) => {
-      this.inOutUpdates = updates;
+      if (!updates) {
+        return;
+      }
+      const newInOutUpdates = _.cloneDeep(this.inOutUpdates);
+      const updatedInterfaces = Object.keys(updates);
+      for (const nic of updatedInterfaces) {
+        newInOutUpdates[nic] = {
```
AlexKarpov98, line 208: the spread operator { ...updates[nic] } is redundant here since updates[nic] is directly assigned to newInOutUpdates[nic]. You can simplify this to `newInOutUpdates[nic] = updates[nic];`

AlexKarpov98 (commented):
Code looks good, @RehanY147 could you tell me what exactly should I test?

Were you able to reproduce the original BUG?

RehanY147:
@AlexKarpov98 When middleware fires an nic update from the reporting subscription, it can skip over interfaces which don't have any currrent acitivity even if they are "UP" or online. So, when we completely replace the old object with the new one, any interface that was up but just wasn't part of the latest update will be considered offline. So, instead, we only want to update interfaces which are included in the update and leave the ones that are not in there as is without making any changes.

src/app/pages/network/components/interfaces-card/interfaces-card.component.ts
RehanY147: Explained in the separate comment. Let me know what you think.

AlexKarpov98 (approved):
Tested with mock data, Two interfaces appeared online.

PR #9420, opened 2024-01-09
NAS-126077 / 24.04 / The directories don't reset when credentials are changed on the Cloudsync Task Form

AlexKarpov98:
Testing: see ticket, you should be able to see `<ix-explorer>` values updated once trigger is called.
Test for any regressions.

Example:
`Cloud Sync Tasks` - Credential -> Folder (as well I reset folder value once Credential is changed)

see:

https://github.com/truenas/webui/assets/22980553/982fe4ee-205a-4aad-bdba-21685c5a5242

as well check how it worked before, to see the difference. As well some optimisations done.

Cloud Sync Task ✅ 

Replication Task ✅ (As well added missing logic for the Wizard)
```
sourceNodeProvider: TreeNodeProvider;
targetNodeProvider: TreeNodeProvider;
```
⬆️ these two were defined but value was not set..

Rsync Task ✔️ 

@RehanY147  as well mentioned `Rsync Task` but I didn't find such logic there.


denysbutenko (approved):
Looks good 👍🏻

PR #9427, opened 2024-01-11
NAS-125453 / 24.04 / Fixed network traffic units on both dashboard widgets and reports dashboard

RehanY147:
On the dashboard on NIC Widget and Network Widget, check network traffic units and conversion.
Also check units and conversion on the reports dashboard for interfaces.


undsoft (changes requested):
1. Values in network card in In/Out section lack `/s` suffix.
2. I was trying to compare values in the interface with values coming from reporting subscription and I couldn't get numbers to work. Please re-check them.

RehanY147:
`fileSize` assumes you're passing in Bytes always. That was causing wrong conversions

src/app/pages/dashboard/components/widget-network/widget-network.component.ts
```diff
@@ -484,15 +481,31 @@ export class WidgetNetworkComponent extends WidgetComponent implements OnInit, A
...
-  private getSpeedLabel(value: number): string {
-    const converted = filesize(Math.round(Math.abs(value)), { bits: true }).replace('bit', 'b');
-    return `${converted}/s`;
-  }
-
-  private splitValue(value: number): number {
-    if (value < 1024) {
-      return Number(value.toString().slice(0, 4));
+  protected getSpeedLabel(value: number): string {
+    let bits = Math.abs(value);
```
undsoft, line 485: Can we not get filesize pipe to work? I'd rather we not build our own code for this.

undsoft (changes requested):
1. I see 'kb/s/s' in interface card.
2. I see too many decimal places in network chart tooltip.
Снимок экрана 2024-01-17 в 12 04 58
3. I still see unexpected numbers compared to when I calculate them on my own.
My expectation is that `update.interfaces.eno1.received_bytes_rate / 8 / 1000` would yield me kilobits per second.
4. There's code duplication and I'd rather we figure out how to work with `filesize` pipe instead of doing our own thing. If we need to, we can create our own pipe that would wrap around `filesize` with additional settings.

src/app/core/pipes/bit-size.pipe.ts
```diff
@@ -0,0 +1,11 @@
...
+import prettyBytes from 'pretty-bytes';
+
+@Pipe({
+  name: 'bits',
+})
+export class BitSizePipe implements PipeTransform {
+  transform(value: number): string {
+    return prettyBytes(value, { bits: true });
+  }
+}
```
undsoft, line 11: Let's add a couple of simple tests to prevent regressions.

src/app/pages/dashboard/components/widget-nic/widget-nic.component.ts
```diff
@@ -175,4 +175,32 @@ export class WidgetNicComponent extends WidgetComponent implements AfterViewInit
...
+      case multiplier < 1000 * 1000:
+        label += ' kb';
+        break;
+      case multiplier < 1000 * 1000 * 1000:
+        label += ' Mb';
+        break;
+      case multiplier < 1000 * 1000 * 1000 * 1000:
+        label += ' Gb';
+        break;
+      case multiplier < 1000 * 1000 * 1000 * 1000 * 1000:
```
undsoft, line 200: Is this something that is still needed? Can we use the pipe here?

undsoft (changes requested):
Looks better, but still some things to fix:
1. In reporting I get `expected a finite number, got number: NaN` when hovering over network chart.
2. Reporting also shows `mb/s` in table, while dashboard shows `mbits`. Should pick one.
3. Main dashboard -> Network chart. When hovering over outgoing values there is a minus sign in tooltip, that shouldn't be there.

undsoft (approved):
Looks mostly correct. The only issue I could find was a lowercase 'mb/s' in Reporting section.

PR #9428, opened 2024-01-11
NAS-126723 / 24.04 / ix-explorer bug with comma once I clear input and select something 

AlexKarpov98:
Testing: see ticket.
As well I did a small refactoring to remove code duplication for CloudSync form.


bvasilenko (approved):
Works as expected

PR #9429, opened 2024-01-11
NAS-126685 / 24.04 / NAS-126685: Do not do full build when only translation files were changed

AlexKarpov98:
For testing -- see this PR, which is based on this branch in order to test pipeline:

https://github.com/truenas/webui/pull/9433

You should see that if only translation files were updated - we skip some of the steps and run only `Resort Strings and Commit if only translations were added` Step.

Screenshot 2024-01-11 at 15 52 35

As well you can go ahead and update some other non i18n file, to see that the pipeline will now skip `Resort Strings and Commit if only translations were added` Step and will only run previous steps we had.

Screenshot 2024-01-11 at 15 53 33

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -108,20 +108,14 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: ${{ github.event.pull_request.head.ref }}
-          fetch-depth: 0 # fetch all history for all branches and tags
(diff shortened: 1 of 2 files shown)


bvasilenko (approved):
Worked as expected

**Translation updated:**

image

**Code updated:**

image

PR #9434, opened 2024-01-11
NAS-126711 / 24.04 / Fix HA enabled message when loading widget info

denysbutenko:
For testing, ensure no `HA Enabled/HA Disabled` text appears while the sys-info widget is loading. Check the non-HA system (e.g., nightly) and check on the HA system (e.g., m40).


AlexKarpov98 (approved):
Works as expected.

PR #9437, opened 2024-01-12
NAS-126710 / 24.04 / Memory conversion error in apps

AlexKarpov98:
Testing: see m40 Apps for example.


denysbutenko (approved):
No more `PiB and TiB` in the memory usage column.

Looks good.

PR #9438, opened 2024-01-12
NAS-125813 / 24.04 / Copying over enclosure info for disks.query

RehanY147:
The enclosure number and slot info on `disk.query` is not reliable. We should be using a completely new endpoint `enclosure2.query` or at worst `enclosure.query` for that info. We are supposed to have an overhaul for the enclosure view in Electric Eel. This PR ensurers that the disks have the correct and updated enclosure info by copying it over from the response of `enclosure.query` response.


undsoft (approved):
Confirmed fix on the machine in ticket and on our m40.

undsoft:
Nice work 👍🏻

PR #9447, opened 2024-01-16
NAS-126752 / 24.04 / Integrate new endpoints

bvasilenko:
**Testing**

1. **Storage** \> **Manage Devices** (Select any disk)

On card **Hardware Disk Encryption**, **Global SED Password** field

- is **Password is set** when `system.advanced.sed_global_password_is_set` returns `true`
- is **Password is not set** when `system.advanced.sed_global_password_is_set` returns `false`

image

2. **System Settings** \> **Advanced**

- **System Security** card exists when `system.security.info.fips_available` returns `true`
- **System Security** card does not exist when `system.security.info.fips_available` returns `false`

image


denysbutenko (approved):
Looks good

PR #9448, opened 2024-01-16
NAS-126714: Validate iSCSI via API

AlexKarpov98:
Testing:
see ticket, open Target Form, try with invalid name using `=`

You should see:
Screenshot 2024-01-16 at 11 17 51

Or write existing name:
Screenshot 2024-01-16 at 11 16 56


bvasilenko (approved):
Works fine for me!

One minor remark would be that "Add" button stays active, and user can actually click this button before the server-side validation roundtrip will finish. This click gets unnoticed by front-end validator, and is being passed to middleware, where it's rejected.


https://github.com/truenas/webui/assets/20611516/a11df6f8-210e-4389-86c7-e7d0b9cc509a

PR #9449, opened 2024-01-16
NAS-125762 / 23.10.2 / Installed apps sorting unsorts itself

AlexKarpov98:
Testing: see ticket, try start / stop apps and keep sorted by Status.

Result (sorted realtime):

https://github.com/truenas/webui/assets/22980553/58fc01f2-758b-463a-aa3f-51fb12dd8549

### DIFF
--- a/src/app/pages/apps/components/installed-apps/installed-apps.component.ts
+++ b/src/app/pages/apps/components/installed-apps/installed-apps.component.ts
@@ -63,6 +63,11 @@ export class InstalledAppsComponent implements OnInit, AfterViewInit {
   showMobileDetails = false;
   isMobileView = false;
   appJobs = new Map<string, Job<ChartScaleResult, ChartScaleQueryParams>>();
+  sortingInfo: Sort = {
+    active: SortableField.Application,
+    direction: SortDirection.Asc,
+  };
+
   readonly sortableField = SortableField;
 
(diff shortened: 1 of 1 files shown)


RehanY147 (changes requested):
If an app is running, and you click stop, the status changes, however the sorting doesn't fix itself. I sorted by Status. Clicked on "Stop" icon. The app status changed to "Stopping" but now, since the table is sorted by status, this app should move with the other apps with the same status. However, it still stays where it is until the whole table updates when the apps are reloaded.

AlexKarpov98:
> If an app is running, and you click stop, the status changes, however the sorting doesn't fix itself. I sorted by Status. Clicked on "Stop" icon. The app status changed to "Stopping" but now, since the table is sorted by status, this app should move with the other apps with the same status. However, it still stays where it is until the whole table updates when the apps are reloaded.

I think this is fine, because we rely on the STATUS which comes from the middleware, but not our local status.
Do you think we should rely on local status as well for this? (for me it's not critical)

@undsoft - what do you think?

RehanY147:
@AlexKarpov98 I think it's about the user expectation. The user wouldn't know what we are doing behind the scenes. They just see the status change without the sorting fixing itself. If the apps update takes a long while, the row stays there.

RehanY147 (approved):

PR #9451, opened 2024-01-16
NAS-125531 / 24.04 / Minor text updates

undsoft:
Code review is enough.

Some of the translations will be lost, but I don't see any other good alternative.


AlexKarpov98 (approved):
Looks fine, regarding translations - maybe we can go ahead over every translation and update text manually? Only `Legacy feature.` should be translated for each language - rest remains the same.

I counted, there are only max 8 translated .json files (regarding text change)

undsoft:
Yeah, I could do that. I'm just not sure if we can rely on machine translations. 😬

undsoft:
Looked into it some more. There're languages like Korean and Japanese, I don't feel confident enough to auto-translate. Hopefully, contributors will know to use git.
I'll update our translation instructions to hint at this.

PR #9452, opened 2024-01-16
NAS-126785 / 24.04 / Splitting feedback into multiple components

undsoft:
Testing scenarios: 
1. Submitting a review
2. Submitting a bug or improvement on a non enterprise system.
3. Submitting a bug on an enterprise system.
4. Seeing similar tickets when typing in title in a bug for for non enterprise system (requires fresh middleware).
5. Form state when feedback is disabled (you'd need to mock).

Known issues:
1. When clicking on EULA the dialog is not closed.
2. Images may not be uploaded from localhost, but you can check requests.


denysbutenko (approved):
If you type a message and switch from `Rate a page` to `Report a bug` or `Suggestion`, the field becomes empty.
The same goes for `Attach additional images`.

Other than that it looks good.

undsoft:
Yup, this would have to stay.

PR #9455, opened 2024-01-16
NAS-125407 / 24.04 / Polish Add Reporting Exporter Screen

AlexKarpov98:
Testing: See ticket + ticket comments.

Check `Reporting Exporter Screen` and abbreviation handling. (like ID, IP, IPs, IDs, API)
As well check that `boolean` type is rendered as checkbox and `integer` type is an input type=number.
Screenshot 2024-01-17 at 00 58 01

I did the same for `acmedns-form` as well.


src/app/helpers/object-keys-to-human-readable.helper.ts
```diff
@@ -1,13 +1,33 @@
 export function toHumanReadableKey(str: string): string {
+  // Capitalize first letter of each word
```
RehanY147, line 2: Comments are unnecessary as the method name is explanatory enough and can go out of date while the code evolves. Just an observation.

RehanY147 (approved):

PR #9456, opened 2024-01-17
NAS-124892 / 23.10.0 / Filter devices based on selected metrics

RehanY147:
When the "Metrics" select value changes, the "Devices" select options should change based on what disks have the selected metric graphs available.

### DIFF
--- a/src/app/pages/reports-dashboard/components/reports-global-controls/reports-global-controls.component.html
+++ b/src/app/pages/reports-dashboard/components/reports-global-controls/reports-global-controls.component.html
@@ -3,7 +3,7 @@
     <ix-select
       formControlName="devices"
       [label]="'Devices' | translate"
-      [options]="diskDevices$"
+      [options]="diskDevicesOptions$"
       [multiple]="true"
     ></ix-select>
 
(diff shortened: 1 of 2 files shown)


undsoft (changes requested):
1. This operates under assumption that this is why people are reporting that temperatures are missing. I would prefer to confirm this.
2. If it is a problem of expectation, I don't know if this really solves a problem. I can have Disk I/O and Temperatures selected (default state) and I will see my disk on the list, even if temperatures for it are not available.
3. When I navigate reporting, disks dropdown is empty even though metrics are selected.
4. Dropdowns would need to be swapped around.

In total I would prefer not to do it as it requires more thought as to how to communicate to users that certain metrics are not available to all disks.

alexmarkley:
I'm the original reporter of https://ixsystems.atlassian.net/browse/NAS-124892

I would be very happy to test this patch and report back if it resolves the issue as reported. Would that be helpful? If so, could someone point me in the right direction for how to test this?

I have similar reservations to @undsoft based on reading this patch. I am concerned that the underlying issue is being "swept under the rug" here. Namely, disk temperatures were being detected correctly in Bluefin and now they are missing in Cobia.

alexmarkley:
My disk temperature issue (which I reported as [NAS-124892](https://ixsystems.atlassian.net/browse/NAS-124892) is resolved in 23.10.1.1. I suspect the relevant changelog entry was: _Fix disk temperature reporting ([NAS-125841](https://ixsystems.atlassian.net/browse/NAS-125841)).​_

RehanY147:
The PR issue will be handled by a different ticket.

PR #9459, opened 2024-01-17
NAS-126753 / 24.04 / Wrong seconds calculation in WebUI Sessions Settings Widget

AlexKarpov98:
Testing: 
Go to `Advanced Settings -> Access` card
http://localhost:4200/system/advanced

Try to update `Token Lifetime`
Screenshot 2024-01-17 at 12 08 47

Make sure that it works if seconds in total are greater then 24 hours:

Example:
Screenshot 2024-01-17 at 12 04 13
Screenshot 2024-01-17 at 12 04 33

### DIFF
--- a/src/app/pages/system/advanced/access/access-card/access-card.component.ts
+++ b/src/app/pages/system/advanced/access/access-card/access-card.component.ts
@@ -163,7 +163,7 @@ export class AccessCardComponent implements OnInit {
   asDuration(tokenLifetime: number): string {
     const duration = intervalToDuration({ start: 0, end: tokenLifetime * 1000 });
     return formatDuration(duration, {
-      format: ['hours', 'minutes', 'seconds'],
+      format: ['days', 'hours', 'minutes', 'seconds'],
     });
   }
 
--- a/src/app/pages/system/advanced/access/access-form/access-form.component.ts
+++ b/src/app/pages/system/advanced/access/access-form/access-form.component.ts
@@ -29,8 +29,8 @@ export class AccessFormComponent implements OnInit {
   form = this.fb.group({
     token_lifetime: [defaultPreferences.lifetime, [
       Validators.required,
-      Validators.min(30),
-      Validators.max(Math.floor(2 ** 31 / 1000 - 1)), // Max value for setTimeout
+      Validators.min(30), // Min value allowed is 30 seconds.
+      Validators.max(2147482), // Max value is 2147482 seconds, or 24 days, 20 hours, 31 minutes, and 22 seconds.
     ]],
     ds_auth: [false],
   });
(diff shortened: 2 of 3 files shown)


src/app/pages/system/advanced/access/access-form/access-form.component.ts
```diff
@@ -29,15 +29,16 @@ export class AccessFormComponent implements OnInit {
   form = this.fb.group({
     token_lifetime: [defaultPreferences.lifetime, [
       Validators.required,
-      Validators.min(30),
-      Validators.max(Math.floor(2 ** 31 / 1000 - 1)), // Max value for setTimeout
+      Validators.min(30), // Min value allowed is 30 seconds.
+      Validators.max(2147482), // Max value is 2147482 seconds, or 24 days, 20 hours, 31 minutes, and 22 seconds.
```
razvanphp, line 33: just curious, how is this fixing it? I tried it in the console, and the `Math.floor` function returns the same thing...

src/app/pages/system/advanced/access/access-card/access-card.component.ts
```diff
@@ -163,7 +163,7 @@ export class AccessCardComponent implements OnInit {
   asDuration(tokenLifetime: number): string {
     const duration = intervalToDuration({ start: 0, end: tokenLifetime * 1000 });
     return formatDuration(duration, {
-      format: ['hours', 'minutes', 'seconds'],
+      format: ['days', 'hours', 'minutes', 'seconds'],
```
AlexKarpov98, line 166: @razvanphp - actual fix is here, we missed `days`, so the problem was that instead of `1 day 1 second` it omitted day and showed `1 second` for example.

src/app/pages/system/advanced/access/access-form/access-form.component.ts
AlexKarpov98: > just curious, how is this fixing it? I tried it in the console, and the `Math.floor` function returns the same thing...

I don't know why do we need this calculation `Math.floor(2 ** 31 / 1000 - 1)` if documentation clearly says max is `2147482` (basically the same)

undsoft: Max for setTimeout is a 32bit value, hence 2 to the power of 32 - 1 second was used.

razvanphp: Maybe good to have this documented with `Math.floor`, everybody was asking this in the forum...

RehanY147 (approved):

src/app/pages/system/advanced/access/access-form/access-form.component.ts
AlexKarpov98: > Maybe good to have this documented with `Math.floor`, everybody was asking this in the forum...

@truenas/docs-team

PR #9460, opened 2024-01-17
NAS-125547 / 24.04 / The Disks Are Displayed In A Random Order On The Disk I/O Reporting Page

AlexKarpov98:
Testing: 
go to Disk Reporting
http://localhost:4200/reportsdashboard/disk

See that Disks are sorted in alphabetical order in the `Select` and In `Cards`

Example:
Screenshot 2024-01-17 at 12 33 24


undsoft (approved):
Looks okay, but one ? is missing which breaks the page when you click Reporting on main menu.

PR #9466, opened 2024-01-18
NAS-125604 / 24.04 / Poor error handling in apps

AlexKarpov98:
Testing: 
See ticket.
for example try changing request to get MiddleWare errors.

You should see improved Error Handling For Apps. 
Check that loaders are stopping on errors.

Example:

https://github.com/truenas/webui/assets/22980553/35826242-0036-4efe-a20f-ec60dd2be4a7

### DIFF
--- a/src/app/pages/apps/store/apps-store.service.ts
+++ b/src/app/pages/apps/store/apps-store.service.ts
@@ -3,7 +3,7 @@ import { UntilDestroy } from '@ngneat/until-destroy';
 import { ComponentStore } from '@ngrx/component-store';
 import {
   EMPTY,
-  Observable, catchError, combineLatest, switchMap, tap,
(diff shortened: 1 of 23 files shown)


src/app/pages/apps/store/apps-store.service.ts
```diff
@@ -92,20 +93,15 @@ export class AppsStore extends ComponentStore<AppsState> {
           };
         });
       }),
-      catchError(() => {
-        this.handleError();
+      catchError((error) => {
```
RehanY147, line 96: Don't think removing individual error handling for those methods is better here. These methods are handling success case by mutating state. They should also handle error cases individually. Although, they are not called individually right now, so I can approve this, but if they are called individually in the future, it is assumed that they will handle all cases on their on because they are handling success case.

RehanY147 (approved):

PR #9468, opened 2024-01-19
NAS-126815 / 24.10 / Fix invalid group values in ix-chips and user form

denysbutenko:


denysbutenko:
> Well, the reason tests do not work is because behaviour of ix-chips changed. Previously you could type `admin` in search and press Enter and it would add `admin`. This is no longer works. I suggest we keep this old behaviour.
> 
Yeah, it is. Fixed it in another way.

> We also need to do something with `resolveValue` and `options`. It's vague as what they do and what the difference between `autocompleteProvider` and `options` is.

Added explanation

undsoft (approved):
Looks reasonable, thanks.

PR #9470, opened 2024-01-19
NAS-126848 / 24.10 / Export to CSV Not Working

AlexKarpov98:
Testing: 
See ticket's comment about request update.

Testing for file downloading is limited (locally). You need to verify that it issues the correct downloading request. For this request to work, the WebUI should be built and hosted from the middleware (not from the angular development server), and such testing is outside of scope of this code review.

As well I fixed loading indicator and button **disabled** state

https://github.com/truenas/webui/assets/22980553/6eb1dbf8-1d6c-40af-9d44-3abe6784c79a


bvasilenko (changes requested):
I've noticed that inputs of the `method` input and `coreDownloadMethod` input look a bit not settled: it worth checking if `method` and `coreDownloadMethod` are actually duplicates, and in that case we need to keep only one. It's safe to edit the interface of `ix-export-button`, since it's used solely by audit page.

As for additional inputs, in future we'll may want to override `.csv`  and `text/csv` , it might be a good idea to extract these as new `fileType` input and as  new `fileMimeType` input.

AlexKarpov98:
@bvasilenko - @anodos325 provided an example:
```
def test_audit_export(request):
    depends(request, ["AUDIT_OPS_PERFORMED"], scope="session")
    for backend in ['CSV', 'JSON', 'YAML']:
        report_path = call('audit.export', {'export_format': backend}, job=True)
        assert report_path.startswith('/audit/reports/root/')
        st = call('filesystem.stat', report_path)
        assert st['size'] != 0, str(st)

        job_id, path = call("core.download", "audit.download_report", [{
            "report_name": os.path.basename(report_path)
        }], f"report.{backend.lower()}")
        r = requests.get(f"{url()}{path}")
        r.raise_for_status()
        assert len(r.content) == st['size']
```

So I think we should first call `audit.export` and after that `"core.download", "audit.download_report"`, that's why I provided   `coreDownloadMethod`.

bvasilenko:
> Do you think we can somehow test downloading locally?

@AlexKarpov98 one idea is to adjust bash command, which you can get using **Copy > Copy as cURL** from the context menu of the `/_download` request on **Network** tab opened at your local development server. The adjustment you need to make: replace host of your local development server with the host of your truenas box, and also adjust `Origin` and `Referrer` HTTP headers, and last detail to adjust is auth token at `Cookie` HTTP header. When you have a command, you can obtain the file and examine it's contents.

AlexKarpov98:
@bvasilenko - can't make it work on my side, were you able to tweak it and actually test it locally?

AlexKarpov98:
@bvasilenko - we may try merging and test on the latest nightly and if any problem - create another PR 😄

bvasilenko (approved):
Looks ok from the code perspective
We're going to test it on the Nightly

PR #9471, opened 2024-01-19
NAS-126789 / 24.10 / Update Timeout Dialog to Match Access Widget

AlexKarpov98:
Testing: see ticket, make Sure you see correct text and navigated correctly to the **Advanced Settings** & **Focused** on the **Access** card.

Screenshot 2024-01-19 at 16 47 18


undsoft (approved):
Looks good and is a good idea, but in the future we should make this a directive. I'll create a ticket for this.

PR #9473, opened 2024-01-22
NAS-125605 / 24.10 / Do not reload system.info every 30 seconds

denysbutenko:
For testing, ensure uptime values are still updating on the system information widgets.

### DIFF
--- a/src/app/pages/dashboard/components/widget-sys-info/widget-sys-info.component.ts
+++ b/src/app/pages/dashboard/components/widget-sys-info/widget-sys-info.component.ts
@@ -9,12 +9,14 @@ import { Router } from '@angular/router';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { Store } from '@ngrx/store';
 import { TranslateService } from '@ngx-translate/core';
-import { filter, repeat, take } from 'rxjs/operators';
+import { filter, take } from 'rxjs/operators';
 import { JobState } from 'app/enums/job-state.enum';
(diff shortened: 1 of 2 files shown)


AlexKarpov98 (commented):
@undsoft named ticket: `Do not reload system.info every 30 seconds`
@denysbutenko - you did that it now updates every 5 seconds 😄 

So what's the correct behaviour? Shall we remove at all that interval?

undsoft:
Auto-ddos. We shouldn't send recurring requests to middleware.

denysbutenko:
> @undsoft named ticket: `Do not reload system.info every 30 seconds` @denysbutenko - you did that it now updates every 5 seconds 😄
> 
> So what's the correct behaviour? Shall we remove at all that interval?

I'll try to explain the difference. The problem is when we call `system.info` in the global scope of the app store, it causes an issue described in https://ixsystems.atlassian.net/browse/NAS-125603. Now, it is moved to widget scope and works similarly to a network widget where data is updated with a short delay.

denysbutenko:
> Auto-ddos. We shouldn't send recurring requests to middleware.

Okay, we can switch back to the internal counter implementation again because necessary information is available in a single query.

AlexKarpov98 (changes requested):
Interval is not stopped when component is destroyed.
see:

https://github.com/truenas/webui/assets/22980553/32382cc5-3366-49d9-b78e-ac4d900666a8

AlexKarpov98 (approved):
Works fine and interval is cleared once component is destroyed.

PR #9478, opened 2024-01-22
NAS-126909 / 24.10 / Bug in the Workflows - If I created PR with translation files and with other files - it skips required steps

AlexKarpov98:
Testing:
Create new temporary PR based on this branch (like I did here https://github.com/truenas/webui/pull/9488) and make sure that if translation files were changed together with any other files, we should NOT skip build, test and other workflow steps.

Screenshot 2024-01-23 at 03 12 19

If only translation files were changed - then we skip build, test and so on.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,6 +47,7 @@ jobs:
               echo "translations-only=true" >> $GITHUB_ENV
             else
               echo "translations-only=false" >> $GITHUB_ENV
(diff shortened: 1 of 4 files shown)


undsoft (approved):
https://github.com/truenas/webui/pull/9469 didn't have any jobs reported, but translations otherwise were resorted and worked correctly in other scenarios. To be tested further.

PR #9483, opened 2024-01-22
NAS-126843: Fix max concurrent calls bug

bvasilenko:
**Summary**

Until login request request is finished, I've disabled sign-in form

**Testing**

On **Login** page,

1. Enter login and password (try both correct and incorrect)
2. Press **Enter** key multiple times
3. After \~20 quick subsequent presses of **Enter** key, it displays

- **Expected result:** it should not be possible to submit a new attempt to login, until there's unfinished login request .

### DIFF
--- a/src/app/views/sessions/signin/signin-form/signin-form.component.ts
+++ b/src/app/views/sessions/signin/signin-form/signin-form.component.ts
@@ -6,7 +6,7 @@ import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { TranslateService } from '@ngx-translate/core';
 import _ from 'lodash';
 import {
-  distinctUntilChanged, EMPTY, switchMap,
+  distinctUntilChanged, EMPTY, firstValueFrom, switchMap,
 } from 'rxjs';
 import { LoginResult } from 'app/enums/login-result.enum';
 import { WINDOW } from 'app/helpers/window.helper';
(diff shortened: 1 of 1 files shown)


RehanY147 (changes requested):
This does what the PR is meant to do. However, I can't reproduce the issue unless I REALLY try to reproduce it by clicking the enter really quickly for many many times. The user is reporting that they can see this issue but casually just pressing enter while the form is loading.

bvasilenko:
@RehanY147 I confirm that issue is not reproducible the way it's described. The only way to replicate it is to actually misuse the login form and try to submit incorrect (or correct) password 20+ times.

So this PR turns out to be a purely cosmetic improvement, which might however affect the automated tests.

@undsoft should we close it as "Cannot reproduce" ?

undsoft:
@bvasilenko Can we not prevent from submission until result for previous password has been obtained? If it's easy, let's do it, otherwise we can close the ticket.

RehanY147 (approved):
@undsoft The PR does that. Just the reproduction steps are different from what the user is reporting. We might be fixing an issue. But, I'm not sure if we are fixing the issue the user has reported.

undsoft:
Well, the reporter is @AlexKarpov98, perhaps he can help.
I cannot reproduce as well, neither on this branch, not on master.

undsoft:
@AlexKarpov98 Can you reproduce it on the branch?

AlexKarpov98:
@undsoft -  I can't reproduce it anymore with these changes here.
So this PR resolves the issue.

PR #9495, opened 2024-01-23
NAS-126919 / 24.10 / On Dragonfish BETA.1  the licence agreement appears multiple times.

AlexKarpov98:
See ticket.

### DIFF
--- a/src/app/helpers/operators/filter-async.operator.ts
+++ b/src/app/helpers/operators/filter-async.operator.ts
@@ -4,8 +4,6 @@ import {
 
 /**
  * Similar to filter(), but accepts an observable.
- * The first() operator will take the first emitted value from asyncPredicate$
- * and then complete, which means it won't emit multiple times for the same source item.
  */
 export function filterAsync<T>(asyncPredicate$: Observable<boolean>): OperatorFunction<T, T> {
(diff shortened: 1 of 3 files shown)


undsoft (changes requested):
Only hides the problem, let's go for solutions that fix the underlying issue.

AlexKarpov98:
> Only hides the problem, let's go for solutions that fix the underlying issue.

Problem is that filterAsync may emit multiple times, not sure if we need that behaviour, or just get first() value from the successful observable.

By adding the first() operator, asyncPredicate$ will only be considered once per each item from the source observable.

undsoft (changes requested):
`filterAsync` does not emit multiple times. `hasRole` does. I would suggest checking usages for it and seeing if we can/should add `take(1)` there. 
If it's possible let's add it there and test.
If not possible, we can add `take(1)` in this effect in `hasRole` pipe.

Also note about `first`: https://stackoverflow.com/questions/42345969/take1-vs-first

AlexKarpov98:
@undsoft - yeah, it's possible to add take(1) in the `hasRole`. Now It only emits once.

undsoft (approved):
Alright, that's fine too.

PR #9507, opened 2024-01-24
NAS-126904 / 24.10 / Directive for highlighting parts of UI

AlexKarpov98:
Testing: see ticket.
Try with `session-expiring-dialog.component` and with Email settings:

https://github.com/truenas/webui/assets/22980553/c5e8a444-1c4b-4714-bc4a-33f288357cea


undsoft (changes requested):
There was a misunderstanding about the task.
The task is not to have a directive for <a> styling, the task is to take out the behaviour we have for when we click on the link (focusing on a specific portion of the page).

AlexKarpov98:
> There was a misunderstanding about the task. The task is not to have a directive for styling, the task is to take out the behaviour we have for when we click on the link (focusing on a specific portion of the page).

I'm sorry, please check it now :)

PR #9509, opened 2024-01-24
NAS-126826 / 24.10 / On the latest Dragonfish nightly build Pool Wizard all mat-option data-test  are missing the option text.

AlexKarpov98:
Testing: 
See ticket, go to http://localhost:4200/storage/create

And see that data-test for `Layout` is now as expected, example:
**data-test="option-layout-RAIDZ1"**

So the Idea here is to append original data-test last part value to overwritten data-test attribute value.
Because for select->option - we can't get option.value from the parent component and append it dynamically via `ixTestOverride`

But with this update, we can override it:
before: **option-RAIDZ1**
with ixTestOverride: **option-layout**
with ixTestOverride && keepLastPart: **option-layout-RAIDZ1**

### DIFF
--- a/src/app/modules/test-id/test/test.directive.ts
+++ b/src/app/modules/test-id/test/test.directive.ts
@@ -36,7 +36,7 @@ export class TestDirective {
       .filter((part) => part)
       .map((part) => kebabCase(String(part)));
 
-    if (this.overrideDirective?.appendInitialDataTestLastPart) {
+    if (this.overrideDirective?.keepLastPart) {
       const normalizedInitialDescription = Array.isArray(this.description) ? this.description : [this.description];
       normalizedDescription.push(normalizedInitialDescription[normalizedInitialDescription.length - 1]);
     }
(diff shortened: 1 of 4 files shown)


src/app/modules/test-id/test/test.directive.ts
```diff
@@ -30,11 +30,18 @@ export class TestDirective {
...
     const description = this.overrideDirective?.overrideDescription ?? this.description;
-    const normalizedDescription = Array.isArray(description) ? description : [description];
+    let normalizedDescription = Array.isArray(description) ? description : [description];
 
-    return normalizedDescription
+    normalizedDescription = normalizedDescription
       .filter((part) => part)
       .map((part) => kebabCase(String(part)));
+
+    if (this.overrideDirective?.appendInitialDataTestLastPart) {
```
bvasilenko, line 39: We might improve the naming here `this.overrideDirective?.keepLastPart`

bvasilenko (approved):
Works as expected

PR #9525, opened 2024-01-25
NAS-126848 / 24.10 / Export to CSV Not Working

AlexKarpov98:
Testing: 
First on UI - we get 504 error:

Screenshot 2024-01-25 at 15 54 35

Then we need to grab that URL, and use with the env host instead of localhost, for i.e I use m40.

Example:
We have: http://localhost:4200/_download/7010?auth_token=sxDVmUAZUq1I82SWLEToyg_BFn9ITdMtGhmuSrYrfF62_HMl7se0kMxiGbYS5R7h

We need to make it:
**<<<ENV_HOST>>>**/_download/7010?auth_token=sxDVmUAZUq1I82SWLEToyg_BFn9ITdMtGhmuSrYrfF62_HMl7se0kMxiGbYS5R7h

Then try to open that link in browser, then the file is downloading, and it should not be empty.

I had this for example in the end:
[_audit_reports_root_3f15016c-0612-40bb-9594-9fda301ce49e.csv](https://github.com/truenas/webui/files/14052986/_audit_reports_root_3f15016c-0612-40bb-9594-9fda301ce49e.csv)

@bvasilenko - please check direct message for detailed video.


bvasilenko (approved):
Code looks good, so I'm approving for integration testing on Nightly

AlexKarpov98:
> Code looks good, so I'm approving for integration testing on Nightly
@bvasilenko  - I tested on latest nightly - works as expected, file is downloaded with data inside.

PR #9527, opened 2024-01-25
NAS-126827 / 24.10 / Add different success messages for feedback

bvasilenko:
**Testing**

Please see ticket


src/app/modules/feedback/services/feedback.service.ts
```diff
@@ -182,6 +190,26 @@ export class FeedbackService {
     );
   }
 
+  showSnackbar(ticketUrl?: string): MatSnackBarRef<SnackbarComponent> {
```
RehanY147, line 193: This method name and implementation is ambiguous. We should have two methods instead of one.
`showTicketSuccessMsg(url: string): MatSnackBarRef<SnackbarComponent>`
and
`showFeedbackSuccessMsg(): MatSnackBarRef<SnackbarComponent>`

RehanY147 (changes requested):

src/app/modules/feedback/components/file-ticket-licensed/file-ticket-licensed.component.ts
```diff
@@ -117,11 +115,13 @@ export class FileTicketLicensedComponent {
 
     this.prepareTicket().pipe(
       switchMap((ticket) => this.feedbackService.createNewTicket(ticket)),
-      switchMap((createdTicket) => this.addAttachmentsIfNeeded(createdTicket.ticket)),
+      switchMap((createdTicket) => {
+        return this.addAttachmentsIfNeeded(createdTicket.ticket).pipe(switchMap(() => of(createdTicket)));
```
RehanY147, line 119: In fact, a better approach would be to move all of this chained method calling inside the `FeedbackService` itself. There's no reason apparent why the component needs to make these decisions. The service can do it with one method call. And it shouldn't be up to the component to make the call to show the success messages either. The service can and should be responsible for that too. The component should simply be doing
`this.feedbackService.createNewTicket(ticket).pipe(untilDestroyed(this)).subscribe();` or something similar.
Same applies for other places where we are trying to arrange method calls inside the component. Or, we can move just the snackbar showing part back to the component as it makes sense to leave it up to the component what to do in case of errors or success. That might be different for different components.

RehanY147 (changes requested):
I think a better approach would be

RehanY147 (approved):
Nice!

PR #9532, opened 2024-01-26
NAS-127001 / 24.10 / Fix resort translations action for external contributors

AlexKarpov98:


AlexKarpov98:
After a lot of check.
It seems impossible to do it for external contributors who can make changes via forked project PR.
Workflow will not have an access to the external branch and adding tokens for all external contributors - seems like overkill.

Another idea - do resorting once we merge into master.
Gonna think about this approach together with @undsoft

PR #9538, opened 2024-01-26
NAS-126992 / 24.10 / Mini-R image missing from dashboard widget

AlexKarpov98:
Testing:
you may need to mock **isIxHardware** to be **true** (see `loadIsIxHardware`)
Only in this case you will be able to see image.
Screenshot 2024-01-26 at 17 03 51


undsoft:
@denysbutenko Has this been unblocked?

AlexKarpov98:
> @denysbutenko Has this been unblocked?

why was it blocked?

undsoft (approved):
The image is shown, although I think it would be better to fix the issue at assignment and not in template.

PR #9541, opened 2024-01-28
NAS-123855 / 24.10 / Sublight (H-series) Enclosure Support - UI

bvasilenko:
**Testing**

Use the server from the ticket.

On the **System Settings > Enclosure** page:

1. Ensure that the front panel of the new enclosure is visually matching the one attached to the ticket (two controllers at both sides and general look of the front-panel)
2. Ensure that the page is showing overview of the server rack in the right sidebar
3. ...showing server's disks:
   - disk images are positioned at correct coordinates at the front panel image
   - individual disk can be clicked, resulting in highlighting and showing the details of this disk in the bottom 
3. ...showing sensor list:
   - each sensor shows it's value and this value is updated


AlexKarpov98 (approved):
It works:

Screenshot 2024-01-29 at 09 13 32

PR #9542, opened 2024-01-29
NAS-126856 / 24.10 / Improve similar tickets search results

denysbutenko:
Check ticket description


bvasilenko (approved):
Yes it became better, and works as expected

PR #9544, opened 2024-01-29
NAS-127039 / 24.10 / Feedback form - Rate page error handling is missing, i.e I try to upload 10mb image

AlexKarpov98:
Testing:
See ticket & try to upload large image for the feedback form.

### DIFF
--- a/src/app/helptext/system/support.ts
+++ b/src/app/helptext/system/support.ts
@@ -133,5 +133,6 @@ export const helptextSystemSupport = {
   attachmentsFailed: {
     title: T('Attachments not uploaded'),
     message: T('Ticket was created, but we were unable to upload one or more attachments.'),
+    reviewMessage: T('Review was created, but we were unable to upload one or more attachments.'),
   },
 };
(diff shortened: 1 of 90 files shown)


AlexKarpov98:
@denysbutenko - what do you think about it?
I don't really like the idea of just silently ignoring image uploading.
If that fine, we can close the PR and ticket.

denysbutenko:
> @denysbutenko - what do you think about it? I don't really like the idea of just silently ignoring image uploading. If that fine, we can close the PR and ticket.

I think it would be better to prevent adding files if the size is exceeded - raise validation error. Some time ago, I assumed that JIRA and feedback service have equal limits, which was wrong - it needs to differentiate or create a ticket to sync limits. Regarding the current behavior of silently ignoring errors - I think that's okay, but it would be better to ask @undsoft

AlexKarpov98:
I like the idea of limit sync and raising error if file is too big - before upload.
@undsoft

undsoft:
Both behaviours are okay. The reason I disabled error reporting for reviews is because it's not too critical in reviews if image was uploaded, but having an error could work too.
Validation error sounds good.

denysbutenko:
@AlexKarpov98 would you like to merge this PR or plan to change something?

AlexKarpov98:
> @AlexKarpov98 would you like to merge this PR or plan to change something?

@denysbutenko - I think we can close this one and create new ticket. --> `Feedback form: File Size Limit sync & raise error if file is too big - before upload.`

PR #9547, opened 2024-01-29
NAS-127050 / 24.10 / Audit header on load time is partially invisible

AlexKarpov98:
Testing:
see ticket & audit page header on page reload.


bvasilenko (approved):
Works as expected!

PR #9557, opened 2024-01-30
NAS-126983 / 24.10 / Share Page: iSCSI Service Status Label and Configure button UI overlap in different browser window sizes

AlexKarpov98:
Testing:
See Shares page on smaller screens and test other cards with toolbar headers.
As well some colouring fixes were made, I switched theme and detected them.

**Result:**
Screenshot 2024-01-31 at 11 24 47

Screenshot 2024-01-31 at 11 24 15

**Before, it was like 👇 :**
Screenshot 2024-01-30 at 10 24 25

### DIFF
--- a/src/app/modules/entity/entity-table/entity-table.component.scss
+++ b/src/app/modules/entity/entity-table/entity-table.component.scss
@@ -378,6 +378,6 @@ div.mat-mdc-card-table {
   justify-content: space-between;
 
   h3 {
-    color: var(--primary-txt);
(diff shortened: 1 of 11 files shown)


RehanY147 (changes requested):
Let's align buttons to the right instead of left. That seems more natural. Especially, with the three dots menu button.

AlexKarpov98:
@RehanY147 - ready 😄

RehanY147 (approved):

PR #9559, opened 2024-01-30
NAS-126987 / 24.10 / The Services Toggles are not working correctly, HA is Disabled, and the WebUI disconnects

AlexKarpov98:
Testing: see ticket and see that it's no longer reproduceable.
Problem was that we didn't update `enable` value in the store, but only in the component, so once we toggle `Running` switch - we returned earlier state with `enable` for each item. 
And since it was not updated in store before - we got old values back.

### DIFF
--- a/src/app/pages/services/services.component.ts
+++ b/src/app/pages/services/services.component.ts
@@ -28,6 +28,7 @@ import { IscsiService } from 'app/services/iscsi.service';
 import { IxSlideInService } from 'app/services/ix-slide-in.service';
 import { UrlOptionsService } from 'app/services/url-options.service';
 import { WebSocketService } from 'app/services/ws.service';
+import { serviceDisabled, serviceEnabled } from 'app/store/services/services.actions';
 import { ServicesState } from 'app/store/services/services.reducer';
 import { waitForServices } from 'app/store/services/services.selectors';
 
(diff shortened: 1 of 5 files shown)


AlexKarpov98:
@undsoft please as well check my comment:

Problem was that we didn't update `Enable` value in the store, but only in the component, so once we toggle `Running` switch - we returned earlier state with enable for each item.
And since it was not updated in store before - we got old values back.

undsoft:
`subscribeToUpdates$` is supposed to update the store.

AlexKarpov98:
> subscribeToUpdates$

@undsoft - it only runs for one service a time for `STARTED` prop, not for `ENABLE` prop (as I tested it.)

But anyway - problem with store update is still the case, if I only run `ENABLE` override locally - it will be reset once `STARTED` process has ended.

undsoft:
> not for ENABLE prop (as I tested it.)

So a potential middleware issue.

AlexKarpov98:
@undsoft - I am trying to reach @yocalebo at this time for middleware fix.
But actually - it can make a glitch on the UI side without my current fix (blinking enable checkbox [setting/unsetting]).

undsoft (approved):
Okay, code changes make sense.

PR #9560, opened 2024-01-30
NAS-127086 / 24.10 / Missing permission checks in Apps

undsoft:
Testing:
Check permissions in Apps section as a readonly user.

Note:
- installed apps may not load because of a middleware issue. In this case login as full admin and mock AuthService.hasRole to return false.
- pod logs button is not working, but I believe it should. I have created a middleware ticket for it.


AlexKarpov98 (approved):
Permissions logic part works as expected. 👍

PR #9566, opened 2024-01-31
NAS-127002 / 24.10 / Refactor AlertServiceListComponent to ix-tables2

AlexKarpov98:
Testing: check for any regressions.

Before:
Screenshot 2024-01-31 at 13 26 49

After:
Screenshot 2024-01-31 at 13 26 56


src/app/enums/role.enum.ts
```diff
@@ -101,8 +102,9 @@ export enum Role {
 
 export const roleNames = new Map<Role, string>([
   [Role.AlertListRead, T('Alert List Read')],
+  [Role.AlertListWrite, T('Alert List Write')],
   [Role.AuthSessionsRead, T('Auth Sessions Read')],
-  [Role.AuthSessionsWrite, T('Alert Sessions Write')],
+  [Role.AuthSessionsWrite, T('Auth Sessions Write')],
```
undsoft, line 107: Nicely spotted, please create a PR for Dragonfish for fixing labels as well.

src/app/interfaces/api/api-call-directory.interface.ts
```diff
@@ -294,7 +294,7 @@ export interface ApiCallDirectory {
   'alertclasses.config': { params: void; response: AlertClasses };
   'alertclasses.update': { params: [AlertClassesUpdate]; response: AlertClasses };
   'alertservice.create': { params: [AlertServiceEdit]; response: AlertService };
-  'alertservice.delete': { params: number; response: boolean };
+  'alertservice.delete': { params: number[]; response: boolean };
```
undsoft, line 297: If it only supports one id when deleting, it's better to write `[number]`

src/app/pages/system/alert-service/alert-service-list/alert-service-list.component.html
```diff
@@ -0,0 +1,41 @@
+<mat-card>
+  <mat-toolbar-row>
+    <h3 class="card-title">{{ 'Alert Services' | translate }}</h3>
+
+    <div>
+      <ix-search-input [value]="filterString" (search)="filterUpdated($event)"></ix-search-input>
+      <ix-table-columns-selector [columns]="columns" (columnsChange)="columnsChange($event)"></ix-table-columns-selector>
+      <button
+        *ixRequiresRoles="requiresRoles"
```
undsoft, line 9: We don't add roles checks on buttons here, but instead add them in the form. We need to allow users to open the edit form because we may not be showing all the fields in the table. And then to keep UI consistent between edit and add scenarios, we should not have checks on this button.

src/app/pages/system/alert-service/alert-service-list/alert-service-list.component.ts
```diff
@@ -16,93 +24,129 @@ import { WebSocketService } from 'app/services/ws.service';
...
+      }),
+    }).pipe(
+      filter(Boolean),
+      switchMap(() => this.ws.call('alertservice.delete', [alertService.id])),
+      untilDestroyed(this),
+    ).subscribe({
+      next: () => {
+        this.getAlertServices();
+      },
+      error: (error) => {
```
undsoft, line 143: You can also do `this.errorHandler.catchError` inside of the pipe.

src/app/pages/system/alert-service/alert-service-list/alert-service-list.component.ts
```diff
@@ -16,93 +24,129 @@ import { WebSocketService } from 'app/services/ws.service';
...
+      getValue: (service) => (service.enabled ? this.translate.instant('Yes') : this.translate.instant('No')),
+    }),
+    actionsColumn({
+      cssClass: 'wide-actions',
+      actions: [
+        {
+          iconName: 'edit',
+          tooltip: this.translate.instant('Edit'),
+          onClick: (row) => this.editAlertService(row),
+          hidden: (row) => of(!this.providerList.includes(row.type)),
```
undsoft, line 67: What does this check do? Are there alert services we want to hide?

AlexKarpov98: @undsoft - it's some logic from the old code, gonna remove it 👍

PR #9576, opened 2024-02-01
NAS-126976 / 24.10 / Slide In When Closed Is Not Setting Focus On The Button Which Triggered Slide In To Open

AlexKarpov98:
Testing: see ticket 

Result: Focus is restored once slide in is closed.

https://github.com/truenas/webui/assets/22980553/06a44b5b-e046-4383-90ee-e7cffe41cbe6


denysbutenko (approved):
Nice improvement. Works great.

PR #9577, opened 2024-02-01
NAS-126957 / 24.10 / SMB Status Screen Improvements

AlexKarpov98:
Testing: see ticket


RehanY147 (commented):
Change tilte from `Smb` to `SMB`

RehanY147 (changes requested):
Also, breadcrumb for `NFS Sessions` should be `Sharing -> NFS`

AlexKarpov98:
@RehanY147 - okay, but please go ahead with the ticket borders, if we need to fix another things - it's always better to create a separate ticket!

AlexKarpov98:
@RehanY147 - check it out now.

RehanY147 (approved):

PR #9584, opened 2024-02-01
NAS-127146 / 24.10 / Enabling change detection onPush for more components

undsoft:
Test some of the updated components.


src/app/modules/ix-tables/components/ix-table/ix-table.component.ts
```diff
@@ -39,6 +39,7 @@ import { MatTable } from '@angular/material/table';
   ],
   encapsulation: ViewEncapsulation.None,
   // See note on CdkTable for explanation on why this uses the default change detection strategy.
+  // eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
AlexKarpov98, line 42: Why do we need it?

AlexKarpov98 (approved):
Works good. Was not able to find any problems regarding these updates.

src/app/modules/ix-tables/components/ix-table/ix-table.component.ts
undsoft: Because apparently according to the comment above Default change detection strategy is required for `CdkTable`. i don't know specifics, but decided not to override it.

PR #9591, opened 2024-02-01
NAS-127153 / 24.10 / Upgrade node in CI

undsoft:
No review necessary.


denysbutenko (approved):
Dockerfile is still using version 16. Looks good.

PR #9592, opened 2024-02-02
NAS-127071 / 24.04-RC.1 / Add additional check for dialog display

bvasilenko:
**Testing**

See ticket


undsoft (approved):
Looks good, but let's rename `isCreatingSmb` and `isCreatingNfs` to something like `canCreateSmb` and `canCreateNfs`.

PR #9597, opened 2024-02-02
NAS-126976 / 24.10 / Focus service updates to correctly handle table reloading and loosing html ref & other improvements

AlexKarpov98:
Testing:
I noticed that focus is lost if we update form and return to the table (html element is rerendered and it's reference is lost, I got an Idea to depend on data-test attribute, since it's unchanging)
As well did some fixes for Directory Services.

### DIFF
--- a/src/app/services/focus.service.ts
+++ b/src/app/services/focus.service.ts
@@ -19,7 +19,7 @@ export class FocusService {
       if (this.lastFocusedElement) {
         setTimeout(() => {
           const dataTestValue = this.lastFocusedElement.getAttribute('data-test');
-          const hasOverlayWithChildren = this.document.querySelector('.cdk-overlay-container').hasChildNodes();
+          const hasOverlayWithChildren = this.document.querySelector('.cdk-overlay-container')?.hasChildNodes();
 
           if (dataTestValue && !hasOverlayWithChildren) {
             const dataTestElement = this.document.querySelector(`[data-test="${dataTestValue}"]`);
--- a/src/app/services/focus.service.ts
+++ b/src/app/services/focus.service.ts
@@ -21,7 +21,7 @@ export class FocusService {
           const dataTestValue = this.lastFocusedElement.getAttribute('data-test');
 
           if (dataTestValue) {
-            const dataTestElement = document.querySelector(`[data-test="${dataTestValue}"]`);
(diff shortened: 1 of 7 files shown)


src/app/services/focus.service.ts
```diff
@@ -17,8 +17,18 @@ export class FocusService {
 
     restoreFocus(): void {
       if (this.lastFocusedElement) {
-        this.lastFocusedElement.focus();
-        this.lastFocusedElement = null;
+        setTimeout(() => {
+          const dataTestValue = this.lastFocusedElement.getAttribute('data-test');
```
denysbutenko, line 21: Is it possible when the last focused element does not have a `data-test` attribute?

src/app/services/focus.service.ts
```diff
@@ -17,8 +17,18 @@ export class FocusService {
...
+
+          if (dataTestValue) {
+            const dataTestElement = this.document.querySelector(`[data-test="${dataTestValue}"]`);
+            (dataTestElement as HTMLElement)?.focus();
+          } else {
+            this.lastFocusedElement?.focus();
+          }
+
+          this.lastFocusedElement = null;
+        }, 200);
```
denysbutenko, line 31: Is it related somehow to animation duration?

denysbutenko (approved):
I noticed that focus was not restored when pressing `Cancel` in the dialog. Perhaps that is an edge case since the `Show` button isn't present when the dialog is opened. See screen recording:

https://github.com/truenas/webui/assets/351613/d28aae17-5cd8-43f4-b7fb-0703315a8b5e

Works great in other cases. Good to go 👍🏻

src/app/services/focus.service.ts
AlexKarpov98: > Is it possible when the last focused element does not have a `data-test` attribute?

we have a linter for such elements, they should have data-test.

src/app/services/focus.service.ts
AlexKarpov98: > Is it related somehow to animation duration?

possibly, it's just timing when element will be available after table re-rendered.

AlexKarpov98:
> I noticed that focus was not restored when pressing `Cancel` in the dialog. Perhaps that is an edge case since the `Show` button isn't present when the dialog is opened. See screen recording:
Currently this works only with slide ins, so the trigger which opens slideIn gets focus back.

PR #9598, opened 2024-02-02
NAS-127170 / 24.10 / ix-table2 jumps when something is updated because of data hiding and spinner

AlexKarpov98:
Testing: 
See ticket

Result:

https://github.com/truenas/webui/assets/22980553/363a1f5a-4639-414e-bd4a-ff569adb86e2


undsoft (approved):
Looks good, would be good to add some sort of loading indication in the future when table is being reloaded after a change.

PR #9600, opened 2024-02-02
NAS-127105 / 24.10 / Update SysInfo Widget

denysbutenko:
Tested on different systems.

M60 HA is enabled:
image

M40 HA is disabled:
image

non-HA nightly instance:
image

For testing, try to find edge cases where it can be broken.


bvasilenko (approved):
Worked without issues for me

PR #9606, opened 2024-02-04
NAS-127201 / 24.10 / Removing element when component gets destroyed

RehanY147:
To test, go to the `Data Protection` dashboard and click the `Add` button for `Cloud Sync Tasks`. Leave the form open and let the web session expire so UI goes to the sign in page automatically. Once that happens, you should be able to interact with the sign in page and continue using the UI. On `master` that was not possible.


AlexKarpov98 (approved):
Slide In is closed after token expiration. 👍  Works as described.

RehanY147:
backport

PR #9608, opened 2024-02-04
Update fr.json

MDVAurelien:
Fix some translations that should not be translated in this context. It's better.


undsoft (approved):
Thanks.

PR #9626, opened 2024-02-06
NAS-127042 / 24.10 / Refactor SnapshotTaskListComponent to ix-tables2

AlexKarpov98:
Testing: See ticket (refactoring)

Screenshot 2024-02-08 at 12 52 37

### DIFF
--- a/src/app/interfaces/periodic-snapshot-task.interface.ts
+++ b/src/app/interfaces/periodic-snapshot-task.interface.ts
@@ -39,4 +39,6 @@ export interface PeriodicSnapshotTaskUi extends PeriodicSnapshotTask {
   when: string;
   frequency: string;
   next_run: string;
+  last_run: string;
+  legacy: boolean;
 }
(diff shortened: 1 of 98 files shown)


bvasilenko (changes requested):
Looks like something is not right with the formatting of **Next Run** within expandable section. Same for **Last Run**.

image

Also, after choosing **Last Run** column to be shown, the formatting of column value is also missing

image

AlexKarpov98:
@bvasilenko - we have an issue with **relativeDateColumn** - it only works for the table cell, not inside of the details panel - I currently replaced it with textColumn, so that value transformation will also work inside of the details as well.

I updated `Last Run` as well:

Screenshot 2024-02-08 at 13 06 07

AlexKarpov98:
@bvasilenko - would be good to find a way to still use `relativeDateColumn`, it adds nice tooltip.

Screenshot 2024-02-08 at 13 07 37

Or we should use relativeDateColumn for card only, and for list page - we can use textColumn with correct text

AlexKarpov98:
@bvasilenko - as well a new bug (ticket) was created https://ixsystems.atlassian.net/browse/NAS-127279

bvasilenko (approved):
Looks good to me

PR #9630, opened 2024-02-06
NAS-127022 / 24.10 / Fix `Identify Drive` visible

bvasilenko:
**Testing**

On **Enclosure** page, when connected to the box specified in the ticket,

The visibility of **Identify Drive** button should depend on the `has_slot_status` property of the response from **enclosure.query** endpoint. `false` - hidden, `true` - visible.


undsoft (approved):
While this can go in `master`.The fix for Dragonfish has to be as described in the comment, because middleware is not returning correct values for  `has_slot_status` and fixing it would be too complicated now.

PR #9635, opened 2024-02-07
NAS-126853 / 24.10 / `fillForm` now refreshes available controls after setting each value

RehanY147:
Code review should be enough. This new method allows parts of the form to fill partially so conditional inputs can show up before we try to fill them without having to call the original method `fillForm` repeatedly.


undsoft:
Why not just fix `fillForm` to trigger a detection cycle after each control?

RehanY147:
@undsoft The order of the inputs is important. So, if `checkbox1` enables `select1` and the fillForm method is called like this `...fillForm({ select1: 'value', checkbox1: true });`, we don't know in what order to set the values. We'd have to change from `Record<string, unknown>` to maybe `[label: string, value: unknown][]` as method param to ensure array sorting to maintain in the upcoming loops of keys.

undsoft:
I believe order of keys in objects is now guaranteed https://stackoverflow.com/questions/30076219/does-es6-introduce-a-well-defined-order-of-enumeration-for-object-properties

bvasilenko (approved):
Looks good to me!

src/app/modules/ix-forms/components/ix-list/ix-list-item/ix-list-item.harness.ts
```diff
@@ -27,7 +27,14 @@ export class IxListItemHarness extends ComponentHarness {
   }
 
   async fillForm(values: Record<string, unknown>): Promise<void> {
-    const controls = await this.getControlHarnessesDict();
-    return fillControlValues(controls, values);
+    const labels = Object.keys(values);
+    for (const label of labels) {
+      const controlsDict = await this.getControlHarnessesDict();
+      const control = controlsDict[label] as IxFormControlHarness;
```
undsoft, line 33: Maybe introduce a `fillControl` helper?

undsoft (approved):
Looks good, thanks.

src/app/modules/ix-forms/components/ix-list/ix-list-item/ix-list-item.harness.ts
RehanY147: `getControlHarnessesDict` is a situation specific method. Creating a helper method would require passing the method in as a param to be called. We can create class member methods but that's not reusable anywhere else so I think this is good enough.

PR #9640, opened 2024-02-07
NAS-127108 / 24.10 / Add narrower checks for sharing roles

undsoft:
Testing: test services page. You may need to create a privilege with READONLY + role to test. Or just mock.
If your new cannot login, you may need to mock `processLoginResult` in `AuthService`.

### DIFF
--- a/src/app/directives/common/requires-roles/requires-roles.directive.ts
+++ b/src/app/directives/common/requires-roles/requires-roles.directive.ts
@@ -2,6 +2,7 @@ import {
   ComponentRef, Directive, HostBinding, Input, TemplateRef, ViewContainerRef,
 } from '@angular/core';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
+import { isEqual } from 'lodash';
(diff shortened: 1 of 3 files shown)


undsoft:
@bvasilenko

src/app/directives/common/requires-roles/requires-roles.directive.ts
```diff
@@ -2,6 +2,7 @@ import {
   ComponentRef, Directive, HostBinding, Input, TemplateRef, ViewContainerRef,
 } from '@angular/core';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
+import { isEqual } from 'lodash';
```
bvasilenko, line 5: Maybe it worth using `isMatch` here

bvasilenko (approved):
Works as expected!

src/app/directives/common/requires-roles/requires-roles.directive.ts
undsoft: Would't `isMatch` perform a partial comparison or something like it?

undsoft:
backport

PR #9643, opened 2024-02-08
NAS-127110 / 24.10 / ix-table-columns-selector columns -> reset to default hides all columns/rows values & Select/Unselect all works with glitches

AlexKarpov98:
Testing: (see ticket & videos inside the ticket - for the bugs I've found)
See results 👇 

https://github.com/truenas/webui/assets/22980553/a8512a23-5d4c-4c5c-ad0d-88791e88517b

### DIFF
--- a/src/app/modules/ix-table2/components/ix-table-columns-selector/ix-table-columns-selector.component.ts
+++ b/src/app/modules/ix-table2/components/ix-table-columns-selector/ix-table-columns-selector.component.ts
@@ -20,8 +20,12 @@ export class IxTableColumnsSelectorComponent<T = unknown> implements OnChanges {
   hiddenColumns = new SelectionModel<Column<T, ColumnComponent<T>>>(true, []);
   private defaultColumns: Column<T, ColumnComponent<T>>[];
 
-  get isAllChecked(): boolean {
+  get isOnlyOneColumnSelected(): boolean {
     return this.columns.filter((column) => !column.hidden && !!column.title).length === 1;
+  }
+
+  get isAllSelected(): boolean {
+    return !this.columns.filter((column) => column.hidden && !!column.title).length;
   }
 
   constructor(private cdr: ChangeDetectorRef) {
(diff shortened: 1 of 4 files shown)


src/app/modules/ix-table2/components/ix-table-columns-selector/ix-table-columns-selector.component.ts
```diff
@@ -63,8 +68,13 @@ export class IxTableColumnsSelectorComponent<T = unknown> implements OnChanges {
 
   private setInitialState(): void {
     this.columns = _.cloneDeep(this.defaultColumns);
-    this.hiddenColumns.clear();
-    this.hiddenColumns.select(...this.columns.filter((column) => column.hidden));
+    this.hiddenColumns.select(...this.columns);
+
+    this.defaultColumns
+      .filter((column) => !column.hidden)
+      .map((column) => this.defaultColumns.findIndex((column2) => column.title === column2.title))
```
undsoft, line 75: Can we rewrite this in some other way, this is a bit of hard to follow.

undsoft (changes requested):
So, what is `unselect all` supposed to do? Show empty table or show first column? On `Replication Tasks` 3 random columns that are not on the list.

Also would be good to update tests so that they alert us in case of future problems.

src/app/modules/ix-table2/components/ix-table-columns-selector/ix-table-columns-selector.component.ts
AlexKarpov98: > Can we rewrite this in some other way, this is a bit of hard to follow.

sure, but only `findIndex` works, that's weird

AlexKarpov98:
> So, what is `unselect all` supposed to do? Show empty table or show first column? On `Replication Tasks` 3 random columns that are not on the list.
> 
> Also would be good to update tests so that they alert us in case of future problems.

I think `unselect all` should give us at least first column, because if we show nothing - that's not looking good and not reasonable I think.
For me - how it works now - it's `good` especially comparing to the previous glitchy version of it.

AlexKarpov98:
@undsoft  - I think we can rewrite logic for `Select all` & `Unselect all` 
Currently we have `Unselect all` when some items are selected, I think we shall see `Select all` unless we selected all items - then we can show `Unselect all` 

What do you think?

undsoft (approved):
Nice, selector looks good.

PR #9645, opened 2024-02-08
NAS-127275 / 24.10 / Update Development requirements section of README.md along with formatting

nekomeowww:
Currently the `README.md` documentation stated `truenas/webui` depends on

- yarn >= 1.12
- Node.js >= 14

However, according to the actual building `Dockerfile`, `truenas/webui` uses `node@16`, `yarn@1.22`.

And with `yarn@1.12`, the command execution to `yarn install` will never success since the existing `.yarnrc` isn't compatible to `yarn@1.12` where it complains about

```
error An unexpected error occurred: "Invalid value type 1:16 in <path/to/yarnrc>"
```

This issue suggests updating the `README.md` to align the `node@16`, `yarn@1.22` with Dockerfile.


bvasilenko (approved):
Looks good to me!

PR #9646, opened 2024-02-08
NAS-127276 / 24.10 / guard user.privilege access when processing login results to prevent undefined property read

nekomeowww:
An error dialog with `Cannot read properties of undefined (reading 'webui_access')` prompt when login successful but without privilege field:

image

The correct implementation should respond users with `User is lacking permissions to access WebUI.` when have no permissions or failed to authenticate:

image

## Notes

1. I am not quite sure why the field `privilege` is `undefined` when the authentication process has been done. Did I missed something about environment configurations?
2. I have also checked the unit test spec files that located at `src/app/services/auth/auth.service.spec.ts`, and found that there are no current possibilities to add unit tests for such case since the `processLoginResult` method is private, how do you folks usually do with it when it comes to black box testing?

### DIFF
--- a/src/app/services/auth/auth.service.ts
+++ b/src/app/services/auth/auth.service.ts
@@ -205,7 +205,7 @@ export class AuthService {
         // Check if user has access to webui.
         return this.getLoggedInUserInformation().pipe(
           switchMap((user) => {
-            if (!user.privilege.webui_access) {
+            if (!user.privilege || !user.privilege.webui_access) {
               this.isLoggedIn$.next(false);
               return of(LoginResult.NoAccess);
             }


RehanY147:
@nekomeowww The described issue is seen on Cobia (23.10 and so on). This PR is to merge into master which is electric eel (24.10 and so on). Please wait for the Jira ticket to pass through the team processes so it can be decided what to do about this issue.

RehanY147 (commented):
^

undsoft (approved):
Thanks, the original issue won't happen on Dragonfish, but it won't hurt to have the check.

PR #9647, opened 2024-02-08
Update fr.po

MDVAurelien:
Fix and add some French translations


AlexKarpov98 (changes requested):
@MDVAurelien - please edit `src/assets/i18n/fr.json` instead.

AlexKarpov98 (approved):
Works for 13.1

PR #9650, opened 2024-02-08
NAS-127022 / 24.04-RC.1 / Fix `Identify Drive` visible

bvasilenko:
**Summary**

Implemented as per request https://github.com/truenas/webui/pull/9630#pullrequestreview-1866069657 , which references https://ixsystems.atlassian.net/browse/NAS-127022?focusedCommentId=241103

**Testing**

Go to **Enclosure** page, when connected to the box specified in the ticket

Check visibility of **Identity Drive** button , it should not be visible on this box

### DIFF
--- a/src/app/pages/system/view-enclosure/components/enclosure-disks/enclosure-disks.component.ts
+++ b/src/app/pages/system/view-enclosure/components/enclosure-disks/enclosure-disks.component.ts
@@ -254,7 +254,7 @@ export class EnclosureDisksComponent implements AfterContentInit, OnDestroy {
 
   get hideIdentifyDrive(): boolean {
     const selectedEnclosureView = this.selectedEnclosureView;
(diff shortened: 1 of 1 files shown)


RehanY147 (changes requested):
While the code looks Ok, are we sure these are the only models that should be excluded? Caleb said there are other platforms that can't be lit up. Need clarification on what models we want excluded.

bvasilenko:
@RehanY147 Caleb replies [here](https://ixsystems.atlassian.net/browse/NAS-127022?focusedCommentId=241103) that the drive identification button should be hidden (disabled) on anything else besides R-series, in other words the R-series is the only models where it should be displayed, and I've double checked that the code is written accordingly. would you please confirm?

bvasilenko:
replaced with https://github.com/truenas/webui/pull/9674

PR #9652, opened 2024-02-09
NAS-127273 / 24.10 / Data Protection → Next Run shows value for disabled tasks miss-match between card and list

AlexKarpov98:
Testing: see ticket.
Main idea is to make SAME behaviour across `Enabled & Next Run` - to show `Disabled` for Next Run if task is NOT Enabled.

As well in this ticket I fixed some issues with LIST components in Data Protection & Cron, to show same value for table's column and expandable details.

More fixes on listing & expandable rows issue will be done here: https://ixsystems.atlassian.net/browse/NAS-127279


undsoft (approved):
Nice catch 👍🏻

PR #9654, opened 2024-02-09
NAS-127055: Read only user : Truecommand cloud services > Setting > SAVE enabled

AlexKarpov98:
Testing: see ticket


RehanY147 (changes requested):
Are we sure only `FULL_ADMIN`s should be allowed to perform this action?

AlexKarpov98:
> Are we sure only `FULL_ADMIN`s should be allowed to perform this action?

@RehanY147 - what do you think we else need?

RehanY147 (approved):
I was just questioning if maybe other roles/privileges should also be allowed to perform this action. But I'm not sure who else should be. Maybe talk to middleware to see if any other roles can make this middleware call other than `FULL_ADMIN`

AlexKarpov98:
@RehanY147 - I talked about it with @yocalebo, MW gonna provide new Roles for us.

**Caleb St. John**
We didn't add any type of roles to the truecommand plugin it seems. So this is probably something we'll have to do on our side.

AlexKarpov98:
@RehanY147 There are `TRUECOMMAND_READ` and `TRUECOMMAND_WRITE` added.

PR #9656, opened 2024-02-09
NAS-127289 / 24.10 / Enable linter rules for ChangeDetection.OnPush

undsoft:
Test some of the affected components for regressions.


bvasilenko (changes requested):
Here's a few issues I've noticed:
1. Go to **System Settings > Enclosure** on m60-100.dc1.ixsystems.net. **Actual:** After loading is complete, enclosure page is blank 
2. Go to **System Settings > Advanced > Global Two Factor Authorization > Configure > Enable 2-factor authorization globally**. (**Note:** Scan QR code not to loose access.) Create another admin user. Login with this newly created user. **Expected:** full-screen dialog "Two-Factor Authentication Setup Warning!" with "Continue" button. **Actual**: regular login to dashboard, without any dialog

undsoft:
1. The first item is due to missing `truenas.is_ix_hardware` on that machine.
2. Could not reproduce, I also don't see any code that would require `cdr.markForCheck` or anything similar around full screen dialog.

bvasilenko:
1. thanks for clarification, this was a false alarm. Enclosure works fine.
2. to me it's not reproducible any more.

bvasilenko (approved):
no regressions found

PR #9657, opened 2024-02-09
NAS-127291 / 24.10 / Disallow warnings in tests

undsoft:
This installs `jest-fail-on-console`, which will not allow tests to be passed if they emit warnings or use `console.error`.
It also clears tests from currently warnings that we have.

Code review is enough.


denysbutenko (approved):
Looks good

PR #9669, opened 2024-02-12
NAS-126964 / 24.10 / Improve JobProgressDialog

undsoft:
You can test this in Disks -> Wipe.

Note that full disk wipe is likely to not complete and report status properly (middleware issue).
Also if you wipe the same disk multiple times quickly this may lead to an error.


bvasilenko (approved):
Looks good to me

PR #9674, opened 2024-02-13
NAS-127022 / 24.04-RC.1 / Fix `Identify Drive` visible for Dragonfish

bvasilenko:
**Summary**

Implemented as per request https://github.com/truenas/webui/pull/9630#pullrequestreview-1866069657 , which references https://ixsystems.atlassian.net/browse/NAS-127022?focusedCommentId=241103

**Testing**

Go to **Enclosure** page, when connected to the box specified in the ticket

Check visibility of **Identity Drive** button , it should not be visible on this box


RehanY147:
I can't test on the mentioned system. I think it doesn't have the privileges API in the middleware. Need another system to test live.

undsoft (changes requested):
I think the condition should include MINI-R and all models of R series. See `createEnclosure` for list of models.

PR #9683, opened 2024-02-15
NAS-127370 / 24.10 / Refactor some usages of EntityJobDialog

undsoft:
Testing: regression testing for some of the affected dialogs.


bvasilenko (approved):
Looks good, no regression found

PR #9684, opened 2024-02-15
NAS-127177 / 24.10 / System Audit > Event Data, "Copy JSON" option is not in JSON format

AlexKarpov98:
Testing: see ticket.
Problem was with the `copyViaDeprecatedExecCommand` - it always received plain text instead of desired value.

https://github.com/truenas/webui/assets/22980553/3c5bab73-fedf-4ebd-809d-1fa14c7bf64b


denysbutenko (approved):
Looks good

PR #9688, opened 2024-02-15
NAS-126998 / 24.10 / Refactor `Advanced Settings` to use `IxChainedSlideInService`

RehanY147:
This PR does the following things,
1. Refactor fetching information that is already available and can instead be passed on by the parent component that's calling the form component in
2. Remove coupling between different cards where one form submission on one of the cards can trigger card refresh on multiple cards
3. Move all forms to `IxChainedSlideInService`.
4. Do some other clean up
For testing, check for regressions. Also, test the issue reported in the ticket.


bvasilenko (commented):
@RehanY147 Looks good overall, however it's a bit tedious to type `@Inject(CHAINED_COMPONENT_REF)` for every new editable card.

I'd suggested to encapsulate token-based dependency 

`@Inject(CHAINED_COMPONENT_REF) private chainedRef: ChainedComponentRef,`

 into a typed dependency

`private chainedRef: SomeNewType,`

I think something similar is done at Angular Material dialogs.

PR #9693, opened 2024-02-16
NAS-127049 / 24.10 / Read only user > SMB shares table > Delete - confirmationbox Unshare is not disabled

AlexKarpov98:
Testing: see ticket


bvasilenko (approved):
Works as expected

PR #9703, opened 2024-02-19
Update ar.json

AlexKarpov98:


AlexKarpov98:
now this job is failing if some invalid translation was added:
Screenshot 2024-02-19 at 12 00 19

PR #9704, opened 2024-02-19
NAS-121641 / 24.10 / Fix container image update dialog

denysbutenko:
Fix condition and updated tests.


AlexKarpov98 (approved):
Image modal disappears as expected.

PR #9705, opened 2024-02-19
NAS-127047: Read only user: SMB share list view, "enabled" checkbox enabled.

AlexKarpov98:
Dragonfish only. On master it works as expected already (it uses ix-table2)

Results:
Screenshot 2024-02-19 at 12 45 24
Screenshot 2024-02-19 at 12 45 21


bvasilenko (approved):
Works as expected

PR #9708, opened 2024-02-19
NAS-127417 / 24.04-RC.1 / Lock icon for restricted elements

undsoft:
Strike-through on restricted elements was replaced with a lock icon as per feedback we received.

Testing: check if lock icon is not looking nice somewhere.


denysbutenko (approved):
image
image
image
image
image
image

Sometimes, multiple tooltips appear at the same time. Looks good.

PR #9710, opened 2024-02-19
NAS-127425 / 24.10 / Extracting entity-empty to its own module

undsoft:
Test for regressions around entity-empty component.


undsoft:
Thanks, should be fixed.

bvasilenko (approved):
Looks good, no regressions found

PR #9727, opened 2024-02-22
NAS-127419 / 24.10 / Disable restricted forms earlier

AlexKarpov98:
Testing: see ticket & do a global application overview. (a lot of files were changed)


undsoft (approved):
Looks good, nice work.
I've noticed some issues, most of which are not related to this PR, but it would be good to fix them now:

- Storage -> Upgrade button on a Pool is not disabled.
- Datasets -> Roles card -> Create SMB and Create NFS share links are not disabled.
- Apps -> App -> Install button is not full width
- Enclosure UI -> Edit Label is not disabled

in VMs:
- is special permission required to download logs in VMs?
- feels like Edit button should be allowed.
- in VMs -> Devices -> NIC -> Edit, Generate button is misaligned.

AlexKarpov98:
@undsoft 
is special permission required to download logs in VMs?
-- seems - yes
https://github.com/truenas/webui/assets/22980553/d0c5c6a2-e730-43e2-a993-39e4fef700d7

src/app/modules/dialog/components/start-service-dialog/start-service-dialog.component.ts
```diff
@@ -27,6 +28,8 @@ export interface StartServiceDialogResult {
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class StartServiceDialogComponent implements OnInit {
+  readonly requiredRoles = [Role.ServiceWrite];
```
RehanY147, line 31: This check feels redundant. The ticket requires that this check is performed earlier. Meaning the user shouldn't even get to this dialog if the user doesn't have the required role.

src/app/modules/ix-forms/components/ix-explorer/create-dataset-dialog/create-dataset-dialog.component.ts
```diff
@@ -22,6 +23,8 @@ import { WebSocketService } from 'app/services/ws.service';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class CreateDatasetDialogComponent implements OnInit {
+  readonly requiredRoles = [Role.DatasetWrite];
```
RehanY147, line 26: Redundant check. Same as above.

RehanY147 (approved):
Some dialogs have redundant role checks. We shouldn't get to the dialogs in the first place according to the ticket. We should ensure that at least. That all dialogs have a check first that doesn't allow button clicks that bring those dialogs up.

PR #9731, opened 2024-02-23
NAS-127507 / 24.10 / Shell UI under System Settings is missing for restricted …

undsoft:
…users

Change: Shell menu option is now disabled and not hidden for readonly users.


bvasilenko (approved):
Works as expected!

PR #9735, opened 2024-02-26
Update nl.json

renediepenbroek:
Please put my nl.json in your main folder as soon as possible!!! I have had to update nl.json so many times, merge it, delete sentences, go through the entire file again, etc. That takes a lot of (unnecessary) time and that can be partly avoided if my nl.json is in your hands as quickly as possible. main folder. Thanks for that!!! Rene


undsoft (approved):

PR #9736, opened 2024-02-26
NAS-127515 / 24.10 / Add codename to system info widget

denysbutenko:
Changes

- Replace `TrueNAS-SCALE` with release codename (when available)
- Fix regression on standby widget when HA is disabled
- Added tooltip to see full version (no need to copy anymore)

image


undsoft (approved):
Looks good.

PR #9737, opened 2024-02-26
NAS-127505 / 24.04-RC.1 / UI submitting duplicate issue reports

undsoft:
Previously, oauth subscriptions were incorrectly cleared, which lead to possible duplicate ticket submittions.
Also, filters out websocket close events from Sentry.

Testing:
1. Update `addTicket` in `FeedbackService` to not create actual tickets:
```
   return of(ticket).pipe(
      map(() => {
        console.log('creating ticket');
        return { } as NewTicketResponse;
      }),
    );
```

2. Open feedback form on an non-enterprise system and fill in a bug report.
3. Click Login To Jira.
4. Close the window.
5. Click Login to Jira again and submit the ticket.

You should only see one `create ticket` message in console.


AlexKarpov98 (approved):
Works as described. Nice 👍

PR #9744, opened 2024-02-27
NAS-127400 / 24.10 / Cloud Sync Tasks: State does not refresh after stopping

AlexKarpov98:
Testing: see ticket & try with CloudSync Tasks


undsoft (approved):
Altering state of data provider is something we should refactor in the future, but the scope of this ticket is complete.

PR #9754, opened 2024-02-28
NAS-127471: Allow subnets to be entered in Allowed Addresses

AlexKarpov98:
Testing: see ticket & try:
Screenshot 2024-02-28 at 11 27 37


src/app/pages/system/advanced/allowed-addresses/allowed-addresses-form/allowed-addresses-form.component.html
```diff
@@ -11,6 +11,7 @@
         [formArray]="form.get('addresses')"
```
undsoft, line 11: Let's also update the label here to make it more clear.

undsoft (approved):
Looks good, thanks.

src/app/pages/system/advanced/allowed-addresses/allowed-addresses-form/allowed-addresses-form.component.html
AlexKarpov98: @undsoft - what do you mean? 
Which label should I use as an updated value? 😄

undsoft: I mean that it would clearer for the user if instead of `IP Address *` we said `IP Address/Subnet *`

PR #9755, opened 2024-02-28
NAS-127371 / 24.10 / Refactor Apps module to use new jobDialog

denysbutenko:
Changes are scoped to the Apps section only. Check affected components for regressions.


bvasilenko (approved):
Works good to me! No regressions found.

PR #9757, opened 2024-02-29
NAS-127582 / 24.10 / Performing advanced query in webui audit form appears to submit query payload twice

AlexKarpov98:
Testing: 
See that load() is not called multiple times (upon initial load request)

inside of the `load(): void {` 

you can alert `alert(this.method);` to make sure it's only called once.
Before it was called twice because of the `setSorting` and `setPagination` logic.


denysbutenko (approved):
I've checked in websocket messages. Looks good.

PR #9758, opened 2024-03-01
NAS-127596 / 24.10 / Investigate how to implement a UI elements search provider

AlexKarpov98:
Testing: 
Some of the code is added to see it in action and will be removed if we merge it.

`So we have a json file and a provider will take data from there`

I used the first idea: 
`Mark UI elements with some directives and write a parser to create a tree of elements at compile time.`

So in the end we have json file `ui-searchable-elements.json`
example item:
```
{
    "hierarchy": [
      "System Settings",
      "General",
      "Localization",
      "Language"
    ],
    "synonyms": [
      "Translate App"
    ],
    "requiredRoles": [
      "FULL_ADMIN"
    ],
    "routerLink": [],
    "anchorRouterLink": [
      "/system/general"
    ],
    "anchor": "language-input",
    "triggerAnchor": "localization-settings"
  }
  ```
  
  We can add `| translate` strings for hierarchy or synonyms and then use `translate.instant` to get translation - when we gonna render results inside of the search box hints.
  _______
  
  See 3 examples in `dashboard.component.ts`
  
  Example 1:
  Focusing the desired element (If user searches for some action or card maybe)
  

https://github.com/truenas/webui/assets/22980553/32506952-1c50-40f3-8ff7-7c09a0d5a87e

Example 2: 
Navigating to the URL (If user searches for Pool Creation for example [has routerLink])

https://github.com/truenas/webui/assets/22980553/0b4342b2-28b0-4ec5-97ba-c8a793515c02

Example 3: (we can have anchor - one we will focus and triggerAnchor - one we can focus and click)
Navigating to the anchor page, clicking trigger anchor and focus on anchor (used for forms)

https://github.com/truenas/webui/assets/22980553/4bbcb952-05bb-44c0-805c-05fb4da8bc34

### CODE REVIEW

**Comment by AlexKarpov98** [comment]
@undsoft Provider will be a service which will interact with `json` file

**Review by undsoft — COMMENTED** [review]
Looks pretty good, nice work 👍🏻 

1. Maybe we can create a ts file next to html and keep most of text stuff there and then import it in a component similarly to how `helptext` works.
2. Would be good not to have to specify `id`s on elements and for example generate it based on combination of attributes.

**Comment by AlexKarpov98** [comment]
> Looks pretty good, nice work 👍🏻
> 
> 1. Maybe we can create a ts file next to html and keep most of text stuff there and then import it in a component similarly to how `helptext` works.
> 2. Would be good not to have to specify `id`s on elements and for example generate it based on combination of attributes.

Problem is that parser is converting everything to string and it's not possible to read data outside of the html or .ts file. 
We can define config in .ts file for example, but it all should be accessible from .ts file, not exported somewhere.

**Comment by AlexKarpov98** [comment]
@undsoft  maybe we can use one prop instead of many?

```
uiSearchConfig =   {
    "hierarchy": [
      "Storage",
      "Import Pool"
    ],
    "synonyms": [
      "Storage"
    ],
    "requiredRoles": [
      "ACCOUNT_WRITE",
      "APPS_WRITE"
    ],
    "anchorRouterLink": [
      "/storage"
    ],
    "anchor": "import-pool-button",
    "triggerAnchor": ""
  },
  ```
  
  and use it `[uiSearchConfig]="importPoolSearchConfig"`

undsoft:
What if we enforce some naming convention on this ts file that can be imported in the component?
For example if the file is always `pools-dashboard.elements.ts` and it always exports `elements` const, you'll know to check for existence of the file and you'll be able to match keys to the html file.

AlexKarpov98:
@undsoft - you mean only text values to be extracted there or configs for search elements?
Storing configs sounds better for me, and will result in less .html `[`props]`

undsoft:
I mean that we could have `user-form.elements.ts`:
```
export const elements = {
  firstName: {
    hierarchy: [],
    synonyms: [], 
    etc,
  }
};
```

Developers would then import it in component file and use in `user-form.component.html` template:

```
<ix-input
  [ixSearch]="searchElements.firstName"
```

AlexKarpov98:
@undsoft 
Code is updated now.
Screenshot 2024-03-04 at 12 14 25

elements config:
Screenshot 2024-03-04 at 12 15 00

In some places we'll need to add `requiredRoles` manually, if there is no `ixRequiresRoles` directive applied to the element.

End object in config is:
```
{
    "hierarchy": [
      "System Settings",
      "General",
      "Localization",
      "Language"
    ],
    "synonyms": [
      "Translate App"
    ],
    "requiredRoles": [
      "FULL_ADMIN"
    ],
    "routerLink": null,
    "anchorRouterLink": [
      "/system",
      "general"
    ],
    "anchor": "language-input",
    "triggerAnchor": "localization-settings"
  }
  ```
  
  ID prop is automatically attached from `anchor`.
  `triggerAnchor` - should be set manually to the desired element, so we can click it and then focus on the `anchor`

undsoft (approved):
Let's add some documentation to the script file explaining what's going on.
Other than that looks pretty good.

PR #9759, opened 2024-03-01
NAS-127438 / 24.10 / Refactor DockerImagesListComponent to use ix-table2

denysbutenko:
Check Manage Container Images


AlexKarpov98 (approved):
It works for me. 👍 
Screenshot 2024-03-01 at 17 47 07

PR #9774, opened 2024-03-05
NAS-127364 / 24.10 / Update to Angular 17

undsoft:
Testing: regression testing, especially for styles.

Known changes:
- Font in main menu.
- Checkboxes and selects now all use the same font.
- Don't know why TreeNodePaddingDirective was needed, but it worked better without it.


bvasilenko (approved):
Looks good! No regressions found

PR #9778, opened 2024-03-06
NAS-127594: Search component for global search

AlexKarpov98:
Testing: test UI search component.


undsoft (changes requested):
Looks good, but some improvements are needed.
1. Not a big fan of creating a custom backdrop. It doesn't look similar to the backdrop we have and we are reimplementing some of the functionality, like closing overlay on navigation.  
There's also an issue with focus isolation. Right now I can tab underneath elements behind a backdrop. 
I would try to use overlay or dialog. Perhaps we can cheat and render _another_ search input once overlay has been opened. Or render original search input separately with css `isolation` property and always keep it over the overlay.
If those approaches fail, we should try to use native material backdrop and implement cdkFocusTrap.
2. Would be good to also allow to use arrows up and down to select a search result.
2. If I tab on a search result and type, I think those keystrokes should be registered in the input.
2. As discussed on Slack, last portion of the search result should show the item I was looking for. So if I'm search for "Advanced", I should just get "System Settings -> Advanced" as the first result. I should see any of the results nested in Advanced.
3. Need test case for various search scenarios.
4. Let's bring it closer to designs: input is too narrow vertically, results should have less margin on top and more padding, check fonts, make results box wider, remove border and update how we highlight items.
5. When I search for `dd` I get 'Сховище → Імпортувати пул' and I'm not really sure why.

AlexKarpov98:
@undsoft - updated:

https://github.com/truenas/webui/assets/22980553/382814ba-e193-4d94-900d-3a86e315e5cd

🤩

undsoft (approved):
Yes, very nice.

I don't know if it's worth showing all item on focus. Maybe we'll add recent searches in the future 🤔 

Also, I think we should only search by the beginning of each word, e.g. "board" shouldn't find "dashboard".

PR #9782, opened 2024-03-07
NAS-127721 / 24.10 / Switch to esbuild

undsoft:
69s => 37s build time on my machine

Testing: check for regressions.


src/assets/styles/mixins/loader.scss
```diff
@@ -3,7 +3,7 @@
 @mixin background-logo-loader($zoom: 1) {
   &::after {
     animation: loading 2s linear infinite;
-    background: transparent url('~assets/customicons/truenas_scale_logo_full.svg') no-repeat center center;
+    background: transparent url('/src/assets/customicons/truenas_scale_logo_full.svg') no-repeat center center;
```
denysbutenko, line 6: There is regression on main dashboard and reporting page where loading state does not show truenas logo.

denysbutenko (commented):
Support card on `/system/general` is missing m40 image

denysbutenko (approved):
Other than that looks good.

PR #9783, opened 2024-03-07
NAS-123838 / 24.04.0 / Add support for legacy model ES24N (additions)

bvasilenko:
**Testing**

1. Connect to machine specified in the ticket's commentaries
2. Go to **System Settings > Enclosure** page, and select **ES24N** from the right sidebar 
3. Verify visibility of **Identify Drive** button for model of enclosure (`ES24N`). For this enclosure, on this particular model, button should be hidden, since it is not supported in Dragonfish.

4. Verify the enclosure image displayed on **Dashboard** page at **System Information** widget. **Expected image:**  
image


undsoft (approved):
Looks okay to the extent I was able to test it (no disks in expansion shelf).

Also, I'm assuming image would need to be ported to Electric Eel.

PR #9784, opened 2024-03-07
NAS-127694 / 13.3 / Removes s3 service

undsoft:
Code review is enough.


AlexKarpov98 (approved):
Bye Bye `s3` service 👋

PR #9787, opened 2024-03-08
NAS-127615 / 24.10 / Do not re-request count during shifts in pagination (do request recount if someone shifts to last offset)

AlexKarpov98:
Test that `count` request is only set: 
1. When we have initial request.
2. If we entered last page

You can test using console.log in affected `if()` check.
See ticket.


undsoft (approved):
Looks good, everything appears to be working correctly.

PR #9788, opened 2024-03-08
NAS-127670 / 24.10 / Save debug is disabled but debug can still be downloaded

AlexKarpov98:
`Save Dubug` is allowed operation.
Code review is enough.

Screenshot 2024-03-08 at 13 06 54


bvasilenko (approved):
LGTM!

PR #9793, opened 2024-03-11
NAS-127719 / 24.04.0 / Do not offer service dialogs if user does not have access

undsoft:
Testing: see ticket. Start service dialog will no longer be shown is user won't be able to do anything in it.


undsoft:
Thanks, good catch.

PR #9799, opened 2024-03-12
NAS-127794 / 24.04.0 / Dashboard - Backup Tasks - Infinite loading on m40

AlexKarpov98:
Testing: see ticket
I as well asked @yocalebo about the error.
Result - no more infinite `spinner`:
Screenshot 2024-03-12 at 12 31 44

### DIFF
--- a/src/app/pages/dashboard/components/widget-backup/widget-backup.component.scss
+++ b/src/app/pages/dashboard/components/widget-backup/widget-backup.component.scss
@@ -5,6 +5,10 @@
   width: 100%;
 }
 
+.widget mat-toolbar-row {
+  justify-content: flex-start;
+}
+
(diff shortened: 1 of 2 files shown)


undsoft (changes requested):
Wouldn't it be better to show an error message if any of the tasks were not loaded.
This could theoretically result in user seeing 0 tasks when having connectivity issues – an unexpected result.

AlexKarpov98:
> Wouldn't it be better to show an error message if any of the tasks were not loaded. This could theoretically result in user seeing 0 tasks when having connectivity issues – an unexpected result.

then it would be good to show "Block" but with some error text instead, I mean not use modal, because it's an error in the card.

AlexKarpov98:
This issue was fixed on the middleware side.

PR #9805, opened 2024-03-12
NAS-127736 / 24.04.0 / Fix failed and success states on backup tasks widget

bvasilenko:
**Testing**

Follow replication steps described in the ticket

### DIFF
--- a/src/app/pages/dashboard/components/widget-backup/widget-backup.component.ts
+++ b/src/app/pages/dashboard/components/widget-backup/widget-backup.component.ts
@@ -58,6 +58,9 @@ export class WidgetBackupComponent extends WidgetComponent implements OnInit {
   backups: BackupRow[] = [];
   isLoading = false;
 
+  successStates = [JobState.Success, JobState.Finished];
+  failedStates = [JobState.Failed, JobState.Error, JobState.Aborted];
+
   readonly ScreenType = ScreenType;
 
   trackByTile: TrackByFunction<BackupTile> = (_, tile) => tile.title;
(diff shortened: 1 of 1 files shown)


src/app/pages/dashboard/components/widget-backup/widget-backup.component.ts
```diff
@@ -58,6 +58,9 @@ export class WidgetBackupComponent extends WidgetComponent implements OnInit {
   backups: BackupRow[] = [];
   isLoading = false;
 
+  successStates = [JobState.Success, JobState.Finished];
+  failedStates = [JobState.Failed, JobState.Error, JobState.Aborted];
```
RehanY147, line 62: Instead of having two arrays, where we can accidentally add the same state twice, we can just create one array and check the absence of state in that array to determine other states.
```
  get failedCount(): number {
    return this.backups.filter((backup) => !this.successStates.includes(backup.state)).length;
  }

  private getTile(title: string, tasks: BackupRow[]): BackupTile {
    const successfulTasks = tasks.filter((backup) => this.successStates.includes(backup.state));
  ...
```

RehanY147 (changes requested):

src/app/pages/dashboard/components/widget-backup/widget-backup.component.ts
bvasilenko: I created two arrays because the absence of "success" in the array does not necessarily mean that the task is "unsuccessful," and vice versa. There are `JobState` statuses that are not related to either success or failure. 

image

Am I missing something?

RehanY147 (approved):

PR #9808, opened 2024-03-13
NAS-127720 / 24.10 / Webui does not handle validation errors properly when creating SMB shares via "Add Dataset"

AlexKarpov98:
Testing:

See ticket, try to create dataset.
You should not be able to submit with invalid name or empty name now.

Screenshot 2024-03-13 at 10 11 49


undsoft (changes requested):
This is a workaround that breaks scope of responsibilities of nested component and add an extra special case. 
We should fix this properly and have child components communicate their validation status up to the parent.

undsoft (approved):
Code looks okay, but I think you forgot to add a validator for SMB share name. I can tick `Create SMB Share` and remove share name without Save button getting disabled.

See `SmbValidationService`.

PR #9812, opened 2024-03-13
NAS-126795 / 24.04.0 / Updated package to handle more timezones

RehanY147:
Updated package can parse more timezones returned from middleware without causing errors.


AlexKarpov98 (commented):
@RehanY147 - can you show some places where it caused errors please?

bvasilenko (approved):
Tested on m60-100a. 
bug with timezone is fixed

PR #9816, opened 2024-03-13
NAS-127836 / 24.10 / Add "Netdata" button to reporting section

william-gr:
This button will make a WS call to generate a password and make a request to /netdata/ for the browser to cache the credentials. Once that is done a new tab will be opened with netdata.

### DIFF
--- a/src/app/pages/reports-dashboard/reports.service.ts
+++ b/src/app/pages/reports-dashboard/reports.service.ts
@@ -1,6 +1,6 @@
 import { Inject, Injectable } from '@angular/core';
 import {
-  map, Observable, shareReplay, BehaviorSubject, Subject,
+  filter, map, Observable, shareReplay, BehaviorSubject, Subject,
 } from 'rxjs';
 import { ReportingGraphName } from 'app/enums/reporting.enum';
 import { WINDOW } from 'app/helpers/window.helper';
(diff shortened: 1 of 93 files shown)


src/app/pages/reports-dashboard/reports.service.ts
```diff
@@ -162,4 +167,18 @@ export class ReportsService {
...
     );
   }
+
+  openNetdata(password: string): void {
+    // Thats the only way I could manage to have the browser save the password
+    // to re-use when the new tab is opened.
+    // HttpClient seems to do it through HttpHeaders which won't cache in browser.
+    const xhr = new XMLHttpRequest();
+    this.loggedInUser$.subscribe((user) => {
+      xhr.open('GET', '/netdata/', true, user.pw_name, password);
```
undsoft, line 177: It looks like doing this just puts username and password in the url: 
`http://root:password@localhost:4200/netdata/`.
If we want to do this (security concerns?), maybe you can form the URL yourself and use HttpClient (it will be easier in tests).

undsoft (changes requested):
Thanks, looks okay.
CI won't let you through without a test case for this. You can see how other component tests are implemented. You'd generally try to locate the button via MatButtonHarness, press it and check that a mocked HttpClient or something else called the correct url.

undsoft (approved):

PR #9824, opened 2024-03-15
NAS-125411 / 24.10 / Added check for vdev capacties difference

RehanY147:
Mixed Vdev Capacities warning should only be shown if the difference is greater than 2 Gib and the percentage difference between the highest and lowest vdev capacities is more than 5%.

### DIFF
--- a/src/app/services/storage.service.ts
+++ b/src/app/services/storage.service.ts
@@ -1,7 +1,6 @@
 import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
-import { GiB } from 'app/constants/bytes.constant';
 import { TopologyItemType, TopologyWarning, VdevType } from 'app/enums/v-dev-type.enum';
 import { FileSystemStat } from 'app/interfaces/filesystem-stat.interface';
 import { Option } from 'app/interfaces/option.interface';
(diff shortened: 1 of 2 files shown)


src/app/services/storage.service.ts
```diff
@@ -105,8 +105,14 @@ export class StorageService {
 
   // Check to see if every VDEV has the same capacity. Best practices dictate every vdev should be uniform
   isMixedVdevCapacity(allVdevCapacities: Set<number>): boolean {
+    const max = Math.max(...allVdevCapacities);
+    const min = Math.min(...allVdevCapacities);
+    const diffPercentage = 100 - ((min / max) * 100);
     const diff = Math.max(...allVdevCapacities) - Math.min(...allVdevCapacities);
-    return allVdevCapacities.size > 1 && diff >= GiB * 2;
+
+    const isDiffNoticable = diffPercentage >= 5 && diff >= GiB * 2;
```
amotin, line 113: Since the two factors (swap and different number of metaslabs) may coexist, I would lean towards `2GB + 5%` rather than `2GB and 5%` here.  But area when it would make the difference is already pretty shady (very small disks), so I don't insist.

src/app/services/storage.service.ts
```diff
@@ -105,8 +105,14 @@ export class StorageService {
...
+
+    return allVdevCapacities.size > 1 && isDiffNoticable;
```
amotin, line 115: Mostly out of curiosity: why do we need to check the `size > 1` here?  If as I understand it is a number of disks, then for one disk isDiffNoticable should always be false already.  If there are not disks, then I wonder what min/max functions return, or whether they throw an exception.

amotin: Yea, that is what I meant. If there is only one value, it can't be much different from itself. ;)

amotin (approved):

PR #9825, opened 2024-03-15
NAS-127824 / 24.10 / Added `'/s'` to the graph tooltip and legend for Network widget on the dashboard

RehanY147:


undsoft (changes requested):
See my comment in the original ticket, there is also Disks IO in Reporting to fix.

PR #9826, opened 2024-03-15
NAS-127811 / 24.10 / Request:  Update help text for "Application name" of Custom App to reference name restriction of lower case letters

AlexKarpov98:
Testing: 
Check updated validation rules and text. Text is taken from the middleware response.


RehanY147 (changes requested):
Let's show only the first validator that is failing out of the three conditions at a time. Also, let's add a tooltip that explains all of the rules together.

AlexKarpov98:
> Let's show only the first validator that is failing out of the three conditions at a time. Also, let's add a tooltip that explains all of the rules together.

@RehanY147  - we have a pattern validator. 
I talked to ticket creator, we decided to change text to smth that we already have:
image

RehanY147 (approved):

PR #9828, opened 2024-03-15
NAS-127866 / 24.10 / Remove license extraction from webui

undsoft:
Confirmed that this is no longer necessary.


AlexKarpov98 (approved):
👌

PR #9839, opened 2024-03-18
NAS-127439 / 24.10 / Refactor dataset-quotas-grouplist.component List to ix-table2

AlexKarpov98:
Now we have One Component for Dataset Quotas, similarly to the Add & Edit form.

You can test both `user-quotas` and `group-quotas` pages.


bvasilenko (approved):
Works as expected!

PR #9840, opened 2024-03-18
NAS-127778 / 24.04.0 / Added check to open Advanced Options on error

RehanY147:
This is a quick solution to opening the Advanced Options if there's a validation field for one of the fields hidden in Basic Options mode. Proper solution will be provided by https://ixsystems.atlassian.net/browse/NAS-127892


denysbutenko (approved):
image

Works well

PR #9843, opened 2024-03-18
NAS-127022: Revert drive identification condition

undsoft:
Reverts the condition because it turned out to be an incorrect solution.


AlexKarpov98 (approved):
👍

william-gr:
backport

PR #9850, opened 2024-03-20
NAS-127796 / 24.04.0 / On Dragonfish nightly and RC.1 the reordering of the dashboard changes when going somewhere else and going to the dashboard.

AlexKarpov98:
Testing: See ticket.


denysbutenko (approved):
Works well

PR #9855, opened 2024-03-20
NAS-126760 / 13.3 / Disable autocomplete for user and password inputs on disk edit

RehanY147:


AlexKarpov98:
@RehanY147 - does it fix the original issue?

AlexKarpov98 (commented):
I am not sure it's fixing the original issue.
I tried without your changes - it works fine for me (same as with your changes)

How did you reproduce the original issue?

As well I get compilation error:

Screenshot 2024-03-20 at 18 11 45

RehanY147:
This issue seems to be unrelated to what the UI is doing. Will creating another PR

PR #9864, opened 2024-03-21
NAS-127920 / 24.10 / Fix tree node padding on Devices & Boot Status pages

denysbutenko:
To test, open the Devices and Boot Status pages. Ensure tree node takes full width.

Expected no left offset outside of tree node.
image

Actual:
image


AlexKarpov98 (approved):
Works as described.

PR #9869, opened 2024-03-22
NAS-127745 / 24.10 / Accessibility improvements for `Cron Scheduler`

RehanY147:
Check ticket for testing


src/app/modules/scheduler/components/scheduler-modal/scheduler-preview-column/scheduler-preview-column.component.html
```diff
@@ -1,7 +1,13 @@
 <section class="preview-header">
   <h4 class="crontab-header">
     {{ 'Schedule Preview' | translate }}
-    <button class="close-btn" mat-icon-button ixTest="close-scheduler" [matDialogClose]>
+    <button
+      class="close-btn"
+      mat-icon-button
+      aria-label="Close scheduler"
```
denysbutenko, line 7: aria-labels need to be translatable

src/app/modules/scheduler/components/scheduler/scheduler.component.html
```diff
@@ -13,33 +13,40 @@
...
+      </mat-option>
+      <mat-option
+        class="schedule-option"
+        [value]="customValue"
+        [ixTest]="[controlDirective.name, 'presets', customValue]"
+        (onSelectionChange)="onOptionSelectionChange($event)"
+      >
+        {{ 'Create' | translate }}
+        <span class="preset-description">
+          {{ 'Create custom schedule' | translate }}
```
denysbutenko, line 49: It looks weird having two `Create` words. I think we should keep only the first one.
image

denysbutenko (changes requested):
Works well using keyboard navigation. Just a few minor things

src/app/modules/scheduler/components/scheduler-modal/scheduler-preview-column/scheduler-preview-column.component.html
```diff
@@ -1,9 +1,16 @@
 <section class="preview-header">
   <h4 class="crontab-header">
     {{ 'Schedule Preview' | translate }}
-    <button class="close-btn" mat-icon-button ixTest="close-scheduler" [matDialogClose]>
-      <ix-icon name="cancel"></ix-icon>
-    </button>
+    <div class="close-wrapper" [attr.aria-label]="'Close scheduler' | translate">
```
denysbutenko, line 4: Other buttons has aria-label without a wrapper. Why is it needed here?

denysbutenko (approved):
Looks good

src/app/modules/scheduler/components/scheduler-modal/scheduler-preview-column/scheduler-preview-column.component.html
RehanY147: Don't know why but it doesn't work on the button or icon itself. Only works on them if not used as an input property. But we need to use it as an input property to provide translation logic. So it works there if I use it like `aria-label="Close scheduler"` but not as `[attr.aria-label]="'Close scheduler' | translate"`

PR #9872, opened 2024-03-22
NAS-127798 / 24.04.0 / Unable to save "Enable S.M.A.R.T." settings in Storage > Edit Disk page

AlexKarpov98:
Only exists on dragonfish, I assume it was somehow accidentally backported to dragonfish. 
(on Dragonfish table details row - we show true/false for all of the tables instead of Yes/No as on Electric Eel)


bvasilenko (approved):
Ok, works for me

PR #9877, opened 2024-03-24
NAS-127923 / 24.04.0 / Enclosure graphics/info are not loading for m60s

AlexKarpov98:
Testing: see ticket


denysbutenko (approved):
tested on m60 - it works well

PR #9881, opened 2024-03-25
NAS-127461: Refactor services.component List to ix-table2

AlexKarpov98:
Testing: test `Services` page.

Screenshot 2024-03-25 at 14 05 53


src/app/modules/ix-table2/components/ix-table-body/cells/ix-cell-actions/icon-action-config.interface.ts
```diff
@@ -4,9 +4,10 @@ import { Role } from 'app/enums/role.enum';
...
   iconName: string;
   tooltip?: string;
+  requiredRoles?: Role[];
   onClick: (row: T) => void;
-  dynamicTooltip?: (row: T) => Observable<string>;
   hidden?: (row: T) => Observable<boolean>;
   disabled?: (row: T) => Observable<boolean>;
-  requiredRoles?: Role[];
+  dynamicTooltip?: (row: T) => Observable<string>;
+  dynamicRequiredRoles?: (row: T) => Observable<Role[]>;
```
denysbutenko, line 12: I'm curious about maintainability with two inputs for a single parameter. I see two tooltips already there, but I wouldn't say I like this approach, which keeps growing. Maybe we could check if it is observable and pass it forward, and if it is not, make it observable on the fly to keep a single input. What do you think?

denysbutenko (approved):
Generally, it works well. No regressions are seen.

src/app/modules/ix-table2/components/ix-table-body/cells/ix-cell-actions/icon-action-config.interface.ts
AlexKarpov98: let's consider it for the future ticket and leave it as it is here.

src/app/pages/services/services.component.ts
```diff
@@ -35,209 +40,274 @@ import { serviceChanged } from 'app/store/services/services.actions';
...
-        return services.map((service) => {
-          return {
-            ...service,
-            name: serviceNames.has(service.service) ? serviceNames.get(service.service) : service.service,
-          } as ServiceRow;
-        });
-      }),
+      map((services) => services.map((service) => ({
+        ...service,
+        name: serviceNames.has(service.service) ? serviceNames.get(service.service) : service.service,
```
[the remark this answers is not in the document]
AlexKarpov98: @denysbutenko - I checked it, we need that `name` prop.
To search eligibly by visible name at least.

PR #9883, opened 2024-03-25
NAS-127892 / 24.10 / show validation errors for hidden controls

bvasilenko:
**Summary**

Adding `IxFormService`, which holds a collection of only those form fields, which are visible for a user.

That said, adding and removing control should is implemented using the `AfterViewInit` and `OnDestroy` methods of each `ix-` form component, such as `ix-checkbox`, `ix-input`, etc.

**Testing**

1. Go to **Shares > Windows SMB shares > Add SMB**
2. Enter **Name** and **Path** so that **Save** button become active
3. Click **Advanced** and check **Legacy AFP** checkbox, and click **Save.** The error will be displayed next to **Legacy AFP** checkbox, and **Save** button will become disabled. This behavior matches the expectations.
4. Click **Basic** , then go back to **Advanced**, and then click **Basic** again. This will make **Save** button clickable. This behavior also matches the expectations.
5. Now click the **Save** button. After a call to API, it will become become disabled, without displaying any error message.

   **Expected result:** error message is presented to the user, and it mentions to the field causing the **Save** button to be disabled


src/app/modules/ix-forms/services/form-error-handler.service.ts
```diff
@@ -53,8 +55,9 @@ export class FormErrorHandlerService {
       const errorMessage = extraItem[1];
 
       let control = this.getFormField(formGroup, field, fieldsMap);
+      const controlNames = this.ixFormService.getControls().map((ctrl) => ctrl.name);
```
denysbutenko, line 58: Let's move this into a new method in the form service, e.g. `getControlsNames` or update an existing method to return it.

src/app/modules/ix-forms/services/form-error-handler.service.ts
```diff
@@ -4,13 +4,15 @@ import { ResponseErrorType } from 'app/enums/response-error-type.enum';
...
 import { DialogService } from 'app/modules/dialog/dialog.service';
+import { IxFormService } from 'app/modules/ix-forms/services/ix-form.service';
 import { ErrorHandlerService } from 'app/services/error-handler.service';
 
 @Injectable({ providedIn: 'root' })
 export class FormErrorHandlerService {
   constructor(
     private dialog: DialogService,
     private errorHandler: ErrorHandlerService,
+    private ixFormService: IxFormService,
```
denysbutenko, line 15: Usually, we don't use the `ix` prefix inside components if such name is not taken

denysbutenko (commented):
When `Advanced` fields are hidden:
image

When `Advanced` fields are visible:
image

Overall, it works, and it still needs to be clarified how to fix the error.

As a user, I expected more guidance on fixing errors. Ideally, the system would show advanced fields and smoothly scroll to the first error field. Perhaps we could incorporate highlight element directive from the global search component.

bvasilenko:
@denysbutenko all the requests should be addressed now

src/app/modules/ix-forms/services/form-error-handler.service.ts
```diff
@@ -46,38 +56,64 @@ export class FormErrorHandlerService {
...
       let control = this.getFormField(formGroup, field, fieldsMap);
+      let controlsNames = this.formService.getControlsNames();
 
-      if (!control) {
-        console.error(`Could not find control ${field}.`);
-        // Fallback to default modal error message.
-        this.dialog.error(this.errorHandler.parseError(error));
-        return;
-      }
+      const showValidationError: () => void = (): void => {
```
denysbutenko, line 70: Wondering if make sense to move out of the loop

src/app/modules/ix-forms/services/form-error-handler.service.ts
```diff
@@ -46,38 +56,64 @@ export class FormErrorHandlerService {
...
+          control.setErrors({
+            manualValidateError: true,
+            manualValidateErrorMsg: errorMessage,
+            ixManualValidateError: { message: errorMessage },
+          });
+          control.markAsTouched();
+
+          const element = this.formService.getElementByControlName(field);
+          if (element && !isOnErrorFocused) {
+            element.scrollIntoView();
```
denysbutenko, line 97: Let's try add options to make it smooth

denysbutenko (approved):
Somehow, it does not work on the first try - when click save and nothing happens due to an error, but I can't repeat it anymore. There is also a case when I saw both errors in dialog and under control, which I can't reproduce. 

image

Looks good to me, thanks

PR #9893, opened 2024-03-28
NAS-127966 / 24.10 / Doing actions colapses the Dataset Tree

AlexKarpov98:
Testing: see ticket. When we do some update - opened datasets shall not collapse.

Result: 

https://github.com/truenas/webui/assets/22980553/e927c232-510e-491d-9a7d-d465bfa28e39


bvasilenko (approved):
Works good

PR #9899, opened 2024-03-30
Update and fix German translations

Leseratte10:
All the tooltips that start with "Set to ..." in English all start with different translations in German, some of which just sound weird and some of which have an entirely different meaning.

Technical terms like "iSCSI extent", "Domain", "zvol", "Debug", "Scrub" or "Flag" are translated with multiple different words depending on who translated it.

I haven't added too many new translations, but I did correct quite a few existing translations. I'd rather see an english text here or there than a badly-translated German one.

PR #9900, opened 2024-03-30
Don't force-capitalize every word on button labels

Leseratte10:
While adding some German translations (#9899) I noticed that the labels on all the buttons in the WebUI have the first letter of every single word capitalized due to the use of the CSS filter `text-transform: capitalize`. 

This isn't really grammatically correct, and having every single word on a button capitalized looks weird, in my opinion. At least in German, maybe it's different in english. I'm not sure why the capitalization is even there in the first place (in the CSS) when it might as well be solved by capitalizing these words in the translation files. 

Looking at the english translation file, nearly every button label is already included in the translation file with all words capitalized

With this PR, the CSS for the buttons is changed to remove that auto-capitalization. For the english language this probably won't change anything since all/most of the button labels have every single word capitalized in the i18n file anyways, but it allows other languages to follow proper grammar rules with regards to case-sensitivity and have some words on a button label that aren't capitalized. 

Or am I missing something and there's a good reason why these buttons have CSS that auto-capitalizes every single word?

### DIFF
--- a/src/assets/styles/other/_material-reduction.scss
+++ b/src/assets/styles/other/_material-reduction.scss
@@ -247,7 +247,6 @@ ix-tooltip button {
 .mat-flat-button {
   border-radius: 0 !important;
   font-weight: 400 !important;
-  text-transform: capitalize !important;
 
   &:focus-visible {
     outline: 1.5px solid var(--alt-fg2) !important;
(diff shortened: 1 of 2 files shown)


denysbutenko (approved):
Sounds reasonable. Thanks!

Leseratte10:
Thanks for the approve. Looks like the auto-merge didn't work, though, because the codecov-step got stuck.

denysbutenko:
@undsoft take a look please

PR #9906, opened 2024-04-01
NAS-126760 / 13.3 / Replace 'SED Password' with 'SED Key'

RehanY147:
There is no concrete way to stop password manager extensions like Bitwarden from forcefully filling in a field that has any mention of a "password". Setting autocomplete off or other solutions from scale don't work. On scale this problem doesn't exist because we handle forms completely differently from core (slide-ins instead of full-page forms).

### DIFF
--- a/src/app/helptext/storage/disks/disks.ts
+++ b/src/app/helptext/storage/disks/disks.ts
@@ -83,9 +83,9 @@ export default {
   disk_form_informational_tooltip: T('Report if drive temperature is at or\
  above this temperature in Celsius. <i>0</i> disables the report.'),
 
-  disk_form_passwd_placeholder: T('SED Password'),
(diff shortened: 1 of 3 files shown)


AlexKarpov98 (commented):
I was not able to reproduce the issue. 
But it you say that problem can be with just `password` word, can you show it please?

I am fine with `SED Key` though.

@undsoft - please have a look at it as well.

undsoft (approved):
I'm okay with field being renamed.

RehanY147:
@AlexKarpov98 For reference, when using the 'Password' keyword in the label, the field is forced filled in by the password manager extension.

https://github.com/truenas/webui/assets/24204846/e8424f45-51b0-4dd4-842b-00cd9eee73c9

PR #9907, opened 2024-04-01
NAS-127442 / 24.10 / Refactor JobsList to use ix-table2

denysbutenko:
For testing check Jobs full page table at `/jobs`


bvasilenko (approved):
Looks good

PR #9909, opened 2024-04-02
NAS-127948 / 24.10 / Dragonfish: Restore Snapshot dialog rendering not consistent with other dialogs

AlexKarpov98:
Testing: see result 👇 

Screenshot 2024-04-01 at 19 39 08


undsoft (approved):
Dialog looks good.

PR #9916, opened 2024-04-02
NAS-127963 / 24.04.0 / Fix reporting units again

bvasilenko:
**Testing**

Check units at **Reporting > Network** and at **Dashboard > Network Widget**

**Expected result:**

* Network speeds - **bits** per second `b`, `Kb/s`, `Mb/s`, `Gb/s` etc.\
* Memory  - **bytes** `B`, `KiB`, `MiB`, `GiB`, etc.
* Disk - **bytes** per second `B/s`, `KiB/s`, `MiB/s`, `GiB/s`.


src/app/pages/reports-dashboard/components/line-chart/line-chart.component.ts
```diff
@@ -268,6 +269,14 @@ export class LineChartComponent implements AfterViewInit, OnDestroy, OnChanges {
 
   axisLabelFormatter = (numero: number): string => {
     if (this.report?.name === ReportingGraphName.NetworkInterface) {
+      if (numero < 1000) {
```
denysbutenko, line 272: `< Kb`?

denysbutenko (changes requested):
image

`/s` missing in the tooltip
`Sent/Received` order is swapped between `In/Out`. 
Maybe let's do something like `Received .../s / Sent .../s`

Other than that, it looks good to me

src/app/helpers/file-size.utils.ts
```diff
@@ -53,8 +53,8 @@ function normalizeFileSizeBase10(value: number, baseUnit: 'b' | 'B'): [formatted
   }
   formatted = Math.round((formatted + Number.EPSILON) * 100) / 100;
   switch (increment) {
-    case kb:
-      return [formatted, 'k' + baseUnit];
+    case Kb:
```
undsoft, line 56: This is incorrect.
https://en.wikipedia.org/wiki/Data-rate_units#:~:text=Kilobit%20per%20second%20(symbol%20kbit,1%2C000%20bits%20per%20second

src/app/pages/reports-dashboard/utils/report.utils.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  it('returns value for bits', () => {
+    expect(formatValue(500, 'bits')).toBe('500 B');
+  });
+
+  it('returns value for bytes', () => {
+    expect(formatValue(500, 'bytes')).toBe('500 B');
+  });
+});
```
undsoft, line 31: 👍🏼

undsoft (changes requested):
Lower case for network units was correct.
If we have time to fix spacing between "k" in decimals to make it consistent, it would good. I'd prefer 100k written together.
Снимок экрана 2024-04-03 в 11 22 48

undsoft (approved):
I think it's pretty good for now.
There are some minor issues that we can improve later.

PR #9917, opened 2024-04-02
NAS-128005 / 24.10 / Main dashboard architecture

undsoft:
Changes:
* Renames current dashboard to dashboard-old
* Adds new dashboard module with general architecture for new widgets.
* Adds two dummy widgets.

How to review:
Review description for the new main dashboard in comments here https://ixsystems.atlassian.net/browse/NAS-127217 and see if new proposed architecture will be able to support designs.


RehanY147 (changes requested):
It looks fine. I do feel like from the screenshots, each widget group needs a grid of columns and rows. And each widget will need to remember it's placement in the group grid. We should decide the grid size in this phase as well. Max rows x columns. 2 x 2 seems like the way to go but we should decide that and add it to the architecture.

undsoft:
See `WidgetGroup`. Each will have 4 slots with either a widget or null. For order of slots see `WidgetGroupLayout`.

RehanY147 (approved):

PR #9918, opened 2024-04-02
NAS-127951 / 24.10 / Fix apps network usage

denysbutenko:
image


bvasilenko (approved):
Looks good to me

undsoft (changes requested):
Doesn't look like it's really "all" according to last Rehan's comment.

denysbutenko:
> Doesn't look like it's really "all" according to last Rehan's comment.

@undsoft does it mean UI changes are not needed due to middleware changes because data becomes bytes per second rather than in/out summary?

denysbutenko:
Middleware has changes to provide data `bytes per second`, no need to change UI.

PR #9919, opened 2024-04-03
NAS-127369 / 24.10 / Using SVG template for enclosure template

RehanY147:


src/app/pages/system/system.routing.ts
```diff
@@ -48,7 +48,7 @@ export const routes: Routes = [
       children: [
         {
           path: '',
-          component: ViewEnclosureComponent,
+          component: ViewEnclosure2Component,
```
undsoft, line 51: Let's add a new menu item for new enclosure.

undsoft (approved):
Looks good so far, I'm assuming in the future we'll have separate components for each model and their slots.
Also, let's create a new module for this.

RehanY147:
Creating a new PR

PR #9920, opened 2024-04-03
NAS-128138 / 24.10 / Add router pages as searchable elements (global search)

AlexKarpov98:
Testing: see updated items in the Global Search.

https://github.com/truenas/webui/assets/22980553/05d6a745-133f-46fb-a824-87b7252bedd3


denysbutenko (approved):
image

Works well. There are two minor style imperfections - the heading row doesn't take full width. Some text is visible under the `Show Less` button.

PR #9925, opened 2024-04-05
NAS-128070 / 24.10 / Show recently clicked searches when opening global search

AlexKarpov98:
Currently we save recent searches to localstorage.
We keep unique elements there and update/resort them once user selected something.

I added new `global-search-sections.service.ts` where we keep all search sections and results.

https://github.com/truenas/webui/assets/22980553/654ca9fc-4c86-4ce2-a56a-3417d96e1b6d


RehanY147 (approved):
Nice!

PR #9927, opened 2024-04-05
NAS-128187 / 24.04.1 / FIPS reboot improvements

undsoft:
TBD


bvasilenko (approved):
@undsoft Works good overall, just a few notes:

1. Tests look like need fixing

2. A minor typo in the Reboot prompt

   image

3. Not sure is this a bug, but other node reboots even when user chooses not to reboot. Replication steps: when user enables FIPs, then in Reboot prompt user chooses "Cancel", reboot job still added to the queue, and other node is rebooted.

4. Minor adjustment. When user does not change the value of "FIPs" toggle, and clicks **Save** button is enabled, and the form is being submitted to the middleware. Ideally the **Save** button should be disabled until FIPs is actually changed.

PR #9928, opened 2024-04-07
NAS-128165 / 24.10 / New feature indicator

bvasilenko:
**Testing**

Try using the new `ixNewFeatureIndicator` directive

Acceptance criteria - see ticket


AlexKarpov98 (commented):
`[ixNewFeatureIndicator]="{key: 'add-group', message: 'Add Group'}"`
Screenshot 2024-04-08 at 12 23 35
1. For me `Add` button is gone after adding `ixNewFeatureIndicator` to it.
@bvasilenko - am I doing it correctly?

2. `This dot is only shown if user hasn’t seen the message yet.` & `Do not show this new indicator for users that have already seen it.`
Will that dot & indicator **reappear** once I reload the page?

src/app/directives/common/new-feature-indicator/new-feature-indicator-wrapper.component.html
```diff
@@ -0,0 +1,37 @@
...
+            {{ 'Help' | translate }}
+          </h5>
+        </div>
+      </div>
+      <div>
+        <ix-icon
+          class="close-icon"
+          name="mdi-close"
+          ixTest="tooltip-close"
+          (click)="popperTooltip.hide()"
```
denysbutenko, line 17: this is not good from accessibility point of view. let's wrap icon with material icon button.

src/app/directives/common/new-feature-indicator/new-feature-indicator.service.ts
```diff
@@ -0,0 +1,33 @@
...
+
+@Injectable()
+export class NewFeatureIndicatorService {
+  onShown = new EventEmitter<NewFeatureIndicator>();
+  private shownIndicatorKeys: string[] = [];
+
+  constructor(
+    private store$: Store<AppState>,
+  ) {
+    this.store$.pipe(select(selectPreferencesState)).subscribe((prefs) => {
```
denysbutenko, line 16: Since this is not a singleton class it requires the until destroyed pipe and decorator.

denysbutenko (changes requested):
Code looks okay. Some example would be useful for testing.

src/app/directives/common/new-feature-indicator/new-feature-indicator.directive.ts
```diff
@@ -6,6 +6,16 @@ import { NewFeatureIndicatorWrapperComponent } from 'app/directives/common/new-f
 import { NewFeatureIndicator } from 'app/directives/common/new-feature-indicator/new-feature-indicator.interface';
 import { NewFeatureIndicatorService } from 'app/directives/common/new-feature-indicator/new-feature-indicator.service';
 
+/**
+ * Usage: adding an indicator with a hint about a new feature.
+ *
+ * @example
```
denysbutenko, line 12: This is good to have but I meant an example to use new directive somewhere in the project, so it can be tested via docker image. Otherwise it's okay if no plan to use it now.

PR #9929, opened 2024-04-08
NAS-128045 / 24.10 / Apps Install UI Dropdown field show "x" instead of dropdown indicator

AlexKarpov98:
Testing, see:
Screenshot 2024-04-08 at 13 43 23

instead of 
Screenshot 2024-04-08 at 13 44 28


RehanY147 (changes requested):
This input is not meant to be a dropdown. It's an autocomplete input. It is not supposed to show dropdown options like a select or a combobox. If it acts so much like a combobox, maybe we should just use a combobox in its place.
@undsoft Opinion?

RehanY147:
Google search example
image

RehanY147:
@AlexKarpov98 I think we should resist this change. We want the input for combobox or autocomplete input to feel like an input. We want it to behave like a searchable field. You are supposed to have to type something into it for options to show up. It doesn't make sense even in a combobox to not show the `x` icon. Besides, just clicking inside the input already shows all of the available options. So, it's not like the down icon is needed. I think we should just switch to combobox maybe and leave the icon as is. So, it can be used like a scrollable, searchable select.

AlexKarpov98:
@undsoft - let us now the final decision please 👍

undsoft:
Makes sense to use combobox instead of adding combobox functionality into `ix-input`.

AlexKarpov98:
@undsoft - what about the dropdown icon instead of "x" ?
This ticket is about it 🤔

undsoft:
Would ix-select work here?

AlexKarpov98:
> Would ix-select work here?
@undsoft 
no, it's required to have a searchable component, like `ix-combobox`

undsoft:
What happens if you click the (x) on this field? Is this field required? Should it be?
I don't know if I want to globally replace (x) icon with a dropdown icon in all comboboxes right now.

AlexKarpov98:
@undsoft it will cleanup the field, if it's required- you'll see required error message, so no problem at all for me.

RehanY147:
@ksimm1 There's a misunderstanding I think. With that example, what I meant to show was the behavior for the autocomplete-input that we have as it is meant to be used.

-Intended uses
input with `autocomplete="true"` - An input where you can type your value in but it will also show you complete values as you are typing so you don't have to type the whole thing. You can either type into the input field to search for a complete value or to enter a new value. This is what we currently use. The Google search is an example of this input type. If we indicate in anyway that it is a dropdown (with icons for example), then it becomes too similar to a combobox.

combobox - A drop-down where you cannot type in a new value. You have to select from one of the options already made available by the system. You can search the options by typing into the input field. But only the options already available are accepted as a value for said field.

My opinion regarding this issue is that we should be using a combobox instead of an autocomplete-input here as that seems more appropriate and how the input field is meant to be handled. But the PR has updated since my comment so consider the original comment outdated.

RehanY147 (changes requested):
@undsoft  @AlexKarpov98 My point here is that `<ix-input autocomplete="true">` is exactly the same to `<ix-combobox>` visually. And the problem that the user described is still there if we switch to a `<ix-combobox>`. The ticket is about replacing `x` with a down arrow. That remains to be addressed.

RehanY147:
To be clear, I'm not saying that we should do that. I'm saying that that's what the ticket is about. So, while this PR addresses the issue of the wrong input field being used, it doesn't address the original gripe that the user had.

undsoft:
@AlexKarpov98 Can you figure out why we are not using a simple dropdown (instead of a combobox) for those fields?

AlexKarpov98:
> @AlexKarpov98 Can you figure out why we are not using a simple dropdown (instead of a combobox) for those fields?
@RehanY147 
@undsoft we use it because there may be a lot of elements on the ENUM array, this is the original ticket where it was required https://ixsystems.atlassian.net/browse/NAS-123257
The only problem is that I updated `ix-input` instead of using `ix-combobox` 

But the problem with `x` still remains though. 
I guess that's NOT a problem at all, we can add input prop to ix-combobox to show input with dropdown icon instead of x.

`@Input() showWithDropdown` for example.

undsoft:
No, we shouldn't add obscure inputs to our core components.

Here's what I think we can do:
1. If the field is empty and we are going to show select with all options on focus, show arrow icon.
2. If the field has full value and we are going to show select will all the options on focus, show arrow icon.
3. If user started typing and list of suggestions shows partial matches, show (x) icon.

AlexKarpov98:
@RehanY147 @ksimm1 - Result 👇 

https://github.com/truenas/webui/assets/22980553/9125591f-8ebf-4975-aa1b-be72fefc8486

src/app/modules/ix-forms/components/ix-combobox/ix-combobox.component.ts
```diff
@@ -247,6 +247,10 @@ export class IxComboboxComponent implements ControlValueAccessor, OnInit {
     return this.inputElementRef?.nativeElement?.value && this.inputElementRef.nativeElement.value.length > 0;
   }
 
+  hasExactValueSelected(value: string): boolean {
```
RehanY147, line 250: Can be renamed to `isValueFromOptions`

RehanY147 (approved):
Looks good to me.

PR #9934, opened 2024-04-08
NAS-128173 / 24.10 / Add websocket callAndSubscribe method

denysbutenko:
For testing, use team m40. Open `Storage` or `Dashboard (old)` page and try to import/export `NAS-128173` pool from another tab. Ensure changes are reflected on the first tab.


src/app/services/ws.service.ts
```diff
@@ -112,21 +133,22 @@ export class WebSocketService {
...
   }
 
   clearSubscriptions(): void {
     this.clearSubscriptions$.next();
-    this.eventSubscriptions.clear();
+    this.subscriptions.clear();
   }
 
-  private callMethod<M extends ApiCallMethod>(method: M, params?: ApiCallParams<M>): Observable<ApiCallResponse<M>>;
+  // eslint-disable-next-line max-len
```
undsoft, line 150: Can't you break call method into multiple lines instead?

src/app/services/ws.service.ts
```diff
@@ -164,11 +186,7 @@ export class WebSocketService {
   }
 
   private subscribeToJobUpdates(jobId: number): Observable<Job> {
-    return this.subscribe('core.get_jobs').pipe(
-      filter((apiEvent) => apiEvent.id === jobId),
-      map((apiEvent) => apiEvent.fields),
-      takeUntil(this.clearSubscriptions$),
-    );
+    return this.store$.select(selectJob(jobId)).pipe(takeUntil(this.clearSubscriptions$));
```
undsoft, line 189: This doesn't look right. WebsocketService is at lower level conceptually, it shouldn't know anything about Store.

src/app/services/ws.service.spec.ts
```diff
@@ -85,6 +92,22 @@ describe('WebSocketService', () => {
     });
   });
 
+  describe('callAndSubscribe', () => {
+    it('should call and subscribe to updates', async () => {
+      const observablePools$ = of(pools);
+      jest.spyOn(service, 'callAndSubscribe').mockImplementationOnce((method: string) => {
```
undsoft, line 98: This doesn't really test `callAndSubscribe`, because you've overridden it are just checking that your mock was called correctly.

undsoft (changes requested):
So I'm a bit confused about how would `callAndSubscribe` would work?
What happens when an update event is received?
I think I should receive a new updated collection, so if I `callAndSubscribe` to `interface.query` and one of the interfaces is removed, I should receive a new array of interfaces minus the one that was removed.

src/app/services/ws.service.ts
denysbutenko: Yeah, agreed. This is unrelated change.

denysbutenko:
> So I'm a bit confused about how would `callAndSubscribe` would work? What happens when an update event is received? I think I should receive a new updated collection, so if I `callAndSubscribe` to `interface.query` and one of the interfaces is removed, I should receive a new array of interfaces minus the one that was removed.

A bit stuck with type inheritance over functionality. The method was updated. Tests are improved.

undsoft (changes requested):
The approach looks correct. I would extract all that code to adjust the collection in some helper method.
Please request review again once the PR is ready.

RehanY147 (changes requested):
^ What Evgeny said

src/app/helpers/operators/handle-api-event.operator.ts
```diff
@@ -0,0 +1,23 @@
+import { OperatorFunction, map } from 'rxjs';
+import { IncomingApiMessageType } from 'app/enums/api-message-type.enum';
+import { ApiCallAndSubscribeMethod, ApiCallAndSubscribeResponse } from 'app/interfaces/api/api-call-and-subscribe-directory.interface';
+import { ApiCallResponse } from 'app/interfaces/api/api-call-directory.interface';
+import { ApiEventTyped } from 'app/interfaces/api-message.interface';
+
+export function handleApiEvent<
```
undsoft, line 7: How about `applyApiEvent`?

undsoft (approved):
Looks good, nice work.

src/app/helpers/operators/handle-api-event.operator.ts
denysbutenko: Sounds good

src/app/pages/storage/stores/pools-dashboard-store.service.ts
```diff
@@ -107,11 +106,7 @@ export class PoolsDashboardStore extends ComponentStore<PoolsDashboardState> {
   }
 
   getPools(): Observable<Pool[]> {
-    return this.ws.call('pool.query', [[], { extra: { is_upgraded: true } }]);
-  }
-
-  listenForPoolUpdates(): Observable<ApiEvent<Pool>> {
-    return this.ws.subscribe('pool.query');
+    return this.ws.callAndSubscribe('pool.query', [[], { extra: { is_upgraded: true } }]);
```
AlexKarpov98, line 109: Should we handle subscriptions `untilDestroyed` or smth?
For cases when I leave `Pools Dashboard` page and then return back?

denysbutenko: Good point. I assumed that it managed by store effect and there is general protection in web socket to unsubscribe when session is recreated. I'll double check it.

denysbutenko: The `PoolsDashboardStore` instance is part of the `StorageModule` and remains active until the tab is closed. The primary function, `loadDashboard()`, is called from various locations, including when navigating to the Pools Dashboard. As such, it calls `pool.query` and other endpoints and then restores the "subscribe" observable (API event updates) from subscriptions in the websocket service. In this case, we do not need to unsubscribe.

It is unclear whether the `loadDashboard()` method unsubscribes from previous subscriptions or if the `PoolsDashboardStore` architecture has any memory leaks. These are excellent questions that require additional investigation, but unfortunately, they are beyond the scope of this ticket. Feel free to log a ticket for further investigation.

AlexKarpov98: @undsoft - please check these comments as well

undsoft: You need to unsubscribe where you subscribe (as in rxjs subscribe). This happens in the consuming component.
As long as `callAndSubscribe` stops listening to middleware events when last subscriber unsubscribes, this is okay.

RehanY147: Maybe we should replaced the "subscribe" keyword for these method names with "getEvents" or "subscribeToEvents" or something more declarative so we can avoid confusions like this in the future. Can't say this makes a very strong case for renaming though. Just a thought.

RehanY147 (approved):

src/app/pages/storage/stores/pools-dashboard-store.service.ts
denysbutenko: This name directly states what the method does, which is first to make a `call` and then to `subscribe` to updates. In context of websocket service those are known terms. I'd prefer to keep it short and clear.

PR #9935, opened 2024-04-09
NAS-127551 / 24.10 / Added error catch methodology for forms

RehanY147:
Check the "Dataset Form" and "GUI Form". Both have different methods of causing failure. The pages should show fallback html.

### DIFF
--- a/src/app/pages/datasets/components/dataset-form/dataset-form.component.html
+++ b/src/app/pages/datasets/components/dataset-form/dataset-form.component.html
@@ -23,7 +23,7 @@
         [advancedMode]="isAdvancedMode"
         (formValidityChange)="isEncryptionValid = $event"
       ></ix-encryption-section>
-      {{ test.test }}
+      {{ test.test.test }}
       <ix-other-options-section
         [existing]="existingDataset"
         [parent]="parentDataset"
--- a/src/app/pages/datasets/components/dataset-form/dataset-form.component.html
+++ b/src/app/pages/datasets/components/dataset-form/dataset-form.component.html
@@ -4,7 +4,7 @@
   [loading]="isLoading"
 ></ix-modal-header>
 
-<mat-card>
+<mat-card *ngIf="!guiService.hasError">
   <mat-card-content>
     <form class="ix-form-container" [formGroup]="form" (submit)="onSubmit()">
       <ix-name-and-options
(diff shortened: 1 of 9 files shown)


src/app/services/ix-graceful-updater.service.ts
```diff
@@ -0,0 +1,6 @@
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class IxGracefulUpdaterService {
```
undsoft, line 4: Seems like we can live without this extra service.

undsoft (changes requested):
Looks promising, some things to consider:
1. We should still call global error handler to report this error to Sentry.
2. Some errors may be expected, e.g. backend errors. We may need to try to differentiate whether this was a rendering error or something else.
3. How would it work if we have multiple elements that can throw an error, e.g. 3 cards on the same page?
4. It would be ideal to still implement the error boundary component/directive, so that it's easier to reuse.

RehanY147:
@undsoft 
> We should still call global error handler to report this error to Sentry.
We are still using the global ErrorHandlerService. We're just creating a new instance on a component level to localize the context of the error state. So, it is still reporting to Sentry.

> How would it work if we have multiple elements that can throw an error, e.g. 3 cards on the same page?
The way it works is by creating a new instance of the ErrorHandlerService for each component and a new instance of the "Updater" service. So, each component will only handle its own errors. So, you would just need to break it down into smaller components in this scenario, and give each component it's own error handler.

src/app/pages/datasets/components/dataset-form/dataset-form.component.html
```diff
@@ -4,7 +4,7 @@
   [loading]="isLoading"
 ></ix-modal-header>
 
-<mat-card>
+<mat-card *ngIf="!guiService.hasError">
```
denysbutenko, line 7: It would be nice to keep templates clean and handle error cases using directive.

denysbutenko (changes requested):
image
image

Looks like working solution. However, I'd like to minimize the code changes as much as possible to add support for new error handler for every component (e.g. avoid error template definition). Perhaps it worth to consider retry mechanism if error occurs.

undsoft:
> We are still using the global ErrorHandlerService. We're just creating a new instance on a component level to localize the context of the error state. So, it is still reporting to Sentry.

Are you sure it doesn't override global handler for the component?

RehanY147:
This approach is also flawed and unreliable. Closing PR for now.

PR #9941, opened 2024-04-10
NAS-128289 / 24.10 / Replace some usages of EntityJobComponent

undsoft:


RehanY147 (approved):
Tested pool creation and certificate deletion. Works ok.

PR #9945, opened 2024-04-12
NAS-128285 / 24.10 / Rename ix-table2 to ix-table & fix usage

AlexKarpov98:
Testing: see that tests passing & code review is enough.


bvasilenko (approved):
Looks good to me!

PR #9947, opened 2024-04-15
NAS-128270: UI elements parser improvements

AlexKarpov98:
Testing: see ticket


AlexKarpov98:
> Code looks good. Is there anything on the UI side that I need to test for regressions? There are a lot of changes.

@RehanY147 
nope, on the UI side you should only see pretty much the same `src/assets/ui-searchable-elements.json`.
So the only changes are about improving parser, so check the end result the parser gives to us. 😄

RehanY147 (approved):

PR #9962, opened 2024-04-17
NAS-128379 / 24.10 / Update Storj logo

mobyvb:
This is the new Storj logo as of this week. I am not sure how to check how it looks in the web UI, so to be safe, I preserved the exact dimensions of the old png.
However, I can provide a logo with alternate dimensions or as an SVG if that is desirable.


RehanY147 (approved):
Thank you

AlexKarpov98 (approved):
👍

PR #9969, opened 2024-04-18
NAS-128344 / 24.10 / Calculation for estimated storage capacity is incorrect in pool manager and dashboard when capacity is > 999TiB

bvasilenko:
**Testing**

Setup mocks to emulate total pool capacity > 999TiB

**Expected result:**

At Create Pool wizard, At the review stage of the wizard,

Display correct units of measurement (PiB) for **Est. Usable Raw Capacity** field at **Topology Summary** block


AlexKarpov98 (approved):
Nice, it works 👍 
Screenshot 2024-04-19 at 10 34 28

Tests failed, but if you pull latest master changes - they should pass! ✅ 

BTW - would be good to add tests to `file-size.utils.ts` as well. ✔️

PR #9972, opened 2024-04-19
NAS-128397 / 24.04.0 / Restore change password form for limited users

undsoft:
Not sure why changing user password was limited to a full admin, but this is incorrect.


AlexKarpov98 (approved):
Change Password is now available to any user.

PR #9981, opened 2024-04-22
NAS-128167 / 24.10 / Widget group in the editor

undsoft:
Testing: press edit on a widget group on a main dashboard.


AlexKarpov98 (approved):
Looks good to me. Nice job.


small improvement 👇 
I think we should make in form layout elements focusable.
Currently once I open the form I can only focus on `x` icon

undsoft:
Good idea.

PR #9986, opened 2024-04-23
NAS-128168 / 24.10 / Implement datapoint widget

AlexKarpov98:
Testing: see ticket & results.
Text will be dynamically resized based on the input, we can set maxFontSize and minFontSize.
As well it can be dynamically resized on window size change, can be tested by switching to mobile view.

With some longer texts:
NAS-128168

What is currently set up:
Screenshot 2024-04-23 at 11 14 04


bvasilenko (approved):
Text sizing worked good for me

PR #9987, opened 2024-04-23
Ensure multiline App "Section help" can be rendered

StevenMcElligott:
Currently we cannot write multi-line section-help for the App install/edit GUI.
As those get removed, leading to one big blob of text.

This fix, ensures that multiline descriptions get respected accordingly.

Fixes the following issue:

image

To render as:

image

Thanks for your attention!

### DIFF
--- a/src/app/modules/ix-dynamic-form/components/ix-dynamic-wizard/ix-dynamic-wizard.component.html
+++ b/src/app/modules/ix-dynamic-form/components/ix-dynamic-wizard/ix-dynamic-wizard.component.html
@@ -20,7 +20,9 @@
       </div>
       <div class="help" [hidden]="!section.help">
         <div class="title">{{ 'Section Help' | translate }}</div>
-        <div class="value"> {{ section.help }}</div>
+        <div class="value">
+          <pre>{{ section.help }}</pre>
+        </div>
       </div>
     </div>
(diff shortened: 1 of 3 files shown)


AlexKarpov98 (changes requested):
Thanks for it.

One note - We need to add some styling here:

```
    white-space: pre-wrap;
    font-family: var(--font-family-body);
```
we need this:
Screenshot 2024-04-23 at 16 12 39

Currently it looks like this:
Screenshot 2024-04-23 at 16 11 53

AlexKarpov98 (approved):
Works good! Thank You. 👍

StevenMcElligott:
> Works good! Thank You. 👍

Np!

undsoft (changes requested):
Hi,
Unfortunately this is the file with global styles, we are trying to get away from setting styles globally.
It's better to set those styles in `ix-dynamic-wizard.component.scss`.

AlexKarpov98:
I'm sorry, I missed that this is a global change, I was in `ix-dynamic-wizard.component.scss` context. 
Thanks @undsoft .

AlexKarpov98:
@undsoft - now it is fine.

undsoft (approved):
Thanks.

PR #10002, opened 2024-04-27
NAS-127483 / 24.10 / Removed redundancy check for Dedup and Special vdevs with Data vdevs

RehanY147:


AlexKarpov98 (commented):
@RehanY147 @undsoft - do we need to fully remove `RedundancyMismatch` check?

RehanY147:
Adding a quote from Caleb here for his opinion on the matter since its relevant

> Any type of business logic related to "warning" a user on vdevs should be removed
> Before this type of logic gets put into the UI, the UI team needs to consult with the OS team to make sure it makes sense

RehanY147:
@AlexKarpov98 We can merge this PR. If more changes are required we can do it later.

AlexKarpov98 (approved):
@RehanY147 said we can merge this one, even though we removed `RedundancyMismatch` check.

PR #10007, opened 2024-04-29
NAS-128625 / 24.10 / Recent searches in global search shows unusual results

AlexKarpov98:
Testing: see ticket, try to update local storage`recentSearches`, add unexisting search result. (update hierarchy item)

You should only see items which are available in `ui-searchable-elements.json`.


undsoft (approved):
Old search results are gone.

RehanY147 (approved):

PR #10010, opened 2024-04-30
NAS-128655 / 24.10 / Reducing usages on EntityJobComponent

undsoft:
Test some of the affected places.


denysbutenko (approved):
I checked the `Import Pool`, `Replace Disk`, and `Save Debug`—there were no regressions. It looks good!

PR #10013, opened 2024-05-01
NAS-128689 / 24.10 / Fix layout preview issues for WidgetGroupFormComponent

RehanY147:
Go to the WidgetGroupForm on the new dashboard and switch between different possible layouts.
on `master`, the preview doesn't update properly. This PR fixes that.


AlexKarpov98 (approved):
It works now.
The only thing we can fix - is a background color of an empty element, white is not good.

https://github.com/truenas/webui/assets/22980553/61eb7dfb-c9b3-4230-a048-f29d47f8be4a

RehanY147:
I can fix that in a separate PR.

PR #10014, opened 2024-05-01
NAS-128386 / 24.10 / Port WidgetNetworkComponent to new dashboard

denysbutenko:
Check network widget at new dashboard


src/app/pages/dashboard/widgets/network/widget-network/widget-network.component.ts
```diff
@@ -0,0 +1,212 @@
...
+  ));
+
+  protected reportingData = toSignal(this.interface$.pipe(
+    filter(Boolean),
+    switchMap((networkInterface) => this.resources.networkInterfaceUpdate(networkInterface.name)),
+    map((response) => {
+      const updatedResponse = response[0];
+      (updatedResponse.data as number[][]).forEach((row, index) => {
+        // remove first column and convert kilobits/s to bytes
+        (updatedResponse.data as number[][])[index] = row.slice(1).map((value) => value * KiB);
```
RehanY147, line 59: Network traffic must be calculated in bps (bits per second) and must use base-10 units (Kb, Mb, Gb) instead of base-2 units (KiB, MiB, GiB).

RehanY147 (changes requested):

src/app/pages/dashboard/services/widget-resources.service.ts
```diff
@@ -28,7 +37,34 @@ export class WidgetResourcesService {
...
+    combineLatestWith(timer(0, 10000)),
+    map(([serverTime]) => {
+      serverTime.setSeconds(serverTime.getSeconds() + 10000 / 1000);
+      return serverTime;
+    }),
+  );
+
+  networkInterfaceUpdate(interfaceName: string): Observable<ReportingData[]> {
+    return this.serverTime$.pipe(
+      skipWhile(() => !interfaceName),
```
RehanY147, line 52: Don't think this will work when `interfaceName` is null/empty string. string is passed by value so that variable will never update and this observable will keep skipping emits because of the `skipWhile`.
Also, I think computing time values on the spot are a better approach. We don't need this information all the time. Just when making the calls. Why keep timers on when they are not needed.

src/app/pages/dashboard/widgets/network/widget-network/widget-network.component.ts
```diff
@@ -0,0 +1,214 @@
...
+          pointBackgroundColor: currentTheme.orange,
+          pointRadius: 0,
+          tension: 0.2,
+          fill: true,
+        },
+      ],
+    };
+  });
+
+  protected chartOptions: ChartOptions<'line'> = {
```
RehanY147, line 124: We should break this large object down into smaller chunks stored in variables with declarative names so it's more readable what we're doing.

src/app/pages/dashboard/widgets/network/widget-network/widget-network.utils.ts
```diff
@@ -0,0 +1,75 @@
...
+export function processNetworkInterfaces(interfaces: NetworkInterface[]): DashboardNetworkInterface[] {
+  const dashboardNetworkInterfaces = [...interfaces] as DashboardNetworkInterface[];
+  const removeNics: Record<string, number | string> = {};
+
+  // Store keys for fast lookup
+  const nicKeys: Record<string, number | string> = {};
+  interfaces.forEach((networkInterface, index) => {
+    nicKeys[networkInterface.name] = index.toString();
+
+    // Process Vlans (attach vlans to their parent)
```
RehanY147, line 14: This comment is out of place or no longer applicable.

src/app/pages/dashboard/widgets/network/widget-network/widget-network.utils.ts
```diff
@@ -0,0 +1,75 @@
+import { NetworkInterfaceType, NetworkInterfaceAliasType } from 'app/enums/network-interface.enum';
+import { NetworkInterface } from 'app/interfaces/network-interface.interface';
+import { DashboardNetworkInterface, DashboardNetworkInterfaceAlias } from 'app/pages/dashboard-old/components/dashboard/dashboard.component';
+
+export function processNetworkInterfaces(interfaces: NetworkInterface[]): DashboardNetworkInterface[] {
+  const dashboardNetworkInterfaces = [...interfaces] as DashboardNetworkInterface[];
+  const removeNics: Record<string, number | string> = {};
+
+  // Store keys for fast lookup
```
RehanY147, line 9: Comment not needed

src/app/pages/dashboard/widgets/network/widget-network/widget-network.utils.ts
```diff
@@ -0,0 +1,75 @@
...
+          (alias as DashboardNetworkInterfaceAlias).interface = nic;
+        });
+        if (dashboardNetworkInterfaces[nicKeys[nic] as number]) {
+          const concatenatedAliases = dashboardNetworkInterfaces[index].state.aliases.concat(
+            dashboardNetworkInterfaces[nicKeys[nic] as number].state.aliases,
+          );
+          dashboardNetworkInterfaces[index].state.aliases = concatenatedAliases;
+        }
+
+        // Consolidate vlans
```
RehanY147, line 44: Not needed

src/app/pages/dashboard/widgets/network/widget-network/widget-network.utils.ts
```diff
@@ -0,0 +1,75 @@
...
+        dashboardNetworkInterfaces[index].state.vlans.forEach((vlan) => { vlan.interface = nic; });
+        dashboardNetworkInterfaces[index].state.vlans = dashboardNetworkInterfaces[index].state.vlans.concat(
+          dashboardNetworkInterfaces[nicKeys[nic] as number].state.vlans,
+        );
+
+        // Mark interface for removal
```
RehanY147, line 50: Not needed

src/app/pages/dashboard/widgets/network/widget-network/widget-network.utils.ts
```diff
@@ -0,0 +1,75 @@
...
+        removeNics[nic] = nicKeys[nic];
+      });
+    }
+  });
+
+  // Remove NICs from list
```
RehanY147, line 56: All of the following comments are not needed

RehanY147 (changes requested):

src/app/pages/dashboard/services/widget-resources.service.ts
denysbutenko: Yeah, that's right. It looks like leftover since we are switching to `networkInterfaceUpdate` once the interface is loaded

src/app/pages/dashboard/widgets/network/widget-network/widget-network.component.ts
denysbutenko: It sounded like a good idea at first. When I did, it became less readable, and understanding what and where became more complex. So, I prefer to keep it as is. 

Here is what it might look like after splitting:
image
image

RehanY147 (approved):

PR #10017, opened 2024-05-02
NAS-128352: Search elements for System Settings

AlexKarpov98:
Testing: see ticket


bvasilenko (changes requested):
I've noticed that a few blocks are missing from search results:
- **System > Advanced > Cron Jobs**  
- **System > Advanced > Init/Shutdown Cron Scripts**
Apart from that works good!

AlexKarpov98:
@bvasilenko - `Cron Jobs` are there, it leads us to the Cron Jobs page.
Screenshot 2024-05-06 at 10 36 03

Same for `Init/Shutdown Cron Scripts`
Screenshot 2024-05-06 at 10 36 51

bvasilenko (approved):
Looks good, since I've realized that it matches what was expected

PR #10018, opened 2024-05-02
NAS-128676 / 24.04.1 / Fallback for failed Netdata login

undsoft:
Automatic password setup doesn't work for some users for some reason, so I built a fallback.

Снимок экрана 2024-05-02 в 16 35 03

Testing: Reporting -> Netdata. If you are testing on localhost, there will be an expected 404 error when opening dialog.


william-gr (approved):
Looks reasonable to me

AlexKarpov98 (approved):
Screenshot 2024-05-03 at 11 50 09
It works well for me!

PR #10022, opened 2024-05-03
NAS-128557 / 24.10 / Finishing WidgetGroupFormComponent

RehanY147:


src/app/pages/dashboard/components/widget-group-form/widget-group-form.component.ts
```diff
@@ -1,47 +1,161 @@
...
+      const widget = widgetRegistry[type];
+      if (widget.category === category && widget.supportedSizes.includes(slotSize)) {
+        return true;
+      }
+
+      return false;
+    });
+
+    this.widgetTypesOptions$ = of(sizeSuitedTypes.map((type) => {
+      return { label: this.getEnumKeyByEnumValue(WidgetType, type), value: type };
```
undsoft, line 152: Name should come from widgetDefinition and be translated.

src/app/pages/dashboard/components/widget-group-form/widget-group-form.component.ts
```diff
@@ -1,47 +1,161 @@
...
+
+  getEnumKeyByEnumValue<T extends Record<string, string>>(
+    myEnum: T,
+    enumValue: string,
+  ): keyof T | null {
+    const keys = Object.keys(myEnum).filter((x) => myEnum[x] === enumValue);
+    return keys.length > 0 ? keys[0] : null;
+  }
+
+  setCategoryOptions(): void {
```
undsoft, line 116: This is complicated and you are doing similar things here and in `setWidgetTypeOptions`.
Instead, why not first filter for widgets that fit the slot size in a private method:

```ts
    const widgets = Object.values(widgetRegistry);
    const layout = this.layout.value;
    const slotSize = layoutToSlotSizes[layout][this.selectedSlot];

    const supportedWidgets = widgets.filter((widget) => widget.supportedSizes.includes(slotSize));
```

and then extract categories:
```ts
    const categories = uniq(supportedWidgets.map((widget) => widget.category));
    return categories.map((category) => {
      return {
        label: widgetCategoryLabels.get(category) || category,
        value: category
      };
    });
```

or types when you need them.

src/app/pages/dashboard/components/widget-group-form/widget-group-form.component.ts
```diff
@@ -1,47 +1,161 @@
...
+  widgetCategoriesOptions$: Observable<Option[]>;
+  widgetTypesOptions$: Observable<Option[]>;
 
   // TODO: Implement template options
   templateOptions$ = of([]);
 
   constructor(
-    private formBuilder: FormBuilder,
-    private chainedRef: ChainedRef<WidgetGroup>,
+    protected chainedRef: ChainedRef<WidgetGroup>,
```
undsoft, line 47: Editing existing widgets groups is not supported.

src/app/pages/dashboard/components/widget-group-form/widget-group-form.component.ts
```diff
@@ -1,47 +1,161 @@
...
 export class WidgetGroupFormComponent {
   protected group: WidgetGroup;
   protected selectedSlot = 0;
-  protected form = this.formBuilder.group({
-    template: [''],
-    layout: [WidgetGroupLayout.Full],
-  });
-  readonly layoutsMap = widgetGroupIcons;
+
+  protected readonly template = new FormControl('');
```
undsoft, line 33: I think component store makes your life harder here, because instead of having one source of truth you things stored in three places. There are form controls, there's `group` inside of the component and there's a store.
It's really hard to trace what changes what here. 

Having a separate place for logic is a good idea, but then we should come up with a way to utilize it more.

We should avoid doing what we did in pool creation wizard, because it really makes for some complicated logic. 

Options for dropdowns should react to changes in data store whereever it is.

See this video https://www.youtube.com/watch?v=skOTEbGwncE
It focuses on rxjs, but you could do the same thing easier with signals.

You need to pick one place for source of truth (could be a store, could be a form or signals here or in a service) and then have other things derive from it.

undsoft (changes requested):
Needs more work.

undsoft:
Also you are storing validation errors for every slot, which is fine, but then we need to come up with an error state for those widgets, so that user knows where the error is if he switches to another slot.

RehanY147:
Will add tests once the approach is approved.

src/app/pages/dashboard/types/widget-component.interface.ts
```diff
@@ -18,13 +18,8 @@ export type WidgetComponent<Settings extends SomeWidgetSettings = null> =
...
     : WidgetComponentWithSettings<Settings>;
 
-export interface WidgetSettingsComponent<Settings extends SomeWidgetSettings = null> {
-  // TODO: Figure out how to communicate with parent component to read and write settings and get validation state.
-  something: Settings;
-}
-
 type WidgetSettingsComponentType<Settings> = Settings extends SomeWidgetSettings
-  ? WidgetSettingsComponent<Settings>
+  ? Settings
```
undsoft, line 22: The original intent was to tie widget settings to widgets, so that developers don't make mistakes when they define their components.
This is now in broken state.

You need to:
1. Make `WidgetSettingsRef` accept a `Seettings` generic.
2. Recover back `WidgetSettingsComponent` and have it require `WidgetSettingsRef<Settings>` as one of the fields (instead of `something`).
3. Have `WidgetInterfaceIpSettingsComponent` implement `WidgetSettingsComponent` correctly.

src/app/pages/dashboard/types/widget-settings-ref.interface.ts
```diff
@@ -0,0 +1,7 @@
+import { ValidationErrors } from '@angular/forms';
+
+export class WidgetSettingsRef {
+  getData: () => { slot: number; settings: object };
```
undsoft, line 4: I don't really see why settings components need to know what slot number they are in.
You can add slot number to `WidgetSettingsRef` in a way that would hide this information or you can just rely on currently selected slot.

src/app/pages/dashboard/widgets/network/widget-interface-ip/widget-interface-ip-settings/widget-interface-ip-settings.component.ts
```diff
@@ -1,14 +1,67 @@
...
+    this.widgetSettingsRef.updateValidity(this.slot, this.getAllFormErrors());
+    this.form.valueChanges.pipe(untilDestroyed(this)).subscribe({
+      next: (settings) => {
+        this.widgetSettingsRef.updateSettings(this.slot, { interface: settings.interfaceIp });
+        this.widgetSettingsRef.updateValidity(this.slot, this.getAllFormErrors());
+      },
+    });
+  }
+
+  private getAllFormErrors(): Record<string, ValidationErrors> {
```
undsoft, line 57: Needs to go to some utils function or be part of `WidgetSettingsRef` class.

src/app/pages/dashboard/components/widget-group-form/widget-group-form.component.ts
```diff
@@ -12,36 +26,204 @@ import { WidgetGroup, WidgetGroupLayout, widgetGroupIcons } from 'app/pages/dash
...
+      return {
+        label: widgetTypeLabels.get(type) || type,
+        value: type,
+      };
+    });
+    this.widgetTypesOptions$ = of(options);
+  }
+
+  getLayoutSupportedWidgets(): { type: WidgetType; category: WidgetCategory; [key: string]: unknown }[] {
+    const widgetsEntires = Object.entries(widgetRegistry);
```
undsoft, line 214: typo

undsoft (changes requested):
Practical issues:
1. I see `null widget is not supported`, when I should just see `Empty` and it shouldn't be white.
2. When I update settings, they should update the preview.
3. We need to clean up widget settings when user presses save.

src/app/pages/dashboard/components/dashboard/dashboard.component.ts
```diff
@@ -75,7 +75,7 @@ export class DashboardComponent implements OnInit {
   protected onAddGroup(): void {
     const newGroup: WidgetGroup = {
       layout: WidgetGroupLayout.Full,
-      slots: [],
+      slots: [{ category: null, type: null }],
```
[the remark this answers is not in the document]
RehanY147: Technically, we should be adding a new group here at all. But we're doing that and we are adding the new group to the dashboard right here. But since, we're doing that, a full layout widget has at least 1 slot and the object should represent that. I will remove the whole logic from here.

src/app/pages/dashboard/types/widget-group.interface.ts
```diff
@@ -38,6 +40,36 @@ export interface WidgetGroup {
...
+  layout: WidgetGroupLayout;
+  slots: [
+    (FormWidget | null)?,
+    (FormWidget | null)?,
+    (FormWidget | null)?,
+    (FormWidget | null)?,
+  ];
+}
+
+export function widgetGroupToFormWidgetGroup(widgetGroup: WidgetGroup): FormWidgetGroup {
```
undsoft, line 53: Optional change. If you were to determine category from widget type, you would not need to convert and de-convert into another object.

undsoft (changes requested):
Don't have any complains about architecture, just need to finish the PR now:
1. Fix button styles.
2. Clean up widgets in wrong slots on save.
3. Update preview when user updates settings.
4. Let's update Category and Type selects take 50% of the width.
5. Add tests.

src/app/pages/dashboard/components/widget-group-form/widget-editor-group/widget-editor-group.component.html
```diff
@@ -14,7 +14,13 @@
     class="inert-container"
     inert
   >
-    <ng-container *ngComponentOutlet="slot.component; inputs: slot.inputs;" />
+    <div *ngIf="hasErrors(slotIndex); else component" class="error">
+      <ix-icon name="error" class="error-icon"></ix-icon>
```
undsoft, line 18: WidgetErrorComponent?

src/app/pages/dashboard/components/widget-group-form/widget-group-slot-form/widget-group-slot-form.component.ts
```diff
@@ -0,0 +1,216 @@
...
+  templateUrl: './widget-group-slot-form.component.html',
+  styleUrl: './widget-group-slot-form.component.scss',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class WidgetGroupSlotFormComponent implements AfterViewInit, OnChanges {
+  slotConfig = input.required<WidgetGroupSlot<object>>();
+  slot = signal<WidgetGroupSlot<object>>(null);
+  categorySubscription: Subscription;
+  typeSubscription: Subscription;
+  settingsValidionErrors = signal({} as ValidationErrors);
```
undsoft, line 40: typo

src/app/pages/dashboard/components/widget-group-form/widget-group-form.component.ts
```diff
@@ -13,36 +29,163 @@ import { WidgetGroup, WidgetGroupLayout, widgetGroupIcons } from 'app/pages/dash
...
+      if (!this.group().slots[i]?.type) {
+        hasEmptySlot = true;
+        break;
+      }
+    }
+    return hasEmptySlot || this.validationErrors().some((errors) => !!Object.keys(errors).length);
+  });
+
+  protected readonly layoutsMap = widgetGroupIcons;
+  protected readonly widgetRegistry = widgetRegistry;
```
undsoft, line 68: unused

undsoft (changes requested):
1. User is not required to fill in all the slots. An empty slot is a valid state.
2. There should be a way to make slot empty, perhaps by selecting Empty in Category dropdown.
3. When I save form with a widget that `doesn't support size` it becomes empty (which is fine), but then it should still let me save.
4. Not sure about two column layout. I'd rather we tried to reduce vertical margins and then show Category and Type one after another in one 50% column.
5. If I have a group with one full slot and then I open this group for edit, Widget Category doesn't show current values.
6. I think we can automatically select first entry from Type when user changes Category.
7. If you trigger validation error in one slot and then switch to a layout where this slot is no longer visible, you won't be able to save the form.

undsoft (approved):
Looks good 👍🏼

undsoft:
Looked into failing tests a little bit.
I think this kind of error usually happens when there is an unexpected network request, but I cannot be sure.
The easiest current way to diagnose this is to update `package.json` to only run a few tests like so:
```
"test:pr": "yarn run check-env -s && echo 'Setting up temporary environment file...\\n' && yarn run ui remote -i 'headless.local' && jest --coverage --maxWorkers=2 src/app/pages/dashboard/components",
```

and then to start removing parts of template / component to see where the issue is.

My suspicion is that `MatIconTestingModule` doesn't stop network requests completely or that there is an issue with another dependency.

Also `checks layout selector` doesn't check anything. It should at least validate that changing the layout changes the preview by doing something like:
```
      const editor = spectator.query(WidgetEditorGroupComponent);
      expect(editor.group).toMatchObject(...);
```

It would be good to update other tests to also test what user sees / actual behaviour of the component, instead of calling component functions that should be marked as `protected`.

PR #10030, opened 2024-05-04
NAS-128488 / 24.10 / Cleanup `initialOptions` after adding

bvasilenko:
**Testing**

Please see ticket for replication steps and expected result


undsoft (commented):
Can you explain why this would help? I don't see `initialOptions` being used at all.

undsoft (changes requested):
I think we need to find another solution, something that would be more obvious.
Current fix is not obvious depends on how many times and behaviour of `mapOptions` will be different depending on how many times you call it.

PR #10037, opened 2024-05-08
NAS-128703 / 24.04.1 / Only 11 Cloud sync tasks are shown, "items per page" is missing

AlexKarpov98:


undsoft (commented):
Please check other tables as well.

undsoft (approved):
Pages is there. Thanks.

PR #10040, opened 2024-05-08
NAS-128623 / 24.10 / Port Storage widget to new dashboard

bvasilenko:
**Testing**

See ticket


denysbutenko (changes requested):
Please resolve the merge conflict to create a docker image for review.

denysbutenko (commented):
image

image

image

image

image

image

denysbutenko (changes requested):
It works great for me. However, I think we need to find a way to display the full pool name when multiple pools have similar names. The `Pool Status` button is not specific to a pool, so we could move it to a shared space or remove it to free up more valuable space for the pool name. I don't know which approach fits better here: multiline name on mouse over or tooltip or *put your idea*.

denysbutenko:
Other ideas: reduce padding and/or adaptive font size

denysbutenko (changes requested):
image

image

It looks okay when pool tile takes half of widget. When it takes full width looks weird.

Let's change tooltip position to `above`. Also, would be good to show tooltip above exact pool name not whole container. For example when I mouse over on the empty space opposite side to pool name it also shows tooltip. Pool name shifted up after button was removed.

denysbutenko:
The pool name is still shifted up. The vertical space between the pool name and pool details is huge. See screenshots.

Old Dashboard:
image

New Dashboard:
image

denysbutenko (approved):
Tooltip works fine. It looks good to me. Thanks for improvements

bvasilenko:
@AlexKarpov98 I've made a [follow-up PR](https://github.com/truenas/webui/pull/10098) where this issue is addressed

PR #10042, opened 2024-05-09
NAS-128846 / 24.10 / Do not execute actions when selecting results for Top Panel

AlexKarpov98:

https://github.com/truenas/webui/assets/22980553/3376edd7-6284-4e49-bf16-f6876906b73b

I think we don't need to click other buttons as well but only highlight them for the user, to help search where they are.
So what we should click - is only `triggerAnchor` if it's provided.


RehanY147 (changes requested):
Ticket also asks to add shut down synonyms like "turn off", "reboot" etc.

RehanY147 (approved):

PR #10047, opened 2024-05-10
NAS-128451 / 24.10 / Port WidgetSysInfoComponent to new dashboard

denysbutenko:
There are some TODOs to resolve but it is ready for review and testing.

The widget has some changes:

- Show scale image placeholder even if it is Unsupported Hardware.
- When a system has HA but HA is disabled the triangle exclamation icon and tooltip was moved into the button:
image

- When a system has running update `UPDATE IN PROGRESS` string was moved into the button:
image

- The system reports icon was added on widget for active node that brings user to `/reportsdashboard/system`
image


denysbutenko:
@RehanY147 you're lucky :)

src/app/constants/time.constant.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Milliseconds in an hour
+ */
+export const oneHour = 60 * 60 * 1000;
```
RehanY147, line 4: Should add `'Millis'` as a postfix to both variable names. Makes it more obvious what value the variable holds

src/app/interfaces/system-info.interface.ts
```diff
@@ -3,7 +3,7 @@ import { LicenseFeature } from 'app/enums/license-feature.enum';
 import { ApiDate, ApiTimestamp } from 'app/interfaces/api-date.interface';
 
 export interface SystemInfo {
-  platform: string;
+  platform: string; // Check if this presents in the API
```
RehanY147, line 6: Should be a proper TODO comment or removed if not needed

src/app/pages/dashboard/widgets/system/common/product-image/product-image.component.html
```diff
@@ -0,0 +1,30 @@
...
+</div>
+
+<div class="product-image-text">
+  <strong *ngIf="productImage() && isHaLicensed()">(<ng-content></ng-content>)</strong>
+  <strong *ngIf="isUnsupportedHardware()">({{ 'Unsupported Hardware' | translate }})</strong>
+</div>
+
+<ng-template #generic>
+  <ix-icon
+    *ngIf="!isCertified() && !isHaLicensed()"
```
RehanY147, line 21: Feels like we should have fallback for all the other cases. Currently only two cases are covered.
`!isCertified() && !isHaLicensed()` and `isCertified() && !isHaLicensed()`. If other cases like `isCertified() && isHaLicensed()` or `!isCertified() && isHaLicensed()` are not possible, then never mind. Still, we should use `else` instead of two ifs just in case.

src/app/pages/dashboard/widgets/system/common/product-image/product-image.component.spec.ts
```diff
@@ -0,0 +1,3 @@
+describe('ProductImageComponent', () => {
+  // TODO: Write unit tests
```
RehanY147, line 2: Tests are required to be in the same PR. I think we should at least have a ticket linked in the PR so it's in the queue and doesn't get ignore for a while.

src/app/pages/dashboard/widgets/system/common/product-image/product-image.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  @HostBinding('class.truenas') isEnterprise = input.required<boolean>();
+  product = input.required<string>();
+  productImage = input.required<string>();
+  hasEnclosureSupport = input.required<boolean>();
+  isHaLicensed = input.required<boolean>();
+  isUnsupportedHardware = input.required<boolean>();
+  isCertified = input.required<boolean>();
+
+  productImageSrc = computed(() => {
+    if (!this.productImage()) {
```
RehanY147, line 27: I think it's better to call the signal once and assign the value to a variable const and reuse that instead of calling it over and over for readibility. Also, although not applicable here, conditional signals can sometimes be ignored in certain cases.

src/app/pages/dashboard/widgets/system/widget-sys-info-passive/widget-sys-info-passive.component.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+    return getProductEnclosure(this.systemInfo().system_product);
+  });
+
+  isUnsupportedHardware = computed(() => {
+    return this.isEnterprise() && !this.productImage() && !this.isIxHardware();
+  });
+
+  systemVersion = computed(() => {
+    return getSystemVersion(this.systemInfo().version, this.systemInfo()?.codename);
```
RehanY147, line 64: `this.systemInfo()` has a null check some times and sometimes it doesn't have that. We should decide if null check is needed or not. An alternative would be to add a filter pipe to the observable select so it never emits null.

src/app/pages/dashboard/widgets/system/widget-sys-info-active/widget-sys-info-active.component.ts
```diff
@@ -0,0 +1,82 @@
...
+  isEnterprise = toSignal(this.store$.select(selectIsEnterprise));
+  isHaLicensed = toSignal(this.store$.select(selectIsHaLicensed));
+  isCertified = toSignal(this.store$.select(selectIsCertified));
+  hasEnclosureSupport = toSignal(this.store$.select(selectEnclosureSupport));
+  isUpdateRunning = toSignal(this.store$.select(selectUpdateJobForActiveNode));
+  updateAvailable = toSignal(this.resources.updateAvailable$);
+  systemInfo = toSignal(this.resources.systemInfo$);
+
+  product = computed(() => {
+    return getServerProduct(this.systemInfo()?.system_product);
```
RehanY147, line 39: Same thing here.

src/app/pages/dashboard/widgets/system/widget-sys-info-active/widget-sys-info-active.component.html
```diff
@@ -0,0 +1,126 @@
...
+        </mat-list>
+      </div>
+    </div>
+  </mat-card-content>
+</mat-card>
+
+
+<ng-template #generic>
+  <ix-icon
+    *ngIf="!isCertified() && !isHaLicensed()"
```
RehanY147, line 116: Same thing here as well. Please review comment on passive component

RehanY147 (changes requested):

denysbutenko:
@RehanY147 okay I made bunch of adjustments and ready for another review.

src/app/constants/time.constant.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Milliseconds in an hour
```
RehanY147, line 2: I don't think the comments are needed anymore. The variable name is clear enough.

src/app/pages/dashboard/widgets/system/common/product-image/product-image.component.ts
```diff
@@ -0,0 +1,45 @@
...
+    return this.productImage().includes('ix-original.svg');
+  }
+  @HostBinding('class.truenas')
+  isEnterprise = input.required<boolean>();
+  systemProduct = input.required<string>();
+  hasEnclosureSupport = input.required<boolean>();
+  isHaLicensed = input.required<boolean>();
+  isIxHardware = input.required<boolean>();
+
+  isUnsupportedHardware = computed(() => this.isEnterprise() && !this.isIxHardware() && !this.isHaLicensed());
```
RehanY147, line 26: Prefer writing this like the following
```
isUnsupportedHardware = computed(() => {
  const isEnterprise = this.isEnterprise();
  const isIxHardware = this.isIxHardware();
  const isHaLicensed = this.isHaLicensed();
  return isEnterprise && !isIxHardware && !isHaLicensed;
});
```
Because of the way JavaScript handles comparison operators, it can ignore a signal dependency in a computed signal if during the initial execution, the signal was never fetched. For example, `computed(() => this.isEnterprise() || this.isHa())` will ignore `this.isHa()` every time if `this.isEnterprise()` is always true. So, `this.isHa()` is never considered a dependency for the computed signal.
This applies to all computed signals. As a best practice, any dependency signals should be fetched at the start of the compute callback function into variables so it's clear that they are all dependencies.

RehanY147 (approved):

PR #10063, opened 2024-05-14
NAS-128968 / 24.10 / TrueCloud Backup shows non-Storj credentials

AlexKarpov98:
Testing: see ticket
Screenshot 2024-05-14 at 16 49 08


denysbutenko (approved):
👍🏻

src/app/modules/custom-selects/cloud-credentials-select/cloud-credentials-select.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+        {
+          provide: CloudCredentialService,
+          useValue: cloudCredentialServiceMock,
+        },
+        {
+          provide: TranslateService,
+          useClass: MockTranslateService,
+        },
+      ],
+      schemas: [NO_ERRORS_SCHEMA],
```
undsoft, line 41: No bueno.

src/app/modules/custom-selects/cloud-credentials-select/cloud-credentials-select.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    const options = await fixture.componentInstance.fetchOptions().toPromise();
+    expect(options).toEqual([{ label: 'AWS S3 (Amazon S3)', value: '1' }]);
+    expect(cloudCredentialService.getCloudSyncCredentials).toHaveBeenCalled();
+  });
+
+  it('should return the correct form component type', () => {
+    expect(component.getFormComponentType()).toBe(CloudCredentialsFormComponent);
+  });
+});
```
undsoft, line 69: Was this test auto-generated? It needs to rewriten to match our code base by using spectator, removing unnecessary mocks and changing test to not use methods that should be protected like  `fetchOptions`.

You should create a test using `createHost` in spectator and test things that user/other components will interact with. This mostly means changing inputs and seeing that `ix-select` got populated with correct options.

src/app/modules/custom-selects/cloud-credentials-select/cloud-credentials-select.spec.ts
```diff
@@ -0,0 +1,69 @@
+import { NO_ERRORS_SCHEMA } from '@angular/core';
```
undsoft, line 1: Filename should be `cloud-credentials-select.component.spec.ts`

undsoft (changes requested):
Test needs to be rewritten and I would also try to figure out why codecov views most of the component as not being covered.

undsoft (approved):
Thanks.

PR #10064, opened 2024-05-14
NAS-128457 / 24.10 / Reorganizing enclosure pages

undsoft:
Reorganizes enclosure pages using router into different views, which can be seen by clicking in Elements menu.


src/app/pages/system/enclosure/components/enclosure-header/view-elements-menu/view-elements-menu.component.ts
```diff
@@ -0,0 +1,56 @@
...
+      EnclosureElementType.TemperatureSensors,
+      EnclosureElementType.Cooling,
+      EnclosureElementType.VoltageSensor,
+      EnclosureElementType.PowerSupply,
+      EnclosureElementType.SasConnector,
+      EnclosureElementType.EnclosureServicesControllerElectronics,
+    ] as const;
+
+    return supportedViews
+      .filter((view) => this.enclosure().elements[view])
```
RehanY147, line 33: Should declare a constant at the start and save the value and reuse it so it's clear this is a dependency for the computed signal.

src/app/pages/system/enclosure/enclosure.routing.ts
```diff
@@ -0,0 +1,46 @@
...
+    path: '',
+    component: EnclosureDashboardComponent,
+    data: { title: T('View Enclosure'), breadcrumb: null },
+    children: [
+      {
+        path: '',
+        component: EnclosureViewComponent,
+        data: { title: T('View Enclosure'), breadcrumb: null },
+      },
+      {
```
RehanY147, line 32: duplicate

RehanY147 (changes requested):
Architecture looks good. A couple issues in code pointed out. Also, when elements select changes value, the "Disks" option stays highlighted.

RehanY147 (approved):
Please address the above suggestions. Otherwise looks good.

undsoft:
Doesn't seem like there is an easy solution to item highlighting in the menu.

PR #10065, opened 2024-05-15
NAS-128977 / 24.10 / TrueCloud | After new Provider is created no `Bucket & Folder` is available

AlexKarpov98:
Testing: see ticket


bvasilenko (approved):
Works as expected now!

PR #10072, opened 2024-05-16
NAS-128973 / 24.10 / Add `force` flag boolean to `jbof.delete` call

RehanY147:
Also test for regressions in deleting certificates and CSR's. And check for the force flag in the jbof delete call.
This PR creates a new dialog for confirming delete with an optional force flag.


src/app/interfaces/confirm-force-delete-dialog-config.interface.ts
```diff
@@ -0,0 +1,12 @@
+import { Role } from 'app/enums/role.enum';
```
undsoft, line 1: Can be kept close to the dialog.

undsoft (changes requested):
So, I started looking at the code and noticed that `ConfirmDialogComponent` already has a lot of similar functionality that we may have reimplemented :)

PR #10073, opened 2024-05-16
NAS-129032 / 24.10 / TrueCloud | Task is reselected to the first one after update of any other selected task

AlexKarpov98:

https://github.com/truenas/webui/assets/22980553/6a18ad45-1317-4397-a075-627cbcc6c624


undsoft (approved):
Looks good, row stays selected.

PR #10082, opened 2024-05-19
NAS-129086: Update some inputs

undsoft:


bvasilenko (approved):
Smoke testing revealed no regressions

PR #10092, opened 2024-05-21
NAS-129134 / 24.10 / Add keyboard shortcut to global search

undsoft:
See hint and test shortcut for global search.


undsoft:
Denys is adding escape behaviour here https://github.com/truenas/webui/pull/10096

PR #10095, opened 2024-05-21
NAS-129092 / 24.10 / Experiment with fuzzy search for global search

AlexKarpov98:
Testing: see ticket

Now we should be able to find search results even with a typo in a query. `advancd sett`


RehanY147 (approved):
Search works with typos

PR #10102, opened 2024-05-23
NAS-128675 / 13.3 / It is not possible to detach attached drive from boot poo…

undsoft:
See ticket.

Not sure why those lines were added.


denysbutenko (approved):
Due to limitations, I couldn't attach or replace a disk from boot pool on test VM. The system from the ticket also does not work with this image.

I assume it works fine.

PR #10105, opened 2024-05-23
NAS-129184 / 24.10 / Global search improvements & bug fix

AlexKarpov98:
as well I returned global search results sorting (hierarchy first / synonyms matches next)


undsoft (approved):
Not sure what second part does, but pressing Enter works correctly now.

PR #10111, opened 2024-05-27
NAS-129179 / 24.10 / Uptime widget

AlexKarpov98:
Testing: see ticket & new widget
Screenshot 2024-05-27 at 11 49 26


undsoft (changes requested):
1. `fiveSecondsRefreshInterval$` is a poor name, because if want to change it to 10 seconds in the future, now we need to rename everything.
2. Updating uptime via increment will not work correctly if user puts tab in the background. Timers will get suspended after a while and you'll get incorrect value shown.

I think instead of incrementing values a better approach would be to save original uptime from middleware, `Date` when it was queried and then calculate difference with now when timer emits.

undsoft (approved):
Looks good, but similar code can be extracted somewhere else out of widgets.

PR #10114, opened 2024-05-27
NAS-128468 / 13.3 / Enclosure reports all disks as not attached to any pool

undsoft:
I'll attach in the ticket the patch to mock request.

Looks like original issue happened because condition changed it's meaning over time.


denysbutenko (approved):
image

Looks good

PR #10120, opened 2024-05-28
NAS-129253 / 24.10 / Add script to fetch prebuilt nodejs

anodos325:
This commit adds a simple script to download the specified pre-built nodejs binary. The intention for this script is to be called from within scale-build while chrooted during the package build and so hammering the into /usr/bin and /usr/lib is okay, but probably shouldn't be done on developer systems.


RehanY147 (approved):

undsoft (approved):
👍🏼

PR #10122, opened 2024-05-29
NAS-129096 / 24.10 / Global search QA 1

AlexKarpov98:


denysbutenko (approved):
It works great, except a few minor things:

> Remove Manage Disks and Manage Datasets from Storage search results, because it's similar to Datasets and Disks.

I think it would be useful to keep `Manage Disks`/`Manage Datasets` as synonyms for Disks/Datasets

> Add "add/new/create share" as synonyms for Shares page

For those keywords, only `Shares` appear in search results. The `Create NFS Share` and `Create SMB Share` seem helpful. Let's add synonyms to specific shares.

Also, `iSCSI share` is not found. According to [documentation](https://www.truenas.com/docs/core/coretutorials/sharing/iscsi/addingiscsishare/), it is used term. Let's add synonyms to iSCSI Wizard.

Also, I think it makes sense to keep `Sharing` as a synonym for `Shares` pages.

denysbutenko:
Also, I saw duplicate results. not sure if this is related to the PR.
image

PR #10128, opened 2024-05-31
NAS-129265 / 24.10 / Enable sorting on more tables

AlexKarpov98:
Testing: see ticket & check tables.

### DIFF
--- a/src/app/modules/ix-table/components/ix-table-head/ix-table-head.component.spec.ts
+++ b/src/app/modules/ix-table/components/ix-table-head/ix-table-head.component.spec.ts
@@ -25,6 +25,7 @@ const columns = createTable<TestTableData>([
   textColumn({
     title: 'Boolean Field',
     propertyName: 'booleanField',
+    disableSorting: true,
   }),
 ], {
   rowTestId: (row) => 'row' + row.numberField.toString(),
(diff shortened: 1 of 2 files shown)


RehanY147 (changes requested):
Some sorting isn't working. Don't know why. Please check.

https://github.com/truenas/webui/assets/24204846/a9b443d9-d6b5-4200-b2a0-792e3a416ff0

RehanY147 (changes requested):
With so many `sortable: true` maybe that should be the default. We should be intentionally turning it off when we don't want sorting

src/app/pages/audit/components/audit/audit.component.ts
```diff
@@ -79,17 +78,15 @@ export class AuditComponent implements OnInit, OnDestroy {
...
     dateColumn({
       title: this.translate.instant('Timestamp'),
       propertyName: 'message_timestamp',
-      sortable: true,
       getValue: (row) => row.message_timestamp * 1000,
     }),
     textColumn({
       title: this.translate.instant('Event'),
-      sortable: true,
+      disableSorting: true,
```
RehanY147, line 89: This looks like a mistake. Column was sortable before.

RehanY147 (approved):
Looks good

PR #10133, opened 2024-06-03
NAS-129160 / 24.10 / Improvements for System Info widget

denysbutenko:
- Fixed text on the Update button. It was showed `Updates Available` even if the update is unavailable.
- Added button to license row. It will redirect to the support page where the user can add a license.
- Updated standby widget to show text on systems without an HA license.
- Added unit tests

image


denysbutenko:
> Looks good, but is that OK that on m40, I see this? Screenshot 2024-06-04 at 11 31 39

Definitely not OK :) Thanks, I will check it

denysbutenko:
also failover button:

image

PR #10135, opened 2024-06-03
NAS-129241 / 24.10 / Mock enclosure rework

undsoft:
Simplifies enclosure mocking, reworks some commands. Only one model is supported.

Testing:
* Test yarn ui command and all the options.
* See how enclosure mocking works in old enclosure UI.


denysbutenko:
I did a quick overview, and the chunked code looks much better.

`yarn ui` has `error Command failed with exit code 1.` message in output. It is probably better to alias `yarn ui === yarn ui help`

Also, system info overrides needed to be recovered. They are used on the `webui.main.dashboard.sys_info` and `system.info` endpoints. Currently, the product image component can't correctly identify m40.

undsoft:
@denysbutenko Fixed.

RehanY147 (changes requested):
`yarn ui remote -i "[2400:adc7:125:4800:a00:27ff:fef2:c326]"` doesn't work. That's a IPv6 version. It doesn't pass the match for the IPv4 ip that's in the code now. Previous solution was to just add a `--force` flag to remove any validation checks and pass it as is.

src/app/core/testing/mock-enclosure/mock-storage-generator.utils.ts
```diff
@@ -1,984 +1,54 @@
...
-    const generateName = (index: number): string => {
-      let result = '';
-      do {
-        result = (index % 26 + 10).toString(36) + result;
-        index = Math.floor(index / 26) - 1;
-      } while (index >= 0);
-      return result;
+      ...response,
+      system_product: this.config.controllerModel,
+      system_manufacturer: 'iXsystems',
```
denysbutenko, line 30:
```suggestion
      ...response,
      platform: `TRUENAS-${this.config.controllerModel}`,
      system_product: `TRUENAS-${this.config.controllerModel}`,
      remote_info: {
        ...response.remote_info,
        platform: `TRUENAS-${this.config.controllerModel}`,
        system_product: `TRUENAS-${this.config.controllerModel}`,
      },
```
`TRUENAS-` prefix is used to determine enclosure in system features.
`system_product` called `platform` in `webui.main.dashboard.sys_info`, so needs both
`remote_info` is system info for standby node in `webui.main.dashboard.sys_info`
`truenas.is_ix_hardware` is used to determine the manufacturer
`system_manufacturer` is not used anymore

undsoft:
@RehanY147 
> yarn ui remote -i "[2400:adc7:125:4800:a00:27ff:fef2:c326]" doesn't work.
--force flag still works, at least for me.

src/app/core/testing/mock-enclosure/mock-storage-generator.utils.ts
undsoft: I only see one check for `TRUENAS` in system product. I don't think we should rely on it.
`truenas.is_ix_hardware` is already mocked in `MockEnclosureWebsocketService`.
Added remote_info and platform.

denysbutenko: System Info widgets rely on the `selectEnclosureSupport` selector, whose value depends on the prefix in `system_product`. This selector's source could be changed, but otherwise, some functions do not work correctly.

undsoft: Yeah, I saw it, I think we'd need to update it.

PR #10137, opened 2024-06-04
NAS-129359 / 24.10 / NFS sessions page has no selected NFS type on initial load

AlexKarpov98:
See ticket.


undsoft (approved):
Page loads correctly now.

PR #10138, opened 2024-06-04
NAS-129360 / 24.10 / Language Change Is Not Updating UI Correctly

AlexKarpov98:
Testing: see ticket


denysbutenko (changes requested):
Global search works well when language is changed.

`Yes/No` table column still shows untranslated string:
image

`Вибух: No` 🙂

denysbutenko (approved):

PR #10140, opened 2024-06-04
NAS-129264 / 24.10 / Add CPU Recent Usage widget

bvasilenko:
**Summary**

FreeBSD does not have a single CPU usage chart. So we're going to display `System` and `User` CPU usage charts - these are the closest analogues of overall CPU usage

**Testing**

Please test the widget on the new dashboard


RehanY147 (changes requested):
CPU stats don't update once you navigate away and come back to the dashboard.

https://github.com/truenas/webui/assets/24204846/d58bf4d7-daa0-48a9-9fab-d1f84b6bf093

bvasilenko:
> CPU stats don't update once you navigate away and come back to the dashboard.

@RehanY147  Turned out that this is not caused by the WebUI, and root cause is middleware. I've moved your notice into a [separate issue](https://ixsystems.atlassian.net/browse/NAS-129552). 

So, I suggest to go ahead and merge this PR, if there's no other issues.

RehanY147 (approved):

PR #10143, opened 2024-06-05
NAS-129078 / 24.10 / iX Table Search Fix

AlexKarpov98:
Testing: see ticket & test some of the tables.
I fixed a weird logic when you start typing something it converts to lower case and is not case sensitive anyway.
Screenshot 2024-06-05 at 14 31 37


RehanY147 (changes requested):
I think it's better to declare one function that takes in a query string and an array of comparison strings and compares them after applying lowercase function on them. Makes our code cleaner. And it will be reusable.

PR #10146, opened 2024-06-05
NAS-128542 / 24.04.0 / Update certificate list when creating acme certificate

bvasilenko:
**Testing**

For proper end-to-end testing you'll need to be the domain owner, and update it's DNS records.

Since the scope is limited to the frontend it will be easier for you to make a mock , i.e. temporarily add code that simulates the successful completion of a job that creates an ACME Certificate. Then, you need to test that the Certificates block receives fresh data from the server.


undsoft:
24.04.0 was released more than a month ago.

denysbutenko (changes requested):
The changes looking good but targeted wrong branch

bvasilenko:
replaced by https://github.com/truenas/webui/pull/10161

PR #10148, opened 2024-06-06
NAS-128803 / 24.10 / Removed quota items from the card and edit button for Zvol

RehanY147:


undsoft (approved):
Looks good, zvols no longer show quota info.

PR #10153, opened 2024-06-06
NAS-129016 / 24.10 / Add loading state for dashboard

denysbutenko:


bvasilenko (approved):
Works good in both cases

PR #10156, opened 2024-06-07
NAS-129443 / 24.10 / ui-searchable-elements.json is changed on each `yarn extract-ui-search-elements` command

AlexKarpov98:
Try calling `yarn extract-ui-search-elements` locally and see that file is not added to the git Source Control if nothing has changed.

Without this fix it regenerates randomly our .json file and we have a git conflict on each merge into master.


denysbutenko (approved):
It's really annoying

scripts/ui-search/find-component-files.ts
```diff
@@ -1,5 +1,15 @@
-import { glob } from 'glob';
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable import/no-extraneous-dependencies */
+import glob from 'glob';
```
undsoft, line 3: Where is this package coming from? Some types should exist for it somewhere.

PR #10165, opened 2024-06-10
NAS-129444 / 24.10 / Static Search Improvements

AlexKarpov98:
Testing: see ticket & test local searches over the WEB UI application.

### DIFF
--- a/src/app/modules/ix-table/classes/async-data-provider/async-data-provider.ts
+++ b/src/app/modules/ix-table/classes/async-data-provider/async-data-provider.ts
@@ -2,7 +2,7 @@ import { Observable } from 'rxjs';
 import { EmptyType } from 'app/enums/empty-type.enum';
 import { BaseDataProvider } from 'app/modules/ix-table/classes/base-data-provider';
 import { TableFilter } from 'app/modules/ix-table/interfaces/table-filter.interface';
-import { filterTableColumns } from 'app/modules/ix-table/utils';
+import { filterTableRows } from 'app/modules/ix-table/utils';
 
 export class AsyncDataProvider<T> extends BaseDataProvider<T> {
   private loadedRows: T[] = [];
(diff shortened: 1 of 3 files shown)


src/app/modules/ix-table/interfaces/table-filter.interface.ts
```diff
@@ -0,0 +1,6 @@
+export interface TableFilter<T> {
+  query: string;
+  columnKeys: (keyof T)[];
+  list?: T[];
+  preprocessMap?: Partial<Record<keyof T, (value: T[keyof T]) => string>>;
+}
```
undsoft, line 6: Currently a key in `preprocessMap` is any key of T, meaning that any type is allowed on function argument as long as it exists in T.

If you change to:
```
export interface TableFilter<T> {
  query: string;
  columnKeys: (keyof T)[];
  list?: T[];
  preprocessMap?: {
    [K in keyof T]?: (value: T[K]) => string;
  };
}
```
it would require key to match function type.

undsoft (approved):
Nice improvement 👍🏼 

I get:
```
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at _a31.onListFiltered (smb-list.component.ts:209:26)
```
on SMB shares table when I navigate to the page. 
And the same error on empty NFS table when I search.

src/app/pages/vm/devices/device-list/device-list/device-list.component.ts
```diff
@@ -140,11 +140,14 @@ export class DeviceListComponent implements OnInit {
...
-      const deviceTypeLabel = this.getDeviceTypeLabel(device);
-      return String(device.id).includes(this.filterString)
-        || deviceTypeLabel.toLowerCase().includes(this.filterString);
-    }));
+    this.dataProvider.setFilter({
+      list: this.devices,
+      query,
+      columnKeys: ['id', 'dtype'],
+      preprocessMap: {
+        dtype: (dtype: number) => this.getDeviceTypeLabel({ dtype } as unknown as VmDevice),
```
undsoft, line 148: Once you update the `TypeFilter` interface, you can set `dtype` to `VmDeviceType` and remove `unknown`.

PR #10168, opened 2024-06-11
NAS-129334 / 24.04.2 / Fix Dashboard is not loading

denysbutenko:
Check instructions in ticket


bvasilenko (approved):
It successfully resolves the error mentioned in the ticket

PR #10169, opened 2024-06-11
NAS-129438 / 24.10 / White Box surrounds iXsystems Logo in UI after Clicking enter in the Search UI

AlexKarpov98:
Testing: see ticket, now we are focusing on the first element in the sidenav content, it's needed to not loose focus.
As well search element in global search is fixed to focus from first to second if search input is focused (previously we had to double tab to move to second element)


denysbutenko (approved):
Seems to work fine

PR #10172, opened 2024-06-11
NAS-129078 / 24.04.2 / Search Field in Cloud Sync Task List only finds lower case elements

AlexKarpov98:
Testing: see ticket
This fix is only for **Dragonfish** branch.
Master branch has a different fix & improvement (https://github.com/truenas/webui/pull/10165)


denysbutenko (approved):
👍🏻

PR #10173, opened 2024-06-11
NAS-129425 / 24.10 / Hide standby node info widget for a non-HA system

bvasilenko:
**Summary**

In general, the **System Info** widget that needs to be hidden can be any size. And, if all the widgets are hidden inside a group of widgets (dark gray tile below),

image

in this case one, then the group itself may not be closed.

I made this conclusion on the basis that now it is possible to add an empty group of widgets.

**Testing**

See ticket for replication steps

For Non-HA system use any CI instance. For HA system use 10.220.16.103


undsoft (changes requested):
1. WidgetResources is not an appropriate place for this logic.
2. Current approach means that there is going to be a service which is going to be smart. It's going to know conditions for hiding any widget that needs to be hidden. This violates O in SOLID. Is there a way to have this logic contained in widget/widget definition?

RehanY147 (commented):
What Evgeny said

undsoft (changes requested):
I was thinking more about moving this logic:
```
if (!isHaLicensed) {
  widgets.push(WidgetType.SystemInfoPassive);
}
```
here.
I know that it may be tricky, but I think there should be a way to do it.
For example, see if we can do injection like in Angular dependency containers:

```
visibility: {
  deps: [Store],
  isVisible$: (store: Store) => store.select(selectIsHaLicensed),
}

```

See if this helps https://angular.dev/guide/di/dependency-injection-context#run-within-an-injection-context

src/app/pages/dashboard/types/widget-component.interface.ts
```diff
@@ -27,6 +29,9 @@ type WidgetSettingsComponentType<Settings> = Settings extends SomeWidgetSettings
   ? WidgetSettingsComponent<Settings>
   : null;
 
+export type WidgetVisibilityProvider = Store;
```
undsoft, line 32: Better to specify as generics or at least unknowns. Adding visibility condition to a widget definition should not necessitate adding types here.

src/app/pages/dashboard/components/widget-group-form/widget-group-slot-form/widget-group-slot-form.component.ts
```diff
@@ -148,6 +164,27 @@ export class WidgetGroupSlotFormComponent implements AfterViewInit, OnChanges {
     this.setValuesFromInput();
   }
 
+  ngOnInit(): void {
+    runInInjectionContext(this.environmentInjector, () => {
```
undsoft, line 168: Large and complicated function needs to be split into methods/utility functions. Perhaps you can also figure out a way to extract similar code here and in `widget-group` into a shared function.

undsoft (changes requested):
The code works, but let's add test cases for new behaviour.

undsoft (approved):
Okay, it works correctly.

PR #10177, opened 2024-06-12
NAS-129106 / 24.10 / Traceback on webdav install

AlexKarpov98:
Testing: see ticket.
The main idea here is to update validation, add missing validators, so that user will not see errors after submit if he did fill all the required fields.
I asked @stavros-k to set up an app for testing, you can do the same, just provide the machine where it's possible.

See video of the result:

https://github.com/truenas/webui/assets/22980553/190de7ab-f18a-42ee-a068-a5f091ca0ee2


stavros-k (approved):

bvasilenko (commented):
Works as expected,

However, I've noticed that for MinIO app, it actually was possible to install the app before this addition. But after this addition, there's a weird ambiguity , shown below 

**Domain**  field has  **(Optional)** in it's name,
However, after these changes, validation message `Domain is required` is displayed below this field - see screenshot below

image


So, in general, to be able to install the app, user has to fill in much more fields, even the ones which were optional before.

bvasilenko (approved):
Works good! No regressions found

PR #10178, opened 2024-06-12
NAS-129528 / 24.04.2 / ACL forms are broken when directory services cache is dis…

undsoft:
…abled.


denysbutenko (approved):
Looks good

PR #10180, opened 2024-06-13
NAS-129501 / 24.10 /  NAS-129501: Fix expansion shelves in old enclosure UI

AlexKarpov98:
Testing: see ticket:
Screenshot 2024-06-12 at 17 54 16


undsoft (approved):
Thanks, enclosures selector looks good.

PR #10181, opened 2024-06-13
NAS-129551 / 24.10 / Advanced search some Reporting link to pages that does not exist, click goes nowhere

AlexKarpov98:
Testing: see ticket


denysbutenko (approved):
No evidence exists for `NFS, Partitions or Target` in [graphs.py](https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/plugins/reporting/netdata/graphs.py). LGTM

PR #10191, opened 2024-06-15
NAS-129104 / 24.10 / Add remaining seconds back

RehanY147:


undsoft (approved):
Impossible to test right now, but code is the same as it was before previous PR.

PR #10200, opened 2024-06-17
NAS-129625 / 24.10 / Dashboard - Editing - I can click and focus on widget buttons

AlexKarpov98:
Testing: see ticket.

Current result:

https://github.com/truenas/webui/assets/22980553/a93fb440-7c53-4924-b672-bc31e89baf21

### DIFF
--- a/src/app/pages/dashboard/components/dashboard/dashboard.component.ts
+++ b/src/app/pages/dashboard/components/dashboard/dashboard.component.ts
@@ -2,8 +2,10 @@ import {
   animate, group as groupAnimations, style, transition, trigger,
 } from '@angular/animations';
 import { moveItemInArray } from '@angular/cdk/drag-drop';
+import { DOCUMENT } from '@angular/common';
 import {
-  ChangeDetectionStrategy, Component, HostListener, OnInit, computed, signal,
+  ChangeDetectionStrategy, Component, ElementRef, HostListener,
+  Inject, OnInit, QueryList, ViewChildren, computed, signal,
 } from '@angular/core';
 import { toSignal } from '@angular/core/rxjs-interop';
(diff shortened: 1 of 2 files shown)


src/app/pages/dashboard/components/dashboard/dashboard.component.ts
```diff
@@ -145,4 +150,24 @@ export class DashboardComponent implements OnInit {
...
       });
   }
+
+  private updateIsEditingState(value: boolean): void {
+    this.isEditing.set(value);
+    this.updateElementsAvailability();
+  }
+
+  private updateElementsAvailability(): void {
+    const groupElements = this.document.querySelectorAll('.group *');
```
denysbutenko, line 160: Unsure if modifying DOM here is a good approach.

src/app/pages/dashboard/components/dashboard/dashboard.component.ts
```diff
@@ -145,4 +150,24 @@ export class DashboardComponent implements OnInit {
         this.renderedGroups.set(groups);
       });
   }
+
+  private updateIsEditingState(value: boolean): void {
```
denysbutenko, line 154: It may help avoid unnecessary wrapper functions https://angular.dev/guide/signals#effects

undsoft (changes requested):
1. Let's try to stick to signal syntax.
2. We have the same problem with restricted roles where user can tab into a restricted button. It would be good to try to build a reusable solution (something like a directive) this time. Angular material has `cdkTrapFocus`, we need essentially the opposite of that. See if something already exists on npm or check source code for `cdkTrapFocus`, maybe it'll give you ideas.

PR #10201, opened 2024-06-17
NAS-129626 / 24.10 / Root User with FullAdmin Role Can't Restart & Shut Down the system

AlexKarpov98:
Testing: see ticket.

Now once you log in you should see all described buttons as enabled for the `root` user.
Screenshot 2024-06-17 at 13 27 14

FYI: It's not the case for the Dragonfish.


denysbutenko (approved):
I've seen this issue in other places 👍🏻

PR #10204, opened 2024-06-18
NAS-129584 / 24.10 / Update tab order for elements in search results

AlexKarpov98:
Testing: see ticket.

You should be able to focus on the search results first - then on the remove from the list button.

result ⬇️ 

https://github.com/truenas/webui/assets/22980553/51fb35df-14ee-4a7b-8471-fc3e01c7fa09


RehanY147 (changes requested):
We should also add `aria-label` that describes what the search items are and what the 'x' will do. Currently, using voiceover on macos, no indication about what the x does or what it even is that is in focus.

RehanY147 (changes requested):
The aria-label is added but voiceover isn't reading the label. When I focus on the cross icon with tab, it doesn't describe the label you've put. Please check.

RehanY147 (changes requested):
It works on Chrome. but doesn't work on Firefox.

RehanY147:
@AlexKarpov98 You can create a new ticket if you want. It doesn't have to be done in this ticket. I just mentioned it because working using tabs is mostly used by users who need voice over tools at the same time. And if you don't know what's is focused, you don't know if you want to click it or not etc.
Regarding your question, firefox just re-reads the search items again next to the x icon. Instead of reading the remove aria-label.

RehanY147 (approved):

PR #10205, opened 2024-06-18
NAS-129391 / 24.04.2 / Can't edit iSCSI Target on SCALE

AlexKarpov98:
Testing: see ticket.

Try to edit some field on a iSCSI Target except name, it should not block


undsoft (approved):
I can save the form now. 👍🏼

PR #10206, opened 2024-06-18
NAS-129582 / None / Enclosure architectural improvements

undsoft:
Notable changes:
* Renamed some components, because we had "view" referring to multiple concepts.
* SVG is now loaded via httpClient to let us use normal CSS.
* Introduced more components and components are now split by their level of abstraction. E.g. `EnclosureSvgComponent` is only responsible for rendering an svg and does not know or care about pools.
* Moved more items into enclosure state to better handle scenarios with multiple enclosures.

Testing:
Use `yarn ui mock-enclosure`. Select M50 and fill some slots. You can optionally also type M40 or M50 in expansion shelves.

Known issues:
- Pool colors are not properly implemented.
- Mock some slots in environment.ts does not work correctly with multiple enclosures.


RehanY147 (approved):
Everything works alright. Nice changes.

PR #10071, opened 2024-05-16
NAS-128980 / 24.10 / TrueCloud | Not all possible fields are shown to the user on the form and on the details card

AlexKarpov98:
Testing: see ticket & more fields to be available for the user

**Before:**
Screenshot 2024-05-16 at 12 38 58

**After:**
Screenshot 2024-05-16 at 12 39 20

**Before:**
Screenshot 2024-05-16 at 12 39 09

**After:**
Screenshot 2024-05-16 at 12 39 32

PR #10207, opened 2024-06-18
NAS-129598: Fix Cloud Sync Bandwidth Limit is not an integer

denysbutenko:
Manual backport of workaround from https://github.com/truenas/webui/pull/10071
The goal is to allow users to edit the cloud sync tasks.

There is another issue with conversion limits between `buildNormalizedFileSize` and `CloudCredentialService.getByte` that increases limits on every save. It'll be addressed [separately](https://ixsystems.atlassian.net/browse/NAS-129624).


undsoft (approved):
Form can be saved, even though the number changes every time.

PR #10213, opened 2024-06-19
NAS-129642 / 24.10 / Added more enclosure model SVGs

RehanY147:


undsoft (changes requested):
So, I've tried to test it with the following config and I'm running into errors:
```
  mockConfig: {
    enabled: true,
    controllerModel: 'F60',
    expansionModels: ["ES24F","ES24N","ES60G2","ES102G2","F100","F130"],
    scenario: MockStorageScenario.FillAllSlots,
  },
```

Also worth checking if number of slots with doc here https://docs.google.com/spreadsheets/d/1qhy_SMARhBZK4xNLDK1e9kv0uTmN3sI6CgcZvVpWc5M/edit?gid=0#gid=0. 
Something feels wrong when I try to compare number of slots I see vs spec.

RehanY147:
@undsoft The svg for ES60G2 had the drive cage numbering start from 1 when UI expects it to start from 0. Fixed it. Also, for both ES60 and ES102, mock utility gave them front slots when they are supposed to have top slots. Fixed that as well. It should work now.

src/app/pages/system/enclosure/components/pages/enclosure-page/enclosure-side/enclosure-svg/enclosure-svg.component.ts
```diff
@@ -87,17 +87,6 @@ export class EnclosureSvgComponent {
     });
   });
 
-  private highlightSelectedSlot = effect(() => {
```
undsoft, line 90: Why did this get removed?

undsoft (approved):
Something wrong with something.
If I have:
```
    enabled: true,
    controllerModel: 'M40',
    expansionModels: ['ES24F', 'ES60G2', 'ES24N', 'ES102G2', 'F60', 'F100', 'F130'],
```

I get the following image. Except this is not an M40 svg shown.
Снимок экрана 2024-06-20 в 11 22 08

PR #10218, opened 2024-06-19
NAS-129198 / 24.10 / Update enabled field after change

bvasilenko:
**Testing**

**Steps to Reproduce**

        
1. Navigate to Shares page

        
2. toggle share 'enabled' slider to 'disabled'

        
3. At this share click 'pencil' icon

**Expected Result**

right panel displays and 'enabled' checkbox is disabled

### DIFF
--- a/src/app/pages/sharing/components/shares-dashboard/smb-card/smb-card.component.ts
+++ b/src/app/pages/sharing/components/shares-dashboard/smb-card/smb-card.component.ts
@@ -6,10 +6,11 @@ import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { Store } from '@ngrx/store';
 import { TranslateService } from '@ngx-translate/core';
 import {
-  tap, map, filter, switchMap,
+  tap, map, filter, switchMap, BehaviorSubject,
 } from 'rxjs';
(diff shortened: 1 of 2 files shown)


src/app/pages/sharing/components/shares-dashboard/smb-card/smb-card.component.ts
```diff
@@ -36,27 +37,29 @@ import { selectService } from 'app/store/services/services.selectors';
 })
 export class SmbCardComponent implements OnInit {
   requiredRoles = [Role.SharingSmbWrite, Role.SharingWrite];
+  changedEnabled = new Map<number, boolean>([]);
```
denysbutenko, line 40: Let's imagine fixing every other table with this approach.
TBH, it looks like a workaround. We have a mechanism called `getSmbShares` to get fresh results on each change. The question is, why does it not work as expected?

denysbutenko (changes requested):

denysbutenko (approved):
Kapture 2024-06-25 at 16 11 11

It works well. However, I think `AppLoaderService.withLoader()` is doing its job.

bvasilenko:
backport

PR #10222, opened 2024-06-20
NAS-129630 / 24.10 / [TrueCloud] Make Name a required field

AlexKarpov98:
Code review is enough.


denysbutenko (approved):
Maybe we could prefill this field with auto-generated value?

PR #10224, opened 2024-06-20
NAS-129657 / 24.10 / Connect new enclosure UI to routing

undsoft:
Add several enclosures via mock-enclosure and see how url updates when you switch between them.


bvasilenko (approved):
Works as expected

PR #10226, opened 2024-06-20
NAS-129681 / 24.10 / Move more components to standalone

undsoft:
Makes more component standalone, moves some components from `core`, replaces one usage of EntityJobComponent.

Testing: regression testing.


denysbutenko (approved):
Huh, a huge one. This tab freezes the system :)

PR #10229, opened 2024-06-21
NAS-129640 / 24.10 / Newly added cloudsync task doesn't appear when added via Advanced

AlexKarpov98:
Testing: see ticket & try switching between wizard & advanced creation. You should see data updated eventually once you added new item.


RehanY147 (changes requested):
This is some kind of a race condition. Let's try to solve the race condition. I feel like adding another observable is not what we want from an architecture point of view. We want to return the observable that will be emitting the response of the slide-in as a return value of the `open` method on `ix-chained-slide-in.service`.

If you delete or comment out line# 99 on `ix-chained-slide-in.service.ts`, it starts to work.
`newMap.delete(id);` talking about this line.

PR #10237, opened 2024-06-24
NAS-129646 / 24.10 / Fix issues in DisksOverviewComponent

bvasilenko:
**Testing**
Code review should be enough


RehanY147 (changes requested):
I think we can make smaller components for better separation of concerns and code grouping here.

undsoft (approved):
Looks okay to me.

PR #10240, opened 2024-06-25
NAS-129649 / 24.10 / Add loading states to new enclosure

AlexKarpov98:
Testing: see ticket

Results: (delayed)

https://github.com/truenas/webui/assets/22980553/25fef1e1-23b1-4c3d-866a-518d1589f263


undsoft (changes requested):
Look is okay, but store needs to be removed out of EnclosureSvgComponent.
I couldn't get progress bar to appear without mocks in browser, even with throttling in inspector.

Switching between enclosures results in jumping, maybe we could something better there, for example by not showing skeleton in those scenarios and applying some transitions.


https://github.com/truenas/webui/assets/1481270/7489af35-d2a6-43a0-a753-509a032aba1c

I'm also open to the idea of animating models or slots in some way when they appear.

src/app/pages/system/enclosure/components/enclosure-side/enclosure-svg/enclosure-svg.component.ts
```diff
@@ -36,7 +48,7 @@ export class EnclosureSvgComponent implements OnDestroy {
   private keyDownListener: () => void;
   private clickListener: () => void;
 
-  protected svg = signal<SafeHtml | undefined>(undefined);
+  svg = signal<SafeHtml | undefined>(undefined);
```
undsoft, line 51: Marking something as public because it does not work in tests is a good indication that you shouldn't interact with it in the test.

Do this instead:
```ts
const svgCache = spectator.inject(SvgCacheService);
jest.spyOn(svgCache, 'loadSvg').mockReturnValue(of()); // Simulate long network request.
spectator.setInput('svgUrl', '/assets/m2.svg');
...
```

undsoft (approved):
Looks and feel is good, but there is still a flash of skeleton when switching between previously loaded svgs. (likely because of `delay(0)`/`setTimeout`)

Try adding animation to skeleton as well, so that it does not appear immediately.

https://github.com/truenas/webui/assets/1481270/7faea6c5-5216-47ba-9251-9f06ddf9fa99

RehanY147 (approved):

PR #10250, opened 2024-06-27
NAS-129732 / 24.10 / Truenas Cloud Snapshots listed in wrong chronological order

AlexKarpov98:
Testing: see ticket.


undsoft (approved):
Snapshots are in correct order now.

PR #10251, opened 2024-06-27
NAS-129716 / 24.10 / Search UI takes User to Wrong Category for Shutdown Scripts

AlexKarpov98:
Testing: see ticket.
It's recommended to add custom `id` attributes to the ADD buttons, if we don't add them we'll generate from hierarchy - and it's just `add` -> leading to duplicate ids.

Result:

https://github.com/truenas/webui/assets/22980553/4fe68d17-f756-4407-b48f-6bfb839e41c0


undsoft (approved):
👍🏼

PR #10254, opened 2024-06-27
NAS-129773 / 24.10 / Remove some flex layout usages

undsoft:
As said before flex-layout is no longer maintained and we shouldn't use it.

Testing: Test some of the changes places for regressions.


RehanY147 (changes requested):
Storage usage card looks different from how it is on `master`.

PR #10257, opened 2024-06-28
NAS-129788 / 24.10 / Cloud Backup | Snapshots list is not updating after Job Run

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/truenas/webui/assets/22980553/a751da85-340d-466d-8ace-7e875ed57e70


bvasilenko (approved):
Now it works as expected

PR #10264, opened 2024-07-01
NAS-129010 / 24.10 / Buttons are not aligned in Data Protection

AlexKarpov98:
Testing: see ticket.

result:
Screenshot 2024-07-01 at 11 28 18


bvasilenko (approved):
Buttons look well now!

PR #10268, opened 2024-07-02
NAS-129039 / 24.10 / Support Ticket doesn't include attached screenshots

AlexKarpov98:
Testing: see ticket.


RehanY147 (changes requested):
Instead of disabling the close button out right, I think we should instead at least allow to close the dialog by pressing the close button (x button top right) and show a confirmation dialog.

Furthermore, the problem here also is that we are not communicating to the user that ticket submission and file upload are separate processes. We should be doing that. When the user submits the ticket with attachments, there should be a loading dialog that says something like "Submitting ticket" and show progress for the job/call to websocket. Once the call completes and the file upload starts, the loading dialog should be switched to another dialog which shows a vertical list of all the files to be uploaded and show loading in front of all of them and show progress for those files as they are uploaded so the user understands what's happening.

AlexKarpov98:
Sound like a good idea, I may try one more solution and you will check it.

AlexKarpov98:
@RehanY147 - see my updated solution. I think it's the easiest one.
If we confirm what you say, I think we can open up a new ticket for that improvement.

RehanY147 (changes requested):
If you don't add the recommended dialogs for file upload, then I don't think we need to add the loading indicator. The fake progress bar on the original dialog already shows that something is loading.

My recommendation was that if we completely disallow closing the dialog, the user can skip the process by reloading their page anyway. But that way, we will lose control of the situation. Instead we should allow the user to click the cross button and show a confirmation dialog that warns the user that the process is incomplete.

AlexKarpov98:
> If you don't add the recommended dialogs for file upload, then I don't think we need to add the loading indicator. The fake progress bar on the original dialog already shows that something is loading.
> 
> My recommendation was that if we completely disallow closing the dialog, the user can skip the process by reloading their page anyway. But that way, we will lose control of the situation. Instead we should allow the user to click the cross button and show a confirmation dialog that warns the user that the process is incomplete.

@RehanY147 - this ticket is more about preventing user to close the form while it's loading.
I think "Loader" is the most suitable option for this.

Let's ask @undsoft as well

AlexKarpov98:
I think for the User it's easier to see just one LOADER, which means that his ticket is creating, he may not be interested is partial processes like 
1. Ticket is created 
2. Images are uploaded

Once user sees spinner, he knows - he needs to wait, and not care what's actually happening under the hood. That's my idea.

undsoft:
I'm okay with preventing dialog from being closed, but there is already a progress bar on top of the dialog. I don't think we need to add another one.

AlexKarpov98:
> I'm okay with preventing dialog from being closed, but there is already a progress bar on top of the dialog. I don't think we need to add another one.

@undsoft 
maybe we can remove fake progress bar and leave Spinner Loader -> which already prevents user to close the dialog?

In my previous commits I fully disabled user to close the dialog once it's loading (DisableClose = True & Disabled Close Button)

Which option may be better?

undsoft:
I'd rather we use fake loader instead of spinner.

AlexKarpov98:
@RehanY147 - now user is not able to close the dialog once form is loading (again).
Evgeny Approved - `I'm okay with preventing dialog from being closed`

src/app/modules/feedback/components/feedback-dialog/feedback-dialog.component.ts
```diff
@@ -46,15 +46,15 @@ export class FeedbackDialogComponent implements OnInit {
     this.loadFeedbackTypes();
   }
 
-  protected onIsLoadingChange(newValue: boolean): void {
+  onIsLoadingChange(newValue: boolean): void {
```
RehanY147, line 49: The param name could be better.

PR #10271, opened 2024-07-02
NAS-129686 / 24.10 / Fix SMB ACL form

bvasilenko:
**Summary**

1. when AD cache is empty (see ticket), when **Group** combobox does not contain any suitable values to choose from,
I've changed to code to add the ability to type in a custom value from keyboard . When user presses **Save**, every value is validated for being a valid existing entry in the Active Directory 

2. Middleware returns special type of AD entry `BOTH` , and when AD cache is empty (see ticket) it's not possible to distinguish a user from a group. 
   It was agreed that, it would be unreasonable to try to squeeze in last-second middleware update for 24.04.2.
   So, in such cases, the WebUI code was updated to make an attempt of detecting the type, otherwise it treats an AD entry as `GROUP`.

**Testing**

1. Log into WebUI
2. Setup the Active directory
4. Disable AD User / Group Cache and rebuild cache
5. Save the configuration
6. Create an SMB share
7. Edit the SMB ACL
8. Add a known to exist AD Group using keyboard entry, not picking any of options of the dropdown menu (for example `AD02\domain users`)
  and/or add a User entry by typing on the keyboard (for example type in `AD02\administrator`)
9. Save the ACL
10. Edit the SMB ACL
11. Observe the AD Group and/or User entry


denysbutenko (changes requested):
There were no errors when I typed unexisted user and clicked on Save. The loading state is never-ending. Let's improve error handling so users know about it.

image

Also, when I choose `Who: User` and type `AD02\administrator` when I edit share again, it becomes `Who: Group`. Is it expected?

image

undsoft (commented):
It's kinda weird that my user entry silently becomes a group entry. I understand that we don't have time to make middleware changes, but can't we keep showing/sending BOTH as a value?

undsoft (commented):
It works, but I have the following questions:
1. What does it mean to create a record with BOTH type? Can we only show this option on edit?
2. Can we call it "User or Group" or "Unknown"?

bvasilenko:
@undsoft 

> 1. What does it mean to create a record with BOTH type? Can we only show this option on edit?

We can't show this option on edit. `BOTH` type means middleware is not providing enough information for determining whether this a user or a group. As discussed with @anodos325 , middleware will introduce this capability in the upcoming releases.

> 2. Can we call it "User or Group" or "Unknown"?

Sure, I've updated the PR

src/app/pages/sharing/smb/smb-acl/smb-acl.component.html
```diff
@@ -48,6 +49,17 @@
                 formControlName="group"
                 [label]="'Group' | translate"
                 [provider]="groupProvider"
+                [allowCustomValue]="true"
+                [required]="true"
+              ></ix-combobox>
+
+              <!-- TODO: Delete after the backend has support for distinguishing user from group -->
```
denysbutenko, line 56: Is there a ticket number available?

denysbutenko (approved):
image

image
Steps:
1. Choose `AD02\administrator`
2. Replace slash with `/`
3. Click save
4. Replace slash back to `\`


image

Looks good

undsoft:
Okay, there are more problems with how this form works, but it works to a certain extent which is better than not working completely. I'm going to merge this PR and we can improve on it later.

PR #10275, opened 2024-07-03
NAS-129468 / 24.10 / Handle loading errors on datasets

denysbutenko:
image

For testing, you'll need to simulate network error for `pool.dataset.details` endpoint.


AlexKarpov98 (approved):
Errors are reported now as expected.

PR #10279, opened 2024-07-03
NAS-129003 / 24.10 / Add a warning when configuring the export SMB/NFS of root-level datasets

bvasilenko:
**Testing**

See ticket


AlexKarpov98 (changes requested):
Code looks fine, but I noticed few issues and you missed some thing.

Evgeny added comment:
Let’s add something reusable to both **Dataset form when SMB or Multiprotocol preset is selected** and to **SMB/NFS forms.** Show both on create and edit.

See the bold text ⬆️ 

Issues:
1. SMB create - cancel:

IF I cancel, I still see loading indicator and Save is disabled.

https://github.com/truenas/webui/assets/22980553/72a5eb66-b113-44c3-ba74-8a38b762d6cf

2. SMB create with root (first) entry mountpoint - error (and see that they are stacked, not only one error modal appears - there are 4 error modals)

https://github.com/truenas/webui/assets/22980553/a9c16364-220f-44cf-bfaa-e71850d4db9a

3. SMB Save error & multiple error modals:

https://github.com/truenas/webui/assets/22980553/3b89d386-515e-4335-a0e6-e4234003f824

4. You missed Evgeny's comment, `**Dataset form when SMB or Multiprotocol preset is selected**` -- here we should warn user as well.

bvasilenko:
> You missed Evgeny's comment, **Dataset form when SMB or Multiprotocol preset is selected** -- here we should warn user as well.

@AlexKarpov98 this was done intentionally. I realized that **Dataset form when SMB or Multiprotocol preset is selected** does not need this functionality. The situtation when this is applicable will never occur:

1. When adding: datesets can be added within the root pool, but not in the pool itself, as per ticket description: "for instance /mnt/tank (root-level dataset)." So user has no ability to add a dataset at the root-level.

2. Editing: SMB and NFS cannot be configured at all.

src/app/services/dataset-service/dataset.service.ts
```diff
@@ -51,4 +57,12 @@ export class DatasetService {
       );
     };
   }
+
+  rootLevelDatasetWarning(path: string): Observable<boolean> {
+    const isRootLevelDataset = !path.replace(`${mntPath}/`, '').includes('/');
```
undsoft, line 62: There is an `isRootDataset` utility function that you can use.

undsoft (changes requested):
1. We need to improve text to explain why it's a bad idea to share a root dataset. Please ask Andrew in the ticket for description.
2. I don't think it's worth showing this warning on edit if path has not changed. We should only show it if it's a new share or if I'm updating path.

undsoft (approved):
Warning works correctly. I've updated some strings.

PR #10281, opened 2024-07-04
NAS-127486 / 24.10 / Accessibility - Buttons without labels for Screen reader users in TrueNAS Scale

AlexKarpov98:
Testing: see ticket. 
Accessibility Improvements / Improvements for Screen Reader Users
.


denysbutenko (approved):
Why sometimes `[aria-label]` and `[attr.aria-label]`?

LGTM

AlexKarpov98:
> Why sometimes `[aria-label]` and `[attr.aria-label]`?
> 
> LGTM

some of angular components needs `[aria-label]` prop to be passed, if I try with `[attr.aria-label]` it's just not showing up eventually.

PR #10286, opened 2024-07-04
NAS-129913 / 24.10 / Control flow syntax migration 1

undsoft:
Code review is enough.


bvasilenko (approved):
New syntax looks good!

PR #10287, opened 2024-07-04
NAS-127520 / 24.10 / SMB validator: do not prevent user from continuing in one particular case

bvasilenko:
### **Testing**

**Replication steps:**

1. Make sure there are no local or domain users with SMB enabled. (See ticket for details on this step.)

   Or as alternative, add mocks to the `SmbValidationService` to simulate "TrueNAS server must be joined to Active Directory or have at least one local SMB user before creating an SMB share" error.

3. Go to Shares > Add SMB

4. Select a Path

**Expected result:** 

A non-intrusive warning pop-up with the text "TrueNAS server must be joined to Active Directory or have at least one local SMB user before creating an SMB share",

which does not block the user from continuing.

### DIFF
--- a/src/app/pages/sharing/smb/smb-form/smb-validator.service.ts
+++ b/src/app/pages/sharing/smb/smb-form/smb-validator.service.ts
@@ -2,26 +2,31 @@ import { Injectable } from '@angular/core';
 import {
   AbstractControl, ValidationErrors,
 } from '@angular/forms';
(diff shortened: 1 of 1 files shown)


bvasilenko:
@RehanY147 To reproduce the bug, add the following code to **line 41** in the **src/app/pages/sharing/smb/smb-form/smb-validator.service.ts** file:

```ts
map(() => ({ reason: this.noSmbUsersError })),
```

image

RehanY147 (approved):

PR #10289, opened 2024-07-05
NAS-129535 / 24.10 / Uptime time zone does not update with time zone localization change

AlexKarpov98:
Testing: see ticket.


src/app/pages/dashboard-old/components/widget-sys-info/widget-sys-info.component.ts
```diff
@@ -134,6 +134,11 @@ export class WidgetSysInfoComponent extends WidgetComponent implements OnInit, O
       });
   }
 
+  getDateTimeByTimezone(timeZone: string): number {
```
bvasilenko, line 137: I noticed that this function is not used anywhere else. I'd suggested to reuse existing 
 `<ix-date` component here, if possible

bvasilenko (changes requested):

src/app/pages/dashboard-old/components/widget-sys-info/widget-sys-info.component.ts
AlexKarpov98: > I noticed that this function is not used anywhere else. I'd suggested to reuse existing `<ix-date` component here, if possible

I think we can't use `ix-date` here since we need custom text, not just plain date.
currently we show: `Uptime: 26 minutes 11 seconds as of 02:15 `

AlexKarpov98:
@bvasilenko 

I think we can't use `ix-date` here since we need custom text, not just plain date.
currently we show: `Uptime: 26 minutes 11 seconds as of 02:15 `

bvasilenko (approved):
Ok, works as expected

PR #10293, opened 2024-07-05
NAS-129913 / 24.10 / Control syntax migration 2

undsoft:
Code review is enough.


AlexKarpov98 (approved):
looks like a new syntax 😄 👍 
I used to see an old one still, huh

PR #10296, opened 2024-07-08
NAS-128951 / 24.10 / If you edit VM to use Custom CPU, then try to switch back to host passthrough it gives error

AlexKarpov98:
Testing: see ticket.


denysbutenko (approved):
Switching CPU for a VM works well. LGTM

PR #10300, opened 2024-07-08
NAS-128858 / 24.10 / Fix & improve http error handler

bvasilenko:
**Summary**

This PR contains 2 changes:

1. The error modal window depicted in the ticket (`Error(0)`) was shown due to the fact that the handler received not an HTTP error, but a Job error from the `websocket`. I've fixed this handler.

2. Adjust HTTP error handler to show error title and message from HTTP errors, which was preventing from determining the root cause of the issue

**Testing**

On any box, for example on http://10.220.16.103

1. When user clicks **Data Protection > Cloud Sync Tasks > Green "Success" Pill**

3. At the error modal window, user clicks **Download Logs**
   - **Important:** you will need to adjust the code to simulate downloading HTTP error, similar to one shown below


undsoft (approved):
Changes look reasonable, thanks.

PR #10301, opened 2024-07-09
NAS-129643 / 24.10 / Pool tint in new enclosure

undsoft:
Mocking utility will now create multiple pools and vdevs.
Pools should be correctly colored on both rackmount and mini enclosures.

Note: colors may change in the future.


RehanY147 (approved):
Looks good. When choosing how many vdevs and pools should be added, the text should be fixed.

PR #10310, opened 2024-07-11
NAS-129911 / 24.10 / Added more SVGs to new Enclosure module

RehanY147:


undsoft (commented):
So I added `expansionModels: Object.values(EnclosureModel),` to `environment.ts` and some models that were added don't seem to have clickable slots in them.

undsoft (changes requested):
Something's wrong with MINI-3.0-X and XL.

undsoft (approved):
Looks good, thanks.

PR #10313, opened 2024-07-12
NAS-129880 / 24.10 / Incorrect id sent when editing a group

undsoft:
Testing: see ticket.


bvasilenko (approved):
Works perfect! Tested with my staging Active Directory environment

PR #10319, opened 2024-07-16
NAS-129928 / 24.10 / Better fallback for missing enclosures

undsoft:
Testing: there are new "fake" enclosure models in `yarn ui mock-enclosure`. You can use it to test how UI responds to those.


RehanY147 (approved):
Works OK

PR #10321, opened 2024-07-17
NAS-130067 / 24.10 / Introduce OrNotAvailable pipe

undsoft:
And additional control flow syntax migration.

Testing: code review should be sufficient.


AlexKarpov98 (approved):
Looks and works good 👍

PR #10322, opened 2024-07-17
NAS-130077: Add PR template

undsoft:
There is no way to check this until this is merged, so I'll merge as is.


denysbutenko (commented):
Does it possible to add doc_impact label over bugclerk? Or maybe we could add this label here and bugclerk sync with ticket?

undsoft:
Those ideas are in the works too by solutions team.

PR #10325, opened 2024-07-18
NAS-130050 / 24.10 / Adds higlight styles for vdev disks

RehanY147:
**Changes:**

Added a new class to the `overlay-rect` that adds the same highlight styles as the selected slot but with a white border color to all other disks in the same vdev as the currently selected slot.

**Testing:**

Mock some enclosures and test by selecting different slots. Other slots that are also part of the same vdev should also get the highlight styles but with a white border.

### DIFF
--- a/src/app/pages/system/enclosure/components/enclosure-side/enclosure-svg/enclosure-svg.component.ts
+++ b/src/app/pages/system/enclosure/components/enclosure-side/enclosure-svg/enclosure-svg.component.ts
@@ -203,7 +203,7 @@ export class EnclosureSvgComponent implements OnDestroy {
   private handleOverlayKeyNavigation(event: KeyboardEvent, slot: DashboardEnclosureSlot): void {
     switch (event.key) {
       case 'Enter':
-        this.selectedSlot.set(slot);
+        this.slotSelected(slot);
         break;
       case 'ArrowUp':
       case 'ArrowLeft':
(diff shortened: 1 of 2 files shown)


AlexKarpov98 (commented):
It works, but visually it's hard to understand what is highlighted as well.
White border is not really visible.

maybe we can use `var(--red)` or `var(--green)`?

src/app/pages/system/enclosure/components/enclosure-side/enclosure-svg/enclosure-svg.component.ts
```diff
@@ -175,7 +233,28 @@ export class EnclosureSvgComponent implements OnDestroy {
   private addInteractionListeners(slot: DashboardEnclosureSlot): void {
     const overlay = this.overlayRects[slot.drive_bay_number];
 
-    this.clickListener = this.renderer.listen(overlay, 'click', () => this.selectedSlot.set(slot));
+    this.clickListener = this.renderer.listen(overlay, 'click', () => {
```
undsoft, line 236: Time to move to a separate arrow method.

src/app/pages/system/enclosure/components/enclosure-side/enclosure-svg/enclosure-svg.component.ts
```diff
@@ -113,16 +122,65 @@ export class EnclosureSvgComponent implements OnDestroy {
...
+      }
+    }
+
+    for (const vdevDiskDriveBayNumber of selectedVdevDisksDriveBayNumbers) {
+      const vdevDiskSlotOverlay = this.overlayRects[vdevDiskDriveBayNumber];
+      this.renderer.addClass(vdevDiskSlotOverlay, 'selected-vdev-disk');
+    }
+    for (const notVdevDiskDriveBayNumber of notSelectedVdevDisksDriveBayNumbers) {
+      const vdevDiskSlotOverlay = this.overlayRects[notVdevDiskDriveBayNumber];
+      this.renderer.addClass(vdevDiskSlotOverlay, 'not-selected-vdev-disk');
```
undsoft, line 180: Feels like you can simplify this for example by not just updating overlay classes in the first for loop, or by iterating over `overlayRects`.

undsoft (changes requested):
Overall approach was approved.
1. When in vdev mode, I would keep showing a bit more color. 
2. I think we should also consider letting user switch to another vdev directly by clicking on another slot, without user having to click twice.
3. Clicking on the same slot twice should exit vdev mode.
4. Behaviour needs to be consistent whether I'm using mouse or keyboard.

undsoft (approved):
Looks good. Let's add some tests.

PR #10332, opened 2024-07-19
NAS-130048 / 24.10 / Implement identity light functionality in new enclosure

undsoft:
**Changes:**

Identify light buttons added to new enclosure. See ticket for detailed description.

**Testing:**

Middleware is not fully ready yet. You need to replace call in `IdentifyLightComponent` line 42 with `of(true)`.

### Downstream

|Affects         |Reasoning
|Documentation   | New functionality in UI.


AlexKarpov98 (approved):
It works good 👍 
Please fix tests.

Maybe we can add some indication on the svg graphics when some slot/drive is identified? Currently it's only below the svg graphics. Maybe some sort of lamp like below.


https://github.com/user-attachments/assets/fba697dc-8c78-4f37-a3be-0035fbf79819

undsoft:
> Maybe we can add some indication on the svg graphics when some slot/drive is identified? Currently it's only below the svg graphics. Maybe some sort of lamp like below.

I'll add it as an improvement idea.

PR #10338, opened 2024-07-22
NAS-129775 / 24.10 / Unified logic to get the product image url across different modules

RehanY147:
**Changes:**

Removed logic to get the image url for `widget-sys-info` and `support-card` components. Placed logic in `system-general` service with a method. Also fixes some tests regarding `system.build_time`

**Testing:**

Compare to `master` and check that visually no difference in `support-card` under `System -> General` and `widget-sys-info` on the new main dashboard. Also code review.


denysbutenko (changes requested):
This looks similar to https://github.com/truenas/webui/blob/master/src/app/pages/dashboard/widgets/system/common/widget-sys-info.utils.ts

Please check what can be reused or adapted to use in both cases (support-card and new sys-info widgets).

The old widget-sys-info is unnecessary to change since it will be removed soon.

src/app/store/system-info/system-info.effects.ts
```diff
@@ -83,20 +81,6 @@ export class SystemInfoEffects {
     }),
   ));
 
-  loadSystemBuildTime = createEffect(() => this.actions$.pipe(
```
denysbutenko, line 86: Also check leftovers in actions and reducers

src/app/modules/layout/components/admin-layout/admin-layout.component.ts
```diff
@@ -38,7 +39,7 @@ export class AdminLayoutComponent implements OnInit, AfterViewInit, OnDestroy {
   readonly isAlertPanelOpen$ = this.store$.select(selectIsAlertPanelOpen);
   readonly hasConsoleFooter$ = this.store$.select(selectHasConsoleFooter);
   readonly productType$ = this.sysGeneralService.getProductType$;
-  readonly copyrightYear$ = this.sysGeneralService.getCopyrightYear$;
+  readonly copyrightYear = environment.buildYear;
```
denysbutenko, line 42: There is a `copyright-line` component that relies on `selectBuildTime`. It needs to be updated.

RehanY147: @denysbutenko That component is already fixed. It uses `environment.buildYear` to decide build year.

RehanY147:
@denysbutenko Also test that the buildYear has been extracted into the store as a selector so it's easier to mock in tests.

src/app/pages/system/general-settings/support/support-card/support-card.component.ts
```diff
@@ -46,11 +48,11 @@ export class SupportCardComponent implements OnInit {
   protected readonly searchableElements = supportCardElements;
 
   isProduction: boolean;
-  productImage = 'ix-original-cropped.png';
-  isProductImageRack = false;
+  isWiderImage = false;
```
denysbutenko, line 51: Looks like can be computed property

denysbutenko (changes requested):
Have a difference on non-iX hardware (VM instance) 

- support card displays `iX` logo
- system info widgets display `scale` logo

denysbutenko (changes requested):
Mentioned issue with logo difference remains

RehanY147:
@denysbutenko That's the behavior in `master` as well. This PR doesn't aim to change `master` behavior. Only refactor.

denysbutenko:
Looks good then. Create a ticket to fix it later please

denysbutenko (approved):

PR #10343, opened 2024-07-22
NAS-130087 / 24.10 / EE: Access Key and Secret Access Key of Cloud Credentials do not display unless clicking the Help icon

AlexKarpov98:
**Testing:**

See ticket

### DIFF
--- a/src/app/pages/credentials/backup-credentials/cloud-credentials-form/cloud-credentials-form.component.spec.ts
+++ b/src/app/pages/credentials/backup-credentials/cloud-credentials-form/cloud-credentials-form.component.spec.ts
@@ -302,6 +302,8 @@ describe('CloudCredentialsFormComponent', () => {
     });
 
     it('shows existing values when form is opened for edit', async () => {
+      spectator.component.setCredentialsForEdit();
+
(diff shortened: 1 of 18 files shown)


undsoft (changes requested):
Added code conflicts with previously added code for the same purpose in `setCredentialsForEdit`.
It looks like a change detection issue.
I would try figuring out what happened and fix original code rather than add a workaround.

AlexKarpov98:
@undsoft - It's does not seem to be a change detection issue.
I tried all possible ways and only - `patchValue` worked for me. 🤷‍♂️

undsoft (changes requested):
`form` is not intended to be manipulated on directly.
This PR https://github.com/truenas/webui/pull/7534/files added `getFormSetter$` that you can use to set form changes. It already has `patchValue`. Adding `this.cdr.detectChanges();` after it solves the issue.
I'm not sure why `getFormSetter$` was added exactly, but since it was added, we should keep one way of doing things.

There also seem to be a lot of similar code across provider classes that you may be able to extract to base class.

undsoft (approved):
Looks good, thanks.

PR #10350, opened 2024-07-24
NAS-130209: Add half size for active/passive system node widget

AlexKarpov98:
Screenshot 2024-07-24 at 15 17 40


src/app/pages/dashboard/widgets/system/common/widget-sys-info.scss
```diff
@@ -7,6 +7,44 @@
   height: 100%;
 }
 
+:host ::ng-deep {
```
undsoft, line 10: Seems like it works just fine without `::ng-deep`.

undsoft (approved):
Looks good overall.
Let's move blue button a bit below.

src/app/pages/dashboard/widgets/system/common/widget-sys-info.scss
AlexKarpov98: now it's required, I updated `.product-image-placeholder {` which requires it.

PR #10355, opened 2024-07-25
NAS-130246: Make login page more resilient

undsoft:
**Changes:**

Login page should no longer fail if there are issues with certain calls that are optional to login.

**Testing:**

Code review.


denysbutenko (commented):
Perhaps `system.advanced.login_banner` also can be optional. Currently fails on DF

PR #10356, opened 2024-07-25
NAS-130131 / 24.10 / Fix slot ordering on some MINIs

undsoft:
**Changes:**

Corrects order of slots on some MINI models.

**Testing:**

Check MINI models.


undsoft:
Will fix as part of another PR.

PR #10367, opened 2024-07-28
NAS-127001 / 24.10 / Fix resort translations action for external contributors

undsoft:
**Changes:**

This may fix resort action for our translations.

**Testing:**

Code review only. We'll know if it works once this is merged.

### DIFF
--- a/.github/workflows/resort-translations.yml
+++ b/.github/workflows/resort-translations.yml
@@ -1,0 +1,33 @@
+name: Resort translations
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'src/assets/i18n/**'
+
+jobs:
+  resort-strings:
+    name: Resort strings and commit if translations were changed
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/resort-translations.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Validate Translations
+        run: yarn run validate-translations
+      - name: Commit and Push Changes
+        run: |
+          git config --global user.name 'bugclerk'
+          git config --global user.email 'bugclerk@ixsystems.com'
+          git pull origin ${{ github.head_ref }} --no-edit
+          git add src/assets/i18n
+          if ! git diff --quiet --cached; then
+            git commit -m "NAS-XXXXXX: Auto resort translation strings"
```
denysbutenko, line 29: Curios if we can use head_ref to get ticket number for this

denysbutenko (approved):
Looks good to me

.github/workflows/resort-translations.yml
undsoft: It's a good idea, but maybe later. Parsing strings won't be easy.

PR #10375, opened 2024-07-30
NAS-130322 / 24.10 / Fix null bug

RehanY147:
**Changes:**
Fixes a minor undefined/null JS bug on the pool creation wizard

**Testing:**

On master, when adding a new vdev to an existing pool, you can see the described bug in the console and can't submit the form even though form is valid.

### DIFF
--- a/src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/generate-vdevs.service.ts
+++ b/src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/generate-vdevs.service.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import _ from 'lodash';
+import { isEmpty } from 'lodash';
 import { VdevType } from 'app/enums/v-dev-type.enum';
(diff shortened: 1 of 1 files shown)


src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/generate-vdevs.service.ts
```diff
@@ -86,7 +86,7 @@ export class GenerateVdevsService {
         return a.enclosure ? -1 : 1;
       }
 
-      return a.enclosure.id.localeCompare(b.enclosure.id);
+      return a.enclosure?.id?.localeCompare(b.enclosure.id);
```
undsoft, line 89: Isn't there a check for existence couple of lines above?

RehanY147:
@undsoft In this case, enclosure is truthy but an empty object with no attributes/properties. So, `!a.enclosure` check was not enough. Updated the check with `isEmpty(a.enclosure)` which returns true for `null`, `undefined` or `{}`

AlexKarpov98 (approved):

PR #10379, opened 2024-07-31
NAS-130217 / 24.10 / Fixing rear views on some enclosures

undsoft:
**Changes:**

- Added more rear enclosure views.
- Re-indexed svgs to match middleware responses.

**Testing:**

Test enclosure UI.


bvasilenko (approved):
Rear view on enclosures works good

PR #10380, opened 2024-07-31
NAS-130249 / 24.10 / Adjust blinking animation for search results

AlexKarpov98:
**Changes:**
Search result highlight effect is changed.

**Testing:**
Try using global search.

UPD*:

https://github.com/user-attachments/assets/5f962fa5-75d4-46a0-b5b0-19ee4d58fa1a

### Downstream

|Affects         |Reasoning
|Documentation   | Search result highlight effect is changed.


undsoft:
Comment from Patrick:

> if it can get closer to completely invisible in its off state then i'm sold.

undsoft (approved):
It looks good. Is primary button suppose to change color to default while it's being animated?

PR #10383, opened 2024-07-31
NAS-129910 / 24.10 / Enclosure various improvements

bvasilenko:
**Testing:**

Navigate to **Enclosure** page. And please see ticket for the list of expected fixes


undsoft (changes requested):
I would suggest adding 

```
expansionModels: Object.values(EnclosureModel),
```

to environment ts and checking other models.

I've noticed at least the following issues:
1. ES60 doesn't fit the screen both in main view and on enclosure switcher.
2. Enclosure selector doesn't fit the screen.
3. Tablet view doesn't look too good with selector being too tall, legends not being visible. 
Снимок экрана 2024-08-01 в 09 06 11

4. On some resolutions there are no margins between enclosure image and enclosing box. 
Снимок экрана 2024-08-01 в 09 08 01

AlexKarpov98 (commented):
Agree with Evgeny.

AlexKarpov98 (approved):

PR #10384, opened 2024-07-31
NAS-130055 / 24.10 / Updates catalog management & removes container image pages

RehanY147:
**Changes:**

Updates catalog management. Also removes routes for Container Image settings components and removes all mentions as the API endpoint for those settings is going to change and until the new API is ready, we don't want to show broken UI. I've added TODO comments in each place where we should remember to add back stuff.

**Testing:**

Test catalog settings under the "Settings" button on the installed apps page. Look for "Manage Catalog" in the settings drop down (top right).

### Downstream

|Affects         |Reasoning
|Documentation   | Adds catalog settings (preferred trains) on the installed apps page, removes container images page


bvasilenko (commented):
I haven't noticed any issues yet,

@RehanY147 please clarify, do we expect the **Manage Container Images** page to show some contents?

image

PR #10385, opened 2024-08-01
NAS-129023 / 24.10 / The Time Zone applied to the Periodic Snapshot Tasks UI/Webpages are not consistent

AlexKarpov98:
**Changes:**

Changed tooltip logic.
Evgeny approved:
If we are talking about a tooltip and machine time matches browser time you can just show a date. When it doesn’t you can show both:
```
Machine time: <date>
Browser time: <date>
```

**Testing:**
See it.
Screenshot 2024-07-31 at 14 54 45
Try your timezone and different one.

### Downstream

|Affects         |Reasoning
|Documentation   | UI Updates, Tooltip content update for specific case


denysbutenko (approved):
image

Looks good

PR #10386, opened 2024-08-01
NAS-130349 / 24.10 / Animation glitch when switching between enclosures with the same svg

AlexKarpov98:
Testing: see ticket.
_We need a unique url for each enclosure._


RehanY147 (approved):
The animation issue seems to have been resolved.

PR #10394, opened 2024-08-03
NAS-130226 / 24.10 / Improves App installation, update, deletion

RehanY147:
**Changes:**

Changes interfaces and improves API calls for app installation, update and deletion

**Testing:**

Test general apps functionality. App installation, editing, upgrade should work. Also test rollback function.

### Downstream

|Affects         |Reasoning
|Documentation   |


undsoft (changes requested):
1. Tried to install `plex` on `stable` train. The installation form seems to have an infinite loop on it, because my browser tab freezes.
2. Call to `app.similar` fails with an error.

undsoft (approved):
I was able to install an app, albeit slowly.

PR #10395, opened 2024-08-05
NAS-130259 / 24.10 / Ensure alphabetical order of shares on Shares dashboard

denysbutenko:
**Changes:**

1. Fix sort on `SMB`, `NFS` and `iSCSI` cards
2. Sync behavior between mentioned cards
3. Use `ix-table-pager-show-more` component for `View All` button
4. Update `ix-table-pager-show-more` to use signals
5. Add router support in `ix-table-pager-show-more`
6. Minor cleanup

**Testing:**

Go to the Sharing dashboard. Create 5 or more shares for each card.

### Downstream

|Affects         |Reasoning
|Documentation   | Minor UI changes for `View All` button

### DIFF
--- a/src/app/pages/sharing/components/shares-dashboard/iscsi-card/iscsi-card.component.ts
+++ b/src/app/pages/sharing/components/shares-dashboard/iscsi-card/iscsi-card.component.ts
@@ -123,8 +123,9 @@ export class IscsiCardComponent implements OnInit {
 
   setDefaultSort(): void {
     this.dataProvider.setSorting({
-      active: 0, // index column
+      active: 0,
       direction: SortDirection.Asc,
+      propertyName: 'name',
     });
   }
 }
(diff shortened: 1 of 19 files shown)


src/app/pages/sharing/components/shares-dashboard/iscsi-card/iscsi-card.component.ts
```diff
@@ -118,15 +113,18 @@ export class IscsiCardComponent implements OnInit {
...
         this.dialogService.error(this.errorHandler.parseError(err));
       },
     });
   }
 
-  private getIscsiTargets(): void {
-    this.dataProvider.load();
+  setDefaultSort(): void {
+    this.dataProvider.setSorting({
+      active: 0, // index column
```
RehanY147, line 126: Instead of a comment, it would be better to create a variable and use that instead of just `0`. Like `let activeColumnIndex = 0`

RehanY147 (changes requested):
The sorting isn't applied by default in the smb card. `pool` comes before `ds` it doesn't change if I change sorting from Asc to Desc in the code.

AlexKarpov98 (approved):
Works good 👍 

Please fix SBM list default sorting as well.
See:

https://github.com/user-attachments/assets/421673c2-7d90-4403-908b-3875b4b44d6e

denysbutenko:
Thanks for review, @AlexKarpov98. I have fixed default sorting on list pages for SMB, NFS and iSCSI

RehanY147 (approved):

PR #10396, opened 2024-08-05
NAS-129364 / 24.10 / "There are unapplied network interface changes" window is displayed incorrectly

AlexKarpov98:
**Changes:**

New warning dialog is added to inform user that we have restored the previous network configuration to ensure continued connectivity.

**Testing:**
See ticket.
Try updating network interface Aliases with `9.9.9.9` for example.
You can set 10 seconds to wait less time.

Result:
Screenshot 2024-08-05 at 12 05 09

### Downstream

|Affects         |Reasoning
|Documentation   | New dialog is added in case of `Network Reconnection Issue`


denysbutenko (changes requested):
On nightly HA VM (Failover Disabled). 

When I change the address in `IP Address (This Controller)*,` the `Test Changes` button appears, and I press it. A few moments later, I saw the sign-in page and log in as usual. There were no popups or `test changes` section present.

When I change two fields `IP Address (This Controller)*` and `IP Address (TrueNAS Controller 2)*` it applies without testing changes and connectivity with a system is lost.

AlexKarpov98:
@denysbutenko - for me it seems like a middleware issue. (only for HA system, Vladimir Vinogradenko approved)

Since after test changes we are just locked and no way back.
It only worked fine when i updated all the IP fields to be 1.1.1.1
And then I’ve seen TEST CHANGES as well and IP’s were rolled back and I showed up a message (Network Reconnection Issue)

Vladimir Vinogradenko 👇  confirmed it as well (but it's only an issue for HA system and specific cases @denysbutenko described)
`Hi, thank you for this information, we’ll fix it`
`Looks like a middleware issue, yes`

I tested with `non HA` system - and it works fine there. There we can only set one `IP Address`
See `non HA` system in action ⬇️ 

https://github.com/user-attachments/assets/95e206e9-cbf2-4665-8582-ab4ae6ac6e23

denysbutenko (approved):
Looks good

PR #10397, opened 2024-08-05
NAS-130412 / 24.10 / Check smart supports before showing Run Manual Test button

denysbutenko:
**Changes:**

Fix TODO. Add check if disk supports smart. Update tests.

**Testing:**

Open Devices page and check if `Run Manual Test` button appears correctly.

### DIFF
--- a/src/app/pages/storage/modules/devices/components/smart-info-card/smart-info-card.component.spec.ts
+++ b/src/app/pages/storage/modules/devices/components/smart-info-card/smart-info-card.component.spec.ts
@@ -181,7 +181,7 @@ describe('SmartInfoCardComponent', () => {
     });
   });
 
-  it('checks if disk do not supports SMART', () => {
+  it('does not show Run Manual Test when disks does not support SMART tests', () => {
     spectator.setInput('hasSmartTestSupport', false);
 
     const runTestButton = spectator.query(byText('Run Manual Test'));
--- a/src/app/pages/storage/modules/devices/components/smart-info-card/smart-info-card.component.spec.ts
+++ b/src/app/pages/storage/modules/devices/components/smart-info-card/smart-info-card.component.spec.ts
@@ -67,6 +67,10 @@ describe('SmartInfoCardComponent', () => {
           { id: 1 },
           { id: 2 },
         ] as SmartTestTask[]),
+        mockCall('disk.get_instance', {
+          ...disk,
+          supports_smart: true,
+        } as Disk),
       ]),
       mockProvider(MatDialog, {
         open: jest.fn(() => ({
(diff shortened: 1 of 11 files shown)


denysbutenko:
> Can we update `loadNodes` in `DevicesStore` instead to load extra field?

Getting an error from `disk.query`:

```[EFAULT] `supports_smart` cannot be queried if disk count is greater than 1```

Or did you mean to move getting data from card component into store?

**Review by undsoft — COMMENTED** [review]
No, that's what I meant. Just thought we could avoid making another call.
We seem to be using different method to determine whether Manual test button is available on Devices page and on Disks page.
I've asked Vladimir about it here: https://ixsystems.atlassian.net/browse/NAS-116969

**Review by undsoft — CHANGES_REQUESTED** [review]
After checking with Vladimir, it seems that `supports_smart` in `disk.get_instance` means that disk supports SMART monitoring in general and `smart.test.disk_choices` means that disk may support SMART tests. 

Therefore using `supports_smart` in here to show/hide test button is not correct.

File: src/app/pages/storage/modules/devices/stores/devices-store.service.ts [hunk 1]
```diff
@@ -114,6 +117,18 @@ export class DevicesStore extends ComponentStore<DevicesState> {
     );
   });
 
+  readonly loadDisksWithSmartSupport = this.effect((triggers$: Observable<void>) => {
+    return triggers$.pipe(
+      switchMap(() => {
+        return this.ws.call('smart.test.disk_choices').pipe(
+          tap((disksWithSmartSupport) => {
+            this.patchState({ disksWithSmartSupport: Object.values(disksWithSmartSupport) });
```
undsoft, line 125: Would be good to keep naming more specific. Here for example, it's not just disks with smart support, it's disks that support running a smart test.

src/app/pages/storage/modules/devices/stores/devices-store.service.ts
```diff
@@ -36,6 +38,7 @@ export class DevicesStore extends ComponentStore<DevicesState> {
   readonly error$ = this.select((state) => state.error);
   readonly nodes$ = this.select((state) => state.nodes);
   readonly diskDictionary$ = this.select((state) => state.diskDictionary);
+  readonly diskWithSmartSupport$ = this.select((state) => state.disksWithSmartSupport);
```
undsoft, line 41: disk => disks

src/app/pages/storage/modules/devices/components/smart-info-card/smart-info-card.component.spec.ts
```diff
@@ -179,5 +181,10 @@ describe('SmartInfoCardComponent', () => {
     });
   });
 
-  // TODO: handle cases when SMART is not supported
+  it('checks if disk do not supports SMART', () => {
```
undsoft, line 184: "does not show Run Manual Test when disks does not support smart tests"

undsoft (approved):
Looks good.

PR #10403, opened 2024-08-06
NAS-130389 / 24.10 / Hide some search items when needed

bvasilenko:
**Changes:**

The idea is to manually hide search results in UiSearchProvider (similar to roles), adding, for example, fields hideWhenNotVm, hideWhenNotFailover, hideWhenNotEnclosure to the configuration

**Testing:**

- any non-HA box - license not technically allowing to see failover screen. When user types the term “failover” in the search bar, **Expected result:** user could not gain access to the failover screen.

- http://10.220.16.103 - The license that does not allow seeing the virtualization screen. When user types the term “VM” in the search bar, **Expected result:** user should not gain access to the VM screens.


AlexKarpov98 (approved):
Nice solution, I as well thought about something similar. 👍

src/app/modules/global-search/enums/global-search-visible-token.enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum GlobalSearchVisibleToken {
+  Failover = 'FAILOVER',
+  Enclosure = 'ENCLOSURE',
+  Vms = 'VMS',
```
AlexKarpov98, line 4: Maybe we need 'Apps' support here as well?

Screenshot 2024-08-07 at 10 34 07

PR #10408, opened 2024-08-07
NAS-129590 / 24.10 / Read Only Admin User App Edit screen is blank

AlexKarpov98:
Read ticket details:
_When logged on as a Read-Ony Admin User and I select Edit for an App the edit window opens but it is blank. If the User isn’t supposed to access it the Edit button should not be active and should have a lock icon._

Code review is enough.


undsoft (commented):
Are there calls that prevent edit page from rendering? I'd rather show it if possible because there may be values that are there that are not shown anywhere else.

AlexKarpov98:
> Are there calls that prevent edit page from rendering? I'd rather show it if possible because there may be values that are there that are not shown anywhere else.

@undsoft 
Actually the only way to get to the `EDIT` app page is through that button.
I think middleware requests are blocked for readonly user that's why we see empty screen.

undsoft:
And previously it was possible to access the edit page (as intended). We need to check what requests fail if any now.

AlexKarpov98:
This request works fine for root user:
```
  getApp(name: string): Observable<App[]> {
    return this.ws.call('app.query', [[['name', '=', name]], {
      extra: { include_chart_schema: true, history: true },
    }]);
  }
```

But is not working for readonly user 

currently there is error:
`
OROperator failed to dump all schemas: [Error('query_result', 'Item#0 is not valid per list types: [EINVAL] app_query.active_workloads.container_details: Item#0 is not valid per list types: [EINVAL] container_detail.id: Field was not expected\n\n[EINVAL] app_query.active_workloads.images: Field was not expected\n'), Error('app_query', 'A dict was expected'), Error('count', 'Not an integer'), Error('app_query', 'A dict was expected')]
`
I mocked data for readonly user to make changes, this is final result 👇 

Screenshot 2024-08-08 at 13 58 55

undsoft:
Then ticket needs to be reassigned to middleware with the comment and this PR closed.

undsoft:
Yeah, we can keep it. Can't check it right now, since middleware is broken.

undsoft (approved):
Can be merged now.

PR #10410, opened 2024-08-08
NAS-130470 / 24.10 / Adds back container functionality

RehanY147:
**Changes:**

Adds back docker container functionality

**Testing:**

Test docker container list page. Test bulk delete function. Also test that data is displayed correctly on the page. Also test that `dockerhub_rate_limit` is handled correctly.


denysbutenko (changes requested):
image

Menu item is missing.
Routing is not connected.
How to test the changes?

denysbutenko (changes requested):
image

Checkbox column is missing to test bulk items delete.

undsoft (approved):
I can live without a checkbox column for now if it's too hard to add. In this case create a ticket to improve this later.

I did however try to pull `nginx` image and receive an error:
```
[EFAULT] Failed to pull 'nginx:vlatest' image: 404 Client Error for http+docker://localhost/v1.46/images/create?tag=vlatest&fromImage=nginx: Not Found ("manifest for nginx:vlatest not found: manifest unknown: manifest unknown")
```

RehanY147:
I've created a new ticket to add bulk checkbox for deletion

RehanY147:
https://ixsystems.atlassian.net/browse/NAS-130533

undsoft (approved):
Our code looks okay, but I'd like to know why I wasn't able to install `nginx`.

RehanY147:
@undsoft I talked to middleware about tag and there was a misunderstanding on what tag is. I wrongly assumed that whatever the tag string would be, I would have to prefix it with a `'v'` character so the image name and tag name would be joined like so. `'[IMAGE_NAME]:v[TAG_NAME]'`. However, I didn't need to add any prefixes. Tag need to be joined as is to the image name like so. `'[IMAGE_NAME]:[TAG_NAME]'`. The updated code reflects this change.

PR #10417, opened 2024-08-09
NAS-130333 / 24.10 / Adds roles for app related api endpoint access elements

RehanY147:
**Changes:**

Adds roles for app related api endpoint access elements

**Testing:**

Test from API docs that api related endpoints are not missed. Calls that are allowed for `READ_ONLY_ADMIN` role are skipped as they are always allowed.


src/app/pages/apps/components/installed-apps/app-settings-button/app-settings-button.component.html
```diff
@@ -32,6 +34,7 @@
     {{ 'Manage Container Images' | translate }}
   </button> -->
   <button
+    *ixRequiredRoles="manageCatalogRequiredRoles"
```
undsoft, line 37: This can be allowed, because there is no other way for user to see what trains are currently selected.
Related: https://github.com/truenas/webui/pull/10408

undsoft (approved):
There are console errors on Installed apps page:
Can't bind to 'ixRequiredRoles' since it isn't a known property of 'button'

undsoft (approved):
Looks good, thanks.

PR #10422, opened 2024-08-09
NAS-130379 / 24.10 / Workloads card

undsoft:
**Testing:**

Test workloads card.

### Downstream

|Affects         |Reasoning
|Documentation   | Behaviour of workloads card changed.


RehanY147 (changes requested):
image
This dialog doesn't seem consistent with the other dialogs we show. I think it should have action buttons even if just to click "Ok".

RehanY147 (changes requested):
Also, since the max width for dialogs was removed, I think that made some dialogs look weird. Like the dialog that warns about installing apps is too wide and doesn't look nice.

undsoft:
Okay, added OK button and reverted dialog width change.

RehanY147 (approved):
Looks good. The "Ok" button has weird margins but otherwise it's OK.
image

RehanY147 (commented):
In the dialog, maybe the source column can be truncated. Show full value on hover. Give other columns more width to show their values in one row.

undsoft:
I'll keep as is for now, later we need to figure out the issue that prevents dialog from having width overrides.

PR #10425, opened 2024-08-11
NAS-130503 / 24.10 / Fix rollback and upgrade, as well updates for relevant buttons

RehanY147:
**Changes:**

Fixes proper button hidden/unhidden status for update and rollback buttons. Fixes rollback operation.

**Testing:**
Test by adding `test` train to catalog and install `nginx` version `1.0.5`. Then update to latest version and rollback to version `1.0.5`. You should be able to do this without refreshing the page or navigating away.


undsoft (approved):
Validated on a hot-patched system.

PR #10429, opened 2024-08-12
NAS-130373 / 24.10 / Timestamp & Event Data differ on Audit Logs for sudo

AlexKarpov98:
**Changes:**

Corrected field to be displayed.

**Testing:**

See ticket, Result ⬇️ 
Screenshot 2024-08-12 at 12 40 23

### DIFF
--- a/src/app/pages/audit/components/audit/audit.component.spec.ts
+++ b/src/app/pages/audit/components/audit/audit.component.spec.ts
@@ -28,9 +28,8 @@ describe('AuditComponent', () => {
     {
       audit_id: '1',
       timestamp: {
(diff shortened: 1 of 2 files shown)


undsoft (approved):
I'm assumed you have confirmed that we need to use this field. If this is the case, the date is shown correctly.

AlexKarpov98:
@undsoft - discussing it with Andrew Walker - to be 100% sure about changes here.

AlexKarpov98:
Reassigned to middleware 
https://ixsystems.atlassian.net/browse/NAS-130373

Gonna check if it still needs fix on UI side after middleware merge.

PR #10433, opened 2024-08-12
NAS-130525 / 25.04 / Add back app-notes-card

RehanY147:
**Changes:**

Adds back notes app card to show notes in a separate card

**Testing:**

If there are notes available on an app, they should show up in a notes card in the details card panel.

### Downstream

|Affects         |Reasoning
|Documentation   | Adds back notes card for app details


undsoft (approved):
Notes are back.

PR #10434, opened 2024-08-12
NAS-130552 / 25.04 / Pick first network interface when no settings

denysbutenko:
**Changes:**

When the network widget has no settings (fresh install), select the first available network interface to display instead of `Configure widgets` message.

**Testing:**

Ensure the widget network displays network data and some interface is selected


undsoft (changes requested):
I don't think it should be widget's responsibility to do this. We have code somewhere that is responsible for migrating widgets, it should be the one to select first interface.

denysbutenko:
Okay migration code already handles it.

PR #10465, opened 2024-08-14
NAS-130254 / 25.04 / Setting UK/GB date results in error in Reporting graphs

AlexKarpov98:
**Changes:**
Added locale service tests & added new method to correctly convert string date to Date() object.

**Testing:**
See ticket.
Try affected files. (As well fixed dashboard widgets)

**new Date() Bug logic:**
It's because the Date constructor in JavaScript expects the date string to be in a format that it can parse correctly. The format DD/MM/YYYY HH:MM:SS is not natively supported by the Date constructor, leading to an “Invalid Date” error.

### DIFF
--- a/src/app/pages/dashboard/widgets/system/widget-sys-info-active/widget-sys-info-active.component.spec.ts
+++ b/src/app/pages/dashboard/widgets/system/widget-sys-info-active/widget-sys-info-active.component.spec.ts
@@ -83,6 +83,7 @@ describe('WidgetSysInfoActiveComponent', () => {
       }),
       mockProvider(LocaleService, {
         getDateAndTime: () => ['2024-03-15', '10:34:11'],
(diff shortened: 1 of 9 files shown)


undsoft (changes requested):
So I'm confused about the code that we use in uptime calculations.
Currently we take a date, the format it, then parse it back and then add seconds to it.
Why not take a date, add seconds and then format it?

AlexKarpov98:
> So I'm confused about the code that we use in uptime calculations. Currently we take a date, the format it, then parse it back and then add seconds to it. Why not take a date, add seconds and then format it?

@undsoft  - currently - we have a string date, convert it to real date and adding seconds then.
We don't have a Date() initially, but `string date`

undsoft (approved):
I still think there is something fishy going on with uptime, but we can look into it later.
Using different date formats no longer breaks pages.

PR #10475, opened 2024-08-15
NAS-128620 / 25.04 / Rethink widget group containers when sorting widgets on main dashboard

AlexKarpov98:
Evgeny: 
`I don’t know if we’ll keep it, but let’s make a PR and see how it looks.`

**Before:** 
Screenshot 2024-08-15 at 14 12 50

**After:**
Screenshot 2024-08-15 at 14 07 45

@undsoft - you can check it out now.


undsoft (changes requested):
Thanks, I liked old way better. Let's just close this ticket as No fix.

PR #10477, opened 2024-08-15
NAS-129146 / 25.04 / Add max-height limit to apps page

RehanY147:
**Changes:**

Fixing overflow height to page height

**Testing:**

See ticket description for testing


bvasilenko (changes requested):
I've noticed weird behavior:

When user loads the **Apps** page, it's still possible to scroll down the page. The scrolling seems to be caused by the dynamic calculation of height:


https://github.com/user-attachments/assets/3a23be02-e76f-4c9b-a574-4329f45dbdbb


**Note:** triggering responsivity threshold by resizing the window suddenly fixes this issue.

RehanY147:
@bvasilenko I can't see the scroll issue. Please confirm if you can still see this bug.

https://github.com/user-attachments/assets/0aea6f36-02c3-47bd-b349-e4ed6c5c139a

bvasilenko:
@RehanY147 Yes, the bug is still there, here's the clarified replication steps:

1. (**Important step**) Change the scaling of the browser to <100%,  for example 90% 
2. Fresh-load the page

The scrolling behavior is identical to the one described [earlier](https://github.com/truenas/webui/pull/10477#pullrequestreview-2241400931)

RehanY147:
@bvasilenko please review again

src/assets/styles/mixins/layout.scss
```diff
@@ -48,6 +48,7 @@ $scrollbar-offset: 20px;
     color: var(--fg2);
     display: flex;
     flex: 1;
+    max-height: calc(-200px + 100vh);
```
AlexKarpov98, line 51: why not `max-height: calc(100vh - 200px);` ?

RehanY147:
@bvasilenko @AlexKarpov98 I've made some more changes. Please test for regressions where we are using `ix-tree-node` or `ix-tree-view` or `ix-tree-virtual-scroll-view`

AlexKarpov98 (changes requested):
That's not what we want for virtual scrolling elements.
Such as Audit page, Datasets, Apps. 

@undsoft should approve it.

Now we have strict scrolling container on the left side.

https://github.com/user-attachments/assets/bdda4e58-18af-4d28-b03d-bdd9b2446f69

It should scroll with whole page, not strict local container. (like before)

Example: (we do not scroll list itself as a container but scrolling body)

https://github.com/user-attachments/assets/ff52bd6b-d373-4c06-ad20-0dcc88dd7423

bvasilenko:
@AlexKarpov98 the alternative PR has same issue, at least I could replicate it on Chrome with 90% scale

AlexKarpov98:
> @AlexKarpov98 the alternative PR has same issue, at least I could replicate it on Chrome with 90% scale

Yeah, I seen it. I think I fixed it now , try it once again.

RehanY147:
This issue is resolved by alternative PR.

PR #10483, opened 2024-08-16
Update cs.json

markrieder:
**Changes:**

**Testing:**

### Downstream

|Affects         |Reasoning
|Documentation   |


undsoft (approved):
Thanks, looks good. 

I'll merge as is, but maybe for the future PRs you could review those `plural` statements. If Czech is like other Eastern European languages, it probably has more plural forms than just two like in English.
Maybe something like this:
```
{seconds, plural, =1 {# sekunda} few {# sekundy} other {# sekund}}
```

Additional explanation is available here: https://github.com/truenas/webui/blob/master/TRANSLATING.md#plural-forms

PR #10508, opened 2024-08-20
NAS-130726 / 25.04 / Ban flex-layout module

undsoft:
**Changes:**

Removes more usages of flex-layout module.

**Testing:**

Test some of the affected changes.

Layout in support-card may not be identical to old layout, which is fine.


denysbutenko (approved):
Looks good

PR #10510, opened 2024-08-20
NAS-130678 / 25.04 / Fix dRAID logic for number of children when creating a pool

bvasilenko:
**Changes:**

Previously in the code the comparison operator was not strict and was including the value `p+d+s`. But as per the ticket requirements, the comparison operator is now strict

**Testing:**

Please see ticket for replication steps and expected results


bvasilenko:
@AlexKarpov98 

1. Open **Storege \> Pool Creation Wizard page**.
2. In the **Data** step, select **Layout** (as **dRAID**) and **Disk Size**. (In the example there are `22` such disks).

image

3. Select any **Data Device** and **Distributed Hot Spares**.
4. Expand **Children** dropdown:

image

The values ​​in the list should be `p+d+s < num children <= amount disks (no more than 255)`, where:\
`p` = **Layout** (for dRAID1 = 1, dRAID2 = 2, dRAID3 = 3) \
`d` = **Data Device** \
`s` = **Distributed Hot Spares**

In this case: `1+8+6 < num children <= 22` \<=\> `15 < num children <= 22`

For the record, old result:

image

AlexKarpov98 (approved):
Works as expected. 👌

PR #10512, opened 2024-08-21
NAS-130656 / 25.04 / Portal group IDs on iSCSI target are shown as incorrect

AlexKarpov98:
Testing: see ticket. We used different identifiers - that's the issue.


RehanY147 (approved):
Target form works Ok

PR #10522, opened 2024-08-22
Update zh-hans.json

xream:
This is a common mistake. The pronunciation of "后" and "候" is the same.

"请稍候" means "Please wait"

"请稍后" is grammatically incorrect. It should be paired with what to do, such as "请稍后重试" "Please try again again later."


undsoft (approved):
Thanks.

PR #10532, opened 2024-08-23
NAS-130759 / 25.04 / Use same instance on destroy and on init for `tree-vritual-scroll-view`

RehanY147:
**Changes:**

Instead of recreating the instance to disconnect from event listener on component destroy event, use the same instance that is used to register the callback method.

**Testing:**

Have some datasets setup on a pool. Navigate quickly between different pages and come back to the datasets dashboard and try scrolling. If you do it quickly, you can see lots of instances of the same error on the console. This fix resolves that issue.


RehanY147:
backport

PR #10559, opened 2024-08-26
NAS-130645 / 25.04 / Exclude used disks from automated disk selection

bvasilenko:
Summary:

The problem was that the `findSuitableDisks` method was returning more suitable disks for the checked `treatDiskSizeAsMinimum` than there actually were. This occurs because `removeUsedDisks` for **Metadata** was called later than `findSuitableDisks` for **Dedup**, and the disks occupied by **Metadata** were available for selection.

Testing:

For replication steps, please follow the video attached to the ticket


RehanY147 (changes requested):
I can't reproduce this bug after several tries replicating the exact scenario described in the ticket. How can I reproduce this?

undsoft (changes requested):
> This occurs because removeUsedDisks for Metadata was called later than findSuitableDisks for Dedup, and the disks occupied by Metadata were available for selection.

Can you explain how this could happen? 

It seems incorrect to me that we need to introduce new methods and checks to exclude used disks. 

There is already logic for this. 

Disks should be assigned to a category and become unavailable for the next category (`pickDisks()` and `this.groupedDisks.removeUsedDisks(pickedDisks);`).

I see two potential problems:
1. `const categories = this.excludeManualCategories(topology);` may produce categories in a different order from how they are presented on the screen, resulting in different disks assigned to dedup before being assigned to metadata.
2. It seems that selecting smaller disk and `Treat Disk Size as Minimum` does not pick the smallest disk, but instead just picks first good enough disk. 

https://github.com/user-attachments/assets/bb321727-6bcd-4699-a540-46ba90f176ef

If this is not the cause of this bug, please create a ticket to address the issue later.

bvasilenko:
@undsoft the correct order of categories should resolve the issue, but we cannot guarantee that the user will go through the steps from top to bottom

1. Yes, categories are created in the wrong order. This would solve the problem in this case. But in the UI it is possible to go through the steps in any order and the problem can repeat itself in a mirror way.


2. Yes, any suitable disk is selected, and not the smallest one, but this does not affect this particular bug.


Description of the problem:

Let's assume there are 6 disks of 1Gb and 2Gb. For Metadata we will select 2 disks of 1Gb each. Then the UI will not offer a choice for Mirror Dedup, because only 1 of the 3 available disks is free. This is expected:

image

Check the box `Treat Disk Size as Minimum` for `Dedup`. Now 4 out of 6 available disks are free for `Dedup` in UI.

Since the `Dedup` category is listed earlier than `Metadata`, 6 disks will be offered for it and 2 of them will be selected **randomly**. (Including those intended for `Metadata`):

image

My fix informed the algorithm that disk `1` and `2` were already occupied and the choice was made from the remaining 4 (regardless of the order of the categories).

undsoft (changes requested):
I see. It's a good point that user can work with categories in arbitrary order.

However, I still think we need to reimplement the solution. Previous code is written to be mostly independent from the previously generated vdevs. In this logic vdevs, unless they were manually selected, are treated as an output of topology settings. Hence the logic to regenerate all of the vdevs after any change. 
It also uses `groupedDisks` to first select applicable disks and then consume them.

In new logic:
1. There are now two ways for excluding previously used disks (via `groupedDisks.removeUsedDisks` and `usedDisks`).
2. New vdevs are dependent on vdevs that came before them. Steps 1 and 2 have settings that affect all other categories and may cause disks to be shifted around. For example, user can first configure the layout and then go back to Enclosure Options and select Maximize Enclosure Dispersal.

This may result in different disks being assigned to different categories, which in new logic probably won't work correctly, because it'll view those disks as being in use.

I think the only viable alternative is to store the order of edits made by the user.
Let's add a test case for this as well.

undsoft (approved):
Okay, looks good. Let's add a separate integration test for this (see in `pool-manager/tests`).

undsoft:
I'll ask around to see if we want to backport it to Electric Eel, it may be too risky of a change.

undsoft:
Confirmed that this can be backported.

RehanY147 (approved):

bvasilenko:
@RehanY147 Here's an instruction on reproducing the issue

1. Replace the file **src/app/pages/storage/modules/pool-manager/store/disk.store.ts** with the one attached below.

   [disk.store.ts.txt](https://github.com/user-attachments/files/16928762/disk.store.ts.txt)

2. Follow the steps from the task description:

https://github.com/user-attachments/assets/48bef68a-e1da-46ea-bb28-0d5b727d2a93

Additionally: reproducing the problem:


https://github.com/user-attachments/assets/5b7d9fdc-df62-4c12-b51d-5d4868c463a4

undsoft:
backport

PR #10570, opened 2024-08-28
NAS-130758 / 25.04 / Update disk list after wipe

bvasilenko:
**Testing:**

Connect to machine 10.220.16.85
See ticket for replication steps.


AlexKarpov98 (approved):
Works as expected 👍

PR #10580, opened 2024-08-29
NAS-129783 / 25.04 / After Failover, the secondary controller is not visible on the Dashboard

AlexKarpov98:
Testing: see ticket

Result 👇 

https://github.com/user-attachments/assets/75664467-933f-4c71-8e16-e567a3815c9b


bvasilenko (approved):
Works as expected

PR #10582, opened 2024-08-29
NAS-130830 / 25.04 / Audit Logging status on SMB share tables is always Yes

AlexKarpov98:
Testing: see ticket

Result 👇 

Screenshot 2024-08-29 at 11 58 45


undsoft (approved):
Works correctly now.

PR #10584, opened 2024-08-29
NAS-130629 / 25.04 / Add "Reset" to configure dashboard

denysbutenko:


RehanY147 (changes requested):
I think there should be an extra step before saving the "Reset" dashboard. The reset should only apply locally. The user should still be able to cancel the reset function and undo their reset.

RehanY147:
Once the "Reset" button is clicked, the default dashboard config should be applied and shown the user but not saved. The user should be able to click cancel button and not save those changes.

denysbutenko:
> I think there should be an extra step before saving the "Reset" dashboard. The reset should only apply locally. The user should still be able to cancel the reset function and undo their reset.
> Once the "Reset" button is clicked, the default dashboard config should be applied and shown the user but not saved. The user should be able to click cancel button and not save those changes.

This sounds reasonable. PR is updated.

undsoft (approved):
Looks good 👍🏼

PR #10593, opened 2024-08-30
NAS-130862 / 25.04 / Convert ix-forms to standalone components

undsoft:
**Before:**

`ix-code-editor` imports codemirror and is part of ix-forms module. End result is that codemirror is loaded in initial bundle when user accesses TrueNAS.

Снимок экрана 2024-08-29 в 20 39 04

**Now:**
More chunks, but total initial bundle is 20-25% smaller.

Снимок экрана 2024-08-29 в 20 39 14


denysbutenko (approved):
Nice enhancement!

PR #10595, opened 2024-08-30
NAS-130859 / 25.04 / UI provides incorrect information about dataset ACL mode and does not provide a mechanism to fix it when incorrect

AlexKarpov98:
Testing: 

If ACL Mode is somehow incorrectly set, in edit mode - user should be able to see it and be able change it, or update ACL Type and then we should see revalidation. 

Shortly, validation is turned off in edit mode if user has incorrect ACL Mode related to ACL Type.

Example:
`Posix -> passthrough`  is incorrect 
`Posix -> Discard` is correct

So if we entered edit mode and user has incorrect setup - he should be able to change it.
If user opened form with correct setup, like `Posix -> Discard` - Discard option shall remain disabled.

Result:

https://github.com/user-attachments/assets/3f3ee674-9f27-4a0a-a7e1-7b148ead9405


RehanY147 (approved):
Confirmed with Andrew. This issue is possible because we can have shares outside of the TrueNAS systems and settings can become invalid afterwards.

PR #10602, opened 2024-09-02
NAS-130874 / 25.04 / snapshot batch delete button is deleting ALL snapshots, not just selected

AlexKarpov98:
Testing: see ticket

Result: ⬇️ 

https://github.com/user-attachments/assets/f187e383-bd22-48b0-b369-18fc20c852f3


bvasilenko (approved):
Works as expected

PR #10604, opened 2024-09-02
NAS-130854 / 25.04 / Refactor `trackBy` for `ix-table-body`

RehanY147:
**Changes:**

Defines clear `trackBy` functions for loops inside the template of `ix-table-body` component.

**Testing:**

Test that table rows show up correctly.


src/app/modules/ix-table/components/ix-table-body/cells/ix-cell-state-button/ix-cell-state-button.component.ts
```diff
@@ -31,66 +35,63 @@ interface RowState {
...
   errorHandler: ErrorHandlerService = inject(ErrorHandlerService);
 
+  private cdr = inject(ChangeDetectorRef);
   private store$: Store<JobSlice> = inject<Store<JobSlice>>(Store<JobSlice>);
+  job = signal<Job>(null);
+  jobUpdates$: Observable<Job<ApiJobResponse<ApiJobMethod>>>;
+
+  ngOnInit(): void {
+    this.jobUpdates$ = this.store$.pipe(
+      select(selectJob((this.row as Job).id)),
```
undsoft, line 51: How is the `row` a `Job` here? If I have a replication task for example, then logically the row should be a replication task, while job should be some field within this replication task.

src/app/modules/ix-table/components/ix-table-body/ix-table-body.component.ts
```diff
@@ -79,8 +79,18 @@ export class IxTableBodyComponent<T> implements AfterViewInit {
       && (this.dataProvider?.expandedRow?.[this.detailsRowIdentifier] === row?.[this.detailsRowIdentifier]);
   }
 
-  // TODO: Come up with a proper identifier
-  protected trackByIdentity<X>(item: X): X {
-    return item;
+  protected trackRowByIdentity(item: T): string {
+    return this.getTestAttr(item);
```
undsoft, line 83: We need to rename `getTestAttr` and `rowTestId`, because it has different meaning now. Or new fields need to be introduced.

undsoft (changes requested):

src/app/modules/ix-table/components/ix-table-body/ix-table-body.component.ts
```diff
@@ -79,8 +80,30 @@ export class IxTableBodyComponent<T> implements AfterViewInit {
...
+  }
+
+  protected displayedColumnsForRow(row: T, columns: Column<T, ColumnComponent<T>>[]): Column<T, ColumnComponent<T>>[] {
+    for (const column of columns) {
+      column.setRow(row);
```
undsoft, line 89: How does it relate to `setRow` in `IxTableBodyCellDirective`?

src/app/modules/ix-table/components/ix-table-body/ix-table-body.component.ts
```diff
@@ -79,8 +80,30 @@ export class IxTableBodyComponent<T> implements AfterViewInit {
...
+  }
+
+  protected trackColumnByIdentity(column: Column<T, ColumnComponent<T>>, index: number): string {
+    const trackById = this.getTestAttr(column.getRow());
+
+    if (column.title) {
+      return trackById + '-' + column.title + '-' + index;
+    }
+    const actionsColumn = column as Column<T, IxCellActionsComponent<T>>;
+    if (actionsColumn.actions?.length) {
```
undsoft, line 101: This makes `TableBodyComponent` aware of types of columns and their inner workings. 
This would only be useful if user changes visible columns and we don't want to re-render some of the cells, but addition of index reduces usefulness of it.

I'm fine with letting it keep tracking the object like it used to. Otherwise new system is needed, probably one where each ColumnComponent provides its own column id.

Also `trackSomethingByIdentity` needs to be renamed, because the word `identity` used to refer to tracking the object itself.

undsoft (changes requested):

src/app/modules/ix-table/components/ix-table-body/cells/ix-cell-state-button/ix-cell-state-button.component.ts
```diff
@@ -41,6 +41,19 @@ export class IxCellStateButtonComponent<T> extends ColumnComponent<T> implements
...
+    const job = !row || !this.getJob ? undefined : this.getJob(row);
+
+    if (!job) {
+      return;
+    }
+
+    this.state.set(job.state);
+  }, {
+    allowSignalWrites: true,
+  });
```
undsoft, line 55: We can keep it for now, but in the future we should try to refactor code to avoid having effects when possible and rely on `computed` signals more.

undsoft (approved):
Okay, looks good. 
I would try to do something with `if (key === 'value')` checks. They are too specific, more like a hotfix. Perhaps we can use a different way of assigning inputs or use something like `Object.hasOwn` or something else to make sure it's an assignable property.

PR #10608, opened 2024-09-02
NAS-130953 / 25.04 / Do not attempt to login with old tokens

AlexKarpov98:
**Changes:**
If user logged in. We save `tokenLastUsed` with a Date. We update `tokenLastUsed` on middleware requests.
Once user closed the WEB UI, hanged out somewhere else for 5+ minutes `(default middleware value for token lifetime)`, returned to the WEB UI - token will be removed and user should login again.

If user returned earlier, like 4 minutes, he will be automatically logged in as previously. It will only work if it's more than 5 minutes `(default middleware value for token lifetime)`.

**Testing:**
See ticket, read **Changes** part.

### Downstream

|Affects         |Reasoning
|Documentation   | Token removal logic update


denysbutenko (changes requested):

src/app/services/ws.service.ts
```diff
@@ -139,6 +141,13 @@ export class WebSocketService {
     this.eventSubscribers.clear();
   }
 
+  getDebouncedWebSocketStream$(): Observable<unknown> {
```
undsoft, line 144: This service doesn't know why 5 seconds is needed, thus it shouldn't be part of this service.

src/app/views/sessions/signin/store/signin.store.ts
```diff
@@ -93,6 +95,15 @@ export class SigninStore extends ComponentStore<SigninState> {
         this.loadFailoverStatus(),
         this.updateService.hardRefreshIfNeeded(),
       ]).pipe(
+        filter(() => {
```
undsoft, line 98: I think it's time to extract all login-with-token business into a separate method.

src/app/services/token-lifetime.service.ts
```diff
@@ -25,6 +29,21 @@ export class TokenLifetimeService {
...
+  // Check if token was used no more than 15 minutes ago
+  get isTokenWithinTimeline(): boolean {
+    const tokenLastUsed = this.window.localStorage.getItem('tokenLastUsed');
+
+    if (!tokenLastUsed) {
+      return false;
+    }
+
+    const tokenLastUsedDate = new Date(tokenLastUsed).getTime();
+    const tokenLifetime = 60 * 15 * 1000;
```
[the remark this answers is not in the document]
AlexKarpov98: > Shouldn't we rely on a system-configured token lifetime?

Token `system-configured token lifetime` is only known for the logged in user.
We are dealing with the possibly logged in user, but we wan't to avoid calling middleware with invalid token.
But I agree that we should change the value from 15min to 5min as the default min middleware value for token lifetime!

denysbutenko (approved):
Looks good

denysbutenko:
If a system has an extended token lifetime, users must re-login every 5 minutes of inactivity. It needs to be improved in the future.

AlexKarpov98:
> If a system has an extended token lifetime, users must re-login every 5 minutes of inactivity. It needs to be improved in the future.

yeah, but it's only in case of inactivity (web ui is closed and user is not returning there for 5+ minutes)

PR #10609, opened 2024-09-02
NAS-130801 / 25.04 / Fix `DriveBayLightStatus` enum

bvasilenko:
**Testing:**

At the Enclosure page, 
1. Choose a disk
2. Click **Identify**  button

**Expected result:**
- no errors
- enumeration values are expected to match the ones which are defined in the middleware.


undsoft (approved):
The buttons themselves work, but something is wrong with conditions.
I've updated our m40 to latest middleware and I have a slot that has `drive_bay_light_status: "OFF"`, yet I see Turn Off button.

The logic should be that if the status is known (not null), then we show the appropriate button. If the status is not known, we show both buttons.

PR #10612, opened 2024-09-03
NAS-130956 / 25.04 / Reduce numbers of layout shifts on login page

AlexKarpov98:
**Changes:**
Reduced numbers of layout shifts on login page

**Testing:**
See ticket and try it in action.

See how it works if I log in for the first time, see layout shifts fixes.
As well once I was logged in, I reloaded the page. (see)👇 

**BEFORE:**

https://github.com/user-attachments/assets/f2510e9c-20a3-4c40-ab15-4cae81251990

**AFTER:**

https://github.com/user-attachments/assets/58fb935c-be0f-4aee-b872-9921ba2dc4b2

_Note: No tests added before the final approve._


bvasilenko (approved):
Loading of the login page looks good!

undsoft (changes requested):
1. If you restart middleware with `service middlewared restart` you'll still get `Please wait` with the message, even though you are not being logged in and waiting may not help if there is no connectivity. We should only show that message when we are actively logging user in.
2. We should change `Please wait` to `Logging in...`
3. There is a lot of empty space on main login form now, I think we should do something about it.
4. If the system is managed by truecommand and has HA, there is a scrollbar.
5. I'm fine with changing how `FailoverStatusComponent` works. For example, we could show `HA is enabled (?)` with a tooltip for active IP addresses. We could show `HA is disabled: reason` when there is one reason and `HA is disabled (?)` when there are many.
6. We could also avoid showing validation errors in the form, but we need to make snackbar more prominent then.
7. Page doesn't look correct when you connect to a standby controller. This is how used to look. You can test it by connecting by adding `a` or `b` to hostname of our m40.
Снимок экрана 2024-09-05 в 18 06 12

8. If set password form has scrollbar, we need to make to make adjustments to it as well. I'm fine with it having a different height there.

AlexKarpov98:
@undsoft please give it a try now. It's a bit tricky with that `Logging in` show up.

I think we can remove it, if it created more problems than benefits?

src/app/services/auth/auth.service.ts
```diff
@@ -41,6 +41,8 @@ export class AuthService {
   @LocalStorage() private token: string;
   protected loggedInUser$ = new BehaviorSubject<LoggedInUser>(null);
 
+  instantAuthTokenValue$ = new BehaviorSubject<string>(this.window.localStorage.getItem('ngx-webstorage|token'));
```
undsoft, line 44: This relies on internal implementation of a 3rd party. Is there a way to just read value from `token`?

undsoft (changes requested):
Initial login flow looks fine, however I could live without Logging you in after submitting the form.
When reloading the page, I could live without "Connecting to TrueNAS" message at least for some time. I think there should be some period where we expect connection to happen and we don't show this dialog.
Also "Connecting to TrueNAS" is too wide now for some reason.

AlexKarpov98:
@undsoft  - updatred:

https://github.com/user-attachments/assets/fcf25d49-4c04-4aeb-8346-1dcfd6830060

undsoft (changes requested):
Better, but still some things to fix:
1. When connecting to a standby controller, I think it would be helpful to disable inputs as well, otherwise it's not immediately obvious why I cannot click Login.
2. I've used Reset Configuration option in General Settings and it kept saying Logging in.
3. If you use non-existent url for webui, it will still present a login form for a second before switching to "Connecting to TrueNAS" message. I think we should keep longer in previous loading state before showing login form as we attempt to establish a WS connection.

AlexKarpov98:
> Better, but still some things to fix:
> 
> 1. When connecting to a standby controller, I think it would be helpful to disable inputs as well, otherwise it's not immediately obvious why I cannot click Login.
> 2. I've used Reset Configuration option in General Settings and it kept saying Logging in.
> 3. If you use non-existent url for webui, it will still present a login form for a second before switching to "Connecting to TrueNAS" message. I think we should keep longer in previous loading state before showing login form as we attempt to establish a WS connection.

@undsoft  we shod `LOGGING IN` only if user has token and if token `isTokenWithinTimeline` 
probably there may be cases when token is not cleared up?
for this one: `I've used Reset Configuration option in General Settings and it kept saying Logging in.`

AlexKarpov98:
@undsoft 

```
3. If you use non-existent url for webui, it will still present a login form for a second before switching to "Connecting to TrueNAS" message. I think we should keep longer in previous loading state before showing login form as we attempt to establish a WS connection.
```

Problem is that if we'll show loading screen on invalid url, it as well will be shown before valid TRUE NAS url for web ui. And we'll loose that immediate sign in form rendering.

We show LOGGING IN message only if user has `token` and `isTokenWithinTimeline`

As discussed - `<ix-disconnected-message />` is showed with delay,  currently it's `2 sec`.
SO we only show it if we are not connected yet after 2 seconds.

AlexKarpov98:
```I've used Reset Configuration option in General Settings and it kept saying Logging in.```

I see: 
```Connecting to TrueNAS ... Make sure the TrueNAS system is powered on and connected to the network.```

And if I reload the page I see `Logging in...` - but it's shown because we have `token` and it's `isTokenWithinTimeline`

undsoft:
I understand how it currently works, but I am saying that maybe we could keep showing flashing TrueNAS login animation until websocket connection is established. This way we won't have situations where it says "logging in" for an offline system.

AlexKarpov98:
> I understand how it currently works, but I am saying that maybe we could keep showing flashing TrueNAS login animation until websocket connection is established. This way we won't have situations where it says "logging in" for an offline system.

Yeah, I got it, I think that's the only way.

undsoft (changes requested):
It's a bit worse now. My WS connection took 500ms to establish, yet I still saw establishing connection on initial page load.

https://github.com/user-attachments/assets/7ec6f3a7-08d9-4fe4-a794-b38814b556fa

AlexKarpov98:
@undsoft now it's updated to show True NAS logo as a fallback if we are not logged in or not connected to the WS (for 1 second)

User Logged In

https://github.com/user-attachments/assets/82d653cd-33e4-49cc-b1fb-f04e10af9ecc

And screen like that if we are not connected and more than 1 sec waiting

https://github.com/user-attachments/assets/746679f1-4753-464a-b90e-f0e9d43c0008

undsoft (approved):
Thanks. Nice work 👍🏼

PR #10617, opened 2024-09-04
NAS-130957 / 25.04 / Reduce number of fonts in WebUI

AlexKarpov98:
**Changes:**
For terminal we do need a custom font - Inconsolata :100: 

It’s hard to reduce fonts, I tried with using only `IBM Plex Sans` and `Titillium Web`
Removed: 
- Roboto
- Droid Sans Mono

I cleaned up @font-face declaration. 

As well we may need to convert from `.ttf` to `.woff2` to make font files smaller. ℹ️ 

**Testing:**
See ticket. 
See code updates.

You can check Network Tab with Fonts as well:

____

As well I compared build with this update and without, the diff for `styles.css`:
With this update: **505.31 kB**
Without this update: **509.47 kB**

With this update:
Initial total            |   3.19 MB |               670.76 kB

Without this update:
Initial total            |   3.19 MB |               689.23 kB


src/assets/styles/fonts/roboto.scss
```diff
@@ -1,56 +1,4 @@
-@font-face {
```
undsoft, line 1: I don't think we need to remove all of these. We can reduce the number of font-weights that we support, but we at least need a lighter and bolder variant with italic as well.

undsoft:
Damian: 
> In all the designs I’ve been using Titillium Web for headers and IBM Plex Sans for other normal text.  So limiting to those two would make it consistent with our design system

dszidi (changes requested):
It seems in the last commit `IBM Plex Sans` is still omitted and `Roboto` was reintroduced. It should be the other way around. To summarize, what we should be seeing is... 

- `Tittililum Web` for headers
- `IBM Plex Sans` for body text
- `Roboto` can be removed

src/assets/styles/other/_root-variables.scss
```diff
@@ -14,10 +14,10 @@
   --hover-bg: var(--lines) !important;
   --primary-lighter: var(--primary);
   --font-family-header: 'Titillium Web', sans-serif;
-  --font-family-body: 'IBM Plex Sans', 'Roboto' , 'Helvetica Neue' ,sans-serif;
+  --font-family-body: 'Titillium Web', 'Roboto', 'Helvetica Neue', sans-serif;
```
dszidi, line 17: This should be `IBM Plex Sans`

We don't want body text to be `Titillium Web`. It is too stylized and not consistent with our design system

AlexKarpov98:
@undsoft @dszidi 

With latest updates (discussed with Damian):
Screenshot 2024-09-04 at 18 34 04

Before:
Screenshot 2024-09-04 at 18 36 23

src/app/modules/tooltip/tooltip.component.scss
```diff
@@ -36,7 +36,7 @@
 }
 
 .tooltip-message {
-  font-family: Roboto, sans-serif;
+  font-family: 'IBM Plex Sans', sans-serif;
```
undsoft, line 39: Best to change to variables here and similar places.

src/assets/styles/fonts/titillium-web.scss
```diff
@@ -2,8 +2,8 @@
   font-display: swap;
   font-family: 'Titillium Web';
   font-style: normal;
-  font-weight: 900;
-  src: local('Titillium Web Black'), local('TitilliumWeb-Black'), url('../../fonts/TitilliumWeb/TitilliumWeb-Black.ttf') format('truetype');
+  font-weight: 500;
```
undsoft, line 5: Is this for "normal" font-weight? Normal is 400.

undsoft (approved):
Looks good, I don't see any visual changes 😅

src/app/modules/tooltip/tooltip.component.scss
AlexKarpov98: agree

src/assets/styles/fonts/titillium-web.scss
AlexKarpov98: @undsoft 
Screenshot 2024-09-05 at 18 34 38

AlexKarpov98: I think we need to remove `font-style: normal;` in this case

PR #10623, opened 2024-09-05
NAS-130955 / 25.04 / Alternative for global data stores, part 1

bvasilenko:
**Changes:**

An attempt to establish a new kind of service, which will resolve the problem described in the ticket.

**Testing:**

1. Double check that the ticket's requirements are met
2. Code review

### DIFF
--- a/src/app/pages/dashboard/services/widget-resources.service.spec.ts
+++ b/src/app/pages/dashboard/services/widget-resources.service.spec.ts
@@ -7,6 +7,7 @@ import { mockCall, mockWebSocket } from 'app/core/testing/utils/mock-websocket.u
 import { App } from 'app/interfaces/app.interface';
 import { Pool } from 'app/interfaces/pool.interface';
 import { WidgetResourcesService } from 'app/pages/dashboard/services/widget-resources.service';
+import { appStore, poolStore } from 'app/services/global-store/stores.constant';
 
 const pools = [
   { id: 1, name: 'pool_1' },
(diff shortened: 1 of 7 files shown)


undsoft (changes requested):
Am I missing something? How is this different from what I've reviewed previously?

bvasilenko:
>  Am I missing something? How is this different from what I've reviewed previously?

@undsoft  Sorry, I've pushed same version again. Now this PR contains an updated version.

undsoft (changes requested):
These requirements should be handled by the new store and not left for developers to implement:
> 1. Data is loaded lazily, when it is needed.
> 3. It should be possible to invalidate data.

It would also be good to see examples of these and other requirements, like "mockable and testable".

The output of this ticket is a clear way for developers to use this global data.

src/app/pages/dashboard/services/widget-resources.service.ts
```diff
@@ -31,6 +32,9 @@ import { waitForSystemInfo } from 'app/store/system-info/system-info.selectors';
   providedIn: 'root',
 })
 export class WidgetResourcesService {
+  poolStore = globalStore('pool.query');
+  appStore = globalStore('app.query');
```
undsoft, line 36: Let's move these to constants somewhere so that new injectors are not created every time.

src/app/pages/dashboard/services/widget-resources.service.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+        mockCall('cloudsync.query'),
+        mockCall('webui.main.dashboard.sys_info'),
+        mockCall('interface.query'),
+        mockCall('update.check_available'),
+      ]),
+    ],
+  });
+
+  beforeEach(() => {
+    mockGlobalStore({
```
undsoft, line 38: It would be good to refactor `mockGlobalStore`, so that it's mock like other provider mocks and not a separate global function.

undsoft (approved):
Overall looks good, let's see how it behaves in action.

PR #10626, opened 2024-09-06
NAS-130879 / 25.04 / Fix netdata URL

undsoft:
**Testing:**

There is no way to test this until this is merged, so code review only.


bvasilenko (approved):
Code looks good, we're going to make further testing after it's merged

PR #10640, opened 2024-09-09
NAS-130980 / 25.04 / ElectricEel HA Dashboard is missing the Standby System Information card.

AlexKarpov98:
**Changes:**
Updated logic of default widgets generation.

**Testing:**
See ticket, try with HA system & on non HA system as well.

### DIFF
--- a/src/app/pages/dashboard/services/get-default-widgets.ts
+++ b/src/app/pages/dashboard/services/get-default-widgets.ts
@@ -63,7 +63,9 @@ const defaultWidgets: WidgetGroup[] = [
 
 export const getDefaultWidgets = (isHaLicensed?: boolean): WidgetGroup[] => {
   if (!isHaLicensed) {
-    return defaultWidgets.filter((_, index) => index !== 1);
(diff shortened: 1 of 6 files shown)


src/app/pages/dashboard/services/get-default-widgets.ts
```diff
@@ -54,3 +60,11 @@ export const defaultWidgets: WidgetGroup[] = [
     ],
   },
 ];
+
+export const getDefaultWidgets = (isHaLicensed?: boolean): WidgetGroup[] => {
+  if (!isHaLicensed) {
+    return defaultWidgets.filter((_, index) => index !== 1);
```
undsoft, line 66: That's fragile. Some time later someone will change the order of widgets and this code will stop working.

AlexKarpov98: > That's fragile. Some time later someone will change the order of widgets and this code will stop working.

@undsoft - yeah, I can override it. But tests will fail if someone change the order as well.

PR #10649, opened 2024-09-10
NAS-131016 / 25.04 / Uptime on Dashboard runs too fast

AlexKarpov98:
Testing: 
Go to dashboard, switch tab or open another app. 
Wait 2-3 minutes, return to the WEB UI tab and check that time is not ahead.

_As well - I tried `visibilitychange` listener and `requestAnimationFrame` - but no luck._

Screenshot 2024-09-10 at 20 02 19


undsoft (approved):
Appears to be working correctly now. I couldn't reproduce the time discrepancy.

PR #10656, opened 2024-09-11
NAS-130943 / 25.04 / Fix apps bulk upgrade

RehanY147:
**Changes:**

Fixes issue where select input exists when the user has not expanded the panel. Makes the form status invalid.

**Testing:**

Add the `test` train. Install older version `1.0.5` or `nginx` app from the `test` train. Install a couple of instances of the same app and same version. Then go to bulk upgrade dialog. Only expand one of the apps panel to choose the exact version to upgrade to. Leave the other app panel alone. Click the upgrade button. Both apps should upgrade correctly.


undsoft (approved):
Works fine for me.

PR #10660, opened 2024-09-11
NAS-131113 / 25.04 / Show block IO stats

undsoft:
**Changes:**

Adds new block IO column to apps.

**Testing:**

Check new column

### Downstream

|Affects         |Reasoning
|Documentation   | New Block IO column in apps.


denysbutenko (approved):
When an app has stopped, Block IO shows `0 B - 0 B` value while other columns are `N/A`.
image

When an app is running it looks good
image

Also, I notice that CPU shows `N/A` when usage is zero

AlexKarpov98 (approved):
It works as expected.

PR #10671, opened 2024-09-12
NAS-131135 / 25.04 / Better indication for true cloud backup restoration

undsoft:
**Changes:**

Previous implementation used incorrect callback (`next` instead of `complete`). Switched to showing a dialog.


bvasilenko (approved):
Works without issues

PR #10685, opened 2024-09-16
NAS-131101 / 25.04 / Help button in the Active Directory menu is not clickable if the translation is too long.

AlexKarpov98:
Testing: see ticket. 
Result: 
Screenshot 2024-09-16 at 13 40 13


undsoft (approved):
Looks good.

PR #10702, opened 2024-09-18
NAS-131143 / 25.04 / Sharing Admin does not have permission to create, modify and delete any shares.

AlexKarpov98:
Testing: see ticket.

### DIFF
--- a/src/app/interfaces/ds-cache.interface.ts
+++ b/src/app/interfaces/ds-cache.interface.ts
@@ -25,9 +25,7 @@ export interface LoggedInUser extends DsUncachedUser {
 }
 
 export interface AuthMePrivilege {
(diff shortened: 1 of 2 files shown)


undsoft:
So middleware returns roles in different formats? Are you sure it's not a middleware issue?

AlexKarpov98:
> So middleware returns roles in different formats? Are you sure it's not a middleware issue?

@undsoft 
Yes, middleware sends it in different format, if we can reassign it to them, it's fine. (I thought that this is a feature not a bug from their side)

I'll ask

AlexKarpov98:
@undsoft  - middleware team gonna handle it. I'll just test that it's working fine.

PR #10704, opened 2024-09-18
NAS-131231 / 25.04 / if you enter an invalid AUTH key, tailscale goes into a tight restart loop

AlexKarpov98:
**Changes:**
Now we are able to stop the app with `Deploying` status.

– The retry of deploying is done by docker.
Only allow stop from the UI is the option to fix this bug.

I've discussed it with @stavros-k .

**Testing:**
See ticket, comments, the only way for now is to allow stopping an app which is in `Deploying` phase.

### Downstream

|Affects         |Reasoning
|Documentation   | We can stop app with `Deploying` status.


bvasilenko (approved):
Works as expected

PR #10705, opened 2024-09-18
NAS-131273 / 25.04 / Update packages

undsoft:
**Testing:**

Test for regressions.


AlexKarpov98 (approved):
Looks good overall, but we need to fix tests.

PR #10711, opened 2024-09-19
NAS-131246 / 25.04 / Apps Bulk Actions are not showing correct options for selected app(s)

AlexKarpov98:
Testing: see ticket.

Result: 

https://github.com/user-attachments/assets/d8f88e92-0bb6-43c2-89fe-b36afa10ca50


src/app/pages/apps/components/installed-apps/installed-apps.component.ts
```diff
@@ -116,21 +116,26 @@ export class InstalledAppsComponent implements OnInit, AfterViewInit {
     return this.selection.selected;
   }
 
+  get checkedApps(): App[] {
+    return this.checkedAppsNames
+      .map((name) => this.dataSource.find((app) => app.name === name))
+      .filter(Boolean);
```
RehanY147, line 122: Why is this `filter` added here?

RehanY147 (changes requested):

src/app/pages/apps/components/installed-apps/installed-apps.component.ts
AlexKarpov98: > Why is this `filter` added here?

To tripple check that it mapped correctly and no undefined is there.

RehanY147: In my opinion, adding general checks like this can hide other bugs. We should fix the issue where an undefined value is possible in this case instead of adding generic checks like this.

AlexKarpov98: > In my opinion, adding general checks like this can hide other bugs. We should fix the issue where an undefined value is possible in this case instead of adding generic checks like this.

there are no other bugs, I can just remove that if you want. It's just a extra check.

RehanY147 (changes requested):
In my opinion, adding general checks like this can hide other bugs. We should fix the issue where an undefined value is possible in this case instead of adding generic checks like this.

AlexKarpov98:
@RehanY147 - it's updated now.

RehanY147 (approved):

PR #10716, opened 2024-09-20
NAS-131036 / 25.04 / Remove use of `chain` method from `'lodash-es'`

RehanY147:
**Changes:**

Fixes uses of `chain` method from `lodash-es`. We should either be calling `chain` via `_.chain` or use individual methods without chaining. Using `chain` from `import { chain } from 'lodash-es';` is not advisable because it doesn't include any of the methods so you can't call methods on the collection returned by `chain` method.

**Testing:**

Test that breadcrumbs in places like datasets/snapshots or shares/smb etc are showing up. Test in enclosure dashboard. Check console for errors.


denysbutenko (approved):

undsoft:
Let's also ban via linter rules.

PR #10717, opened 2024-09-20
NAS-131290 / 25.04 / Add `debounceTime` to `disk.query` subscription update in `enclosure.store.ts`

RehanY147:
**Changes:**

Adds `debounceTime` operator to `enclosure.store.ts` `disk.query` subscription pipe. The method is set to be called every time `disk.query` event emits an event. Sometimes, when enclosure state changes, `disk.query` can emit for all the disks at once. Meaning, you can get dozens of events fired in milliseconds. So, the update method was calling `webui.enclosure.dashboard` method as many times very quickly, causing the `Maximum concurrent calls` error. The `debounceTIme` operator will ensure that any surge of event updates from `disk.query` only cause one eventual `webui.enclosure.dashboard` call.

**Testing:**

Test using the steps in the ticket.

### DIFF
--- a/src/app/pages/system/enclosure/services/enclosure.store.spec.ts
+++ b/src/app/pages/system/enclosure/services/enclosure.store.spec.ts
@@ -1,3 +1,4 @@
+import { fakeAsync, tick } from '@angular/core/testing';
 import {
   createServiceFactory,
   mockProvider,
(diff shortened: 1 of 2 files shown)


undsoft (approved):
I cannot quite reproduce the original issue, but the fix makes sense. Didn't find regressions on enclosure page.

PR #10724, opened 2024-09-22
NAS-131220 / 25.04 / Add `jobs-module` to `data-protection-module` for job tracking

RehanY147:
**Changes:**

The JobEffects were not working when accessing jobs inside the data-protection module because the JobsModule wasn't included in the data-protection imports.

**Testing:**

Create a new "True Cloud Backup Task" on the "Data Protection" dashboard page. This task won't start automatically. Click on the start task icon on the table and then click on the job state icon button. You shouldn't see any errors in the console and the dialog should show up correctly and show the job progress.
This doesn't work on master


package.json
```diff
@@ -17,7 +17,7 @@
     "test": "jest",
     "test:watch": "jest --watch",
     "test:ci": "jest --runInBand",
-    "test:pr": "yarn run check-env && echo 'Setting up temporary environment file...\\n' && yarn run ui remote -i 'headless.local' && jest --coverage --maxWorkers=2",
+    "test:pr": "yarn run check-env && echo 'Setting up temporary environment file...\\n' && yarn run ui remote -i 'headless.local' && jest --coverage --maxWorkers=2 ix-cell-state-button.component.spec.ts",
```
bvasilenko, line 20: I've noticed the `ix-cell-state-button.component.spec.ts`  at the package.json, which looks like temporary debugging tweak.

bvasilenko (approved):
Looks good overall, just one minor thing

PR #10727, opened 2024-09-23
NAS-131323 / 25.04 / ddns-updater does not show the IP version value if set to “IPv4 and IPv6” when editing the configuration of the docker

AlexKarpov98:
Testing: see ticket & comments for steps to reproduce.

Fixes https://github.com/truenas/charts/issues/2824


undsoft (approved):
I can see `IPv4 and IPv6` being selected.

PR #10731, opened 2024-09-23
NAS-131244 / 25.04 / Fix Roles card on datasets is missing "System Dataset" text

denysbutenko:
**Changes:**

Fix bug when system dataset is not displayed properly

**Testing:**

Go to Datasets and ensure system dataset row is displayed on roles card. also check system dataset icon in the tree


AlexKarpov98 (approved):
👌 
Screenshot 2024-09-23 at 15 25 22

PR #10734, opened 2024-09-23
NAS-131360 / 25.04 / Remove root user from set initial password form

undsoft:
**Changes:**

Changes to initial password form.

Снимок экрана 2024-09-23 в 14 13 23

**Testing:**
General Settings -> Reset Config.
Easier to use non-HA CI system for that. You may need to wait after config has been reset.

### Downstream

|Affects         |Reasoning
|Documentation   | It is not longer possible to use `root` when setting initial admin password.


AlexKarpov98 (approved):
Works as expected.

PR #10741, opened 2024-09-24
NAS-131130 / 25.04 / Refactor `subscribeToInstalledAppsUpdates` in `installed-apps-store.service`

AlexKarpov98:
Refactoring. (messy code fix)

### DIFF
--- a/src/app/pages/apps/store/installed-apps-store.service.ts
+++ b/src/app/pages/apps/store/installed-apps-store.service.ts
@@ -120,7 +120,7 @@ export class InstalledAppsStore extends ComponentStore<InstalledAppsState> imple
         this.handleAddedOrUpdatedEvent(apiEvent);
         break;
       default:
-        console.error('Unknown API event type (installed-apps-store.service):', apiEvent.msg);
+        console.error('Unknown API event type');
         break;
     }
   }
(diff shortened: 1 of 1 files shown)


denysbutenko:
I have been looking into this code for the last few days. Some points that I highlighted:

- We may need to split the logic for installed apps and available apps into corresponding files.
- `app.query` have support `callAndSubscribe` to handle collection changes
- `recommendedApps` is a subset of `availableApps` like computed property.

Just shared thoughts, this is not a call to action

AlexKarpov98:
@denysbutenko - thanks.
My original thoughts regarding this PR - is just refactoring, to make it much easier to read and understand the code.

src/app/pages/apps/store/installed-apps-store.service.ts
```diff
@@ -177,4 +95,89 @@ export class InstalledAppsStore extends ComponentStore<InstalledAppsState> imple
...
+    switch (apiEvent.msg) {
+      case IncomingApiMessageType.Removed:
+        this.handleRemovedEvent(apiEvent);
+        break;
+      case IncomingApiMessageType.Added:
+      case IncomingApiMessageType.Changed:
+        this.handleAddedOrUpdatedEvent(apiEvent);
+        break;
+      default:
+        console.error('Unknown API event type (installed-apps-store.service):', apiEvent.msg);
```
RehanY147, line 123: We shouldn't expose code details in console logs. Also, this has a potential to expose sensitive info from the api event to the console. We shouldn't be logging the message itself.

RehanY147 (changes requested):

RehanY147 (approved):

PR #10762, opened 2024-09-27
NAS-131378 / 25.04 / Fangtooth iSCSI service data-test tag of the Running toggle is inconsistent with the other iSCSI toggle.

AlexKarpov98:
Testing: see ticket

Screenshot 2024-09-27 at 14 11 40


src/app/pages/services/components/service-state-column/service-state-column.component.ts
```diff
@@ -33,6 +34,10 @@ export class ServiceStateColumnComponent extends ColumnComponent<ServiceRow> {
 
   protected readonly isRunning = computed(() => this.service().state === ServiceStatus.Running);
 
+  get testIdServiceName(): string {
```
RehanY147, line 37: We can make it a signal

RehanY147 (approved):

PR #10773, opened 2024-10-01
NAS-130978 / 25.04 / Sort disks by size when Treat Disk Size As Minimum is selected

bvasilenko:
**Testing:**

Please see ticket for expected algorithm change, replication steps, and expected result


AlexKarpov98 (commented):
I am not sure if it's a correct solution. I'll add @undsoft  as a reviewer as well.

undsoft (approved):
Looks good, correct disk is being selected.

PR #10775, opened 2024-10-01
NAS-131413 / 25.04 / Apps discovery should favor app name on search results

denysbutenko:
**Changes:**

Changed order of apps when a search query is applied.

**Testing:**

Go to apps and try to search an app and check ticket for details.


bvasilenko (approved):
Works as expected

PR #10776, opened 2024-10-01
NAS-131423 / 25.04 / App deletion should ask the user if the ixvolumes should be deleted

AlexKarpov98:
**Testing:**
See ticket, result: (for bulk action and for single button)

👌  Tests added.

Screenshot 2024-10-01 at 11 44 35

|Affects         |Reasoning
|Documentation   | UI Update


undsoft (approved):
Looks good. Checkbox works as expected.

PR #10777, opened 2024-10-01
NAS-131191 / 25.04 / Fix title on Apps Volumn Mounts dialog

denysbutenko:
**Changes:**

Added humanReadableKey pipe to make service name looks better

**Testing:**

Check title on Volumn Mount dialog


stavros-k:
The title in human readable form is already defined in apps side. We shouldn't need to do this here.

I've referenced a ticket in the original one, that might actually cause the issue on the UI

RehanY147 (changes requested):
Please address review by Stevros.

denysbutenko:
> The title in human readable form is already defined in apps side. We shouldn't need to do this here.
> 
> I've referenced a ticket in the original one, that might actually cause the issue on the UI

Okay let's close this then

PR #10791, opened 2024-10-03
NAS-131461 / 25.04 / Listen for apps stats only when have installed apps

denysbutenko:
**Changes:**

Listen for apps stats only when have installed apps

**Testing:**

Check websocket messages

### DIFF
--- a/src/app/pages/apps/store/installed-apps-store.service.ts
+++ b/src/app/pages/apps/store/installed-apps-store.service.ts
@@ -7,6 +7,7 @@ import {
   withLatestFrom,
 } from 'rxjs';
 import { IncomingApiMessageType } from 'app/enums/api-message-type.enum';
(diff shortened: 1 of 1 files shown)


undsoft (changes requested):
`this.installedApps$` will emit every time there is a change to installed apps, leading to more and more subscriptions.

denysbutenko:
> `this.installedApps$` will emit every time there is a change to installed apps, leading to more and more subscriptions.

No extra subscriptions are seen in websocket. Changed operator to `tapOnce` anyway.
image

undsoft (approved):
It's fine, but probably can be optimized further. For example, we won't unsubscribe here when last app is removed, but that's not a big deal.

denysbutenko:
backport

denysbutenko:
backport

PR #10794, opened 2024-10-03
NAS-131020 / 25.04 / Network Widget doesn't show autoconfigured IPv6 address

AlexKarpov98:
Testing: 

See ticket & comments.

Comment from @yocalebo: 👇 

UI is using incorrect logic. The aliases key is the addresses that have been assigned statically via our API and exist in our database. The state.aliases key are the addresses that actually are assigned to the network interface according to the OS.

For a network widget you should only use state.aliases key.

---

Based on that I've updated UI logic.
We should rely on `networkInterface?.state?.aliases` but not on `networkInterface?.aliases`
Screenshot 2024-10-03 at 13 20 09


yocalebo:
This is still using the alias key and not the state.alias key?

AlexKarpov98:
> This is still using the alias key and not the state.alias key?

Oh, sorry, one moment, I need to fix it

PR #10804, opened 2024-10-04
NAS-131147 / 25.04 / Allow to edit yaml installed apps

denysbutenko:
**Changes:**

Allow to edit app that was installed via YAML docker compose

**Testing:**

Install an app via YAML and try to edit it.


undsoft (changes requested):
1. Application name field is empty for me.
2. This doesn't sound like it's part of the wizard. It's better to create new component.
3. Needs tests.
4. Yaml editor doesn't grow or occupy enough space.
Снимок экрана 2024-10-07 в 13 26 11

undsoft (changes requested):
Something wrong with the branch. You are reverting multiple other commits.

undsoft (changes requested):
Looks good, but it appears that installing a custom yaml app or making changes to it does not reload the entry in the table.
If I make edits to my custom app and then click Edit again, my changes are not there and I need to reload.

denysbutenko:
@undsoft There is middleware [issue](https://ixsystems.atlassian.net/browse/NAS-131676) causing this behaviour. I am working on a workaround.

undsoft (commented):
I'm not sure about the workaround, it seems to sometimes work and sometimes when I click quickly enough the old data is still there.
Can we not reload the list manually after custom app is saved without relying on events?

denysbutenko:
> I'm not sure about the workaround, it seems to sometimes work and sometimes when I click quickly enough the old data is still there. Can we not reload the list manually after custom app is saved without relying on events?

There is 5s delay before load new data because response does not contain new app right after creation for some reason (I mentioned this case in bug report). I'll try reduce delay.

undsoft (approved):
Okay, thanks, this feels better.

PR #10812, opened 2024-10-07
NAS-131550 / 25.04 / Add cache busting to icon sprite

AlexKarpov98:
Testing:

See new `sprite-config.json` which holds latest sprite version's url.
The idea here is to update hash if there are any changes to the sprite itself. 
If nothing changed - hash will remain the same.

To see updated hashed url - try to add new/non-used icon somewhere in .html, like `<ix-icon name="mdi-rocket"></ix-icon>`

You will see that `sprite.svg` will be changed as well as hash (`sprite-config.json`).

Screenshot 2024-10-07 at 12 50 14

Screenshot 2024-10-07 at 12 50 08

### DIFF
--- a/scripts/icon-sprite/make-sprite.ts
+++ b/scripts/icon-sprite/make-sprite.ts
@@ -1,3 +1,4 @@
+import crypto from 'crypto';
 import fs from 'fs';
 import * as path from 'node:path';
 import { fileURLToPath } from 'node:url';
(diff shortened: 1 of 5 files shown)


scripts/icon-sprite/make-sprite.ts
```diff
@@ -40,9 +42,15 @@ async function makeSprite(): Promise<void> {
 
     fs.writeFileSync(targetPath, buffer);
 
+    const hash = crypto.createHash('md5').update(buffer).digest('hex').slice(0, 10);
```
denysbutenko, line 45: Any specific reason to prefer `crypto` over `uuid`?

denysbutenko (approved):
Looks good

scripts/icon-sprite/make-sprite.ts
AlexKarpov98: @denysbutenko 

Using UUID from angular2-uuid will indeed generate a unique identifier, but it will not be content-dependent. This means every time you build, it will produce a different value, regardless of whether the content has changed or not. This approach is less ideal for caching purposes, as it will effectively disable caching by creating a new URL each time, even if the sprite.svg content remains unchanged.

denysbutenko: Okay thanks for explaining

PR #10814, opened 2024-10-07
NAS-131484 / 25.04 / Sort disk size column numerically

denysbutenko:
**Changes:**

Fix disk size sorting on disk list

**Testing:**

Go to Storage -> Disks and ensure disks are sorted numerically when disk size sort is active


undsoft (commented):
It may be worth checking other places or building a separate column type that would include correct sorting by default.

denysbutenko:
> It may be worth checking other places or building a separate column type that would include correct sorting by default.

Sorting works as expected it rely on getValue() value which is string in this case. I checked and similar cases no found.

PR #10815, opened 2024-10-07
NAS-131633 / 25.04 / Adding cache to Jest

undsoft:
**Changes:**

Maybe this would make tests faster.

**Testing:**

Once this PR builds, I'll push another commit that should and we can look at whether test job takes less time there.


undsoft:
Second run was 17% faster than the first one. I'm not 100% sure that is not just a fluke, but I'm going to merge it and we'll see what happens.

PR #10816, opened 2024-10-07
NAS-131636 / 25.04 / Loader is not closed when there are errors when stopping …

undsoft:
When I created an app with the following yaml:
```
version: '3.8'
services:
  nginx:
    image: nginx:1-alpine
    ports:
      - 8089:80
    volumes:
      - ./html5up-stellar/:/usr/share/nginx/html
```
I couldn't afterwards stop it and loader won't disappear.
If this yaml doens't work for you, you can just add throws manually.

Known issue: starting an app may flash Crashed status. This will be fixed by middleware.


AlexKarpov98 (approved):
Works well 👍

denysbutenko (approved):

PR #10821, opened 2024-10-08
NAS-131138 / 25.04 / Add dialog for Shutdown/Restart Reason

bvasilenko:
Added a dialog, which now asks the user to enter the reason for shutdown/reboot operation.


undsoft (approved):
It works well overall.
I'd like to make the following adjustments and additional improvements that were not part of the original ticket.
1. Make dialog wider.
2. Instead of having `[] Select from the list` checkbox, let's add `Custom reason` to the list as a top item. When item is selected, show text area.
3. Let's search code for redirects to `system-tasks/reboot` and `shutdown` and add appropriate reasons there.
4. Please check how Audit logs show reboot reason in the table. It always shows `reason: reason` for me.

bvasilenko:
FI blocked by https://ixsystems.atlassian.net/browse/NAS-131689

PR #10824, opened 2024-10-08
NAS-131656 / 25.04 / Better handling for invalid cron schedules

AlexKarpov98:
Testing: see ticket.

Try setting up with invalid cron configuration.

Result:
Screenshot 2024-10-08 at 16 09 48


RehanY147 (changes requested):
Looks good. The table doesn't error out now and you see an error message. Although, since this is an error, we need to make it more apparent with a red color or an error icon next to it in red so it jumps out to the user that there is a problem.

AlexKarpov98:
@RehanY147 
Screenshot 2024-10-09 at 15 49 28

RehanY147 (approved):

AlexKarpov98:
backport

PR #10826, opened 2024-10-09
NAS-131407 / 25.04 / Experiment with alternatives to slide-ins

denysbutenko:
For demo purposes.


undsoft (commented):
Some other challenges that need to be solved:
- Placement of Save button.
- Feels like background dimming should be stronger (at least it is stronger with slide-ins).
- Logging out (for example via timeout) should close open sheets.
- Scrolling. Right now scrollbar is show next to header, but header is sticky. We could try with full page scrolling, where there is a scrollbar on global container and when sheet is scrolled, header sticks to the top of the page. If it's too complicated, we could have scrolling in sheet body only.

Let's focus on one of the forms (user form) for now.

undsoft:
Investigation is complete. Further research is required.

PR #10830, opened 2024-10-09
NAS-131666 / 25.04 / Add support for Stopping state

undsoft:
Known issue: when starting an app you may see Crashed for a bit.


AlexKarpov98 (approved):
Stopping state works fine 👌 

No loaders are shown for Restart or Stop functionality.

denysbutenko (approved):
We may need to check `AppState.Stopped` usage and add new state there if necessary. 

It looks good.

PR #10832, opened 2024-10-09
NAS-131667 / 25.04 / App Notes card does not show View More button until I resize window

AlexKarpov98:
Testing: 
See ticket.

Now you should correctly see View More if it's needed.


undsoft (approved):
View more works fine for me.

PR #10841, opened 2024-10-10
NAS-131693 / 25.04 / Make any required UI repo changes to account for removing option from pool.dataset endpoint

AlexKarpov98:
Code review is enough.


undsoft (approved):
👍🏼 
We can avoid backporting it as there are no visible UI changes.

PR #10843, opened 2024-10-10
NAS-130571 / 25.04 / Impossible to set dialog width

AlexKarpov98:
Tested dialog width setup, it's possible to set width from `matDialog.open()`

I set

```
  volumeButtonPressed(containerDetails: AppContainerDetails): void {
    this.matDialog.open(VolumeMountsDialogComponent, {
      maxWidth: '80vw',
      minWidth: '50vw',
      data: containerDetails,
    });
  }
```

And width is set and it’s working,
we currently have Global Styling to limit modal width to  max-width: 700px;

So I updated `VolumeMountsDialogComponent` view.

Testing:
See ticket.

Result:
Before:
Screenshot 2024-10-10 at 15 44 19

After:
Screenshot 2024-10-10 at 15 44 45


undsoft (approved):
Looks good, but let's make it much wider and do something so that path column occupies more space.

PR #10844, opened 2024-10-10
NAS-125917 / 25.04 / Ask user when they try to close the form after making edits

AlexKarpov98:
**Changes:**
We are using 2 different slideIns. 

First one - regular SlideIn (more used) second one - ChainedSlideIn.
- For wizard we don’t always have formGroup available on the top component level and the logic is tricky where we can swap component to Advanced Form. 

I’ve implemented form-change-guard-for-slide-in.directive directive, and it will only work for SlideIn.
For ChainedSlideIn we gonna need similar solution and I think it worth of a separate ticket.

**Testing:**
See result:

https://github.com/user-attachments/assets/7693faca-a26b-4e45-976b-64bf6ac83d34

### Downstream

|Affects         |Reasoning
|Documentation   | New functionality for the user


RehanY147 (changes requested):
The Yes button shouldn't be primary color. It portrays a positive indication. Meaning clicking Yes is the safe and recommended option. Whereas, No button is the safe bet. Yes should be Red maybe. Or make No primary color and Yes to be default.

AlexKarpov98:
@RehanY147 - that's what's possible:
Screenshot 2024-10-12 at 16 52 50

src/app/pages/network/components/static-route-form/static-route-form.component.html
```diff
@@ -2,7 +2,7 @@
 
 <mat-card>
   <mat-card-content>
-    <form class="ix-form-container" [formGroup]="form" (submit)="onSubmit()">
+    <form class="ix-form-container" formChangeGuardForSlideIn [formGroup]="form" (submit)="onSubmit()">
```
undsoft, line 5: How about just calling it `warnAboutChanges` or `warnAboutUnsavedChanges`.

undsoft (approved):
It works. We still need a version for SlideIn2. Overriding `close` method in slide in ref is creative solution and probably not very kosher, but I can live with it for now.

AlexKarpov98:
For Chained Slide In I've created a separate ticket to implement it:
https://ixsystems.atlassian.net/browse/NAS-131790

RehanY147 (changes requested):
Since we only ever use slide-ins for forms, we should be able to make some assumptions like assume that slide-ins are always forms. And alter the slide-in itself. Instead of creating a new directive that could just be part of the slide-in component itself.

undsoft:
> Since we only ever use slide-ins for forms, we should be able to make some assumptions like assume that slide-ins are always forms.

I disagree. There are slide-ins in certificates that are 90% text, plus I don't really want to end up with a slide-in component that has too many responsibilities.

PR #10845, opened 2024-10-10
NAS-131317 / 25.04 / Replaced Lifetime with Power On Hours

aiden3c:
I replaced the "Lifetime" value in our smart test results with "Power On Hours Ago". This reports how many "power on hours" ago a test happened.
Power on hours are how many hours a disk has been on, the "hours ago" is the disks' total POH - the disks' POH when the test was initiated.

It would be really nice if we could track the actual timestamp of test starts, but smartctl doesn't supply that information. We could track it ourselves, but that becomes very complicated after tests overwrite their own history (scsi only reports the most recent 20 for example). I've thought of some ways to maybe track the actual "hours ago" a test was performed, but it always balloons out. After discussion with Caleb we decided that POH would be the best bet.
Regardless, it's more useful and readable than the raw disk lifetime value.

This has to be paired with [this middleware PR](https://github.com/truenas/middleware/pull/14650), otherwise the key won't exist when getting the results.


aiden3c:
Oh another thing I added was a "toolTip" property to our ColumnComponent, and the ix table header will use that in place of the default tooltip (just reiterating the title) if it exists

src/app/modules/ix-table/interfaces/column-component.class.ts
```diff
@@ -5,6 +5,7 @@ import { DataProvider } from 'app/modules/ix-table/interfaces/data-provider.inte
 export abstract class ColumnComponent<T> {
   propertyName?: keyof T;
   title?: string;
+  toolTip?: string;
```
undsoft, line 8: Better call it `headerTooltip`.

src/app/pages/storage/modules/disks/components/smart-test-result-list/smart-test-result-list.component.spec.ts
```diff
@@ -84,6 +89,7 @@ describe('SmartTestResultListComponent', () => {
       lba_of_first_error: null,
       status: SmartTestResultStatus.Success,
       remaining: 0,
+      poh_ago: 0,
```
undsoft, line 92: Maybe add at least one value, so that it appears in the table below.

src/app/pages/storage/modules/disks/components/smart-test-result-list/smart-test-result-list.component.ts
```diff
@@ -82,10 +82,11 @@ export class SmartTestResultListComponent implements OnInit {
       },
     }),
     textColumn({
-      title: this.translate.instant('Lifetime'),
-      propertyName: 'lifetime',
+      title: this.translate.instant('Power On Hours Ago'),
+      propertyName: 'poh_ago',
+      toolTip: '"Power On Hours" are how many hours have passed while the disk has been powered on. "Power On Hours Ago" is how many power on hours have passed since each test.',
```
undsoft, line 87: Please wrap string with `this.translate.instant` so that it gets extracted into translation messages.

src/app/interfaces/smart-test.interface.ts
```diff
@@ -52,6 +52,7 @@ export interface SmartTestResult {
   status: SmartTestResultStatus;
   status_verbose: string;
   segment_number: number;
+  poh_ago: number;
```
undsoft, line 55: `poh_ago`, such a descriptive name 😅

undsoft (changes requested):
So, I don't think it's going to be obvious to the user that there is a tooltip available over table header. I'd rather we make `headerTooltip` an optional property and when supplied  show tooltip circle (`ix-tooltip`) in the header.

undsoft (approved):
Looks good, tested via mocks. As long as middleware has `power_on_hours_ago` in the response, it'll be shown correctly.

Let's make the following adjustments:
1. Add `(click)="$event.stopPropagation()"` to `ix-tooltip` in `ix-table-head`, so that it doesn't trigger column sorting when tooltip is opened.
2. Align it vertically to be inline with the text.
3. Test fails in part because code picks up test in the tooltip. You can replace `Power On Hours Ago` in expect block with `expect.stringContaining('Power On Hours Ago')` for now.

undsoft (approved):
Looks good, tests need fixing.

PR #10850, opened 2024-10-11
Fix logs header in light theme in Container Logs

p0358:
**Changes:**
The text here is unreadable in the white theme:
obraz

It was due to `color: #fff` being on parent `.logs-wrapper` instead of actual child `.logs` element where it's needed to counteract the black background of terminal window.

**Testing:**
Use light theme and ensure text after header is readable and not white. The logs output should still be white on black.

### Downstream
nothing?


bvasilenko (approved):
Looks good 

image

PR #10851, opened 2024-10-12
NAS-128601 / 25.04 / Use `undefined` instead of `EMPTY`

RehanY147:
**Changes:**

Minor code improvement

**Testing:**

Code review is enough


undsoft:
This is already in 24.10.0.

undsoft (approved):
Also from the list of changes that needs to be brought from 24.10.1 – there is a check for gpu length that prevents user from removing all gpus from the vm.

denysbutenko (approved):

undsoft (commented):
For fix in .0, I've simply removed the `if` condition altogether. If this fix is not correct, then it needs to be fixed in .0 as well.
I do think that we could remove the `if` condition though.

PR #10852, opened 2024-10-12
NAS-131706 / 25.04 / Make `Apps` module components, "stand alone"

RehanY147:
**Changes:**

Changes components to `standalone: true`.

**Testing:**

Test for regressions in the Apps module


undsoft (changes requested):
There is a number of issues, mostly with styles:
- Discover button
- Manage Container Images → Pull Image
- Install via Yaml (the button also doesn't work)
- Show all in Discover filters
- App shell

undsoft (approved):
Looks good, don't see other regressions.

PR #10875, opened 2024-10-16
NAS-128726 / 25.04 / The Scrollbar Position Does Not Reset When Loading New Pages

AlexKarpov98:
Testing: 

See ticket, result:

https://github.com/user-attachments/assets/8bd1e0f4-a810-4fae-9445-fc73e46016ce


bvasilenko (approved):
Scrolling works as expected

PR #10895, opened 2024-10-18
NAS-131270 / 25.04 / Accessibility | Setup pool to install dialog focus

AlexKarpov98:
**Testing:**

See ticket, result:

https://github.com/user-attachments/assets/bdd5840b-f7b5-456e-b482-ee3d38c112e2


undsoft (approved):
It works, although it's weird that we need to do this. I'd imagine material is supposed to handle this.

AlexKarpov98:
> It works, although it's weird that we need to do this. I'd imagine material is supposed to handle this.

There is 

> It works, although it's weird that we need to do this. I'd imagine material is supposed to handle this.

@undsoft - there is **MatDialogConfig** with

```
/**
     * Whether the dialog should restore focus to the
     * previously-focused element, after it's closed.
     */
    restoreFocus?: boolean;
```

But it didn't work for me

PR #10897, opened 2024-10-19
NAS-131802 / 25.04 / Error with some dataset paths in pool creation

undsoft:
**Testing:**

The easier way to reproduce the original issue is to:
1. Go to Shares -> SMB -> Add
2. Select a dataset.
3. Add trailing slash in the input.
4. Click Create Dataset.


AlexKarpov98 (approved):
LGTM 👍

PR #10903, opened 2024-10-21
NAS-131800 / 25.04 / App details page does not scale / cropped on mobile

AlexKarpov98:
Testing: see ticket, 
Result:

Screenshot 2024-10-21 at 12 20 03


bvasilenko (approved):
Works as expected on mobile

PR #10905, opened 2024-10-21
NAS-131843 / 25.04 / Incorrect behavior on alert actions with readonly admin

AlexKarpov98:
Testing: see ticket, 

Screenshot 2024-10-21 at 14 21 18

Error handling added as well.
Screenshot 2024-10-21 at 14 21 59

### DIFF
--- a/src/app/modules/alerts/components/alert/alert.component.html
+++ b/src/app/modules/alerts/components/alert/alert.component.html
@@ -25,6 +25,7 @@
 
   @if (alert().dismissed) {
     <a
+      *ixRequiresRoles="requiredRoles"
       tabindex="0"
       role="button"
       class="action-link"
@@ -35,6 +36,7 @@
     </a>
   } @else {
     <a
+      *ixRequiresRoles="requiredRoles"
       tabindex="0"
       role="button"
       class="action-link"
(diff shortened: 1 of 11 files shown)


undsoft (changes requested):
Снимок экрана 2024-10-21 в 14 09 44


1. Lock icons are far too on the right.
2. Dismiss all alerts button doesn't look disabled.
3. If I remove role checks and click Dismiss / Reopen, because of optimistic updates it stays in new incorrect state. We need to dispatch and action to revert the update on error.
4. Error appears to only be handled once, if I "dismiss" multiple alerts, I will only get an error for the new one.

There also appears to be a middleware issue for `datetime_`: https://ixsystems.atlassian.net/browse/NAS-131910?atlOrigin=eyJpIjoiNzI4ZTAwMDFkNTVjNDY0ZDgzNzg2OGI0MDE2ZDI1NjciLCJwIjoiamlyYS1zbGFjay1pbnQifQ

undsoft (approved):
Looks good, but can we align the lock icon for Dismiss All button as well?

AlexKarpov98:
> Looks good, but can we align the lock icon for Dismiss All button as well?

for `Dismiss all` we can't do it since it's supposed to be a full width button, we can't guess absolute position for the lock icon here.

PR #10913, opened 2024-10-23
NAS-131918 / 25.04 / UI should conditionally show ix-volumes deletion button on app deletion

AlexKarpov98:
Testing: see ticket.
Install some apps.
Try to delete them - with single delete button and with Bulk Delete.
We'll see loader (checking if we should show `Remove IxVolumes`) and then the actual confirm dialog.
Did small refactoring to `onBulkDelete` to split code a little more.

Result:

https://github.com/user-attachments/assets/43512063-2333-4506-8978-530efa1a829b


bvasilenko (approved):
Works as expected

PR #10917, opened 2024-10-24
NAS-131774 / 25.04 / User linked API Keys table

denysbutenko:
**Changes:**

- Add new user linked api keys table. 
- Remove old api keys table.

**Testing:**

Go to user menu and click on `API Keys` to see this page. Or put `/credentials/user-api-keys` in browser manually.

### Downstream

|Affects         |Reasoning
|Documentation   | API Keys page were moved and changed.


undsoft (changes requested):
1. Let's make key textarea readonly, so that I cannot accidentally edit it before copying.
2. Let's use combobox for user select.
3. For most users it complains that user does not have enough privileges to have a key. Please check with Andrew to see if we can filter users here to only show users who can have keys.
4. Even if I create a user and assign him correct privileges, I cannot seem to reassign a key to this user.

PR #10918, opened 2024-10-24
NAS-131722 / 25.04 / Storj Bucket is created before `Add TrueCloud Backup Task` form successful submit

AlexKarpov98:
Now middleware handles new bucket creation if it was provided by UI.
So the idea on UI side is to remove the bucket creation calls for TrueCloud Backup Form.


bvasilenko (approved):
Bucket creation works as expected

PR #10927, opened 2024-10-25
NAS-131839 / 25.04 / Fix cpu core bar

bvasilenko:
**Changes:**

Prefer number of physical cores over number of cores, which is incorrect for most CPUs.

**Testing:**

Ensure that number of cores is correct At the **Dashboard** page, at the **CPU Widget**


undsoft (changes requested):
I removed all widgets from dashboard and add a CPU overview widget.
It fails with:
> Cannot read properties of null (reading '0')
>    at _WidgetCpuComponent.parseCpuData (widget-cpu.component.ts:120:47)
>    at _WidgetCpuComponent.getCpuParams (widget-cpu.component.ts:127:23)

This was seen on a CI machine.

undsoft (approved):
Changes look good, thanks.

bvasilenko:
backport

PR #10932, opened 2024-10-25
NAS-131985 / 25.04 / Add confirmation when restoring default set of widgets

undsoft:
**Changes:**

Because Add and Reset buttons are close together when editing main dashboard, this PR adds a confirmation dialog.


AlexKarpov98 (approved):
Makes sense, nice 👍

PR #10933, opened 2024-10-25
NAS-131862 / 25.04 / Fix reboot call

RehanY147:
**Changes:**

Fixes the way UI calls the reboot command

**Testing:**

Test reboot from the top bar


undsoft (changes requested):
The same issue exists with shutdown and probably with config.reset.

RehanY147:
@undsoft The `config.reset` call is correct. It fails though from middleware but doesn't throw a validation error so I guess it's not UI's fault.

undsoft (approved):
Looks good.

PR #10934, opened 2024-10-26
NAS-132013 / 25.04 / Virtualization state indicator

undsoft:
**Changes:**
Implements state indicator in top right on Containers page.
Partially implements form.

Known issue: state doesn't refresh without reloading the page.

**Testing:**
Easiest is to test with mocks by updating `config.state` in `VirtualizationConfigStore.initialize` to desired value.


AlexKarpov98 (approved):
Works well. 
Known issue is confirmed.

PR #10937, opened 2024-10-27
NAS-131557 / 25.04 / Create new services for separation of concerns

RehanY147:
**Changes:**

Creates new class for separation of concerns

**Testing:**

Code review


undsoft:
Can you explain the intended scope of responsibilities for each of the new major services?

undsoft:
Okay, so I like the idea, but I am somewhat confused by the boundaries between those services.

For example, it's not clear to me why certain things are in `WebSocketHandlerService` and some others are in `WebSocketConnectionService`. I get the intent, but the boundary is fuzzy. E.g. I don't know if it makes sense to have a websocket connection without pings `WebSocketConnectionService`.
If I'm a developer and I need to do something with websocket, do I use `WebSocketHandlerService` or `WebSocketConnectionService`. How do I decide?

Then to Api services. I like the idea of moving events to its own thing. This could work. But I don't understand similar code in `ApiService` and `ApiMethodService`.

I would suggest not paying attention to replacing usages and instead just focus on defining classes first. You could even do this with a diagram or method stubs and comments.

undsoft (commented):
I like `WebSocketConnection` as a class better.
With regard to different api-* services, ultimately I think we should keep one like we used to have. Not that this separation is bad per se, but I don't think the distinction matters that much to consumers and there is going to be additional boilerplate to use separate services.

src/app/services/websocket/websocket-handler.service.ts
```diff
@@ -42,14 +59,64 @@ export class WebSocketHandlerService {
...
+    ).subscribe();
+  }
+
+  private processCall(call: { id: string; [key: string]: unknown }): Observable<unknown> {
+    this.activeCalls++;
+    this.pendingCalls.set(call.id, call);
+    this.wsConnection.send(call);
+
+    return this.responses$.pipe(
+      filter((data: IncomingApiMessage) => data.msg === IncomingApiMessageType.Result && data.id === call.id),
```
undsoft, line 86: What if it's an error?

undsoft (commented):
Haven't tested the code, but the code looks reasonable.

PR #10940, opened 2024-10-28
NAS-132023 / 25.04 / Basic virtualization form

undsoft:
**Changes:**
The form is not finished, but I think it's worth getting it in at this stage to make it easier to work on other tickets.

**Testing:**

You can create an instance in the UI and then verify that it was created via `midclt call virt.instance.query | jq`.

CPU Configuration field accepts numbers.


AlexKarpov98 (approved):
Works good.

`CPU Configuration field accepts numbers.` - maybe we can add `type=number` then?

PR #10941, opened 2024-10-28
NAS-131949 / 25.04 / Finish Global Settings form in containers

AlexKarpov98:
Testing:

see ticket, result:

https://github.com/user-attachments/assets/5a6ca09e-a587-47b0-8b37-6b5ff861690d


undsoft (changes requested):
Okay, so there some issues with our design:

1. Bridge dropdown already contains `Automatic` option.
2. If I save everything with `automatic` for IPv4 and IPv6, and then go to edit, I'll see some values already there and checkboxes unticked.
I'm open to ideas on how to present this in a better way.

undsoft:
Also this PR added `virt.global.config` event, so you can subscribe to changes in the store.

AlexKarpov98:
@undsoft  

```
you: If I save everything with automatic for IPv4 and IPv6, and then go to edit, I'll see some values already there and checkboxes unticked.
I'm open to ideas on how to present this in a better way.
```
Actually it works fine with page reload.

I was able to fix it with `Page reload`, I tried subscribing to `virt.global.config` but I don't get any events.
If you reload the page - you will see up to date form values.

undsoft:
That's not what I mean. What I mean is that it's confusing that I select "Automatic IPv4 Network", save and then see some IP address and "automatic" checkbox is unchecked.

AlexKarpov98:
@undsoft now it should work fine.
Please use latest nightly for testing.

https://github.com/user-attachments/assets/df340552-0618-475a-88d5-c850fb5dfb87

undsoft (approved):
Not sure about the behavior of "automatic" IPv4 and IPv6 checkboxes on first save, but otherwise looks good.

AlexKarpov98:
Final solution agreed with William is to only show `IPv4 Network` & `IPv6 Network` for AUTO bridge.

As well `IPv4 Network` & `IPv6 Network` should be there non required and if it's empty we should send `null` to MV (in this case it will be regenerated)

https://github.com/user-attachments/assets/b903b930-c68a-4662-90a0-93177dfe5e66

PR #10949, opened 2024-10-30
NAS-113704: Filter user options by roles

bvasilenko:
**Testing:**

Please see steps/actual/expected at the ticket


AlexKarpov98 (approved):
`root` is not available on the `User Quotas` `Apply To Users` input.
But I can still type `root` manually and save the form and get the validation error.
Screenshot 2024-11-01 at 13 01 59

We should use `allowNewEntries=false` to disallow new items.
here:
```
          <ix-chips
            formControlName="users"
            [label]="'Apply To Users' | translate"
            [tooltip]="tooltips.users"
            [required]="true"
            [autocompleteProvider]="usersProvider"
            [allowNewEntries]="false"
          ></ix-chips>
```

PR #10951, opened 2024-10-30
NAS-132100 / 25.04 / Migrate eslint to flat config

undsoft:
**Changes:**

* Eslint 9 moves to new flat-config format. This updates our configuration as well.
* Some rules were moved out of eslint to @stylistic package.
* `eslint-config-airbnb` does not support flat config yet and `eslint-config-airbnb-typescript` is not maintained, so I moved individual relevant rules from these configs to our configs.
* Some rules didn't work correctly and I have disabled them for now.

**Testing:**

Review changes, play with the linter.


denysbutenko (approved):
`ignores` paths are missing which causes many errors, other than that looks good.

PR #10959, opened 2024-11-01
NAS-131913 / 24.10.1 / Fix the sequence of state updates

RehanY147:
**Changes:**

We were subscribing to only docker store loading boolean. Should also be listening to docker store state update observable to know when to update installed apps store again.

**Testing:**

Testing system and method described in the ticket


AlexKarpov98 (approved):
Works as expected 👍 

https://github.com/user-attachments/assets/9ad8a03f-577e-4ef2-ade0-01163e2e83a6

PR #10965, opened 2024-11-04
NAS-131761 / 25.04 / Enabled toggle in Shares>NFS does not work

AlexKarpov98:
**Testing:**
See ticket.

Logic is taken from the NFS card.

Result:

https://github.com/user-attachments/assets/084d7cd1-b87b-40ba-b982-ca6ff7f69e70


bvasilenko (approved):
Works as expected

PR #10966, opened 2024-11-04
NAS-131747: "Update available" tooltip for Apps shows wrong version

AlexKarpov98:
**Testing:**
See ticket. 

Middleware updated us with `latest_version` field, so we used it.


undsoft (approved):
Looks good, the version is there.

PR #10969, opened 2024-11-04
NAS-132239 / 25.04 / Make it clearer what ports mean in app workloads

undsoft:
**Changes:**

Previously information about ports was underlined which people might have interpreted as links. This just adds a standard tooltip instead.

Also makes unrelated code quality change in a different place.

**Testing:**

Due to an unrelated issue you may need to replace implementation of `getLacksNvidiaDrivers()` with `of(false)` in `DockerStore`.


AlexKarpov98 (approved):
Looks good 👍

PR #10971, opened 2024-11-04
NAS-131978 / 25.04 / Instances table and layout

denysbutenko:
**Changes:**

Implement instances table and layout

**Testing:**

Go to Containers. Check original ticket.


src/app/enums/docker-nvidia-status.enum.ts
```diff
@@ -0,0 +1,17 @@
+import { marker as T } from '@biesbjerg/ngx-translate-extract-marker';
+
+export enum DockerNvidiaStatus {
```
undsoft, line 3: Should part of a different PR because we would need to backport it.

src/app/pages/virtualization/components/all-instances/instance-list/instance-list.component.ts
```diff
@@ -0,0 +1,171 @@
...
+    const selected = instanceId && this.instances().find((instance) => instance.id === instanceId);
+    if (selected) {
+      this.store.selectInstance(selected.id);
+      return;
+    }
+
+    this.selectFirstApp();
+  }
+
+  private selectFirstApp(): void {
```
undsoft, line 130: Method needs to be renamed.

src/app/pages/virtualization/components/all-instances/instance-list/instance-list.component.ts
```diff
@@ -0,0 +1,171 @@
...
+    if (first) {
+      this.store.selectInstance(first.id);
+    }
+  }
+
+  closeMobileDetails(): void {
+    this.showMobileDetails.set(false);
+  }
+
+  restart(instanceId: string): void {
```
undsoft, line 142: Any particular reason to do it here and not in the child component?

src/app/pages/virtualization/components/all-instances/instance-list/instance-list.component.html
```diff
@@ -0,0 +1,58 @@
+<div class="container">
+  <div class="table-container">
+    <div class="table-header">
+      <h2>{{ 'Instances' | translate }}</h2>
```
undsoft, line 4: Let's try without header here and in details.

undsoft (approved):
Looks good overall.

* The controls column could use more breathing room, right now everything is very glued to the right.
* When I use search and there are no results, the empty text should be updated.

src/app/pages/virtualization/components/all-instances/instance-list/instance-list.component.ts
denysbutenko: The idea is to separate logic from `ix-instance-row` using it solely for representation purposes.

src/app/enums/docker-nvidia-status.enum.ts
denysbutenko: Wrong branch. Changes were reverted

PR #10980, opened 2024-11-05
NAS-132071 / 25.04 / App WebUI does not work after migration

AlexKarpov98:
Testing: see ticket.

Check `WebUI button` inside app info card.


undsoft (changes requested):
So if the URL is coming from middleware and is incorrect, why not just reassign the ticket to middleware?

AlexKarpov98:
Fixed on the MV side. 👌

PR #10981, opened 2024-11-05
NAS-132076 / 25.04 / Interface to start / stop a container/app is hidden at default zoom level

AlexKarpov98:
Testing: 
See ticket.

Before:
before-NAS-132076

After:
after-NAS-132076

Result:

https://github.com/user-attachments/assets/91a263f0-d553-4294-9f55-4f18b9b13c18


bvasilenko (approved):
Works as expected

PR #11001, opened 2024-11-08
NAS-132227 / 25.04 / Fix enable change in SMB table

bvasilenko:
**Testing:**

On the **Shares** page

At the **SMB** block, 

1. User sets a share to **Disabled** state.

1. User clicked **View all** to navigate to the distinguished **SMB** page.

2. User sets this share to **Enabled** state.

2. User goes back to **Shares** page, and observes the **SMB** block (and optionally refreshes the page).

- **Expected result:** share is in **Enabled** state


RehanY147 (approved):
Works as described

PR #11005, opened 2024-11-10
NAS-131790 / 25.04 / Defines `SlideIn2CloseConfirmation` interface for confirmation before closing slide-ins

RehanY147:
**Changes:**

Defines `SlideIn2CloseConfirmation` interface for confirmation before closing slide-ins

**Testing:**

Code review
Open the `ssh-connection-form` under `Credentials -> Cloud Credentials` and edit the form and close without saving. You should see a confirmation dialog asking if you're sure
Test on the `ReplicationWizard` under `Data Protection` and edit the form and close without saving. You shouldn't see a confirmation dialog and the dialog should close without a problem.

### DIFF
--- a/src/app/interfaces/slide-in-close-confirmation.interface.ts
+++ b/src/app/interfaces/slide-in-close-confirmation.interface.ts
@@ -1,0 +1,5 @@
+import { Observable } from 'rxjs';
+
+export interface SlideIn2CloseConfirmation {
+  requiresConfirmationOnClose(): Observable<boolean>;
+}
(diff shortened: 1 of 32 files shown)


undsoft (changes requested):
SshConnectionForm presents the dialog even when I click Save.

Also how about instead of making it a method/interface in parent class, we change it to something more explicit in slideInRef:
```
this.slideInRef.requireConfirmationWhen(() => { ... });
```

undsoft (changes requested):
Waiting for feedback to be addressed.

PR #11011, opened 2024-11-11
NAS-132102 / 25.04 / Use usernames and groups in permissions

undsoft:
**Changes:**

Permissions and ACL editor has been adapted to API changes and now send username/groupname instead of ids.

**Testing**
Test trivial permissions and acl editor.


denysbutenko (approved):
No regressions are found. Looks good.

PR #11012, opened 2024-11-11
NAS-132301 / 25.04 / Adapt changes for boot environments

denysbutenko:
**Changes:**

Adapt new boot environment endpoints.

**Testing:**

Check Boot Environments.


undsoft (changes requested):
Some UI and some middleware changes:

- When I change scrub interval in scrub settings, old value is still shown in the dialog next time I open it until I reload the page.
- Cloning, activating or changing keep status results in `[EINVAL] result: Input should be a valid dictionary or instance of BootEnvironmentEntry`
- Date Created column shows different dates compared to old version
- Space column can be called Used Space like used to be
- Delete dialog is missing text prompt: "The following boot environment will be deleted. Are you sure you want to proceed" and there is no Delete button.
- Clone form shouldn't close on errors.

denysbutenko:
> Some UI and some middleware changes:
> 
> * When I change scrub interval in scrub settings, old value is still shown in the dialog next time I open it until I reload the page.

Fixed

> * Cloning, activating or changing keep status results in `[EINVAL] result: Input should be a valid dictionary or instance of BootEnvironmentEntry`

Fixed in recent nightly middleware

> * Date Created column shows different dates compared to old version

This is possibly due to the date/time format saved in local storage. We still use the same field and table column.

> * Space column can be called Used Space like used to be

Done

> * Delete dialog is missing text prompt: "The following boot environment will be deleted. Are you sure you want to proceed" and there is no Delete button.

Fixed

> * Clone form shouldn't close on errors.

Fixed

PR #11013, opened 2024-11-12
NAS-131595 / 25.04 / Prepare redirect URL for IPv6

bvasilenko:
## Testing

User configures an app to use IPv6 Address for it's web interface. 

**Steps to replicate**

1. When logging into the front-end by entering IPv4 address ( 127.0.0.1:4200 ) into the address bar, 
2. User configures some app to use IPv6 Address for it's web interface
3. User logs out. And then logs in again, but by entering IPv6 address of the server into the address bar
4. Go to **Apps**, choose this App, and click button to access app's web interface.

    Please also see video attachment in the ticket.

**Expected Result**

- The Apps can be accessed using both Ipv4 and IPv6 URL's


AlexKarpov98 (commented):
Seems like a similar fix I did here:
https://github.com/truenas/webui/pull/10980

Probably should be fixed on MV side?

@undsoft, could you please check ?

undsoft:
I don't think this is the same issue. https://github.com/truenas/webui/pull/10980 was about middleware returning incorrect IPv6 URLs when they were set, this one is about opening IPv6 current hostname when app portal ip is not available and user connects to TrueNAS using ipv6.

AlexKarpov98 (approved):
👍

PR #11014, opened 2024-11-12
NAS-132263 / 24.10.1 / Change TrueCloud Backup form to have a Transfer setting

undsoft:
**Changes:**

Bandwidth Limit and Transfers fields are removed from TrueCloud Backup form. There is now a Transfer Setting dropdown instead.

### Downstream

|Affects         |Reasoning
|Documentation   | Field changes in the form.


bvasilenko (approved):
The form works as expected

PR #11017, opened 2024-11-12
NAS-132466 / 25.04 / Remove some usages of unknown

undsoft:
**Changes:**

Using `as unknown as Something` while may sometimes be necessary can also hide errors and should be avoided when possible.


denysbutenko (approved):
Looks good

PR #11018, opened 2024-11-12
NAS-132468 / 25.04 / Fix UI docker build

undsoft:
**Testing:**

Once docker job is finished, run
```
docker container run -it -e TNIP=192.168.0.30 -p 8080:80 ixsystems/truenas-webui:11018
```

replace IP with IP of actual system and verify that UI can connect.


AlexKarpov98 (approved):
👍

PR #11030, opened 2024-11-13
Use the new services for websocket connection

RehanY147:
**Changes:**

Replaces `api-service` with the new `api-service`

**Testing:**

Test for regressions across the system


undsoft (commented):
So far, looks good.
There are commented out tests in `ApiService` and commented out dependency in constructor in `TokenLastUsedService`.
I'll review again once tests are fixed.

undsoft (commented):
> There are commented out tests in ApiService and commented out dependency in constructor in TokenLastUsedService.

This still needs to be fixed.

undsoft (approved):
Looks good, but there is still a bunch of commented-out code in `auth.service.spec.ts`.

PR #11034, opened 2024-11-14
NAS-132484 / 25.04 / Add stop options to containers

undsoft:
**Changes:**

When container is being stopped or restarted, there is now a dialog with options.

Снимок экрана 2024-11-18 в 09 27 50


bvasilenko (approved):
Works as expected

PR #11041, opened 2024-11-14
Fix error translation in fr.json

MDVAurelien:
Fix error translation with NTP Burst. This word doesn't need to translate.


undsoft (approved):
👍🏼

PR #11045, opened 2024-11-15
NAS-132456: Various fixes for containers

AlexKarpov98:
Testing: see ticket.

@dszidi  I used that icon, what do you think about it:
Screenshot 2024-11-15 at 15 30 50

As well we may use some custom icon you many provide?


denysbutenko (approved):
image

Looks good

PR #11052, opened 2024-11-17
NAS-132411 / 25.04 / Sort image size column numerically

bvasilenko:
**Testing:**

Please see ticket for repro steps/expected


AlexKarpov98 (approved):
Sorting works correctly 👌

PR #11053, opened 2024-11-17
NAS-131790 / 25.04 / Add option to add confirmation requirement

RehanY147:
**Changes:**

Adds a method in `chained-ref` to enable showing confirm dialog for forms when they are in a changed state so the user doesn't close the form accidentally.

**Testing:**

Test replication form and wizard. Change values and try to close or swap the forms and the confirmation dialog should appear asking if you're sure you want to close this form.


undsoft (changes requested):
Code looks okay, but there are two UX issues:
1. I can click on backdrop and form closes without a warning.
2. If I press Save, I am still presented with a warning dialog.

src/app/services/chained-slide-in.service.ts
```diff
@@ -30,6 +30,7 @@ export interface ChainedComponent {
 export interface ChainedComponentResponse<T = unknown> {
   response: T;
   error: unknown;
+  cancelled?: boolean;
```
undsoft, line 33: I think we could just look at `response` and check if it's falsy.

undsoft (approved):
It works as expected 👍🏼 
Let's add some tests.

PR #11057, opened 2024-11-18
NAS-132581 / 25.04 / NAS-132550: Restricted admin - cloud sync task details not loaded

AlexKarpov98:
Testing: see ticket.
Code review may be enough.


undsoft (approved):
Indicator stops now. While we are here let's also fix the same issue in cloudsync wizard.

PR #11066, opened 2024-11-19
NAS-132299 / 25.04 / Implement virtualization metrics

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/158a24a9-3405-45cb-b0cc-e3e18a39014e


undsoft (approved):
Very nice. Good job.

Two things I've noticed:
1. I get `error-handler.service.ts:38 TypeError: Cannot read properties of undefined (reading 'cpu_user_percentage')
    at instance-metrics.component.ts:78:30` when I start the instance while looking at details.
2. Times on the chart should be in system timezone.

undsoft:
It may also be a good idea to have some units on the vertical axes.

AlexKarpov98:
@undsoft I see that other charts, like on dashboard are not in timezone as well, So possibly it worth a separate ticket for checking charts?

For other stuff, I updated it all to look like:
Screenshot 2024-11-19 at 15 09 53

What do you think?

undsoft:
Okay, let's solve it in another ticket.
Units look okay, but we'd need to check later what happens when memory usage is Gigabytes.

AlexKarpov98:
> Okay, let's solve it in another ticket. Units look okay, but we'd need to check later what happens when memory usage is Gigabytes.

Memory will always be in MiB for us, it's taken from `mem_usage_ram_mib` field.

PR #11068, opened 2024-11-19
NAS-132297 / 25.04 / Add virtualization view and devices stores

denysbutenko:
**Changes:**

Add virtualization view and device stores.

**Testing:**

Check original ticket for testing.


undsoft (approved):
Looks good overall.
Let's try and see if we can switch to mobile details later. The table doesn't have too many columns and I feel like we could still show details here: 
Снимок экрана 2024-11-20 в 09 31 46

Plus condition for right chevron is not the same as for hiding the details.

PR #11072, opened 2024-11-20
NAS-132583: Add mass actions to containers

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/144dbcad-17af-4483-9a06-6681716ff5f7


src/app/pages/virtualization/components/all-instances/instance-list/instance-list.component.ts
```diff
@@ -28,30 +44,76 @@ import { VirtualizationInstancesStore } from 'app/pages/virtualization/stores/vi
...
+      (instance) => [VirtualizationStatus.Running].includes(instance.status)
+        && this.selection.isSelected(instance.id),
+    );
+  });
+
+  protected readonly stoppedCheckedInstances = computed(() => {
+    return this.instances().filter(
+      (instance) => [VirtualizationStatus.Stopped].includes(instance.status)
+        && this.selection.isSelected(instance.id),
+    );
```
undsoft, line 116: See if you can extract it into it's own component.

AlexKarpov98: @undsoft - not sure what do you mean, exactly?

src/app/pages/virtualization/components/all-instances/instance-details/instance-general-info/instance-general-info.component.ts
```diff
@@ -70,6 +70,7 @@ export class InstanceGeneralInfoComponent {
       title: this.translate.instant('Delete'),
       message: this.translate.instant('Delete {name}?', { name: this.instance().name }),
     }).pipe(
+      filter(Boolean),
```
AlexKarpov98, line 73: Fixed a but when we can delete without confirmation just by closing modal.

src/app/pages/apps/components/installed-apps/installed-apps.component.html
```diff
@@ -48,7 +48,6 @@ <h2>{{ 'Applications' | translate }}</h2>
               mat-button
               ixTest="bulk-actions-menu"
               [matMenuTriggerFor]="menu"
-              [disabled]="!hasCheckedApps"
```
AlexKarpov98, line 51: removed useless code singe we are in condition `@if (hasCheckedApps) {}` above.

src/app/pages/virtualization/components/all-instances/instance-list/instance-list.component.html
```diff
@@ -1,5 +1,16 @@
 <div class="container">
   <div class="table-container">
+    <div class="table-header">
+      <h2>{{ 'Instances' | translate }}</h2>
+
+      @if (checkedInstances?.length > 0) {
+        <ix-instance-list-bulk-actions
```
AlexKarpov98, line 7: We may want the same for the `installed apps`, or even make a one reusable component for both. At this moment I only improved code to have separated component for instances bulk list actions.

Screenshot 2024-11-21 at 14 35 26

AlexKarpov98: @undsoft - you may want to create a ticket for this suggestion?

bvasilenko (approved):
Works as expected
One additional thing I noticed, is that the status of a container won't update until the user navigates away and returns to the Containers page, however I'm not sure if this should be resolved at the scope of this particular PR. Should we create a separate issue for this one?

undsoft (approved):
Looks good, but I noticed an issue that when I select two instances and use start/stop only second instance gets updated status. If I reload the page, both instances get new state.

AlexKarpov98:
> Works as expected One additional thing I noticed, is that the status of a container won't update until the user navigates away and returns to the Containers page, however I'm not sure if this should be resolved at the scope of this particular PR. Should we create a separate issue for this one?

@bvasilenko - that's because we have middleware issues now, update operations are not sending latest data on update. It's a known issue (with a ticket)

PR #11080, opened 2024-11-20
NAS-132648 / 25.04 / query param handling for auth tokens

aervin:
**Changes:**

**Testing:**
One method:
* clear `localStorage`
* you'll have to comment out the `https:` check in `AuthService.setQueryToken`
* `midclt call auth.generate_token`
* append the token to an app route: `localhost:4200/storage/disks?token=<token>`

### Downstream

|Affects         |Reasoning
|Documentation   |

### DIFF
--- a/src/app/pages/signin/store/signin.store.ts
+++ b/src/app/pages/signin/store/signin.store.ts
@@ -1,6 +1,6 @@
 import { Inject, Injectable } from '@angular/core';
 import { MatSnackBar } from '@angular/material/snack-bar';
-import { Router } from '@angular/router';
+import { ActivatedRoute, Router } from '@angular/router';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { ComponentStore } from '@ngrx/component-store';
 import { Actions, ofType } from '@ngrx/effects';
@@ -81,6 +81,7 @@ export class SigninStore extends ComponentStore<SigninState> {
     private authService: AuthService,
     private updateService: UpdateService,
     private actions$: Actions,
+    private activatedRoute: ActivatedRoute,
     @Inject(WINDOW) private window: Window,
   ) {
     super(initialState);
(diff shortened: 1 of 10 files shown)


aervin:
The approach so far is to forward query params on redirect to `/signin` (`AuthGuardService:canActivate`). From there, add a bit of logic to `signin.store` and `auth.service` to handle the token.

However, this isn't working like I'd expect. Seems that `WebsocketHandlerService` is calling its `resetUI` method and interfering with the redirect.

image

undsoft (commented):
Seems to be redirecting fine for me: 

https://github.com/user-attachments/assets/c75045a4-d1b8-496f-b0d2-53d4a0cfce22

undsoft:
BTW, I don't know if tokens in browser history is kosher.

src/app/pages/signin/store/signin.store.ts
```diff
@@ -35,13 +35,15 @@ interface SigninState {
...
   loginBanner: string;
+  queryToken: string | null;
 }
 
 const initialState: SigninState = {
   isLoading: false,
   wasAdminSet: true,
   failover: null,
   loginBanner: null,
+  queryToken: null,
```
undsoft, line 46: I don't think it needs to be part of state here and have a selector. It's not used outside of the component.

src/app/pages/signin/store/signin.store.ts
```diff
@@ -240,8 +248,14 @@ export class SigninStore extends ComponentStore<SigninState> {
   }
 
   private handleLoginWithToken(): Observable<LoginResult> {
```
undsoft, line 250: Method gets a bit convoluted with all the conditions. How about introducing a new method for logging in with query token and either calling `handleLoginWithToken` or that new method depending on whether query parameter set.

undsoft (approved):
The code looks good, but I just couldn't get it to work on a CI machine.
I can see token being sent to middleware, but it just returns `AUTH_ERR` for some reason.

RehanY147 (approved):

aervin:
Went to retest against a CI machine and the `hasOpened` logic was already flaky/broken. As an alternative approach, I've manually preserved query params on redirect in the `resetUi` method.

Seems to work fine against a CI machine with both `root` and non-builtin users. To prove I'm not crazy:
nas-redirect

undsoft (commented):
I still get the same issue:

https://github.com/user-attachments/assets/24c2d79c-44e6-43c4-a666-076923222e1b

PR #11086, opened 2024-11-22
NAS-132019 / 25.04 / Remove DST gaps from chart reports

bvasilenko:
**Changes:**

Screenshot attached to a ticket depicts the **DST gap**, which occurs when server's internal clock does a transition from time point `1:59` to time point `3:00 GMT+0100`, and timezone info (`GMT+0100`) is lost, resulting in representational **1:01 hour gap** 

In pseudocode:

Actual: 

```typescript
> new Date(`Jan 01 2024 01:59:00`) - new Date(`Jan 01 2024 03:00:00`)
  -> -3660000
```

Expected: 

```typescript
> new Date(`Jan 01 2024 01:59:00 GMT+0000`) - new Date(`Jan 01 2024 03:00:00 GMT+0100`) 
  -> -60000
```

This PR removes DST gaps from reporting time series.

**Testing:**

To simulate DST, you need to add the `addDstGaps` method in the **src/app/pages/reports-dashboard/reports.service.ts** file and apply it to `reportingData.data` as shown below:

image

Replace the constant `dstTimestamp` with the value of the DST change time (**in seconds!**).

For example, `1732168676` \<=\> `Thu Nov 21 2024 08:57:56 GMT+0300`.

A gap of one hour should appear on the chart:

image

Using the `removeDstGaps` method should remove this gap.

`addDstGaps` method code:

```typescript
addDstGaps(data: number[][]): number[][] {
    data.forEach((item) => {
      const dstTimestamp = new Date().getTime() / 1000 - 10 * 60; // Set DST time, e.g. 1732168676
      if (item[0] > dstTimestamp) {
        item[0] = item[0] + 3600;
      }
    });
    return data;
}
```

### DIFF
--- a/src/app/pages/reports-dashboard/reports.service.spec.ts
+++ b/src/app/pages/reports-dashboard/reports.service.spec.ts
@@ -38,4 +38,46 @@ describe('ReportsService', () => {
       expect(spectator.inject<Window>(WINDOW).open).toHaveBeenCalledWith('/netdata/');
     });
   });
+
+  describe('removeDstGaps', () => {
+    it('returns normal data without changes', () => {
+      expect(spectator.service.removeDstGaps([
+        [1732000500, 0],
+        [1732000501, 1],
+        [1732000502, 2],
+      ])).toEqual([
+        [1732000500, 0],
+        [1732000501, 1],
+        [1732000502, 2],
+      ]);
+    });
+
+    it('removes gap from data', () => {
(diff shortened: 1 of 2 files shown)


undsoft:
So if I understand the PR correctly, middleware returns timestamps that may jump from some value to another hour. If this is the case, then this is not a UI issue as timestamps don't care about DST.

PR #11088, opened 2024-11-22
NAS-132670 / 25.04 / Add breadcrumbs for Containers (as we have for apps)

AlexKarpov98:
Testing: 
see:

Screenshot 2024-11-22 at 12 16 40
Screenshot 2024-11-22 at 12 16 44
Screenshot 2024-11-22 at 12 16 49


bvasilenko (approved):
Works as expected

PR #11094, opened 2024-11-25
NAS-132643 / 24.10.2 / Rework `ping` logic

bvasilenko:
**Testing:**

Enter login page, authenticate, and then logout. Watch messages sent over websocket in browser's network log

**Expected results:**

1. **ping** should be sent only when user is authenticated. In other words, **ping** should not be sent on "public" pages, such as login screen
2. call `core.ping` should be done instead of sending `{ msg: 'ping' }`


undsoft (approved):
Looks good, clean work.

PR #11100, opened 2024-11-26
NAS-132671 / 25.04 / Initial Containers Loading - Hide empty component until data is received

bvasilenko:
**Changes:**

- Show loader until we know the actual response. 
- Show empty component only if we don’t have instances.

**Testing:**

Ovserve the initial loading of the **Containers** page


RehanY147 (changes requested):
Having a boolean just to identify when data is loaded doesn't feel like a good solution. Maybe instead, we can use `undefined` or `null` initially and then just check for a non-null value to identify if the data has been loaded at least once. Replacing `undefined` with `[]` for example, when the request has completed.

RehanY147 (approved):

PR #11102, opened 2024-11-26
NAS-132750: Minor improvements for containers

undsoft:
**Changes:**

* Clarifies what it means not to set CPU or memory.
* Shows environment variables in a tooltip.


src/app/enums/virtualization.enum.ts
```diff
@@ -13,11 +13,13 @@ export const virtualizationTypeLabels = new Map<VirtualizationType, string>([
 export enum VirtualizationStatus {
   Running = 'RUNNING',
   Stopped = 'STOPPED',
+  Unknown = 'Unknown',
```
william-gr, line 16: UNKNOWN, uppercase?

bvasilenko (approved):

src/app/enums/virtualization.enum.ts
undsoft: Indeed.

PR #11106, opened 2024-11-28
NAS-132372 / 25.04 / Add `/dev/zvol` root for explorer

RehanY147:
**Changes:**

Adds option to load zvols alongside datasets by providing two root paths. `/dev/zvol` alongside `/mnt` when calling `filesystem.listdir` endpoint. `ix-explorer` `root` property should be set to `/` for this. Default `root` for `ix-explorer` is `/mnt`

**Testing:**

Test on the TrueCloud Task Form. The path explorer should show `'/'` as it's root. When expanded, it should show two children paths. `'/mnt'` and `'/dev/zvol'`. When expanded, `'/mnt'` should show the normal datasets and `'/dev/zvol'` should show the zvols.


src/app/services/filesystem.service.ts
```diff
@@ -9,6 +9,12 @@ import { ExplorerNodeData, TreeNode } from 'app/interfaces/tree-node.interface';
 import { TreeNodeProvider } from 'app/modules/forms/ix-forms/components/ix-explorer/tree-node-provider.interface';
 import { ApiService } from 'app/services/websocket/api.service';
 
+export interface ProviderOptions {
+  directoriesOnly?: boolean;
+  showHiddenFiles?: boolean;
+  includeSnapshots?: boolean;
+  dsAndZvols?: boolean;
```
undsoft, line 16: let's not abbreviate

undsoft (changes requested):
We may also want to address the `[ENOENT] Directory /dev/zvol does not exist` error that may happen if user does not have any zvols.

undsoft (approved):
A bit hacky, but it works.
Can we change it so that the icon used for zvols is not a generic file icon?

PR #11110, opened 2024-11-28
NAS-132760 / 25.04 / Use master-detail-view on iSCSI Targets page

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/a13b00f9-bb71-4fa0-b457-a1d2a186f7d7


src/app/modules/master-detail-view/master-detail-view.component.html
```diff
@@ -26,6 +26,8 @@ <h3 class="title">
             <ng-content select="[detail-name]"></ng-content>
           </span>
         </h3>
+
+        <ng-content select="[detail-actions]"></ng-content>
```
undsoft, line 30: Can we just have one slot for both name and actions?

undsoft (approved):
Looks good. Let's not make Delete button red, it draws too much attention.

src/app/modules/master-detail-view/master-detail-view.component.html
AlexKarpov98: > Can we just have one slot for both name and actions?

@undsoft not really, actions should to be outside of the title and back button. Not sure it's possible.

PR #11111, opened 2024-11-28
NAS-132758 / 25.04 / Container Metrics | Limit data aggregator with let's say 300 latest items

AlexKarpov98:
Testing: see ticket.

Result: (now we limited each metric to 300 items max to be shown to the user ~ 10 minutes of time frame)

https://github.com/user-attachments/assets/b505d56f-c275-4d16-a858-3e0cc57a771e


src/app/pages/virtualization/components/all-instances/instance-details/instance-metrics/instance-metrics.component.ts
```diff
@@ -35,6 +35,7 @@ import { ApiService } from 'app/services/websocket/api.service';
 })
 export class InstanceMetricsComponent {
   instance = input.required<VirtualizationInstance>();
+  maxItemsCount = 60 * 10 / 2; // latest 10 minutes with 2 seconds API interval
```
bvasilenko, line 38: I've noticed that `maxItemsCount` is a local variable, however it would be better to make it a constant, to clearly state that it won't be updated in a runtime.

bvasilenko (approved):
Looks good, just one suggestion of minor code improvement

PR #11119, opened 2024-11-29
NAS-128390 / 24.10.2 / Adjust children selection options for draid pools

RehanY147:
**Changes:**

Adjusts min allowed child count when creating dRaid pools

**Testing:**

The children select's minimum value when creating a draid pool should be calculated based on this formula

```
max = Math.min(maxAvailableDisks, 255)
parity = 1 if dRaid1, 2 if dRaid2, 3 if dRaid3
min = (data vdev disks) + (hot spares count) + parity
```


undsoft:
I don't think we need this in Electric Eel.

PR #11121, opened 2024-12-02
NAS-132725 / 25.04 / UI - Do not allow user to delete themselves

AlexKarpov98:
Testing: see ticket. 

- Create another admin user.
- Login as another admin user and try delete that new user.  (it's impossible now)


bvasilenko (approved):
Works as expected

PR #11122, opened 2024-12-02
NAS-132751 / 25.04 / Cloud Credentials: It is not possible to delete cloud credentials

AlexKarpov98:
Testing: 

See custom error as an example:

https://github.com/user-attachments/assets/4b8e9bdb-7885-4989-9db3-80f43a01525c


undsoft (approved):
Error is shown now.

PR #11124, opened 2024-12-02
NAS-132790 / 25.04 / Fix IxSimpleChanges

undsoft:
**Changes:**

Adapts IxSimpleChanges to work with signal inputs and converts some inputs.


AlexKarpov98 (approved):
Looks & Works good.
Tests should be fixed 👌

PR #11126, opened 2024-12-02
NAS-132675 / 24.10.2 / Remove `nvidia` state and event handlers

undsoft:
**Testing:**

There is no suitable machine for testing, but you can create an Electric Eel machine and update `AppsSettingsComponent` to have `protected hasNvidiaCard$ = of(true);`


denysbutenko (approved):
image

Looks good

PR #11146, opened 2024-12-05
NAS-132722 / 25.04 / Add fibre channel ports list and form

denysbutenko:
**Changes:**

Add fibre channel ports list and form

**Testing:**

Check Fibre Channel Ports tab on iSCSI share page


undsoft (approved):
* When I remove an item from the table, I need to refresh the page to see change.
* Create New in target doesn't seem to work. We can remove this option.

I'm also now not sure if this is how the form is supposed to work, but it's okay to merge this and fix later if necessary.

PR #11147, opened 2024-12-05
NAS-132833 / 25.04 / Add Fibre Channel related fields

AlexKarpov98:
Testing: see ticket.

Result:
Screenshot 2024-12-05 at 14 21 09
Screenshot 2024-12-05 at 14 21 14

Screenshot 2024-12-05 at 14 21 04


bvasilenko (approved):
Looks good overall, however there's a noticeble "jumping"  of the UI, during fraction of second right after the form is displayed:

https://github.com/user-attachments/assets/471e41da-26c6-4ae2-a730-7e36a934ee95

AlexKarpov98:
@bvasilenko - yeah, but we won't show that form all the time, so if we show preloader - it will jump for the user who will not be able to see that part of a form.

PR #11152, opened 2024-12-06
NAS-132933: Enable no-implicit-any-catch

undsoft:
**Changes:**

Actual error in `catchError` is not necessarily a websocket error. It may be a plain JS error.

**Testing:**

Code review should be enough.


AlexKarpov98 (approved):
👌

PR #11157, opened 2024-12-09
NAS-132873 / 25.04 / Fix styles for instance list header

denysbutenko:
**Changes:**

Original issue is not reproducible. Fixed instance list header position.

**Testing:**

Create enough instances. Check header position.


AlexKarpov98 (approved):
Works well for me :) 


https://github.com/user-attachments/assets/9af7fbf5-cbee-4020-a955-54829d87f288

PR #11158, opened 2024-12-09
NAS-132932 / 25.04 / Add warnings to Filename Encryption option in Cloudsync Tasks

AlexKarpov98:
**Testing:**
See ticket.

Result:

https://github.com/user-attachments/assets/242b0a83-3332-45c1-baaf-42aebb8a3215

|Affects         |Reasoning
|Documentation   | Updated texts & some logic


undsoft (approved):
1. I'd be fine with showing the warning dialog every time user ticks the checkbox.
2. The warning dialog is also shown when you open an existing task for edit. This is confusing, it's better if it only happened on user change.

Otherwise looks good.

PR #11159, opened 2024-12-09
NAS-132914 / 25.04 / Expose `force_remove_ix_volumes` flag on app deletion

undsoft:
**Changes:**

Exposes new checkbox when removing an app that has ixVolumes.

### Downstream

|Affects         |Reasoning
|Documentation   | New option when removing an app.


bvasilenko (approved):
Works as expected

PR #11166, opened 2024-12-10
NAS-133002 / 25.04 / Minor iSCSI fixes

undsoft:
**Changes:**

* Fixes the issue where iSCSI tabs would not be shown on systems without FC.
* Changes behavior of some cards.
* Opens iSCSI global form instead of redirecting to a separate page.


denysbutenko (approved):
image
image

Looks good

PR #11172, opened 2024-12-12
NAS-131558 / 25.04 / Change observable subscription for apps widget

RehanY147:
**Changes:**

Changes how we get the subscription observable in the component

**Testing:**

Install two apps. plex and photoprism. Go to the dashboard. Add a widget for photoprism. Large widget for the app overview. Then without saving, open the add widget form again and add another widget for the same app. 

Problem:
The stats data doesn't load the second time

Fix:
The stats load every time


undsoft (approved):
Steps in PR description work as expected. 👍🏼

PR #11177, opened 2024-12-13
NAS-133050 / 25.04 / Updates to fibre channel ports tab

undsoft:
**Changes:**

Fibre ports tab in iSCSI now works differently.
Additional improvements like using tree view can be made later.
System to test with is in the ticket.


denysbutenko (approved):
image
When increased virtual ports number. No way to select target. If set previous value it looks good.

undsoft:
@denysbutenko That's intended. Target will be associated with a port on a different page.

PR #11179, opened 2024-12-13
NAS-132906 / 25.04 / Various improvements for containers

AlexKarpov98:
Testing: See ticket for details.

Once `Use default network settings` is selected, we should not send any NIC devices.

Screenshot 2024-12-13 at 12 38 43


undsoft (approved):
Changes look good, but let's see some new test cases for changes in functionality.

PR #11180, opened 2024-12-13
NAS-133066 / 25.04 / Use computed signals more

undsoft:
**Testing:**

Code review is enough.


AlexKarpov98 (approved):
More computed() signals are used 👌

PR #11196, opened 2024-12-17
NAS-132965 / 25.04 / Shell authenticator option for ACME DNS produces invalid fields

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/9045464b-a5d6-40b8-96e0-1ca72cf1d487


undsoft (changes requested):
So, looking at schema that is being returned by middleware, all authenticator providers have `authenticator` field with `const value`, including shell provider.
I think we should avoid showing controls that have const value and just send them as is.

undsoft (approved):
Looks good, thanks.

PR #11225, opened 2024-12-24
NAS-133049: UI does not provide mechanism to remove an invalid NFS bind IP address

bvasilenko:
## Changes:

Fixed the bug when dropdown **Binding IP address**  is not populated with the values obtained from the middleware in 2 places:
   1. At **Services > NFS settings** 
   2. At **Services > SMB settings**

### For NFS:

1. Go to **System > Shell** and view currently saved state:

   ```
   midclt call datastore.query services.nfs
   [{"id": 1, ... }]
   ```

2. Set incorrect IP address with the `update` middleware call shown below:

   **Notes:**

   * the 1st argument of `update` middleware call must match the value of `id` field from the output of `query` 

   * the name of an attribute is `nfs_srv_bindip` , not the one from the original issue

   * incorrect IP address - is an address from the same subnet other than correct. You can find the correct IP address on the **Network** page

   ```
   midclt call datastore.update services.nfs 1 '{"nfs_srv_bindip": ["192.168.1.99"]}'
   ```
3. Go to **Services > NFS settings**  and view list of options in the **Binding IP address** dropdown field:

**Expected results:**

- user is able to see the incorrect IP address in the list of values
- user is able to unselect the incorrect IP address
- user is able to save

### For SMB:

1. query: `midclt call datastore.query services.cifs`

2. update incorrect:

```null
midclt call datastore.update services.cifs 1 '{"cifs_srv_bindip": ["192.168.1.99"]}'
```

**Expected results:** same as for NFS


RehanY147 (approved):
Looks good. Both NFS and SMB works

PR #11232, opened 2024-12-24
NAS-133054 / 25.04 / Improvements to schedule column

undsoft:
**Changes:**

Previously schedule column would not support selecting Hourly and setting Start and End times.
Also standardizes how schedules are rendered in various places.

**Testing:**

Check schedule column in various places.


bvasilenko (approved):
Works as expected, no regressions found

PR #11236, opened 2024-12-25
NAS-133155 / 25.04 / Non-default account alert does not truncate

AlexKarpov98:
**Testing:**
See ticket.

Result:

https://github.com/user-attachments/assets/123f4233-67cc-46a3-ac42-5880215e8e5f

### Downstream

|Affects         |Reasoning
|Documentation   | Added Expand/Collapse logic to the alert


undsoft (changes requested):
Looks good, but the button looks a bit intrusive. Let's try to do something about it to either make it less noticeable or try using a link.

AlexKarpov98:
@undsoft I like how it looks now, what do you think?
Screenshot 2024-12-26 at 13 23 09

undsoft (approved):
Let's try with `background: var(--alt-bg1);`.

PR #11245, opened 2024-12-26
NAS-133049 / 24.10.2 / UI does not provide mechanism to remove an invalid NFS bind IP address

bvasilenko:
Backported

Original PR: https://github.com/truenas/webui/pull/11225


undsoft (approved):
backported

PR #11257, opened 2024-12-30
NAS-131275 / 25.04 / UI Search behavior: side menu does not close when search autonavigates to a different screen

AlexKarpov98:
**Changes:**
Updated interface to require `anchor` or `hierarchy`.
We are attaching direct item from `-elements.ts` (inside of the directive [**UiSearchDirective**], so it's not attached from final `ui-searchable-elements.json` and that's OK) and if it does not have any info, like `dashboard: {}` -- it will not get any ID for the element itself and won't correctly register directive.
Without fix it's not setting correct anchors for the search elements. 

Fix:
So anchor will be available always - explicitly from `.anchor` or generated from `.hierarchy`, but one of these should be on the object itself for correct work.

**Testing:**
See:

https://github.com/user-attachments/assets/47cffa39-8b1d-4056-9786-9171cd0850a4


undsoft (commented):
Why was this change necessary to close the side menu? It seems like more boilerplate for developers to add anchors.

AlexKarpov98:
> Why was this change necessary to close the side menu? It seems like more boilerplate for developers to add anchors.

@undsoft 
To close the menu, it's not necessary, but I found out that it gives us another issues, described in the PR description.

undsoft (approved):
Can't say that I fully understand the changes, but search works and secondary menu seems to be closing.

AlexKarpov98:
> Can't say that I fully understand the changes, but search works and secondary menu seems to be closing.

I think that it's real fix, not just shallow one - where we can close the sidenav.
In this case we'll have correct IDs for building `.json` file and correct IDs for using search .elements in a directive.

We are dealing with 2 files `ui-searchable-elements.json` and `*-elements.ts`, and in `.json` file we may have had id generated from parent hierarchy for example, but in `*-elements.ts` that ID won't be available and if no hierarchy or ID set up explicitly, it's just an `{}` empty object, so IDs will not match, between `ui-searchable-elements.json` and `*-elements.ts`.

So now - I required `.hierarchy` or `.id` to be set up in order to have and save the same IDs. Accessing direct object key in directive does not give us access to the parent data so that we can build something useful for concrete directive.

PR #11267, opened 2025-01-01
Translation to Korean update

SejoWuigui:
**Changes:**
- Improved Korean translation


SejoWuigui:
truenas/ui/jobs > "State" string in table header
truenas/ui/credentials/certificates > "State" string in 'Add Certificate Authority'

Same string but different meaning.
Can you separate them? Each needs a different translation.

undsoft:
> Could you separate them? Each needs a different translation.
Unfortunately this is a limitation of the translation system we are currently using. There is a ticket to resolve this, but it's a low priority at the moment.

undsoft (approved):
Thanks for the translations.

PR #11268, opened 2025-01-01
NAS-133264 / 25.04 / Add more warnings about unsaved changes

AlexKarpov98:
Testing: check affected forms.


src/app/pages/dashboard/components/widget-group-form/widget-group-form.component.ts
```diff
@@ -78,6 +78,9 @@ export class WidgetGroupFormComponent {
     public slideInRef: SlideInRef<WidgetGroup | undefined, WidgetGroup | false>,
     private cdr: ChangeDetectorRef,
   ) {
+    this.slideInRef.requireConfirmationWhen(() => {
```
undsoft, line 81: I can open edit form here and change things in dropdowns without the warning.

undsoft (approved):
Warnings are there, but our controls don't work correctly, because form is marked dirty even if user just clicks on the input. We can solve this in another ticket.

PR #11272, opened 2025-01-02
NAS-132626 / 25.04 / Fix styles for contract mentions on sys-info widgets on the dashboard

RehanY147:
**Changes:**

Fix styles for contract mentions on sys-info widgets on the dashboard

**Testing:**

Change the license to "Silver International" and see that the styles look good for system info widgets.


undsoft (changes requested):
So two things:
1. Text is not aligned for me:
Снимок экрана 2025-01-02 в 09 13 55

2. I don't know how silver international is going to be represented on MW, but if it's going to be `SILVER_INTERNATIONAL` like other enums, this is going to result in `Silver_international Contract` on main dashboard and `Contract Type: SILVER_INTERNATIONAL` on Support card in settings.
Please check with MW on what value we should expect in `contract_type`.

undsoft (changes requested):
First point was addressed, second point – wasn't.
You can find a link to MW PR with a list of all licenses in the ticket. It's my understanding that they'll be changed to UPPERCASE when being sent to UI.

undsoft (approved):
Tested with mocks, as there is some issue with `webui.main.dashboard.sys_info` endpoint.
This is good to merge after enum is updated.

undsoft:
backport

PR #11280, opened 2025-01-03
NAS-133384 / 25.04 / Use red color for destructive actions

undsoft:
**Changes:**

Uses red color for buttons that delete something.


AlexKarpov98 (approved):
👍

PR #11282, opened 2025-01-03
NAS-133365 / 25.04 / Reporting exporters form is broken

undsoft:
**Changes:**

Test Reporting -> Exporters form.


bvasilenko (approved):
Works good, however I've noticed a minor issue with **Enabled** checkbox

1. While viewing **Exporters** table, at the column **Enabled** turn off the toggle.
2. Open the **Edit Exporter** form
**Expected result:** **Enabled** checkbox is disabled
**Actual result:** **Enabled** checkbox is still enabled

PR #11287, opened 2025-01-05
Add translate pipe for hints

MDVAurelien:
Changes:

Add translate pipe for hints

Testing:

Check if hints translated properly in another language.


bvasilenko (approved):
Works as expected

image

AlexKarpov98 (approved):
👍

PR #11289, opened 2025-01-06
NAS-133308 / 25.04 / Remove call for buckets when not correct role. Adds `readonly` to `ix-explorer`

RehanY147:
**Changes:**

Remove call for buckets when not correct role. Adds `readonly` to `ix-explorer`

**Testing:**

See the ticket description for testing. Also test the readonly option on the explorer.


undsoft (approved):
Looks good, the calls are not being made.

PR #11291, opened 2025-01-06
NAS-133383 / 25.04 / Expose checkbox to enable GPOS STIG

AlexKarpov98:
**Changes:**

**Testing:**
See ticket.
result:
Screenshot 2025-01-06 at 14 09 39
Screenshot 2025-01-06 at 14 09 45

### Downstream

|Affects         |Reasoning
|Documentation   | New System Security Option


undsoft (changes requested):
UI issues:
1. The whole system security card is being a `system.security.info.fips_available`, so you can remove additional checks for enterprise license.
2. Once STIG is enabled, you get stuck in the infinite loop that shows "[ENOTSUP] TOKEN_PLAIN: mechanism is not supported at current authenticator level.".
We'll make more fixes in another ticket, but for this ticket let's at least clear token on token login errors, so that user doesn't end up in an infinite loop.

AlexKarpov98:
@undsoft 

Fixed.

https://github.com/user-attachments/assets/7878003b-5e07-4aea-bdf9-865c700b77d4

undsoft (approved):
Looks good for this ticket.

PR #11295, opened 2025-01-06
NAS-132661: Move 2FA text

undsoft:
**Changes:**

Instead of showing a snackbar, shows text in-line.

Снимок экрана 2025-01-06 в 18 32 32


bvasilenko (approved):
Works as expected

PR #11304, opened 2025-01-08
NAS-133128 / 24.10.2 / Simplify CPU widget logic

undsoft:
**Changes:**

Previous UI code assumes that if hyperthreading is enabled, than there must be 2 threads per core. 
This is not the case for newer CPUs that have performance and efficiency cores. In Intel processors efficiency cores are single-threaded.

This PR simplifies the logic to just showing what middleware sends us.

**Testing:**

Test CPU widgets for regressions.
You can also use mock data from the ticket in `WidgetResourcesService`.


denysbutenko (approved):
image

No regressions are found, looks good.

AlexKarpov98 (approved):
Works good 👍

PR #11305, opened 2025-01-08
NAS-133213: Handle null latest_version for a custom app upgrade

RehanY147:
**Changes:**

Handles case where pipe outputs empty string. Should show "Upgrade is available" when latest_version is not available. This issue seems resolved in master. But this PR adds a the "Upgrade is available" string.

**Testing:**

Code review


bvasilenko (approved):
Looks good!

image

PR #11307, opened 2025-01-08
NAS-132428 / 25.04 / Two-Factor Authentication > Give 2fa field focus

bvasilenko:
**Testing:**

Ensure that the 2fa field has focus when the page loads.


denysbutenko (approved):
image

Looks good.

PR #11308, opened 2025-01-08
NAS-133431 / 25.04 / Fix how dirty status is updated in controls

AlexKarpov98:
Testing: see ticket.
I've noticed that problem is only with `ix-input` all other components seems to work fine and not cause dirty: true with just being focused and not changed.
So I've updated `ix-input` only.


src/app/modules/forms/ix-forms/components/ix-input/ix-input.component.ts
```diff
@@ -153,9 +157,15 @@ export class IxInputComponent implements ControlValueAccessor, OnInit, OnChanges
...
   registerOnChange(onChange: (value: string | number) => void): void {
-    this.onChange = (val) => {
-      this.lastKnownValue = val;
-      onChange(val);
+    this.onChange = (changedValue) => {
+      if (this.lastKnownValue !== changedValue) {
+        onChange(changedValue);
+      }
+
+      this.controlDirective.control.markAsTouched();
```
undsoft, line 165: It shouldn't be up to you to trigger this. The way for controls to interact with angular forms is via `onChange`/`onTouch` functions that are provided. If there was no change, we should avoid calling `onChange`.

RehanY147 (approved):
Looks good. We should confirm why the removed code was added originally and this doesn't cause any regressions.

AlexKarpov98:
It was my prev PR where it was added: 🙈 
https://github.com/truenas/webui/pull/9200/files

I checked, it all works fine. 👌

PR #11312, opened 2025-01-08
Fix requirements on creating certificates

ahaczewski:
**Changes:**

TrueNAS allows creating CA certificates and CSRs. These can be used for internal network purposes, but a CSR from TrueNAS is also a way to get ACME-DNS authorized certificate from Let's Encrypt.

Let's Encrypt accepts only two fields in a CSR: Common Name, and a list of SANs. All the other subject fields are stripped from the generated certificate, so it makes no sense to require for a certificate to have a country, state, locality and organization.

Additionally, when creating a CA certificate, requiring SANs makes no sense, as these are to be used for the leaf certificates.

Common Name, on the other hand, should be required for all kinds of certificates.

**Testing:**

- Create certificate authority with only Common Name filled
- Create CSR with only Common Name and SANs filled

### Downstream

|Affects         |Reasoning
|Documentation   | Not affected, as it does not mention the fields are required. I would suggest thou to mention the required field Common Name.

### DIFF
--- a/src/app/pages/credentials/certificates-dash/forms/common-steps/certificate-subject/certificate-subject.component.spec.ts
+++ b/src/app/pages/credentials/certificates-dash/forms/common-steps/certificate-subject/certificate-subject.component.spec.ts
@@ -78,15 +78,17 @@ describe('CertificateSubjectComponent', () => {
         },
         {
           label: 'Subject',
-          value: 'Virus Research Dept, Umbrella Corp, Racoon City, Pennsylvania, US',
+          value: 'virus.umbrella.com, Virus Research Dept, Umbrella Corp, Racoon City, Pennsylvania, US',
         },
       ]);
     });
 
     it('skips some of the fields when they are missing', async () => {
       await form.fillForm({
+        State: '',
+        Locality: '',
         'Organizational Unit': '',
-        'Common Name': '',
+        Email: '',
       });
 
       expect(spectator.component.getSummary()).toEqual([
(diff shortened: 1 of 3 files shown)


src/app/pages/credentials/certificates-dash/forms/common-steps/certificate-subject/certificate-subject.component.spec.ts
```diff
@@ -95,14 +97,57 @@ describe('CertificateSubjectComponent', () => {
...
         },
         {
           label: 'Subject',
-          value: 'Umbrella Corp, Racoon City, Pennsylvania, US',
+          value: 'virus.umbrella.com, Umbrella Corp, US',
         },
       ]);
     });
+
+    it('allows creating Lets Encrypt-compatible subject', async () => {
```
ahaczewski, line 110: This test fails, because I have no idea how to force IxSelectHarness to clear itself of selected value. I would happily fix this if someone would be so nice to chime in and lend me a hand 😃

ahaczewski:
The PR at its current stage will fail, because the test I created fails. More on that in the in-line comment. I could use some help with that 😊

undsoft (changes requested):
Unfortunately, this won't work because API has validation in place for `city`, `email`, `organization` and `state` fields.

@themylogin Would you like to chip in on the discussion?

themylogin:
@sonicaj @Qubad786 will it be possible to disable validation for these fields in case of Let's Encrypt?

ahaczewski:
> Unfortunately, this won't work because API has validation in place for `city`, `email`, `organization` and `state` fields.
> 
> @themylogin Would you like to chip in on the discussion?

I'm happy to remove validation there as well, just need to find it in the API first 😅 

> @sonicaj @Qubad786 will it be possible to disable validation for these fields in case of Let's Encrypt?

But why only for Let's Encrypt? Certificate Authorities do not have e-mail nor SANs, check GitHub, Google, Amazon, Cloudflare or any other root certificate.

sonicaj:
@ahaczewski i have created a ticket here https://ixsystems.atlassian.net/browse/NAS-133533 which will cater to the points you have raised, the changes on the UI side alone won't be enough to account for the points raised and middleware (truenas/middleware repo) will have to make some changes to get this to work.

ahaczewski:
@sonicaj Awesome! If nobody will jump in on the middleware, I'll try to carve some time over the weekend and give it a try myself 😊

undsoft:
I'm going to close the PR for now. Feel free to reopen it, if anything becomes available for testing.

PR #11317, opened 2025-01-09
NAS-132705 / 25.04 / Adopts middleware changes to CPU and memory reporting on the dashboard widgets

RehanY147:
**Changes:**

Adopts middleware changes to CPU reporting on the dashboard widgets. Removes temperature related reporting also temperature for cpu widget

**Testing:**

With a recent build, CPU stats related widgets should work correctly. Also shouldn't show any errors in the console logs


undsoft (changes requested):
There is a number of issues most of them are not because of changes in this PR, but let's still fix them.
1. I get `toSignal` error. Can be fixed by adding `.pipe(startWith(0))` to `refreshInterval$` in `WidgetResourcesService`.
2. I've tried simulating load on CPU to test the widget and noticed that load is reported as System on main chart, but bars don't show much. Looks like an issue with MW reporting. I think users would expect to see total load on bars as well. 
Снимок экрана 2025-01-09 в 11 10 13
3. Memory overview widget doesn't load.
4. Let's remove CPU Usage Per Core from default set of widgets as currently there is a gap there.

package.json
```diff
@@ -80,6 +80,7 @@
     "@lezer/common": "~1.2.1",
     "@lezer/generator": "~1.7.1",
     "@lezer/lr": "~1.4.2",
+    "@marijn/find-cluster-break": "~1.0.2",
```
undsoft, line 83: I don't know why this package is needed. It's not used anywhere and it works for me without it. Sounds like an issue on your end.

undsoft (approved):
Besides the discrepancy between CPU load in various widgets (System vs User), the rest looks good.
Let's ask MW to look into this.

PR #11322, opened 2025-01-10
NAS-133479 / 25.04 / HA notice showing on a single system

AlexKarpov98:
Testing: see ticket comments. 

Code review may be enough.


undsoft (approved):
Let's add a message for single system update as well. Right now, job dialog closes and nothing happens for a bit.
Message should say something along the lines of "Update completed successfully. The system will reboot shortly"

PR #11323, opened 2025-01-10
NAS-133481 / 25.04 / Changes for VMs in new virtualization

denysbutenko:
**Changes:**

1. Makes CPU and Memory attributes required again.
2. Update Disks picker, remove destionation and allows to pick zvols only.
3. Updated ix-explorer allows to list zvols only.

**Testing:**

Checks Instance Wizard on Virtualization page.

### Downstream

|Affects         |Reasoning
|Documentation   | Instance Wizard changes


src/app/pages/virtualization/components/all-instances/instance-details/instance-disks/instance-disk-form/instance-disk-form.component.ts
```diff
@@ -63,6 +63,8 @@ export class InstanceDiskFormComponent implements OnInit {
     return !this.isNew() ? this.translate.instant('Edit Disk') : this.translate.instant('Add Disk');
   });
 
+  protected instance = computed(() => this.slideInRef.getData().instance);
```
RehanY147, line 66: The computed callback doesn't have a signal call. This could just be a signal.
`protected instance = signal(this.slideInRef.getData().instance);` Or maybe doesn't need to be a signal

src/app/pages/virtualization/components/all-instances/instance-details/instance-disks/instance-disk-form/instance-disk-form.component.ts
```diff
@@ -112,7 +117,7 @@ export class InstanceDiskFormComponent implements OnInit {
   }
 
   private prepareRequest(): Observable<unknown> {
-    const instanceId = this.slideInRef.getData().instanceId;
+    const instanceId = this.slideInRef.getData().instance.id;
```
RehanY147, line 120: Can reuse the signal

RehanY147 (changes requested):
The CPU and memory field are not required on the instance wizard. You can leave them empty.

denysbutenko:
> The CPU and memory field are not required on the instance wizard. You can leave them empty.

@RehanY147 did you mean to reset value in those fields when user switches from VM to Container?

RehanY147 (changes requested):
I am unclear when we want to choose zvols, can we also select directories? Or does that user has to select a zvol.
`/dev/zvol` as a value should not be selectable. That value is meaningless when requiring to select zvols.

denysbutenko:
> I am unclear when we want to choose zvols, can we also select directories? Or does that user has to select a zvol. `/dev/zvol` as a value should not be selectable. That value is meaningless when requiring to select zvols.

Middleware does not allow directories to be used and raises validation error. The `ix-explorer` does not support this feature yet or have it been added recently?

undsoft (approved):
It's fine the way it works at the moment, we can improve on it later.

PR #11326, opened 2025-01-13
NAS-133460 / 25.04 / Confirm Password shows error message even if I haven't touched it yet

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/2d105171-8642-4c7d-85ba-b37aa0ebc088


bvasilenko (approved):
Works as expected, red validation message appears only after user has touched to field

PR #11328, opened 2025-01-13
NAS-133442 / 25.04 / hide SED related options in UI

denysbutenko:
**Changes:**

- hide SED related options in UI

**Testing:**

Check ticket and related components.

### Downstream

|Affects         |Reasoning
|Documentation   | hide SED related options in UI for Community Edition users who haven't configured SED

### DIFF
--- a/src/app/pages/credentials/kmip/kmip.component.html
+++ b/src/app/pages/credentials/kmip/kmip.component.html
@@ -99,6 +99,7 @@
         </div>
       </ix-fieldset>
       <ix-fieldset [title]="'Management' | translate">
+        // TODO: Check if it needs to be hidden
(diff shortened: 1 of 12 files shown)


RehanY147 (changes requested):
We should hide the Disk Encryption card altogether since unless it's set via CLI, user can't change it from the UI and we don't want them thinking about that anyway.

RehanY147 (changes requested):
I can still see the encryption card on disks where SED was not set
image

denysbutenko:
> I can still see the encryption card on disks where SED was not set image

Have you tried on non-HA system?

RehanY147 (approved):

PR #11339, opened 2025-01-14
NAS-133566 / 25.04 / UI is not reloaded after license is applied

undsoft:
**Testing:**

Testing steps and example license are in the ticket.


denysbutenko (approved):
Works great!

PR #11340, opened 2025-01-15
NAS-133579 / 25.04 / Highlight dataset name when it's about to be deleted

undsoft:
**Changes:**

Makes dataset name bold here.

Снимок экрана 2025-01-15 в 10 19 21


AlexKarpov98 (approved):
Dataset name is highlighted now.

PR #11342, opened 2025-01-16
NAS-133477 / 25.04 / Allow adding directory services groups manually

bvasilenko:
**Changes:**

Added ability of manual entry into the groups dropdown

**Testing:**

1. Prosivion a new machine with HA and Active Directory
2. Please see ticket for replication steps and expected results


undsoft (approved):
Thanks, works well.

PR #11343, opened 2025-01-16
NAS-133547 / 25.04 / Fibre Channel QA#2

AlexKarpov98:
Testing: see ticket.


src/app/modules/ix-table/components/ix-table-body/ix-table-body.component.ts
```diff
@@ -50,6 +50,7 @@ export class IxTableBodyComponent<T> implements AfterViewInit {
   readonly columns = input<Column<T, ColumnComponent<T>>[]>();
   readonly dataProvider = input<DataProvider<T>>();
   readonly isLoading = input(false);
+  readonly hasDetailsCards = input(false);
```
undsoft, line 53: table-body is a generic table component. I don't think it needs to know about details.
Let's try with something like `templateColumn` instead.

src/app/modules/ix-table/components/ix-table-body/ix-table-body.component.scss
```diff
@@ -30,6 +30,12 @@
     @media(max-width: $breakpoint-sm) {
       padding: 0 8px;
     }
+
+    &.show-details-cell {
+      @media (min-width: $breakpoint-sm) {
```
undsoft, line 35: Page enters into mobile view on different resolutions than when chevron appears.

undsoft (approved):
I've also noticed that when I create a FC with existing port and then open it for edit, "Do not connect to a fibre channel port" will be selected.

Otherwise most fixes look good.

RehanY147 (approved):

PR #11346, opened 2025-01-16
NAS-133593 / 25.04 / Add TPM to new VMs

undsoft:
**Testing:**

TPM can be added when VMs are being created in new virtualization.


bvasilenko (approved):
Looks good!

PR #11350, opened 2025-01-17
NAS-133478 / 25.04 / Allow ISOs to be uploaded in VM wizard

bvasilenko:
**Testing:**

At the new **Virtualization** page

When user clicks **Create New Instance,**

Choose **Upload an ISO image** option, and pick a custom ISO file. For example this one: https://releases.ubuntu.com/23.10/ubuntu-23.10-live-server-amd64.iso

Test UI fields and ensure that correct details are being sent to middleware. 

**Expected UI behavior:**

"VM Image Options" should be shown conditionally when **VM** is chosen:

* Option **Use a linux container**

  _This option should be chosen by default._

  When user chooses **Use a linux container, no UI change is required,** then existing **Image \*** field is displayed.

* Option **Upload an ISO image**

  When user chooses **Upload an ISO image**, then the **Image \*** input field changes to a file location picker

### Downstream

|Affects         |Reasoning
|Documentation   | Newly added **Upload an ISO image** form field for ISO uploading, and UI behavior described above


src/app/pages/virtualization/components/instance-wizard/instance-wizard.component.ts
```diff
@@ -116,6 +129,11 @@ export class InstanceWizardComponent {
     }))),
   );
 
+  imageOptions$: Observable<Option<SelectImageType>[]> = of([
+    { label: this.translate.instant('Use a Linux image (linuxcontainer.org)'), value: SelectImageType.Choose },
+    { label: this.translate.instant('Upload an iso image'), value: SelectImageType.Load },
```
undsoft, line 134: Let's update label to `Upload an ISO image`

undsoft (changes requested):
I get `[EFAULT] Failed creating custom volume from ISO: Cannot create volume, already exists on target storage` when I try to upload an ISO for a VM.
Also I can still see `VM Image Options` even when I'm creating a container.

denysbutenko (commented):
image

There is no way to exit and no information about the progress. At same time jobs panel has some details for this job.

image

However, it stuck at 25% for a while and no changes. Probably it would require middleware changes to have better indications.

bvasilenko:
@undsoft @denysbutenko  there's a new **Image Name** input field , to address one of your commentaries and allow user to create 2nd machine from same ISO

undsoft (approved):
I was able to upload the ISO and use it.

PR #11352, opened 2025-01-18
NAS-133118 / 25.04 / Addes tests for glossary and form-section components, form-service

RehanY147:
**Changes:**

Adds tests

**Testing:**

Code review


AlexKarpov98 (changes requested):
Looks ~fine, but noticed some issues:

1. Tests & Code style failed, Please pull from the master.

2. I tried to play with tests (updated it locally) and found out that it is not working correctly here:
`ix-form-glossary.component.spec`

```
  it('loads input options from form glossary', async () => {
    expect(spectator.component).toBeTruthy();
    const input = await harnessLoader.getHarness(IxInputHarness);
    await input.setValue('Control2');
    spectator.detectChanges();
    const matAutocomplete = spectator.query(MatAutocomplete);
    matAutocomplete.showPanel = true;
    const options = matAutocomplete.options.map((option) => option.getLabel());
    expect(options).toEqual(['Control2']);
  });
```
Screenshot 2025-01-18 at 12 02 56

But I expect to only have `Control2` to be in the array since I explicitly set it in input.

And even more:
Screenshot 2025-01-18 at 12 04 33

RehanY147:
@AlexKarpov98 The setting of value in the input and checking the options the way you have doesn't sync properly. That's why the test fails. This specific behavior is already testing in the `ix-input` so I don't think testing that is needed here. We just mean to test that controls added in the form are registered on the form-service and show up in the glossary component.

AlexKarpov98 (approved):
Would be nice to have it synced correctly though.

RehanY147:
@AlexKarpov98 I improved the test as you suggested

PR #11363, opened 2025-01-21
NAS-133464 / 25.04 / Disable token authentication when STIG is enabled

denysbutenko:
**Changes:**

1. Added `authenticator` field for `LoginExResponse`
2. Added check for authenticator level if `LEVEL_1` token usage is allowed.

**Testing:**

- Create HA instance
- Create a new user with FullAdmin role (in additional to root).
- Login as a new user
- Go to System Settings -> Advanced -> Global 2FA -> Enable Global 2FA
- Setup 2FA for this user
- Re-login with 2FA enabled
- Go to System Settings -> Advanced -> System Security -> Enable STIG checkbox
- Confirm to reload local node.
- Wait until loaded.
- Repeat for next node.
- Check websocket messages and ensure no calls for `auth.generate_token` endpoint.
- Disable STIG and 2FA.
- Ensure no regressions for token generation.


src/app/interfaces/auth.interface.ts
```diff
@@ -14,9 +14,12 @@ export enum LoginExMechanism {
   ApiKeyPlain = 'API_KEY_PLAIN',
 }
 
+export type AuthenticatorLoginLevel = 'LEVEL_1' | 'LEVEL_2';
```
undsoft, line 17: Let's make it an enum.

src/app/modules/auth/auth.service.spec.ts
```diff
@@ -62,6 +62,7 @@ describe('AuthService', () => {
           user_info: {
             privilege: { webui_access: true },
           },
+          authenticator: 'LEVEL_1',
         } as LoginExResponse),
```
undsoft, line 66: Needs a test case for new behavior.

undsoft (changes requested):
1. Login doesn't work for non-STIG systems. There seems to be a race condition where `setupAuthenticationUpdate` with `true` is called before `processLoginResult`.
2. We should attempt to clear existing tokens in local storage when STIG is enabled for the first time, so that user don't get authentication errors.

undsoft (approved):
Appears to be working as expected.

PR #11376, opened 2025-01-24
NAS-133718 / 25.04 / Add tunable type

undsoft:
**Changes:**

Adds Type to tunable form in Advanced settings.

**Testing:**

Note that you may experience MW errors when form is saved. Check if those are covered here: https://ixsystems.atlassian.net/browse/NAS-133718

### Downstream

|Affects         |Reasoning
|Documentation   | Adds Type to tunable form.


bvasilenko (approved):
The **Add Sysctl** form is now able to submit the entered values to the middleware.

The progress bar is displayed, and in the background the machine reboots. I guess it is also expected.

When machine is up again, the progress bar won't disappear, which seems a bit odd.

However, the sysctl module parameter are getting persisted just fine.

PR #11378, opened 2025-01-24
NAS-133634 / 24.10.2.1 / Fix paths when they are not absolute

bvasilenko:
**Testing:**

The machine where the issue was discovered: r40-1.dc1.ixsystems.net
Please see replication steps and expected results on the ticket.


undsoft (approved):

PR #11387, opened 2025-01-25
NAS-133783 / 25.10 / Root disk size setting

undsoft:
**Changes:**

Adds `root disk size` field for VMs.


denysbutenko (approved):
image

Looks good.

PR #11388, opened 2025-01-25
NAS-133792 / 25.10 / NAS0-133792: Only allow local groups for groups in user form

undsoft:
**Changes:**

See ticket.


AlexKarpov98 (approved):
👍

PR #11390, opened 2025-01-26
NAS-133747 / 25.10 / Adds default network settings in network instance

RehanY147:
**Changes:**

Shows the default network settings that will be used if the user chooses that option

**Testing:**

See the Network item on the virtualization instance form


undsoft (approved):
Looks good. I can see default network settings.

PR #11402, opened 2025-01-27
NAS-133746 / 25.10 / CPU/memory optional on edit but required on creation

undsoft:
**Changes:**

See ticket.


bvasilenko (approved):
Validation works as expected

PR #11404, opened 2025-01-27
NAS-133598 / 25.10 / Fix `saveUpdatedPreferences` effect

bvasilenko:
**Testing:**

Steps to reproduce:

1. Go to **Advanced Settings \> Access widget**
2. Click **Configure** and change the value of **Session Timeout**. Save.
3. Refresh the window
4. **Session Timeout** reverts to it's previous value, `300`

**Expected result:** proper value is saved and then loaded


bvasilenko:
@RehanY147 
image

By design, the `auth.set_attribute` call should only be executed when the previous and new **preferences** are different. The filter in **line 76** is responsible for this.

When using `take(1)` there is no easy way to get the previous and new preferences.

RehanY147 (approved):

PR #11415, opened 2025-01-27
NAS-133676 / 25.10 / Check/add search elements for new pages

AlexKarpov98:
Testing: see ticket.
New search elements should be available now.


bvasilenko (approved):
Works as expected

PR #11426, opened 2025-01-28
Nas 133744

AlexKarpov98:

https://github.com/user-attachments/assets/555814d9-d37c-410b-bec8-1d8b34baf43c


kmoore134:
FYI - This needs to be backported into BETA1, otherwise people cannot import their old VMs.

kmoore134:
But aside from that, looks nice! Cant wait to see it land.

src/app/pages/virtualization/components/instance-wizard/instance-wizard.component.spec.ts
```diff
@@ -424,11 +431,54 @@ describe('InstanceWizardComponent', () => {
...
+      expect(spectator.inject(SnackbarService).success).toHaveBeenCalled();
+    });
+
+    it('creates new instance using zvol path when form is submitted', async () => {
+      // TODO: Not sure what's causing the below warning, so I mocked 'warn' to make the test pass:
+      // The configured tracking expression (track by identity) caused re-creation of the entire
+      // collection of size 13. This is an expensive operation requiring destruction and subsequent
+      // creation of DOM nodes, directives, components etc. Please review the "track expression"
+      // and make sure that it uniquely identifies items in a collection. Find more at https://angular.dev/errors/NG0956
+      jest.spyOn(console, 'warn').mockImplementation(() => {});
```
RehanY147, line 453: If there is an expected warning, we should test for that specific warning with expect. This would hide any and all warning calls.

RehanY147 (approved):
Looks good. Test should be specific.

src/app/pages/virtualization/components/instance-wizard/instance-wizard.component.spec.ts
AlexKarpov98: I do not see any warnings in real life, not sure what it means here in tests. I checked all `@for` loops, we have correct `track` implementation there.
Maybe @undsoft has any ideas?

AlexKarpov98: You can run test without that mock `jest.spyOn(console, 'warn').mockImplementation(() => {});` and test will fail.

undsoft: Yeah, the issue is somewhere deep inside of `ix-explorer`. Removing that component from the template helps.
The issue seems to be happening VM Image Options is changed:
```
await form.fillForm({
   ...
   'VM Image Options': 'Use zvol with previously installed OS',
});
```

Let's only disable console warnings around this block if possible.

undsoft (approved):
UI sends correct params, but we'd need to do more testing later to make sure imported zvol actually works.

PR #11435, opened 2025-01-30
Update Korean translation

SejoWuigui:
**Changes:**
- Translate new strings to Korean


AlexKarpov98 (approved):
Nice, always 100% 👍 
There will be more to translate 🧑‍🍳

PR #11436, opened 2025-01-30
NAS-133707 / 25.10 / App icons are not initially rendered when searching until screen redraw

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/1446481b-e63e-4bfe-84ef-ea1ceaaa77a2


bvasilenko (approved):
Works as expected

PR #11439, opened 2025-01-30
NAS-133825 / 25.04-RC.1 / Update backups only once at the end

bvasilenko:
Backported.
Original PR: https://github.com/truenas/webui/pull/11459


AlexKarpov98 (approved):
Interesting solution, not sure how to solve it in a different way 👌

PR #11442, opened 2025-01-30
NAS-133665 / 25.10 / Close job dialog when opened manually on job end state

RehanY147:
**Changes:**

When job progress modal is opened from the jobs panel in the top right, it used to remain open until manually closed by either clicking the close button on the modal or clicking outside the modal to close it. Now, the modal will close automatically when the job reaches a "end" state. Either finished with success or error.

**Testing:**

Run any job and open the job modal from the jobs panel in the top right. For example, have two tabs open. Delete a couple of apps and go to the other tab to open the job progress dialog from the jobs panel. The dialog should close the job completes.


src/app/modules/jobs/components/jobs-panel/jobs-panel.component.ts
```diff
@@ -92,8 +98,21 @@ export class JobsPanelComponent {
...
     const title = job.description ? job.description : job.method;
 
+    const job$ = (
+      this.store$.select(
+        selectJob(job.id),
+      ) as Observable<Job<ApiJobResponse<ApiJobMethod>>>
+    ).pipe(
+      observeJob(),
+      tap((jobUpdate) => {
+        if (jobUpdate.state === JobState.Success) {
```
undsoft, line 108: I think this can be in subscribe block without the condition.

undsoft (approved):
Works as expected.

PR #11459, opened 2025-02-02
NAS-133825 / 25.10 / Update backups only once at the end

bvasilenko:
**Testing:**

For replication steps please see ticket


AlexKarpov98 (approved):
Interesting solution, not sure how to solve it in a different way 👌

PR #11463, opened 2025-02-03
NAS-133774 / 25.10 / Reload node children on toggle in ix-explorer

denysbutenko:
**Changes:**

**IxExplorer**: Reload node children on toggle

**Testing:**

Check any page with `ix-explorer` and then create new zvol/dataset in the second tab.
See ticket for expected results.


undsoft (changes requested):
I think something is off here.
I don't think we should have to be calling refresh manually. I don't see anything caching in `getFilesystemNodeProvider`. As far as I can tell, it should return a function that makes a request every time it's used in a component.
It's best to debug this more.

undsoft (approved):
Looks good. I've included some other fixes from a relevant ticket (NAS-134151) so that it's easier to backport.

PR #11464, opened 2025-02-03
NAS-133891 / 25.10 / Remove old VM section

denysbutenko:
**Changes:**

Remove old VM section and related files

**Testing:**

Code review enough and CI passed.


undsoft (approved):
Thanks, old VM stuff is no longer there.

PR #11471, opened 2025-02-03
NAS-133874 / 25.10 / Get rid of `showTwoFactorWarning` and only use `FirstLoginDialogComponent`

AlexKarpov98:
Testing: see ticket.

Result:
Screenshot 2025-02-04 at 00 14 28

https://github.com/user-attachments/assets/6bc6391e-4c36-41e1-b968-0df09efb917e


undsoft (approved):
Thanks, looks good.
Can we also do something about `twoFactor?.userTwoFactorAuthConfigured` in `FirstLoginDialogComponent`. I don't like the idea of one component reaching inside of another component for essentially a `protected` field.

PR #11474, opened 2025-02-04
NAS-133869 / 25.10 / Fix return value of `user.create` and `user.update`

bvasilenko:
**Summary**

Adapted to changes in middleware: `user.create` and `user.update` requests now return an `id` instead of a `User` object.

**Testing:**

Please see ticket  for replication steps and expected results


RehanY147 (changes requested):
Getting this error when I create a new user

image

PR #11477, opened 2025-02-04
NAS-133970 / 25.10 / Add error handling for removal the extent association from the iSCSI Targets page

AlexKarpov98:
Testing: see ticket.

```
Updated

I understand the issue when I connected the target through the initiator, it was not possible to remove it. This is a valid reason. But it would be good if there were warning texts about that.
```

https://github.com/user-attachments/assets/a39acc2c-28f8-4a77-8b43-27213d8b4ac0


bvasilenko (approved):
@AlexKarpov98 Looks good overall,
I just noticed that the error message in the video says "Error handling is working", and it is misleading, however I think it's a debug message that was captured on the video unintentionally.

AlexKarpov98:
@bvasilenko yeah, it's a custom message added only for the demo video, to show that error handling works.

PR #11495, opened 2025-02-05
NAS-133723 / 25.10 / Adds "Enable VLAN" checkbox and sends null when disabled

RehanY147:
**Changes:**

Adds an "Enable VLAN" checkbox to the ipmi form. If the checkbox is unchecked, UI should send null value in the "vlan" field when calling middleware endpoint. When the checkbox is checked, the vlan id field is required.

**Testing:**

Check that vlan id is disabled, enabled and set correctly.


undsoft:
backport

PR #11505, opened 2025-02-06
NAS-134046 / 25.10 / Auth error in Export Key is not properly handled

undsoft:
Original issue:

> 1. Create a dataset with Key encryption.
> 2. Simulate pool.dataset.export_key failing in loadKey in ExportDatasetKeyDialogComponent.
> 3. Click Export Key.
> ER: I see error dialog and I am not prompted to download the key.
> AR: I see error dialog and then I am prompted to download the key.


AlexKarpov98 (approved):
👍

PR #11507, opened 2025-02-07
NAS-133932 / 25.10 / Ease of use improvements for ISO uploads

bvasilenko:
## **Testing:**

You'll find TODO list in the ticket.

### **Item 1:** 

See replication steps in the ticket. The problem was in the common component `ix-file-input`. I fixed it.

### **Item 2:**

  **Replication steps:**

  1. User click **Select ISO** button, which is initially empty
  2. User uploads .iso file
  3. User closes the **Volumes** modal window
  4. User clicks **Select ISO** button again
  5. User uploads exactly same .iso file

  **Expected result:** front-end does pre-validation and shows message immediately, without uploading the file.

### **Item 3**: 

   **Expected result:** reporting of the upload progress ration in percentage `0% ... 100%`


undsoft (approved):
Provides better feedback in the UI, thanks.

PR #11530, opened 2025-02-10
NAS-134061 / 25.10 / Apps - UI validates `uri` type more strictly than middleware

AlexKarpov98:
Testing: see ticket.

Result, no validation error:

Screenshot 2025-02-10 at 13 41 21


src/app/modules/forms/ix-forms/validators/url-validation.service.ts
```diff
@@ -2,5 +2,5 @@ import { Injectable } from '@angular/core';
 
 @Injectable({ providedIn: 'root' })
 export class UrlValidationService {
-  urlRegex = /^(?:http(s)?:\/\/)?[\w.-]+(?:\.[\w.-]+)+[\w\-._~:/?#[\]@!$&'()*+,;=.]+$/;
+  urlRegex = /^(https?|ftp|mqtt|ws|wss):\/\/([a-zA-Z0-9.-]+)(:[0-9]+)?(\/.*)?$/;
```
undsoft, line 5: Can we update regexp to match `whatever://anything`?
This still won't allow for some URLs.

AlexKarpov98: > Can we update regexp to match `whatever://anything`? This still won't allow for some URLs.

I'll check. In this case its really wild card 😃

AlexKarpov98: I updated it + tests.

undsoft (approved):
👍🏼

PR #11532, opened 2025-02-10
NAS-134095: Allow root disk size to be increased

undsoft:
**Changes:**

After VM is created, allows root disk size to be increased.


denysbutenko (approved):
Kapture 2025-02-12 at 19 37 28

Looks good.

PR #11542, opened 2025-02-12
NAS-133897 / 25.10 / Add disk reporting temps

bvasilenko:
**Testing:**

At the **Reporting > Disks** screen,

**Expected result:** the disk temperature data displayed correctly


undsoft (changes requested):
Disk reports, including Temperatures, are shown on m40, but there is a `Cannot read properties of undefined (reading 'split')` error when trying to access Disk reports on a CI machine.

undsoft (approved):
Reports appear to be functioning correctly now.

PR #11543, opened 2025-02-12
NAS-134042: Add a warning when failover is unhealthy to RebootRequiredDialogComponent

AlexKarpov98:
Testing: see ticket.

I worked with mocked data.

We need to have such data in store:
```
export const selectHaStatus = createSelector(
  selectHaInfoState,
  (state) => ({
    ...state?.haStatus,
    reasons: [FailoverDisabledReason.MismatchDisks],
  }),
);

export const selectCanFailover = createSelector(
  selectHaInfoState,
  () => false,
);
```

Example:
Screenshot 2025-02-12 at 08 46 42

### DIFF
--- a/src/app/modules/dialog/components/reboot-required-dialog/reboot-required-dialog.component.html
+++ b/src/app/modules/dialog/components/reboot-required-dialog/reboot-required-dialog.component.html
@@ -3,6 +3,20 @@
 </h3>
 
 <div mat-dialog-content>
+  @if (isHaLicensed() && !canFailover()) {
+    <div class="failover-reasons">
+      <p class="dialog-message error">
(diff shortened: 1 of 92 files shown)


denysbutenko (commented):
Hi @AlexKarpov98 I did apply the changes but the RebootRequiredDialog does not appear, is it expected? how to show it?

AlexKarpov98:
> Hi @AlexKarpov98 I did apply the changes but the RebootRequiredDialog does not appear, is it expected? how to show it?

@denysbutenko - mocked example will not open the dialog. I only provided data so that it will occur in the dialog.
You can manually open it in `topbar.component.ts` 

using

`this.matDialog.open(RebootRequiredDialogComponent, { minWidth: '400px' }).afterClosed()` onInit for example.

Or find a way for `tap(() => this.hasRebootRequiredReasons.set(true)),` condition to work

PR #11546, opened 2025-02-12
NAS-134147 / 25.10 / Cleanup wording in Virtualization Network section

AlexKarpov98:
Testing: 
Code review is enough.

Screenshot 2025-02-12 at 14 00 38


undsoft (approved):
Second point from the ticket also needs to be addressed.

PR #11549, opened 2025-02-12
NAS-134064 / 25.10 / Changes login status before clearing subscriptions

RehanY147:
**Changes:**

Changes login status first before clearing subscriptions so the decision to not make `core.unsubscribe` calls can be made based on updated state

**Testing:**

Test login and then logout and check browser console for errors.


undsoft:
backport

PR #11550, opened 2025-02-13
NAS-131881 / 25.10 / GUI restrictions/issues present when accessing TrueNAS in Mobile display size

AlexKarpov98:
Testing: see ticket.

Check each 12 steps provided in the ticket.

Note my comment:
```
this (9) issue deserves a separate ticket.

Drag and drop selection for manual disk selection is not possible in Mobile mode. No method in place to perform this.
```


bvasilenko (approved):
Looks good to me,

There's one issue, which not entirely related, 

When user sets language to **French (fr)**,  and enters the **Reporting** page,

**Expected result:** page should not crash and there should be no errors in the console

**Actual result:** page crashes with date-related errors in the console

https://github.com/user-attachments/assets/480205dc-310d-4fe5-bf7b-0912571b1d6b

AlexKarpov98:
@bvasilenko - I've seen that error on some machine. I used `m50` and there was no such error as I remember.
I'll check it now, If it only arises on nightlies, and will create a separate ticket.

PR #11559, opened 2025-02-14
NAS-134195 / 25.10 / Update feedback form text

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/3e9cc471-6de1-42d9-8ca0-dbb7d371d3a6

### DIFF
--- a/src/app/helptext/system/support.ts
+++ b/src/app/helptext/system/support.ts
@@ -25,6 +25,7 @@ export const helptextSystemSupport = {
   review: {
     message: {
       placeholder: T('Share your thoughts on our product\'s features, usability, or any suggestions for improvement.'),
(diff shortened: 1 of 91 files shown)


bvasilenko (changes requested):
I've noticed a weird issue in Chrome, when inactive stars are invisible.




https://github.com/user-attachments/assets/7897d407-be78-4c5e-a1a0-c738884b4b7f

AlexKarpov98:
@bvasilenko - interesting catch. I am not able to reproduce it though. 
If you want you can create a special ticket for this case, describing it a little bit more. 
Let's merge this PR because this ticket is only about the placeholder.

PR #11569, opened 2025-02-17
NAS-134072 / 25.10 / Remove CPU netdata call from main dashboard

denysbutenko:
**Changes:**

Remove call to `netdata` endpoint from CPU Recent Usage widget.

**Testing:**

Ensure no regressions.


denysbutenko:
time 00:30

undsoft (approved):
Thanks, it looks better now.

PR #11573, opened 2025-02-17
NAS-134154: Some fields not cleared when changing creation form from container to VM

AlexKarpov98:
Testing: see ticket & updated tests.


undsoft (approved):
The issue is resolved, thanks.

PR #11580, opened 2025-02-18
NAS-134245 / 25.10 / FR/UK or any other language except en brings `Invalid date format` errors for Reporting page

bvasilenko:
**Testing:**

1. At **System > General settings > Localization** change language to **French (fr)**
2. Go to **Reporting** page

**Expected Result**: page should not crash and there should be no errors in the console


AlexKarpov98 (approved):
Works well 👍

denysbutenko (approved):
The changes look good

PR #11589, opened 2025-02-18
NAS-134213 / 25.10 / Disable username in API keys on edit

undsoft:
**Changes:**

Disables Username field in API keys on edit since it cannot be changed.


AlexKarpov98 (approved):
Works as described 👍

PR #11594, opened 2025-02-19
NAS-134049: Users in STIG mode include `FULL_ADMIN` role

AlexKarpov98:
Testing: 
Use latest nightly.
We should not allow full access for user if he only has **FULL_ADMIN** role.

Middleware now returns full list of user roles, same as when STIG mode is on.

_This is initial PR and will be more (fixing roles usage)._


undsoft (approved):
Nice work, thanks.

PR #11598, opened 2025-02-19
NAS-133011 / 25.10 / Network interface graph on Dashboard does not appear to retain network traffic as seen on Reporting page

bvasilenko:
**Summary** 

This is a fix for a network widget chart.

Apart from the main issue with the chart, this PR contains a fix for an additional issue, mentioned in the last line of "Testing" section below ("...status should properly refresh and display **Success**")

**Testing:**

In this scenario user will copy the files from 1st `m40g3-137` machine to 2nd `m50-143a` machine.

1. (Skip first, do later) On 2nd server clean the destination directory in **System \> Shell**

   ```
   rm /mnt/z/iso/*.iso
   ```

2. On 1st server open and watch the chart at **Dashboard** \> **widget** "**Interface** `eno1np0` **"**
2. On 1st server click **Data Protection \> Rsync Tasks \> `mnt/pewl/iso` item \> "Run job" icon**
3. At the tab opened in pt "2"
   1. ensure you can see a traffic peak displayed on 00:13 second of the video,
   2. leave the page (by clicking **Storage** menu item)
   3. go back to the page **without** pressing F5
   4. notice you see a traffic peak (which was previously missing at 00:25 second of the video)

Video captured before the fix:

   https://github.com/user-attachments/assets/740465f0-15d9-4116-a471-7a13d3731134

**Expected result:**

1. After going to back to the page **without** pressing F5, chart in the widget should display the freshly loaded state
   
   image

2. After user waits more time, chart in widget should update
3. To verify that additional issue is also fixed: For `/mnt/pewl/iso` item in **Rsync Tasks** UI block:\
   status should properly refresh and display **Success** while user is not leaving the page

### DIFF
--- a/src/app/pages/dashboard/services/widget-resources.service.ts
+++ b/src/app/pages/dashboard/services/widget-resources.service.ts
@@ -19,6 +19,7 @@ import { VolumesData, VolumeData } from 'app/interfaces/volume-data.interface';
 import { ApiService } from 'app/modules/websocket/api.service';
 import { processNetworkInterfaces } from 'app/pages/dashboard/widgets/network/widget-interface/widget-interface.utils';
 import { AppState } from 'app/store';
+import { systemInfoLoaded, systemInfoUpdated } from 'app/store/system-info/system-info.actions';
 import { waitForSystemInfo } from 'app/store/system-info/system-info.selectors';
 
 /**
@@ -96,6 +97,8 @@ export class WidgetResourcesService {
   );
 
   networkInterfaceLastHourStats(interfaceName: string): Observable<ReportingData[]> {
+    this.store$.dispatch(systemInfoLoaded({ systemInfo: null }));
+    this.store$.dispatch(systemInfoUpdated());
     return this.serverTime$.pipe(
       take(1),
       switchMap((serverTime) => {
--- a/src/app/pages/dashboard/services/widget-resources.service.ts
+++ b/src/app/pages/dashboard/services/widget-resources.service.ts
@@ -37,7 +37,7 @@ export class WidgetResourcesService {
   readonly realtimeUpdates$ = this.api.subscribe('reporting.realtime');
 
   readonly refreshInterval$ = timer(0, 5000).pipe(startWith(0));
-  private readonly triggerRefreshSystemInfo$ = new Subject<void>();
+  private readonly triggerRefreshDashboardSystemInfo$ = new Subject<void>();
 
   readonly backups$ = forkJoin([
     this.api.call('replication.query'),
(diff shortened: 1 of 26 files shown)


src/app/pages/dashboard/services/widget-resources.service.ts
```diff
@@ -96,6 +97,8 @@ export class WidgetResourcesService {
   );
 
   networkInterfaceLastHourStats(interfaceName: string): Observable<ReportingData[]> {
+    this.store$.dispatch(systemInfoLoaded({ systemInfo: null }));
+    this.store$.dispatch(systemInfoUpdated());
```
undsoft, line 101: I think it's kinda weird that a method called `networkInterfaceLastHourStats` makes changes like this.

undsoft (changes requested):

src/app/pages/dashboard/services/widget-resources.service.ts
```diff
@@ -196,6 +198,11 @@ export class WidgetResourcesService {
   }
 
   refreshSystemInfo(): void {
```
undsoft, line 200: Can you explain what's going on here? 
The confusion around a "get" method like `networkInterfaceLastHourStats` making changes to system store stands.

bvasilenko: the problem with the bug was that the WebUI was sending a request to the middleware with an outdated date stamp. To solve this, I called `systemInfoUpdated` to update it. But the subscription to `serverTime$` was running before this update, and the `take(1)` operator prevented it from running again after `serverTime$` was updated.

I set the `systemInfo` value to null so that `waitForSystemInfo` would wait for new data to arrive and the logic would work correctly.

undsoft: So if the problem is with `serverTime$`, can't we fix it somehow? Or why do we need `serverTime$` at all? Why can't we just take current timestamp and current timestamp minus one hour?

bvasilenko: Getting `serverTime$` from the middleware is necessary because of a time synchronization issue between the user's system and the server.
For example, time zones or DST may be different, and the request requires exact start and end times.

undsoft: But you use timestamps in `reporting.netdata_get_data`, so they are the same regardless of timezone or DST.

undsoft (changes requested):
See previous comment.

src/app/pages/dashboard/services/widget-resources.service.ts
bvasilenko: What I meant was that `serverTime` may not match `new Date().getTime()` for various reasons. `serverTime` ensures that it is the current time for reporting.

undsoft: The only reason it may not match if either system has incorrect local time. I don't think we need to plan for this.

undsoft (commented):
☝🏼

undsoft (changes requested):
☝🏼

undsoft (approved):
Looks good, thanks.

bvasilenko:
backport

PR #11609, opened 2025-02-21
NAS-134298 / 25.10 / Use explicit/specific roles for components instead of Role.FullAdmin

AlexKarpov98:
Testing: see ticket.


denysbutenko (approved):
Looks good

PR #11610, opened 2025-02-21
NAS-132258 / 25.10 / Replication task stays on "Running" although it has finished

bvasilenko:
**Changes:**

Fix displaying of replication task status

**Testing:**

In this scenario user will copy the files from 1st `m40g3-137` machine to 2nd `m50-143a` machine.

1. (Skip first, do later) On 2nd server clean the destination directory in **System > Shell**,

   ```
   rm /mnt/z/iso/*.iso
   ```

2. On 1st server,

   1. go to **Datasets** page, select **`pewl/iso` dataset**, and click **Create Snapshot** in the sidebar
   2. go to **Data Protection** page, and run the `pewl/iso - z/iso` **Replication task**

**Expected result:** while user stays on the page and waits, replication task status should transition to 
image
, and then to 
image


undsoft (approved):
Replication status appears to be working correctly.

PR #11626, opened 2025-02-25
NAS-134337 / 25.10 / Dashboard Network Graph shrinks and disappears

AlexKarpov98:
Testing: see ticket.

Before, how I reproduced it:

https://github.com/user-attachments/assets/62b7c353-61b1-4840-8649-200c7a244e01

After: 

https://github.com/user-attachments/assets/1327178f-a02e-408a-adfc-52ddb704d073


bvasilenko (approved):
It's better now

PR #11627, opened 2025-02-25
NAS-134385 / 25.10 / Cannot retrieve response after adding Authorized access

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/c9d798cc-c513-4f58-9516-340994c716dd


undsoft (approved):
The issue is fixed now. 👍🏼

PR #11631, opened 2025-02-25
NAS-134405 / 25.10 / Incorrect name hyperlink in apps

undsoft:
**Testing:**

Install an app (custom or not) and give it a different name.


denysbutenko (commented):
image

Given two custom apps. When I clicked on the link it resets selection to the first app in the list.

denysbutenko (commented):
image

Given a custom app created via Wizard and then converted to custom app again (shrug). No link is available.

denysbutenko (approved):

PR #11633, opened 2025-02-26
NAS-134393 / 25.10 / Expose IO bus configuration in the UI

AlexKarpov98:
**Changes:**
Instance wizard, Instance Edit Form, Instance Disks Form were updated.

**Testing:**
See ticket.

Preview:

https://github.com/user-attachments/assets/0218d61e-2a21-464a-a57f-20b337f24a2b

### Downstream

|Affects         |Reasoning
|Documentation   | IO bus configuration added


kmoore134:
Looks nice! Will be good to get this knob exposed so Windows VMs are easily viable.

undsoft (changes requested):
Overall looks good, but I would move root disk IO bus to Disks card closer to Root Disk field.
You can just show `Root Disk: 10 GiB (NVME)`.
You can repurpose Increase root disk size dialog to show two fields.

Also `Root Disk I/O Bus` shouldn't be shown in edit form for containers.

AlexKarpov98:
@undsoft `You can repurpose Increase root disk size dialog to show two fields.`

I think it's enough to adjust that value in the edit for and keep `Increase` dialog to do only one thing.

undsoft (changes requested):
Yeah, unfortunately it doesn't work. Root Disk Type is shown under disks (as it should be), but to edit this value you need to know to go to Edit in a different card.
It would be better to rename `Increase` to `Change` and show two fields in the dialog.

undsoft (approved):
Looks good, but another small fix is needed.
Right now if I click Change and only want to change root disk type without changing the size, it won't work, because MW expects a larger size if it is sent. Let's avoid sending size, if it wasn't changed.

PR #11634, opened 2025-02-26
NAS-134415 / 25.10 / Fix copyright line

bvasilenko:
Please see ticket


bvasilenko:
@undsoft Would you please suggest backporting tag? Thanks

AlexKarpov98 (approved):
Works good. 
Add tests case for it.

PR #11637, opened 2025-02-27
NAS-133894 / 25.10 / TrueNAS enterprise hint

AlexKarpov98:
**Testing:**
See ticket.

|Affects         |Reasoning
|Documentation   | System Info Widget Updated


src/app/modules/use-enterprise-marketing-link/use-enterprise-marketing-link.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  getCurrentMessage(lastMessageHash: string | null): string {
+    const currentIndex = this.messages.findIndex((message) => this.hashMessage(message) === lastMessageHash);
+    return currentIndex >= 0 ? this.messages[currentIndex] : this.messages[0];
+  }
+
+  hashMessage(message: string): string {
+    return btoa(encodeURIComponent(message));
+  }
+
+  trackClick(message: string): void {
```
undsoft, line 70: Let's just use [href] so that user can open this in the new tab.

undsoft (changes requested):
Interesting approach with showing different messages.
Let's try with two lines. The first line is a message, and the second, also centered is `Explore TrueNAS Enterprise`.
We can reduce margins to make the text fit.

src/app/modules/use-enterprise-marketing-link/use-enterprise-marketing-link.component.ts
AlexKarpov98: @undsoft, I used one liner because of this comment.
Screenshot 2025-02-27 at 22 34 50
Do you think we should go ahead with 2 lines?

undsoft: Yes, I've discussed this with Patrick.

AlexKarpov98:
@undsoft - updated version:
Screenshot 2025-02-28 at 10 51 38

undsoft (approved):
Look and messaging was approved. Thanks.

PR #11644, opened 2025-02-28
NAS-134464: Update VictorOps alert service name to Splunk On-Call

AlexKarpov98:
Code review is enough.

`SplunkOnCall = 'VictorOps',` is updated in that way because middleware does not have that value updated yet. I added comment and waiting for response from the middleware side.


undsoft (commented):
The code looks okay, but I don't see VictorOps or SplunkOnCall in `cloudsync.credentials.query`. Please confirm with middleware why it's not being shown.

undsoft (approved):
Was looking at the wrong thing. Alert service renamed correctly.

PR #11654, opened 2025-03-03
NAS-134371 / 25.10 / Fix unit tests

bvasilenko:
**Changes:**

The problem is caused by the `repeat` operator, which was added recently (line 149). By adding `untilDestroyed(this)` the tests passed locally.

image

**Note**: I was able to reproduce the issue by spinning up Ubuntu instance, now it should pass in the CI pipeline.


AlexKarpov98 (commented):
Could you pull latest changes please?
I removed that `repeat({ delay: () => this.iscsiService.listenForDataRefresh() }),` maybe it already fixed unit tests?

PR #11662, opened 2025-03-04
NAS-134329 / 25.10 / Show core.download job errors

undsoft:
**Changes:**

`core.download` returns a job, which we didn't wait for to complete previously in most cases.

**Testing:**

Check file downloads in the UI.
You can also check error handling by supplying incorrect arguments in `coreDownload` in some jobs.


RehanY147 (approved):
Looks good

PR #11663, opened 2025-03-04
NAS-134556 / 25.10 / Fix bind_ip for smb service config

RehanY147:
**Changes:**

Adjusts UI to the response schema from middleware. Removes addition to bind_ip choices.

**Testing:**

Test normal function. Set, unset bind_ip for smb config under System -> Services -> Edit Config (SMB)

### DIFF
--- a/src/app/interfaces/bind-ip.interface.ts
+++ b/src/app/interfaces/bind-ip.interface.ts
@@ -1,3 +1,0 @@
-export interface BindIp {
-  bindIp: string;
-}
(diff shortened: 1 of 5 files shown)


src/app/interfaces/bind-ip.interface.ts
```diff
@@ -1,3 +1,3 @@
 export interface BindIp {
```
AlexKarpov98, line 1: `BindIp` is only used in `service-smb.component.ts` - we may remove it from global interfaces.

AlexKarpov98 (approved):
Works as expected 👍

PR #11665, opened 2025-03-04
NAS-134447 / 25.10 / Allow user and group ids to be mapped within instances

undsoft:
**Changes:**

Adds an option to manage user and group id mapping between host and instances.
See ticket for proper description.
Снимок экрана 2025-03-04 в 15 19 55

**Known issues**
You need to create a new user/group to be able to add entry. Builtin users/groups are not supported even though they are shown in the combobox. I will create a separate ticket to address it.

### Downstream

|Affects         |Reasoning
|Documentation   | New functionality in Instances.


bvasilenko (approved):
works good!

PR #11667, opened 2025-03-05
NAS-133871 / 25.10 / Fix virt instance is not selected when it is created

denysbutenko:
**Changes:**

Fix instance is not selected when it is created

**Testing:**

Create two instances. Ensure the second is selected in the tree when created.


denysbutenko:
time 6:30

AlexKarpov98 (approved):
It works for new instances selection 👍 

_small issue:_
Noticed, if I put unexisting id in the url, i am not redirected to the first instance but see an empty screen:
Screenshot 2025-03-05 at 12 25 53

denysbutenko:
> _small issue:_ Noticed, if I put unexisting id in the url, i am not redirected to the first instance but see an empty screen: Screenshot 2025-03-05 at 12 25 53

thanks for highlighting it, fixed

PR #11672, opened 2025-03-05
NAS-132124 / 25.10 / WidgetResourcesService should only fetch data once per load and reload when needed

RehanY147:
**Changes:**

Previously, widget-resource-service was making calls to certain endpoints once per session and then ever updating the data no matter how much time passed and even if the user navigated away. I've changed that so the calls are remade whenever the user navigates to the dashboard page. Furthermore, I've optimized the calls so that multiple widgets asking for the same data only trigger one call. Instead of a new call for every widget.

**Testing:**

Code review. Also, check network logs to ensure that whenever you navigate to the dashboard page via either page refresh, or via navigating away to a different page and then coming back to the dashboard page, interface.query call is made only once and data is updated as per the latest call response. Also, check that every time the user navigates away from the dashboard page or refreshes the tab, the data for backup tasks is fetched and shown updated data on the UI.

Backport

### DIFF
--- a/src/app/pages/dashboard/services/widget-resources.service.spec.ts
+++ b/src/app/pages/dashboard/services/widget-resources.service.spec.ts
@@ -3,8 +3,18 @@ import {
   SpectatorService,
 } from '@ngneat/spectator/jest';
 import { firstValueFrom } from 'rxjs';
+import { TestScheduler } from 'rxjs/testing';
+import { getTestScheduler } from 'app/core/testing/utils/get-test-scheduler.utils';
 import { mockCall, mockApi } from 'app/core/testing/utils/mock-api.utils';
+import { CloudSyncProviderName } from 'app/enums/cloudsync-provider.enum';
+import { Direction } from 'app/enums/direction.enum';
+import {
+  NetworkInterfaceType, NetworkInterfaceFlag, LinkState, NetworkInterfaceAliasType,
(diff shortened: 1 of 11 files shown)


src/app/pages/dashboard/services/widget-resources.service.spec.ts
```diff
@@ -20,24 +30,202 @@ const apps = [
...
+      ],
+      failover_virtual_aliases: [
+        {
+          type: NetworkInterfaceAliasType.Inet,
+          address: '10.220.39.70',
+          netmask: 32,
+        },
+      ],
+    },
+  ] as unknown as NetworkInterface[];
```
undsoft, line 140: Unknown is not needed. You could also remove properties that are not used by the test. Since you're only comparing objects to themselves, you can reduce them to `[{ id: 1 }, { id: 2 }] as NetworkInterface[]`.

undsoft (commented):
So, I'm a bit confused by multiple approaches now used in `WidgetResources`.
1. There is your approach.
2. There is `refreshSystemInfo`.
3. And there is the original approach of using refCounts and bufferSize, for example still used by `updateAvailable$`.

Why can't we use option 3 more? It's still only called once even if there are multiple widgets of the same type and it reload every time you navigate to main dashboard.

RehanY147:
As discussed, this is not needed in `master`

PR #11677, opened 2025-03-05
NAS-134428 / 25.10 / Only reload explorer node children if going through zvols

RehanY147:
**Changes:**

The endpoint we have to load datasets hierarchy only needs to be called once and all datasets are loaded. A recent PR made it so that when loading zvols we reload children when a node is collapsed and expanded again. However, this approach only works for zvols. Doesn't work for datasets and datasets just get loaded again from the root. This will fix that issue so reloading children only happens for zvols and not for datasets.

**Testing:**

See ticket description. Also see the ticket description of https://ixsystems.atlassian.net/browse/NAS-133774 to test that the issue there is still resolved.


AlexKarpov98 (approved):
Now, Explorer node children reload only with **zvols**.

PR #11689, opened 2025-03-06
NAS-134610 / 25.10 / "Input should be a valid string" when reloading ACL editor

undsoft:
**Testing:**

See ticket.


undsoft:
@aervin Can you do some testing here as well please to make sure I didn't break something?

aervin:
@undsoft No regressions found from TNC's perspective.

denysbutenko (approved):
Looks good

PR #11692, opened 2025-03-07
NAS-134605 / 25.10 / TrueNAS hint does not look good when there is a serial number

AlexKarpov98:
Testing: see ticket + comment.
Removed code duplications and used `<ng-container *ngTemplateOutlet="skeletonLoader"></ng-container>` instead.

Result:
Screenshot 2025-03-07 at 09 44 49
Screenshot 2025-03-07 at 09 44 56


denysbutenko (approved):
👍🏻

PR #11693, opened 2025-03-07
NAS-134631 / 25.10 / Fixes reselection when instance is already selected

RehanY147:
**Changes:**

Avoids selecting an instance row when it's already the last selected instance

**Testing:**

See ticket


undsoft (approved):
The issue is resolved.

PR #11707, opened 2025-03-10
NAS-134539 / 25.10 / Re-initialize `VirtualizationInstancesStore` after config form closes

RehanY147:
**Changes:**

Re-initializes store state after config form closes to reload data if pool was changed.

**Testing:**

1- Select pool
2- Create an instance
3- Export selected pool
4- Import the pool back
5- Go to instances page and select the pool again

The instance that was created before exporting the pool should automatically load after the pool is selected again and the form is closed.


AlexKarpov98 (commented):
Screenshot 2025-03-11 at 13 16 55
There are some issues, please pull latest master changes.

AlexKarpov98 (commented):
it will start a new subscription.

`switchMap((instances: VirtualizationInstance[]) => this.api.subscribe('virt.instance.query').pipe(`

I think the previous one will be still active?

src/app/pages/instances/stores/virtualization-instances.store.ts
```diff
@@ -36,6 +42,7 @@ export class VirtualizationInstancesStore extends ComponentStore<VirtualizationI
   }
 
   readonly initialize = this.effect((trigger$) => {
+    this.killSubscription$.next();
```
AlexKarpov98, line 45: let's use: 
`private readonly destroy$ = new Subject<void>();`

	•	“Kill” sounds harsh and doesn’t clearly indicate purpose.

AlexKarpov98 (changes requested):
I noticed one more issue.
Once I swap pools for instances, it's not reselecting existing first instance but in details I see old instance.

See: 

https://github.com/user-attachments/assets/5dc173c7-c814-4dd2-8718-d86d343caf57

If one pool does not have instances, I still see details for the old instance:

see:

https://github.com/user-attachments/assets/f57b9cc7-a4f1-4293-97f6-05ff26f6d5a2

AlexKarpov98 (changes requested):
Now pool swithing is not working, I always see the same instances even for newly created pool where I haven't attached any instance yet.

As well once I close the `Global Settings` form - instances list is reloaded an I am unable to select an instance to see details.

see video:

https://github.com/user-attachments/assets/05531a69-2fcd-4bca-b3b8-b9bad2a2fd2d

AlexKarpov98 (changes requested):
@RehanY147 - it's not working for me still. Have you tested it? Please share a video if it works for you.


https://github.com/user-attachments/assets/960658c9-e262-46e9-8574-8baa6e211ec5

RehanY147:
@AlexKarpov98 It should work now

https://github.com/user-attachments/assets/0e86e288-57de-42f4-9fa5-8f48f01d6b9e

AlexKarpov98 (commented):
@RehanY147 - could you send me the video where it works for you??

RehanY147:
@AlexKarpov98 Yes. Middleware returns the same instances list even after the pool is changed in the global settings form.

RehanY147:
@AlexKarpov98 I think if you export the pool set for virtual instances, then all the instances go missing. If you switch the pool, all the data on the previous pool is moved to the new pool it seems like. So, if you saw switching pools removed all instances before, then we can ask middleware. But if exporting the pools removed the instances but switching to a different pool didn't, I think that's the behavior I've seen from the start.

RehanY147:
Even after reloading the page or delaying the request, the same data is returned by middleware. So, at best this is a middleware issue that we still see the same instances when we switch pools. @AlexKarpov98

RehanY147:
backport

PR #11710, opened 2025-03-11
NAS-134290 / 25.10 / Update TrueNAS logo with input from Marketing

AlexKarpov98:
Testing: see ticket.
As well Logo is fixed for different themes.

Result:

https://github.com/user-attachments/assets/062f600f-9337-441b-b791-331db2af676f

Screenshot 2025-03-11 at 13 11 20


RehanY147 (changes requested):
The Community/Enterprise badge remains grey-ish even when background is dark. Ticket asks for it to be white when there's a dark background in the sidebar.
Also, the logo is simply white for all but two themes. I don't know if that is according to spec. But it is multicolored for Ix Dark theme and High Contrast theme. Maybe it should have some other colors? 
image

AlexKarpov98:
Let's see what @undsoft thinks about it 🙌

undsoft (approved):
I think it looks fine, but I would adjust opacity to 1.

PR #11713, opened 2025-03-11
NAS-133871 / 25.10 / Update redirects on instance list

denysbutenko:
**Changes:**

Forward port of changes from FangTooth.

**Testing:**

Ensure that the instances page redirects:

- to the given route param (change URL manually)
- to the first item when no route param
- to the first item when an invalid route param
- to the newly created instance

The selected instance does not make extra calls when the same row is selected twice.


undsoft (changes requested):
Navigating to list of instances, selecting second row and reloading results in first row being selected.
The code is somewhat convoluted, not sure if we can make it clearer.

denysbutenko:
> Navigating to list of instances, selecting second row and reloading results in first row being selected. The code is somewhat convoluted, not sure if we can make it clearer.

I did try to simplify it here. Also, I created a [ticket](https://ixsystems.atlassian.net/browse/NAS-133871) to simplify this logic and make it reusable.

undsoft (changes requested):
In this version I get `[ENOENT] None: VirtInstance test does not exist` when I delete an instance.

PR #11719, opened 2025-03-12
NAS-134684 / 25.10 / Unify design language of apps and instances

AlexKarpov98:
**Testing:**
See ticket.
We update stats data for each Running instance with 5 seconds interval.

Result:

Screenshot 2025-03-14 at 12 28 53

|Affects         |Reasoning
|Documentation   | New fields added to the table

### DIFF
--- a/src/app/pages/instances/components/all-instances/instance-details/instance-general-info/instance-general-info.component.html
+++ b/src/app/pages/instances/components/all-instances/instance-details/instance-general-info/instance-general-info.component.html
@@ -16,10 +16,6 @@
   <mat-card-content>
     @if (instance(); as instance) {
       <p>
-        <span class="label">{{ 'Status' | translate }}:</span>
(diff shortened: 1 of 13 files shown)


undsoft (commented):
Okay, so I interpreted that ticket as just fixing status pill so that they look the same.
I don't know if subscribing to multiple `virt.instance.metrics`. I've asked Caleb about it and will let you know.
Meanwhile status pill can be updated to match the look on apps page.

undsoft:
Also navigating to another page doesn't stop the subscription.

undsoft:
After speaking to Caleb, subscribing to multiple `virt.instance.metrics` is no bueno. Let's remove this part and only update the status column. I'll create a ticket for MW to provide us with a better endpoint if possible.

undsoft (approved):
Снимок экрана 2025-03-14 в 12 48 25

Looks good.

PR #11720, opened 2025-03-12
NAS-134719 / 25.10 / Adds tooltip to System navigation item in the sidebar

RehanY147:
**Changes:**

Adds tooltip to System navigation item in the sidebar

**Testing:**

See that tooltip is shown when hovered over System

### DIFF
--- a/src/app/interfaces/menu-item.interface.ts
+++ b/src/app/interfaces/menu-item.interface.ts
@@ -14,6 +14,7 @@ export interface MenuItem {
   tooltip?: string;
   sub?: SubMenuItem[];
   isVisible$?: Observable<boolean>;
+  hideTooltipOnSidebarCollapsed: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


src/app/interfaces/menu-item.interface.ts
```diff
@@ -14,6 +14,7 @@ export interface MenuItem {
   tooltip?: string;
   sub?: SubMenuItem[];
   isVisible$?: Observable<boolean>;
+  showTooltipAlways?: boolean;
```
AlexKarpov98, line 17: it's better to use `alwaysShowTooltip`, but it's not about this ticket, please check it once again.

AlexKarpov98 (changes requested):
You missed this part:

Screenshot 2025-03-14 at 17 48 20

This ticket is basically only about the description above.
I think you didn't get it right.

PR #11724, opened 2025-03-13
NAS-124151 / 25.10 / Sentry Integration

undsoft:
**Changes:**

* ApiService and friends now throw classes instead of objects. This is necessary to preserve stack trace for Sentry.
* Split `ErrorHandlerService` in `ErrorParserService` and `ErrorHandlerService`. Moved them in a subfolder.
* Integrated with Sentry.
* Streamlined usages of `ErrorHandlerService`. You should just call `this.errorHandler.showErrorModal(error);` now.

**Testing:**

Open `main.ts` and move `Sentry.init` code so that it is called.

Error handling in different scenarios needs to be tested. It is expected that non-validation related errors are logged into Sentry with minified stack trace. Sentry URL will be in the ticket.


AlexKarpov98 (commented):
Code looks good! 
I was not able to get an error in sentry though.

AlexKarpov98 (approved):
Error Handling Works!
I tested all possible errors which came into my mind.

PR #11725, opened 2025-03-13
NAS-134729 / 25.10 / Change defaults for Export Password Secret Seed

undsoft:
Changes defaults for `Export Password Secret Seed` in `Save Configuration` dialog.


AlexKarpov98 (approved):
👍

PR #11726, opened 2025-03-13
NAS-134326 / 25.10 / Change how alerts are sorted

undsoft:
Alerts will be sorted by level, then by date, then by class.

More alerts are available on m40.


denysbutenko (approved):
image

Looks good.

PR #11734, opened 2025-03-17
NAS-134574 / 25.10 / Add optional vertical height to shorter components to make it easier for horizontal alignment

RehanY147:
**Changes:**

Add optional vertical height to shorter components to make it easier for horizontal alignment for inline controls

**Testing:**

See the user form. The slide toggle should align properly to other horizontal components

### DIFF
--- a/src/app/modules/forms/ix-forms/components/ix-checkbox/ix-checkbox.component.html
+++ b/src/app/modules/forms/ix-forms/components/ix-checkbox/ix-checkbox.component.html
@@ -1,27 +1,30 @@
 <div [class.extended-height]="extendedHeight()">
-  <mat-checkbox
-    color="primary"
-    [checked]="value"
-    [required]="required()"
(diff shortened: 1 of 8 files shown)


AlexKarpov98 (changes requested):
It didn't solve the originally described issue.

That's how it looks now:

Screenshot 2025-03-17 at 23 47 16

I don't think it's a good approach, I can't understand how it will help us, with static height, it's kind of hardcoded workaround.
Inline form items are not only available for user form, we need a global fix.

_______

I've seen that the issue may be related to the `ix-select` label which breaks the alignment.

I added `display: none` for the **form > ix-combobox > div > ix-label > label**
+ removed margins from the checkbox and the button
And `baseline` flex alignment seems to work for the wrapper.
Screenshot 2025-03-17 at 23 56 29

RehanY147:
Need a different approach

PR #11747, opened 2025-03-18
NAS-134868 / 25.10 / Instance shutdown dialog UI doesn't match API

undsoft:
**Changes:**

Instead of showing a separate Force checkbox in the shutdown dialog, it's now one of the options in the dropdown, as this matches API behaviour.
Снимок экрана 2025-03-18 в 13 57 54

### Downstream

|Affects         |Reasoning
|Documentation   |Shutdown/restart dialog in instances updated.


AlexKarpov98 (approved):
Works good 👍

PR #11755, opened 2025-03-19
NAS-134752 / 25.10 / UI should lock ability to add disks to running VMs

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/57870882-d02e-4d67-aa1c-28ce6d31045e


src/app/pages/instances/components/all-instances/instance-details/instance-disks/instance-disks.component.html
```diff
@@ -1,3 +1,6 @@
+@let isVmRunning = instance().status === instanceStatus.Running && instance().type === instanceType.Vm;
```
undsoft, line 1: Better not to overload template with business logic and have these as computed properties in the component.

src/app/pages/instances/components/all-instances/instance-details/instance-disks/instance-disks.component.html
```diff
@@ -22,13 +30,16 @@ <h3 mat-card-title>
...
         </span>
-        <a
-          class="action"
-          ixTest="change-disk-setup"
-          (click)="showRootDiskIncreaseDialog()"
-        >
-          {{ 'Change' | translate }}
-        </a>
+
+        @if (!isVmRunning || isContainer) {
```
undsoft, line 34: Let's keep the same approach and have this and actions below as disabled instead of hiding, so that user knows that the option is there.

src/app/pages/instances/stores/virtualization-devices.store.ts
```diff
@@ -74,6 +74,10 @@ export class VirtualizationDevicesStore extends ComponentStore<VirtualizationIns
     this.loadDevices();
   }
 
+  setSelectedInstance(instance: VirtualizationInstance): void {
```
undsoft, line 77: If there is already exists functionality to select rows, why add this method and call it in `InstanceRow`?

undsoft (changes requested):

src/app/pages/instances/stores/virtualization-devices.store.ts
undsoft: If the logic was already working and rows were getting selected, why add this?

AlexKarpov98: @undsoft 
We had an issue with updating rows once state is changed (stopped, running)

undsoft: Then it would be good to fix the underlying issue with updates.

AlexKarpov98: @undsoft - that's why I added new method to update selected instance with latest updates. We need to have it updated in the store.

undsoft (approved):
I see what the problem is with `setSelectedInstance` and other methods. It's a poor design overall, but it may be tricky to fix now. I will create a separate ticket for it.

For this task, it would be good to match disabled menu icon to other disabled buttons in details. Currently it's black.
Снимок экрана 2025-03-24 в 12 55 07

AlexKarpov98:
Updated;
Screenshot 2025-03-25 at 17 19 03

PR #11756, opened 2025-03-19
NAS-134844 / 25.10 / Discover Search, with Category filter returns broken App icons

AlexKarpov98:
Testing: see ticket.

Now we are using the base working-good functionality from `ng-lazyload-image` instead of custom observable.
See that images are loading lazily on scroll, as we expect it to be.

Additional Fix: 
- preloader image is now showing up instead of broken image (incorrect path to un-existing image)

Result:

https://github.com/user-attachments/assets/557e14b4-ceff-4d88-8126-e58dd70026ee


undsoft (approved):
Images are there, but can we make black placeholder less black for example to match placeholder you see when you go to Reports?

denysbutenko (approved):
Check `index.html` for the opacity animation to address Evgeny's comment

undsoft:
Okay, I've updated to white icon. I'll merge it because of code freeze, it should be okay for most people that use dark theme.
But I'll keep the ticket open, because 1) we need to fix it properly 2) it's also weird to use something called `favicon`, the file probably needs to be renamed or another image used.

PR #11767, opened 2025-03-20
NAS-134583 / 25.10 / Updates to VM volumes

undsoft:
**Changes:**

- It is now possible to create new volumes in Volumes dialog.
- It is now possible to import zfs zvols as volumes.
- It's possible to set boot priority on vm disks.
- User is directed to import zvols instead of using them directly.

**Testing:**

Test new options in Volumes dialog and test different selections for root disk when creating a VM.

### Downstream

|Affects         |Reasoning
|Documentation   |New UI options.


AlexKarpov98 (approved):
It works as expected. During Create&Edit time I was able to work with `virt.volumes` successfully.
Screenshot 2025-03-20 at 11 06 13

PR #11775, opened 2025-03-21
NAS-134945 / 25.10 / Relax component naming requirements

undsoft:
**Changes:**

Angular [moves in the direction](https://github.com/angular/angular/discussions/58412) of not requiring `Component` and other suffixes.

While I'm not sure about doing that universally, this takes the first step in this direction.

1. Disables relevant linter rule.
2. Renames `SomeDialogComponent` to just `SomeDialog`.

**Testing:**

Review.


denysbutenko (approved):
Looks good!

PR #11779, opened 2025-03-24
NAS-134876: Map User And Group IDs does not re-enable Set button

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/599b5711-505b-467b-85f5-af1901ed20b4


denysbutenko (approved):
image

It works great now

PR #11780, opened 2025-03-24
NAS-134880 / 25.10 / Fix dependent checkbox in Security form

undsoft:
**Changes:**

1. Open security settings.
2. Enable STIG.
3. FIPS is enabled too as expected.
4. Disable STIG.
5. FIPS should stay enabled.


AlexKarpov98 (approved):
Works as described 👌

PR #11781, opened 2025-03-24
NAS-134788 / 25.10 / Remove system.general.language_choices call from UI

denysbutenko:
**Changes:**

Remove system.general.language_choices call from UI

**Testing:**

Ensure no regressions in webui language settings


undsoft (changes requested):
PR does not appear to be finished.

undsoft (approved):
Works well, but it's strange that there are so many strings resorted in translation files.

denysbutenko:
> Works well, but it's strange that there are so many strings resorted in translation files.

It is due to changes from `.sort()` to `.toSorted()`

PR #11786, opened 2025-03-25
NAS-134700 / 25.10 / Allow only one disk to be selected for Spare Vdevs when creating/editing pools

RehanY147:
**Changes:**

Removes the normal `ix-layout-step` and adds a simple `ix-select` to select a disk to be added to the STRIPE Spare Vdev when creating a pool.

**Testing:**

Testing creating a pool with a spare vdev by choosing a disk from the new dropdown. Test that the selected disk is the one added to the pool by observing the pool devices. Also, remove the spare disk from the devices page and then click on the "Add Vdev" button and select another disk to add as a Spare Vdev disk. Check that the selected disk is updated correctly for the pool topology. Also, without removing the Spare vdev, go to the "Add Vdev" and you should not see the Spare step if there's already a spare disk selected for the pool.


undsoft (changes requested):
Overall looks good, but some issues.
1. Selecting a spare should not be required. Right now clicking on `--` triggers a validation and console error.
2. Would be good to rename `--` to something like `None` if there is an easy way to do it.
3. It would be helpful if dropdown was to list `disk name – type – size`.
4. Reset Step button doesn't do anything.
5. When adding a spare vdev to an existing pool summary in sidebar doesn't show Spare when it is selected.
6. In other cases summary and review on the last step says `Manual Layout | 1 Vdev`. Would be better to just describe the disk selected.

undsoft (approved):
Looks good, thanks.

PR #11788, opened 2025-03-25
NAS-134790 / 25.10 / Update zfs.snapshot calls

undsoft:
**Testing:**

Test affected components.


AlexKarpov98 (approved):
Works fine. No issues found.

PR #11790, opened 2025-03-25
NAS-134943 / 25.10 / Improvements to NIC form

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/d0b9d1bd-c024-40c9-926d-a8acf2957cdd


src/app/pages/network/components/interface-form/interface-form.component.ts
```diff
@@ -303,12 +303,33 @@ export class InterfaceFormComponent implements OnInit {
     });
   }
 
+  private generateRandomNameByType(type: NetworkInterfaceType): string | null {
+    const random = Math.floor(Math.random() * 10000);
```
undsoft, line 307: Yeah, I don't think random is a good approach. It should be next available.

undsoft (changes requested):

src/app/pages/network/components/interface-form/interface-form.component.ts
```diff
@@ -184,12 +184,15 @@ export class InterfaceFormComponent implements OnInit {
     private matDialog: MatDialog,
     private systemGeneralService: SystemGeneralService,
     private store$: Store<AppState>,
-    public slideInRef: SlideInRef<NetworkInterface | undefined, boolean>,
+    public slideInRef: SlideInRef<{
```
undsoft, line 187: The `| undefined` part is no longer true.

undsoft (approved):
Interface names generated as expected.

PR #11801, opened 2025-03-27
NAS-134638 / 25.10 / If the browser is not refreshed after changing the hostname, the debug file downloads as the old hostname

AlexKarpov98:

Now it gets latest systm info data before generating report.

Result:
Note: I change hostname on the remote machine and Save Debug locally.

`this.store$.dispatch(systemInfoLoaded({ systemInfo: null })` is used in order to wait for system info to be loaded (latest data) for generating Debug.

https://github.com/user-attachments/assets/4508be5c-d416-4c0d-8487-e54695ae56eb

### DIFF
--- a/src/app/pages/system/advanced/save-debug-button/save-debug-button.component.ts
+++ b/src/app/pages/system/advanced/save-debug-button/save-debug-button.component.ts
@@ -2,10 +2,13 @@ import { DatePipe } from '@angular/common';
 import { ChangeDetectionStrategy, Component } from '@angular/core';
 import { MatButton } from '@angular/material/button';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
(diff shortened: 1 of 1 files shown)


undsoft (changes requested):
So if hostname change occurs in a different form, why not dispatch `systemInfoUpdated` there instead of dispatching fake actions in a different place?

AlexKarpov98:
> So if hostname change occurs in a different form, why not dispatch `systemInfoUpdated` there instead of dispatching fake actions in a different place?

I already did it before. Please read the comments in the ticket.
We may simply reject this PR if needed.
Actually request is not that common and valid for UI users.

AlexKarpov98:
For UI part it's DONE --> https://ixsystems.atlassian.net/browse/NAS-134638?focusedCommentId=303165

PR #11800, opened 2025-03-27
NAS-134914 / 25.10 / Simplify selection of bootable disk

AlexKarpov98:
Known Issue:
new API `set_bootable_disk` resets/removes `io_bus` for the selected disk --> M.Rehan is aware of it and prepares a fix.

Note:
new API endpoint accepts `vm_id` and `disk_name` (not `disk_source`).

**Testing:**
See ticket. New logic is available for Instance Wizard & Instance Disks Card.

Result:

https://github.com/user-attachments/assets/ea3e788a-1b38-411e-a5bc-e28ada0a2372

### Downstream

|Affects         |Reasoning
|----------------|-------------------------------
|Documentation   | New Logic For Primary (Boot) Disk Selection

PR #11811, opened 2025-03-29
Update it.json

dany22m:
Re-added and Re-removed strings updated in commit #11800 + additional translation.
Now the file has the same number and the same strings like the other language files.


undsoft (approved):
Thanks!

PR #11812, opened 2025-03-31
NAS-135044 / 25.10 / Too many quotas are reported in Dataset Space Management card

undsoft:
**Changes:**

Previously Dataset Space Management card would report too many quotas.


AlexKarpov98 (approved):
👍 Nice

PR #11814, opened 2025-03-31
NAS-135011 / 25.10 / [BUG] email showing wrong SMTP server than what is configured.

AlexKarpov98:
Testing: see ticket.

Result:
NAS-135011-after
Before:
NAS-135011-before


undsoft (approved):
Looks reasonable, thanks.

PR #11820, opened 2025-03-31
NAS-134324: Present ticket info in a more durable format

undsoft:
**Changes:**

Shows a modal dialog instead of snackbar after reporting a bug.

**Testing:**

When testing, update `onSubmit` in `file-ticket-licensed` or `file-ticket` components to avoid creating junk tickets:

```
onSubmit(): void {
    this.onSuccess('http://google.com');
    return;
```


src/app/modules/feedback/services/feedback.service.ts
```diff
@@ -176,31 +176,27 @@ export class FeedbackService {
...
-          action: () => this.window.open(ticketUrl, '_blank'),
-        },
-      },
-      duration: 10000,
-    });
+  showTicketSuccessMessage(ticketUrl: string): void {
+    this.dialogService.generalDialog({
+      message: this.translate.instant('Thank you. Ticket was submitted successfully.'),
+      icon: iconMarker('check'),
+      title: 'Ticket Created',
```
denysbutenko, line 183: Is it intentionally untranslatable?

denysbutenko (approved):
image

Looks good

src/app/modules/feedback/services/feedback.service.ts
undsoft: No, it isn't.

PR #11822, opened 2025-04-01
NAS-135089 / 25.10 / Merge two profile cards on new users page

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/4bef09f3-a618-4c28-9a95-9c4b0987de6b


undsoft (approved):
👍🏼 Looks good

undsoft:
The reason coverage requirement fails is because DeleteUserDialogComponent test tests component from a different module.

PR #11824, opened 2025-04-01
NAS-134494 / 25.10 / Add user picker component

denysbutenko:
**Changes:**

Added `ix-user-picker` component with ability to create new users.

**Testing:**

Use the `API Key form` on the `My API Keys` page. Try to create a new user. Ensure you add a group with privileges to the user (e.g., `builtin_administrators`).

Known issues:
- Duplicate entries on scroll.
- Empty list when no priveledges group is added.


AlexKarpov98 (commented):
Overall looks good.
But:
- Tests failed
- When I create new user as you described I don't see the full list of users, but only 
`Add New & New user created`. Not sure if it's good for the best UX.
Screenshot 2025-04-02 at 10 55 18

Known issues confirmed.

denysbutenko:
> Overall looks good.
> 
> But:
> 
> - Tests failed
> - When I create new user as you described I don't see the full list of users, but only 
> 
> `Add New & New user created`. Not sure if it's good for the best UX.
> 

This is API Key specific case depends on Query Params passed to ix-user-picker. So it expected to behave like this.

> Screenshot 2025-04-02 at 10 55 18
> 
> 
> 
> Known issues confirmed.
> 
> 
> 
> 

Remaining fixes are in progress. Thanks!

denysbutenko:
Changes:

1. Added hint to inform users. The form won't allow the creation of users with no privileges in the future.
2. Fixed duplicates on scroll.
3. The failed test case was replaced with a to-do item to add it back.

PR #11831, opened 2025-04-02
tenant_id should not required for Swift credentials

dedebenui:
**Changes:**

Removed `tenant_id` as required for OpenStack Swift credentials. The [Rclone doc](https://rclone.org/swift/#swift-tenant) says that at least one of `tenant` (i.e. `tenant_name`) or `tenant_id` is required, but not necessarily `tenant_id`.

Related: https://forums.truenas.com/t/cannot-add-cloud-credentials-because-optional-key-is-required-in-ui/38163


AlexKarpov98 (approved):
Thanks.
@undsoft  FYI

PR #11844, opened 2025-04-03
NAS-135036 / 25.10 / Fix http request called twice

denysbutenko:
**Changes:**

Fix for 401 Unauthorized error.

**Testing:**

See ticket.


undsoft (approved):
Thanks, fix makes sense.

PR #11852, opened 2025-04-04
NAS-135183 / 25.10 / Remove support for BOTH in share ACLs

undsoft:
**Testing:**

Create a VM connected to Active Directory, create a share and edit share acl.
You can try getting permissions in a state where the form is incorrect.


AlexKarpov98 (approved):
`Both` option is removed carefully 👌

PR #11854, opened 2025-04-04
Updated Irish translations

aindriu80:
**Changes:**

Updated Irish/Gaeilge translations, also fixed some strings which were inconsistent.

**Testing:**

Examined the JSON file in Poedit for errors.  All ok.


undsoft (changes requested):
Please review the errors in the CI job that fails.
I can see that there some issues with how plural strings are translated.
This may help https://github.com/truenas/webui/blob/master/docs/contributing_translations.md#plural-forms-in-different-languages

For example in this string:
```
{n, plural, =0 {No Errors} one {# Error} other {# Errors}}
```

`plural`, `one` and `other` are keywords that don't need to be translated.

Translated string should probably look something like this:

```
{n, plural, =0 {Gan Earráidí} one {# Earráid} other {# Earráidí}}
```

denysbutenko (approved):
CI passed. Looks good.

undsoft (approved):
Thanks for the contribution.

PR #11858, opened 2025-04-07
NAS-135157 / 25.10 / Remember user selected columns in tables

AlexKarpov98:
Testing: see ticket.
p.s _I don't see a reason for a separate service for these updates._

Current possible issue: We only know `title` or `propertyName`, but they are not always available.
maybe we can provide additional key for each column, especially to know how we save that column to the API.
Currently I use title. (so the selection changes when we change the language) 

Result:

https://github.com/user-attachments/assets/1dc7d022-4986-4fd2-b255-5a516273c9aa

### DIFF
--- a/src/app/interfaces/preferences.interface.ts
+++ b/src/app/interfaces/preferences.interface.ts
@@ -1,16 +1,8 @@
 import { SidenavStatusData } from 'app/interfaces/events/sidenav-status-event.interface';
 
-interface Column {
-  name: string;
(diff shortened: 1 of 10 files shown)


denysbutenko (changes requested):
`Reset to defaults` does nothing and causes max concurrent calls when clicking multiple times.

denysbutenko (approved):
image

Looks good!

PR #11860, opened 2025-04-08
NAS-135156 / 25.10 / Fix missing space between sentences

denysbutenko:
**Changes:**

Minor fix.

**Testing:**

See ticket for details


undsoft (approved):
The issue is fixed, but these messages are not currently translated. Each part of `sysUpdateMessage + ' ' + sysUpdateMsgPt2` needs to be translated separately.

PR #11863, opened 2025-04-08
NAS-134938 / 25.10 / Physical ports in Fibre Channel Ports do not display WWPNs unless target is mapped

AlexKarpov98:
Testing: see ticket.

Before:
Screenshot 2025-04-08 at 16 37 55

After:
Screenshot 2025-04-08 at 16 37 58


undsoft (approved):
Looks good.

PR #11884, opened 2025-04-14
NAS-135258 / 25.10 / Improvements for new user form v2

AlexKarpov98:
Testing: see ticket.
Fixed:
- Now Save button is disabled if any part of the form is invalid, not only the username part.
- Now list of users is updated once new user is created + new user is pre selected. (it required middleware fix, you can check linked ticket)

Added :
`builtin=false` filter to show users the same way as on old users screen (for now)


denysbutenko (approved):
Looks nice!

PR #11885, opened 2025-04-14
NAS-135297 / 25.10 / Actions column disappears when columns are removed in Column Selector

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/c752b47b-81f3-42ac-bc11-123687ffad41


src/app/modules/ix-table/components/ix-table-columns-selector/ix-table-columns-selector.component.ts
```diff
@@ -80,6 +80,8 @@ export class IxTableColumnsSelectorComponent<T = unknown> implements OnChanges,
       untilDestroyed(this),
     ).subscribe((displayedColumns) => {
       this.columns().forEach((column) => {
+        if ('actions' in column) return;
```
undsoft, line 83: Maybe extract `actions` in a constant to use here and in other places.

undsoft (approved):
The issue is fixed. 👍🏼

src/app/modules/ix-table/components/ix-table-columns-selector/ix-table-columns-selector.component.ts
AlexKarpov98: > Maybe extract `actions` in a constant to use here and in other places.

@undsoft I don't see other places where it's used

undsoft: Well, there is some connection to the `IxCellActionsComponent`. It would be good to make sure it's not fragile. For example if you can compare type to `IxCellActionsComponent` that may be better than checking for a key.

AlexKarpov98: @undsoft, yeah - sounds good.

Gonna use `if (column instanceof IxCellActionsComponent) return;`

PR #11886, opened 2025-04-14
NAS-135046 / 25.10 / Make login page more resilient

denysbutenko:
**Changes:**

- Add reconnect message component
- Add error handler for login banner call

**Testing:**

Wait on login page for 10 secs and connection will be dropped. Expected to see reconnect message with a button. If session is expired, expected to see login form (no reconnect message).


denysbutenko:
Demo 
https://github.com/user-attachments/assets/b9317ab1-824a-4a9c-9beb-3699d34a3f2c

undsoft (changes requested):
The ticket was about fixing user experience on the login page. As you sit on the login page, MW closes the connection which led to awkward behaviour.
This is not about changing what happens after you login, so not sure how this relates to frontend session timeout.

1. If I go to signin page and wait without logging in, I should see the new manual reconnect dialog. Right now this is not the case.
2. If I login and wait for session to expire, I should just be redirected to login page like it used to work.

undsoft (approved):
Looks good overall.
One issue is that if you are offline and you press Reconnect, nothing really happens. Would be good to provide at least some indication that it was trying to do something, e.g. a loading bar or at least a disabled button for a bit.

PR #11890, opened 2025-04-15
NAS-135313 / 25.10 / Errors are not reported when exporting a pool

undsoft:
**Testing:**

See ticket.


AlexKarpov98 (approved):
Error is now reported 👍

PR #11896, opened 2025-04-17
NAS-135311 / 25.10 / Remove search results for unlicensed items

denysbutenko:
**Changes:**

Remove search results for unlicensed items

**Testing:**

Edit `system-info.effects.ts#L22`
```
        map((systemInfo) => {
          systemInfo.license.features = [];
          return systemInfoLoaded({ systemInfo });
        }),
        ```

### CODE REVIEW

File: src/app/services/navigation/navigation.service.ts [hunk 1]
```diff
@@ -23,6 +23,8 @@ export class NavigationService {
   readonly hasEnclosure$ = this.store$.select(selectHasEnclosureSupport);
   readonly hasVms$ = new BehaviorSubject(false);
   readonly hasApps$ = new BehaviorSubject(false);
+  readonly hasFibreChannel$ = new BehaviorSubject(false);
+  readonly hasDedup$ = new BehaviorSubject(false);
```
undsoft, line 27: I think it's weird to keep adding new fields here. They are not used in this service and are not really about Navigation.
It's probably time to extract all of this into a separate service.

src/app/services/navigation/navigation.service.ts
```diff
@@ -149,13 +151,16 @@ export class NavigationService {
...
-        const hasApps = systemInfo.license && Boolean(systemInfo.license.features.includes(LicenseFeature.Jails));
-        this.hasApps$.next(hasApps);
-      });
+    this.store$.pipe(
+      waitForSystemInfo,
+      filter((systemInfo) => Boolean(systemInfo.license)),
+      map((systemInfo) => systemInfo.license.features),
+      untilDestroyed(this),
+    ).subscribe((features) => {
+      this.hasVms$.next(features.includes(LicenseFeature.Vm));
```
undsoft, line 160: I think something happened to `hasVms$` check here at some point because Instances menu item is not disabled when it is false.

undsoft (approved):
Search results are fixed.

undsoft:
I've extracted these conditions to a `LicenseService` here:
[LicenseService.patch](https://github.com/user-attachments/files/19834191/LicenseService.patch)

PR #11897, opened 2025-04-17
NAS-135230 / 25.10 / Inline edit controls

undsoft:
**Testing:**

You can go to `/credentials/users-new/demo` for a simple form using new controls.
I will remove demo at later date.


denysbutenko (approved):
image

It works great for me. Great job!

RehanY147 (changes requested):
As discussed, "Not set" seen even when value is present.

RehanY147 (approved):

PR #11898, opened 2025-04-17
NAS-134778 / 25.10 / Adds option to import iso files for volumes dialog under instances

RehanY147:
**Changes:**

Adds option to import iso files for volumes dialog under instances

**Testing:**

You can download a iso for ubuntu with a command like `wget https://releases.ubuntu.com/22.04/ubuntu-22.04.5-desktop-amd64.iso`. Go to the instances page, select a couple pools. Go to create an instance and select vm and then import/upload an iso. Then click the import button and test the dialog. Select the iso and it should show up in the volumes dialog table after a while. It doesn't load immediately because I imagine it takes a minute to be loaded.

### Downstream

|Affects         |Reasoning
|Documentation   |Adds option to import an iso instead of uploading it every time for volumes dialog under instances


AlexKarpov98 (approved):
Works for me.
Please fix tests.

PR #11908, opened 2025-04-21
NAS-135467 / 25.10 / Stop showing stack trace in error dialogs by default

undsoft:
**Changes:**
- Asterisk is no longer show on required ix-selects.
- Additional search terms for Privileges
- Clean up and tests for error dialog. Renamed `more info` to `See stack trace`. `See stack trace` button will only be shown on nightlies.


AlexKarpov98 (approved):
Works as expected 👌 

Why do you think it's better to remove `*`?
Maybe we can add empty values for selects if that's the reason.

undsoft:
The idea behind an asterisk is to communicate to the user that the field is required.
But when select is marked as required, it automatically selects first entry. It also won't have an empty option – it only appears when required is false.
So in this case asterisk is not needed, there is no way for the user not to fill in the field.

PR #11911, opened 2025-04-22
NAS-135216 / 25.10 / Build generic search filter presets component

AlexKarpov98:
Testing: see ticket.

Preview: 

https://github.com/user-attachments/assets/b97cc4c1-bd7a-4ce4-806b-afbb50bd46fa


denysbutenko (approved):
Looks good!

RehanY147 (changes requested):
Found two issues
- If I add a preset by clicking on it and then add another the same way. Then remove the later preset but don't remove the `AND` the preset doesn't return to the preset clickable options below.
- If I've already typed `AND` and then I click on a preset, double `AND AND` is added and the query becomes erroneous.

RehanY147 (changes requested):
Now it's removed too early. As soon as I remove one letter, the filter comes back. Also, if I click it again, it adds a new same filter. We should remove filters as a whole if added via a preset ideally. Since they are added as a whole.

https://github.com/user-attachments/assets/87406ae6-92bc-4508-a996-2d184da7dc05

RehanY147 (changes requested):
I can still see the aforementioned issues. I think the approach here should be that when a filter is added by clicking on the chip preset, the filter shouldn't be added as text. It should be added as an ix-chip. And should be removed as a whole. The chips can be added one after another without adding an `AND`. It is understood that those conditions are applied together. Allowing editing such filters in anyway makes it easily possible to make the whole query erroneous. Also, we shoudn't need to press search after adding a filter. If you click a filter, the latest and all previous filters should be applied automatically and the table should be updated.

If we don't want to get to that level in this ticket, I can approve this PR. But it seems like that's where we want to be eventually.

RehanY147:
I have asked @denysbutenko to review this PR as well. Let's see what he thinks.

AlexKarpov98:
@RehanY147 - chips in search input?  Our input currently works as a text and text only. We need to know if it's possible with current configuration - to apply chips to the search input..
Denys already approved it.

1. What issues do you still see?

2. I think we need to help user to build the search query and when user completes - SEARCH button should be submitted.
See how it works below. 👇 

3. The chips can be added one after another without adding an `AND` -- I don't understand this.
AND is required to make query input a valid string!

4. User is aware if query is erroneous, it's visible visually, user can easily fix it, it's not a problem at all I think. If user is allowed to manually type somewhere - 100% mistakes may be made. I think real user won't play with filters like us, removing part of it and saying that it's erroneous 😄 

5. I think current implementation is enough for this PR, if we want to handle CHIPS in the search input on it's own - let's work on the separate ticket, please, I think it may be not that easy. I already spent a lot of time for this one.

https://github.com/user-attachments/assets/26337ac8-f01f-4921-a8d4-4eb1ed081eab

denysbutenko:
I agree that chip-based inline editing and auto-apply would be a stronger UX, but given the current scope and ticket wording, I think the implementation here is acceptable as a first version. Let's merge this and open a follow-up ticket for the full chip-in-input behavior. That change likely needs deeper technical changes anyway.

RehanY147 (approved):

denysbutenko:
Next step ticket https://ixsystems.atlassian.net/browse/NAS-135591

PR #11918, opened 2025-04-23
NAS-135505 / 25.10 / Strongly typed translations

undsoft:
**Changes**
Over the years, we've collected some inconsistencies in how we work with translated strings.
This partially standartizes the approach with the following rules:
- Components that print strings, such as `ix-label`, should not translate strings themselves.
- Translated strings should be supplied as inputs.

This PR makes `this.translate.instant()` and `| translate` return a special `TranslatedString` string. Components like `ix-label` will expect `TranslatedString` and not normal string. You'll get a compilation error if you try to use normal string.
This also introduces `ignoreTranslation` for cases when you definitely don't want to translate the term.

**Testing**
Code review.


RehanY147 (approved):
Looks good

PR #11923, opened 2025-04-24
NAS-135360 / 25.10 / Use new editable fields in user form

denysbutenko:
**Changes:**

- Use new editable fields in user form

**Testing:**

Check user form on `Users (WIP)`


AlexKarpov98 (commented):
Form looks cool.
I noticed a couple of issues:

1. Form counts invalid (submit is disabled) until I pass through all the `Additional Details` block.

https://github.com/user-attachments/assets/73e986e8-c935-4a92-84fa-1309ae74d74d


2. Selecting `Shell` leads to the infinite loading input.

https://github.com/user-attachments/assets/12d60d94-de06-4146-88a4-33c6227e20d5

3. I can open edit mode but can't exit it if I don't want to change any value, is that OK?

denysbutenko:
> Form counts invalid (submit is disabled) until I pass through all the Additional Details block.

Added asterisk to new labels when field is required.

> Selecting Shell leads to the infinite loading input.

It should not be the case now. 

> I can open edit mode but can't exit it if I don't want to change any value, is that OK? (or if I changed the value I need to open another field for editing to close the previous)

This is a component-specific issue, I will open a ticket for that.

AlexKarpov98 (commented):
1. That issue still exists, new ticket will be created.
Screenshot 2025-05-02 at 16 00 45

2. I tried to fill all form items with `*` but the form is still invalid. Seems like if I update `UID` field, it becomes valid.


https://github.com/user-attachments/assets/bbcf2781-cba4-45ba-9c3f-257744ba89d3

denysbutenko:
> 1. That issue still exists, new ticket will be created.
> 

I created a [ticket](https://ixsystems.atlassian.net/browse/NAS-135734) for that.

> Screenshot 2025-05-02 at 16 00 45
> 2. I tried to fill all form items with `*` but the form is still invalid. Seems like if I update `UID` field, it becomes valid.
> 
>  Screen.Recording.2025-05-02.at.16.01.22.mov

I updated form to make those fields non-required. I talked with Evgeny, we won't use editable for required fields.

AlexKarpov98 (approved):
👌 
With it all discussed, we can merge it now.

PR #11928, opened 2025-04-25
Update pt-br.json - partial translation to Brazilian Portuguese.

msbastos:
**Changes:**

Following the project’s rule to translate from end to beginning, I translated from the file’s last line up to line 5170.
Some lines had already been translated by another contributor, and I did not change them.


undsoft (approved):
Thanks for the contribution 👍🏼

PR #11929, opened 2025-04-26
NAS-135493 / 25.10 / Adds option to provide mac address for instance nic devices in both wizard and instance details

RehanY147:
**Changes:**

Adds option to provide mac address for instance nic devices in both wizard and instance details

**Testing:**

When creating an instance, choose to not have default network settings and select an nic checkbox. You should see a dialog that asks for mac address specifics. Either leave it default or provide a custom mac address. Test the dialog validation as well. And whichever choice you make, you should see it in the label of the checkbox you have selected. If you unselect the checkbox, it should have it's original label minus the mac address config chosen.

On the instance details, when clicking the add nic button, test for the same thing.

### Downstream

|Affects         |Reasoning
|Documentation   |Can add nic device mac address for instances/vms now


denysbutenko (approved):
Looks good!

PR #11930, opened 2025-04-26
Update ja.json

ReiKirishima:
**Changes:**

Translate Dash Board -> TrueNAS Help


undsoft (changes requested):
It looks like you are reverting some of the recent changes here. Please make sure that only `ja.json` is being changed.

PR #11933, opened 2025-04-28
NAS-135554 / 25.10 / Fix disabled `Compression` in replication tasks

undsoft:
**Testing:**

1. Create a replication task that uses SSH connection. You can create an SSH connection to the same system.
2. Check that Compression field can be set to disabled.


RehanY147 (approved):
Looks good

PR #11937, opened 2025-04-29
NAS-135388 / 25.10 / A banner for STIG disappeared if the system rebooted

undsoft:
**Testing:**

Original issue was reproducible by setting login banner in Advanced Settings, logging out and breaking websocket connection.
The expectation is that login banner is still shown after WS connection reestablishes.

For connection to break you can just wait for bit on login page.


RehanY147 (approved):
I see the banner now

undsoft:
backport

PR #11938, opened 2025-04-29
NAS-135491 / 25.10 / Allow specifying image OS for virt VMs in UI

AlexKarpov98:
**Testing:**
See ticket.

Quick preview:

https://github.com/user-attachments/assets/b97d025c-6efe-4d67-8c18-2defa91c46f5

### Downstream

|Affects         |Reasoning
|Documentation   | New Field for ISO Volume


src/app/enums/virtualization.enum.ts
```diff
@@ -1,6 +1,20 @@
...
+  FreeBsd = 'FREEBSD',
+  Windows = 'WINDOWS',
+  ArchLinux = 'ARCHLINUX',
+}
+
+export const imageOsLabels = new Map<ImageOs, string>([
+  [ImageOs.Linux, 'Linux'],
+  [ImageOs.FreeBsd, 'FreeBSD'],
+  [ImageOs.Windows, 'Windows'],
+  [ImageOs.ArchLinux, 'Arch Linux'],
```
undsoft, line 15: Why is `Arch Linux` included as a separate entry? Isn't it just linux?

src/app/interfaces/virtualization.interface.ts
```diff
@@ -86,6 +87,7 @@ export interface CreateVirtualizationInstance {
   zvol_path?: string | null;
   storage_pool: string | null;
   volume?: string | null;
+  image_os?: ImageOs | null;
```
undsoft, line 90: also a string.

src/app/modules/forms/ix-forms/components/ix-combobox/ix-combobox.component.ts
```diff
@@ -98,11 +98,16 @@ export class IxComboboxComponent implements ControlValueAccessor, OnInit {
...
     if (!this.value) {
       this.selectedOption = null;
     }
     if (this.value && this.options?.length) {
-      this.selectedOption = { ...(this.options.find((option: Option) => option.value === this.value)) };
+      let existingOption = this.options.find((option: Option) => option.value === this.value);
+      if (!existingOption && this.allowCustomValue()) {
+        existingOption = { label: this.value as string, value: this.value };
+      }
+      this.selectedOption = { ...existingOption };
```
undsoft, line 110: Would be good to add a test case for changes in core components.

src/app/enums/virtualization.enum.ts
```diff
@@ -1,6 +1,20 @@
 import { marker as T } from '@biesbjerg/ngx-translate-extract-marker';
 import { iconMarker } from 'app/modules/ix-icon/icon-marker.util';
 
+export enum ImageOs {
+  Linux = 'LINUX',
```
undsoft, line 5: Values don't have to be uppercase, since this is not a real enum, just a preset for the user.

src/app/pages/instances/components/all-instances/instance-details/instance-general-info/instance-edit-form/instance-edit-form.component.ts
```diff
@@ -177,6 +186,7 @@ export class InstanceEditFormComponent {
 
     if (this.isVm) {
       payload.secure_boot = values.secure_boot;
+      payload.image_os = values.image_os;
```
undsoft, line 189: Please check with Rehan from MW if we need to send `Linux` when user selects `
Use a Linux image (linuxcontainers.org)`.

src/app/helptext/instances/instances.ts
```diff
@@ -88,4 +88,10 @@ Ideal for legacy applications, full-featured desktops, or software with strict O
     moveTooltip: T('Renames the ZFS dataset to a path in the `ix-virt` dataset in which the zvol is located.'),
     description: T('Importing a zvol as Instances volume allows its lifecycle to be managed, including backups, restores, and snapshots. This allows portability between systems using standard tools.'),
   },
+
+  osImage: {
+    tooltip: T('Optionally specify the operating system for VM-based instances.\
+ Common options are Windows, Linux, FreeBSD, or Arch Linux, but you can also enter a custom value.\
```
undsoft, line 94: Same here. Not sure why Arch Linux is special.

Also include this text:
> When creating a Windows VM, make sure to set the this field to `Windows`.
Doing so will tell us to expect Windows to be running inside of the virtual machine and to tweak behavior accordingly.

src/app/pages/instances/components/instance-wizard/instance-wizard.component.ts
```diff
@@ -776,4 +790,27 @@ export class InstanceWizardComponent implements OnInit {
       }
     });
   }
+
+  private detectImageOs(value: string | null | undefined): ImageOs | null {
```
undsoft, line 794: Would be nice to add a test case for this.

src/app/pages/instances/components/instance-wizard/instance-wizard.component.html
```diff
@@ -72,6 +72,16 @@
               (click)="onSelectRootVolume()"
             >{{ 'Select Volume' | translate }}</button>
           </div>
+
+          @if (form.value.volume_type === VolumeContentType.Iso) {
+            <ix-combobox
+              formControlName="image_os"
+              [label]="'OS Image' | translate"
```
undsoft, line 79: Let's call it `OS Type` here and in edit form.

undsoft (approved):
Looks good 👍🏼

src/app/pages/instances/components/all-instances/instance-details/instance-general-info/instance-edit-form/instance-edit-form.component.ts
AlexKarpov98: @undsoft - I tried it, and it says that it should be sent only with `ISO` volume.

src/app/enums/virtualization.enum.ts
AlexKarpov98: @undsoft - I checked the docs, and there I found these enums:
 
Screenshot 2025-04-30 at 12 36 42

src/app/modules/forms/ix-forms/components/ix-combobox/ix-combobox.component.ts
AlexKarpov98: sounds reasonable.

PR #11948, opened 2025-05-01
NAS-135611 / 25.10 / Switch to 3 dot menu for Shares and Data Protection

AlexKarpov98:
**Testing:**
See ticket.
image

For now I introduced a custom component: 
`ix-cell-actions-with-menu.component`

I'll use it for remaining tables once we merge it.

### Downstream

|Affects         |Reasoning
|Documentation   | Table action buttons updated


undsoft (approved):
1. Edit can also go in the menu.
2. I think it would make sense to be consistent on these pages and also have a three dot menu for when there are only edit and delete buttons.

PR #11956, opened 2025-05-04
NAS-135651 / 25.10 / Fix reconnect when system restart

denysbutenko:
**Changes:**

Do not show reconnect dialog when system restart

**Testing:**

Try to restart both systems


undsoft (changes requested):
Created a non-HA system, restarted it, it never reconnected.

undsoft:
BTW there is no progress bar when reconnecting.

undsoft (approved):
Okay, now it works as expected.

PR #11958, opened 2025-05-05
NAS-135508 / 25.10 / Changes to `ix-explorer` to allow disabling nodes

RehanY147:
**Changes:**

Changes `ix-explorer` to expect `Observable<ExplorerNodeData[]>` type object similar to `Observable<Optionp[]>` for `ix-select`. Adds option to provide a callback in the node-provider options in `file-system.service.ts`  that returns `Observable<boolean>` to determine if a node should be disabled when going through the `ix-explorer` tree. Adds ability for `ix-explorer` to disable a node in the tree based on the aforementioned disabled call back. Disabled node can't be selected as a value and clicking it should do nothing. Also, it should be visually different from other "enabled" nodes to indicate that it is disabled.

**Testing:**

On the SMB form, test that you no longer see `/mnt` as the root node of the path explorer. You should see pools as the root nodes of the explorer. Also, if you are creating a new share, you should not be able to select the root nodes (pool names) as a value for the "path". You should be able to select any child datasets as possible value.
Also, code review.

Will fix tests once code review is done.


src/app/modules/forms/ix-forms/components/ix-explorer/ix-explorer.component.ts
```diff
@@ -63,14 +59,15 @@ import { ErrorParserService } from 'app/services/errors/error-parser.service';
...
-export class IxExplorerComponent implements OnInit, OnChanges, ControlValueAccessor {
+export class IxExplorerComponent implements ControlValueAccessor {
   readonly label = input<TranslatedString>();
   readonly hint = input<TranslatedString>();
   readonly readonly = input<boolean>(false);
   readonly multiple = input(false);
   readonly tooltip = input<TranslatedString>();
   readonly required = input<boolean>(false);
-  readonly root = input(mntPath);
+  readonly rootNodes = input<Observable<ExplorerNodeData[]>>(/* of([rootDatasetNode]) */);
```
undsoft, line 69: I can't think of an advantage of this being an observable. Let's just use array of nodes here.

src/app/services/filesystem.service.ts
```diff
@@ -20,28 +20,28 @@ export interface ProviderOptions {
...
-  type: ExplorerNodeType.Directory,
-};
-
 @Injectable({ providedIn: 'root' })
 export class FilesystemService {
   constructor(
     private api: ApiService,
   ) {}
 
+  getTopLevelDatasetsNodes(providerOptions: ProviderOptions): Observable<ExplorerNodeData[]> {
```
undsoft, line 32: If this method is only used to retrieve top level datasets, then most of the options here are not applicable.

undsoft (changes requested):
1. See if we can use the icon we use on Datasets page for root datasets.
2. List is rendered correctly, but the fact that top items are grayed-out may prevent me from clicking on them, even to expand. I also need to be able to click on the top level dataset in order for Create Dataset button to become available.
Let's prevent selection here in some other way, for example by changing cursor and adding a tooltip.
I'm open to other ideas.

src/app/services/filesystem.service.ts
undsoft: What is this method supposed to return?

undsoft: How could it be something different? The top level nodes for the filesystem would always be root datasets, no? Never zvols, never snapshots, never directories or hidden files, so I don't see now options can change anything.

undsoft (changes requested):
1. I should be able to create child datasets under the root dataset. Right now this is not the case.
2. Tooltip just says "This path is not selectable for Path", which is not very helpful. Ideally, it should say something like "Selecting root datasets for shares is not recommended".
3. Can't say that I like how the button changes text.

Maybe it gets a bit too complicated with trying to disable the root dataset, but keeping it selectable. Instead maybe let's implement a validator that will simply produce an error message saying: "Sharing root dataset is not recommended. Please create a child dataset".

PR #11960, opened 2025-05-05
NAS-135635 / 25.10 / It would be good If we add a Cancel button on the password change required screen

AlexKarpov98:
Testing: see ticket.

Preview: 

https://github.com/user-attachments/assets/44f6c6fe-a37a-43bb-97ba-efd0002c602c


undsoft (changes requested):
It shouldn't be a skip button, it should be a logout button. The screen requires user to change the password.

AlexKarpov98:
@undsoft 

https://github.com/user-attachments/assets/f2be6121-5f45-4416-b666-b962f7b486bb

undsoft (approved):
This works, thanks.

PR #11961, opened 2025-05-05
NAS-135617 / 25.10 / Get Call Error when attempt to add a local group member to a AD group

AlexKarpov98:
Testing: see ticket.


denysbutenko (approved):
Looks good!

PR #11965, opened 2025-05-06
NAS-135232 / 25.10 / Consume global `system.boot_id`

AlexKarpov98:
**Changes:**
Boot ID is now taken from the global API 

**Testing:**
See ticket.

for `environment.ts`
make sure you have:

`production: false` 

To be able to avoid CORS errors, make sure you have `proxy.config.json` set up.
you can check the `proxy.config.json.template`

```
{
    "/api/**": {
        "target": "http://_REMOTE_",
        "secure": false,
        "changeOrigin": true,
        "pathRewrite": {"^/api" : "http://_REMOTE_/api"},
        "loglevel": "debug"
    },
    "/_upload": {
        "target": "http://_REMOTE_",
        "secure": false,
        "changeOrigin": true,
        "pathRewrite": {"^/_upload" : "http://_REMOTE_/_upload"},
        "loglevel": "debug"
    },
    "/_download": {
        "target": "http://_REMOTE_",
        "secure": false,
        "changeOrigin": true,
        "pathRewrite": {"^/_download" : "http://_REMOTE_/_download"},
        "loglevel": "debug"
    }
}
```

Replace `_REMOTE_` with the actual remote IP, like `10.220.36.94`

OR run `yarn run ui remote -i '10.220.36.94'`

Request Result:

Screenshot 2025-05-06 at 10 02 58

@undsoft - FYI - this makes an extra step to be able to login to the app locally, not really convenient.


undsoft:
@AlexKarpov98 If by extra step you mean updating proxy config, then it should automatically generated when `yarn ui remote` is run.

src/app/services/global-api-http.service.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+  const mockWindow = {
+    location: {
+      protocol: 'https:',
+    },
+  } as unknown as Window;
+
+  const createService = createServiceFactory({
+    service: GlobalApiHttpService,
+    imports: [HttpClientTestingModule],
+    providers: [
```
undsoft, line 22: You can also use `mockWindow()` fn.

src/app/services/global-api-http.service.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+  beforeEach(() => {
+    spectator = createService();
+    httpMock = spectator.inject(HttpTestingController);
+  });
+
+  it('should call getBootId() and return string', () => {
+    const expectedBootId = 'boot-abc';
+
+    spectator.service.getBootId().subscribe((result) => {
+      expect(result).toBe(expectedBootId);
```
undsoft, line 36: You can put anything in `toBe` and test will pass. You need to wait for returning value, probably with something like `await firstValueFrom(...)`

src/app/services/global-api-http.service.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  constructor(
+    private http: HttpClient,
+    @Inject(WINDOW) private window: Window,
+  ) {
+    const protocol = this.window.location.protocol === 'https:' ? 'https://' : 'http://';
+    this.baseUrl = environment.production ? `${protocol}${environment.remote}/api` : '/api';
+  }
+
+  get<T>(path: string): Observable<T> {
```
undsoft, line 21: private

undsoft (approved):
Works as expected 👍🏼

PR #11970, opened 2025-05-07
NAS-135277 / 25.10 / remove sanitize-html

RehanY147:
**Changes:**

Use DomPurify to sanitize HTML instead of sanitize-html. 

**Testing:**

You should no longer see `'Module has been externalized'` warnings in the browser console when running the code locally. Also test for regressions in the area.


undsoft (approved):
The error is gone.

AlexKarpov98 (approved):

PR #11971, opened 2025-05-07
NAS-135726 / 25.10 / Share `hasFibreChannel$` observable with subscribers

RehanY147:
**Changes:**

Share the observable so it's not re-evaluated for all the subscribers resulting in dozens of calls to `fc.capable` endpoint.

**Testing:**

See in the network tab that when you type one letter in the global search input, `fc.capable` is only called once.


undsoft (approved):
Thanks, looks good. 👍🏼

PR #11975, opened 2025-05-07
Updated Irish translations

aindriu80:
**Changes:**

Translated new empty strings and edited some existing translations (for a better translation).

**Testing:**

Tested plural translations with Poedit and also messageformat-parser, checks include:-

✔ Correct use of {n, plural, one {...} other {...}}
✔ No missing or extra {} braces
✔ No syntax errors like =1 where one should be used
✔ No missing plural forms (e.g., only one but missing other)


undsoft (approved):

PR #11978, opened 2025-05-08
NAS-135729: Rsync tasks are gone after updating to latest nightly and Rsync tasks are gone after updating to latest nightly and replaced with dummy inserts

AlexKarpov98:
Fixes/removes test data.


undsoft (approved):
👍🏼

PR #11979, opened 2025-05-08
NAS-135771 / 25.10 / Require checklist items to be resolved on PRs

undsoft:
**Testing:**
One of the jobs should fails because this checklist item is not done:

- [x] Not done.

The intention is to allow developers to add checklist items to their PR so that they don't forget to do something before merging the PR.

### DIFF
--- a/.github/workflows/require-checklist.yml
+++ b/.github/workflows/require-checklist.yml
@@ -1,0 +1,13 @@
+name: Require Checklist Items To Be Resolved
+
+on:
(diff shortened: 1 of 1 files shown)


AlexKarpov98 (approved):
Screenshot 2025-05-09 at 12 41 17

Failed indeed.

And succeeded indeed once [x] marked:
Screenshot 2025-05-09 at 12 41 57

PR #11982, opened 2025-05-09
NAS-135781 / 25.10 / "Reset Configuration" changed to "refresh" is confusing

AlexKarpov98:
Quick Text Fix.


src/app/pages/system/network/components/interfaces-card/interfaces-card.component.ts
```diff
@@ -108,7 +108,7 @@ export class InterfacesCardComponent implements OnInit, OnChanges {
         {
           iconName: iconMarker('refresh'),
           requiredRoles: this.requiredRoles,
-          tooltip: this.translate.instant('Refresh'),
+          tooltip: this.translate.instant('Reset configuration'),
```
undsoft, line 111: Can't you just remove the whole line? There is a dynamic tooltip on this item.

undsoft (approved):
I suppose we can show the message right in there, although it's not a pattern we normally use.
But in this case, the line-height needs to be adjusted to look good.
Снимок экрана 2025-05-13 в 08 52 38
Снимок экрана 2025-05-13 в 08 53 57

PR #11986, opened 2025-05-12
NAS-135654 / 25.10 / Webui refers to 2 controllers on FC cards on Non-HA systems

AlexKarpov98:
Testing: see ticket.
New directive introduced. `isHa`


src/app/directives/is-ha/is-ha.directive.ts
```diff
@@ -0,0 +1,37 @@
...
+import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
+import { Store } from '@ngrx/store';
+import { AppState } from 'app/store';
+import { selectIsHaLicensed } from 'app/store/ha-info/ha-info.selectors';
+
+@UntilDestroy()
+@Directive({
+  selector: '[isHa]',
+})
+export class IsHaDirective implements OnInit {
```
undsoft, line 17: Tests please.

undsoft (approved):
Card looks correct now.

src/app/directives/is-ha/is-ha.directive.ts
AlexKarpov98: ah, sorry, missed that one, added tests for the components only 🙉

PR #11990, opened 2025-05-12
NAS-135829 / 25.10 / Add additional options to NVMe-oF forms

undsoft:
**Changes:**

- Hides or disables options that are not available in some forms.
- Allows Host key to be generated.
- Fixes error handling.


RehanY147 (changes requested):
Generate key should have a loading state. Also, the button should be disabled if clicked once. Other you can keep clicking it while the previous request is still loading. Nothing happens when you click the button until the key shows up.

undsoft:
Fair enough, fixed.

PR #11991, opened 2025-05-12
NAS-135764 / 25.10 / On Goldeye Data Protection read-only admin users can click on Run job, Dry Run, Delete and Restore

AlexKarpov98:
Code review shall be enough.

Missing `*ixRequiresRoles` added

Result:
Screenshot 2025-05-12 at 16 16 38


undsoft (approved):
Looks good, but there was a comment in Slack where people actually liked that ACL options were available here for readonly admin (so that admins can see current options).
Just make sure things are properly disabled in those forms.

Снимок экрана 2025-05-13 в 08 40 55

PR #11994, opened 2025-05-13
NAS-135720 / 25.10 / Allow migrating apps from one pool to another in the UI

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/34712a66-b4df-4b96-b44e-729a13c98917


src/app/pages/apps/store/docker.store.ts
```diff
@@ -79,9 +79,17 @@ export class DockerStore extends ComponentStore<DockerConfigState> {
     return this.api.call('docker.status');
   }
 
-  setDockerPool(poolName: string | null): Observable<Job<DockerConfig>> {
+  setDockerPool(poolName: string | null, migrateApps?: boolean): Observable<Job<DockerConfig>> {
+    const payload = {
+      pool: poolName,
+    } as DockerConfig;
+
+    if (typeof migrateApps === 'boolean') {
```
RehanY147, line 87: Insignificant but you can just check for a truthy `migrateApps` instead of a type check. The only case where `payload.migrate_applications` should be true is when `migrateApps` is true. If it's `false` or `undefined`, `payload.migrate_applications` should be false anyway.

RehanY147 (approved):

PR #11995, opened 2025-05-13
NAS-135734: Better keypress handler for editable fields

AlexKarpov98:
Testing: see ticket.

Result:

https://github.com/user-attachments/assets/e6601321-e0f0-4d02-9b9c-da5602350a7a

### DIFF
--- a/src/app/modules/forms/editable/services/editable.service.spec.ts
+++ b/src/app/modules/forms/editable/services/editable.service.spec.ts
@@ -134,16 +134,18 @@ describe('EditableService', () => {
       spectator.service.register(component2);
     });
 
-    it('listens to keydown events and tries to close all editables when Escape is pressed', () => {
+    it('listens to keydown events and tries to close all editables when Escape is pressed', fakeAsync(() => {
       const event = new KeyboardEvent('keydown', { key: 'Escape' });
       const listener = registeredListeners.get('keydown');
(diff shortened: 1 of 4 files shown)


denysbutenko (changes requested):
https://github.com/user-attachments/assets/2c05a482-f6c6-4ab6-9c2b-3467de66ceb3

Sometimes it does not work as expected. Search input gets focused on escape when slide-in in opened.

denysbutenko (approved):
It works better.

Still have edge cases:

1. When editable has only one disabled field - Escape does not work.
2. When editable has multiple fields - Escape does no work.

AlexKarpov98:
@denysbutenko @undsoft - updates arrived, please check

denysbutenko (approved):
It works well!

undsoft (approved):
Looks good, but please also check how MW validation errors are shown (may not be related to your changes).
I went to sharing -> nvme-of and created two subsystems. Then I used inline editing to change the name of the second subsystem to the name of the first subsystem. The error appears in the dialog, when it should appear in the editable.

AlexKarpov98:
@undsoft I tried it, it works fine for me. I'll merge it.

https://github.com/user-attachments/assets/3485baac-c5b5-45e8-8f39-500ac7ee801e

PR #11999, opened 2025-05-14
NAS-135864: Refactor error dialog when exporting a pool

undsoft:
**Changes:**

Don't do templates in Typescript files, kids.

**Testing:**

Hard to reproduce MW conditions, so you'll have to mock by putting something like:
```
this.showServicesErrorsDialog({
      stop_services: ['service1'],
      restart_services: ['service2'],
    } as ServicesToBeRestartedInfo);
    return;
```
on top of `startExportDisconnectJob` in `ExportDisconnectModalComponent`.


AlexKarpov98 (approved):
Much better now! 👍

PR #12002, opened 2025-05-14
NAS-135508 / 25.10 / Adds option to provide multiple roots for `ix-explorer`

RehanY147:
**Changes:**

Adds validator for root datasets for shares. Also, adds option to provider multiple roots for `ix-explorer`

**Testing:**

See the SMB form. You shouldn't be able to submit the form while having selected a root dataset. Also, you shouldn't see `/mnt` as the root for the `ix-explorer` field. Also check for regressions in other places where changes were made.


src/app/pages/sharing/utils/root-datasets-validator.ts
```diff
@@ -0,0 +1,13 @@
+import { AbstractControl } from '@angular/forms';
+
+export const forbiddenRootDatasetsSharingValidatorFn = (control: AbstractControl<string | string[]>): boolean => {
```
undsoft, line 3: Let's see some tests here. Also name is intense.

PR #12007, opened 2025-05-15
NAS-135875 / 25.10 / Fix Automated Disk Selection Types

essinghigh:
Fixes [NAS-135875](https://ixsystems.atlassian.net/browse/NAS-135875)

This pull request fixes an issue in the Pool Creation Wizard where automated disk selection always prioritizes equally-sized HDD over SSD, even when SSD is explicitly selected. See the `findSuitableDisks` method in `src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/grouped-disks.class.ts`.

The `findSuitableDisks` method currently decides disk selection based solely on disk size. HDDs are always returned first due to the following logic:
```typescript
return [...this.diskMap.HDD[categoryDiskSize] || [], ...this.diskMap.SSD[categoryDiskSize] || []];
```

This behavior causes the selection of HDDs over SSDs when disks of the same size are available, even if SSD is specified in the Pool Creation Wizard.

The PR modifies `findsuitableDisks` to only return the relevant `diskMap` of the specified disk type, rather than returning a combined array from both types. Previous functionality left intact in the event that `diskType` is unknown / unset for whatever reason.

Spec has also been adjusted to account for this, as previous tests were not sufficient as disks does not have a `DiskType.Ssd` device the same size as any of the `DiskType.Hdd` devices.

Previous behaviour:
image

New behaviour:
image

`yarn:changed` tests are passing:
```
 PASS  src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/enclosure-list.class.spec.ts (32.734 s)
 PASS  src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/grouped-disks.class.spec.ts (32.778 s)
 PASS  src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/generate-vdevs-normal.spec.ts (32.948 s)
 PASS  src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/generate-vdevs-treat-as-minimum.spec.ts (32.952 s)
 PASS  src/app/pages/storage/modules/pool-manager/utils/generate-vdevs/generate-vdevs-maximize-dispersal.spec.ts (32.957 s)
```


essinghigh:
Apologies, only just noticed this was not a draft. I'd prefer feedback on the ticket on whether this is the preferred direction or not before having it ready for review. Entirely possible you may have a completely different implementation in mind.

RehanY147 (commented):
There's no intentional preference between SSDs and HDDs right now. Although, it is there in implementation, it was unintentional. Automated Disk Selection is there to streamline the process in cases where there's dozens of disks of one type and size. If UI prioritizes SSD's, someone can make the same case for HDD's. That if the user doesn't have control over what resource is being spent, UI should pick the less valuable resource.
For specific cases like this, manual selection should suffice.
However, there is a case for UI to provide options to the user to select what type they want to prioritize of if they want to mix disk types. However, that will result in adding more clicks/actions required for a process that we want to keep as simple as possible.
Waiting for internal discussion.

essinghigh:
>  If UI prioritizes SSD's, someone can make the same case for HDD's. That if the user doesn't have control over what resource is being spent, UI should pick the less valuable resource.

Just as a clarification on this point. Given two identically sized disks (20GiB), the user will be given the option of `20GiB (HDD)` or `20GiB (SSD)` in the disk selection dropdown. Current implementation will always use HDD first, even if the user has explicitly selected `20GiB (SSD)`.

It makes sense to only return the type of disk that the user has selected.

In my mind there's no reason the automatic selection should be mixing types (creating a mixed-type vdev is an edge-case that should most likely be handled in manual selection).

Agree with your point on providing more options, considering the ability to select between types in the dropdown already works as expected shouldn't be needed to add additional selections.

william-gr:
@essinghigh sorry, what is the state of this PR? Do you still want to work on it? If so we will need this rebased on latest master, please.

ghost:
@william-gr I haven't touched it while "waiting for internal discussion" on the direction iX wants to go with this per @RehanY147  - not sure what that internal discussion entails.

Assuming there are no issues with just fixing this bug (i.e. no overhauls of the disk selection UI planned which will make this fix redundant), I can review latest changes and reapply patches as/if required.

william-gr:
@essinghigh fair enough. Rehan is no longer with us but I will review it and get back to you next week. Thanks for understanding and willingness to work with us

william-gr:
@essinghigh your changes make sense to me. We will try to finalize review this week.

PR #12009, opened 2025-05-15
NAS-135833 / 25.10 / Share Admin not able to start/stop services through UI

undsoft:
See ticket.


AlexKarpov98 (approved):
works as expected.

PR #12011, opened 2025-05-16
NAS-135651 / 25.10 / Disable reconnect dialog on failover

denysbutenko:
**Changes:**

- Do not show reconnect dialog when system restart was initiated using the Failover button

**Testing:**

- Try to restart system using Initiate Failover button

### DIFF
--- a/src/app/services/websocket-status.service.ts
+++ b/src/app/services/websocket-status.service.ts
@@ -19,6 +19,7 @@ export class WebSocketStatusService {
   }
 
   readonly isReconnectAllowed$ = new BehaviorSubject<boolean>(false);
+  readonly isFailoverRestart$ = new BehaviorSubject<boolean>(false);
   private readonly isLoggedIn$ = new BehaviorSubject<boolean>(false);
   private readonly authStatus$ = new BehaviorSubject<boolean>(false);
   readonly isAuthenticated$ = this.authStatus$.asObservable();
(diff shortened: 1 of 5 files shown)


RehanY147 (changes requested):
Now, after I initiate failover, the UI never connects again until I refresh the browser tab. If we are not giving a button to reconnect, it must happen automatically.

denysbutenko:
> Now, after I initiate failover, the UI never connects again until I refresh the browser tab. If we are not giving a button to reconnect, it must happen automatically.

Alright, I fixed the sign-in store to re-initialize on failover. Now, you may see the sign-in form after the failover dialog disappears.

src/app/services/websocket-status.service.ts
```diff
@@ -19,6 +19,7 @@ export class WebSocketStatusService {
   }
 
   readonly isReconnectAllowed$ = new BehaviorSubject<boolean>(false);
+  readonly isFailoverRestart$ = new BehaviorSubject<boolean>(false);
```
RehanY147, line 22: Should be private

RehanY147 (approved):
Works now. Once failover is initiated, UI ends up on the sign in page and sign in works.

src/app/services/websocket-status.service.ts
denysbutenko: `isFailoverRestart$` is used in the signin component.

PR #12018, opened 2025-05-19
NAS-135613: Simplify Network page – tables

AlexKarpov98:
**Testing:**
See ticket.

Result:

### Downstream

|Affects         |Reasoning
|Documentation   | Text adjustments


undsoft (changes requested):
Okay, I thought new layout would look better. It doesn't look too great. Let's return to old layout for cards that we had.
Also there is a small space before "Service Announcement" in Network Configuration.

PR #12024, opened 2025-05-21
NAS-135937 / 25.04.1 / Add support for denying some network activities

undsoft:
**Changes:**

The problem was that middleware was setting `activity` in network configuration to:
```
{
  "type": "DENY",
  "activities": [
    "usage",
    "support",
    "update"
  ]
}
```

This was not supported in the UI.

**Testing:**

You don't have to be in STIG mode for this, you can just go to Network and play with Outbound network activity.

### Downstream

|Affects         |Reasoning
|Documentation   | There is now a new "Allow all except" option for outbound network activity.


AlexKarpov98 (approved):
Works as expected 👍

PR #12025, opened 2025-05-21
NAS-135953 / 25.10 / Input is not a valid integer in SSH form

undsoft:
**Testing:**

See ticket.


denysbutenko (approved):
It works great for me.

PR #12035, opened 2025-05-22
NAS-135955 / 25.10 / Host management

undsoft:
**Testing:**
See host card in details `/sharing/nvme-of`.

**Known issues**
- Slide in won't be shown correctly when triggered from the dialog. Because of it editing may not work correctly.
- Deleting a host will only work if it's not used in a subsystem.


AlexKarpov98 (approved):
Works as described.
Screenshot 2025-05-23 at 10 51 29

Known issues seen 👌

PR #12037, opened 2025-05-23
NAS-135328: Adapts to the new job api in middleware

RehanY147:
**Changes:**

Adapts to the new api for jobs on middleware. Review ticket and the api docs for jobs to understand what changes are required.

**Testing:**

Test for regressions in cases where jobs are running. Test running jobs in the background. Test starting a job then refreshing the page and seeing the job again in the tasks drop down top right. Also, code review.

Will add tests once code review is done.

### DIFF
--- a/src/app/interfaces/api-message.interface.ts
+++ b/src/app/interfaces/api-message.interface.ts
@@ -56,7 +56,7 @@ export type ApiMethod = ApiCallMethod | ApiJobMethod | ApiEventMethod;
 
 export interface ApiEvent<T = unknown> {
   collection: ApiCallMethod | ApiJobMethod | ApiEventMethod;
-  fields: T;
+  fields: T & { message_ids: string[] };
   id: number | string;
   msg: CollectionChangeType.Changed
     | CollectionChangeType.Added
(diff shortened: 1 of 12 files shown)


src/app/interfaces/api-message.interface.ts
```diff
@@ -56,7 +56,7 @@ export type ApiMethod = ApiCallMethod | ApiJobMethod | ApiEventMethod;
 
 export interface ApiEvent<T = unknown> {
   collection: ApiCallMethod | ApiJobMethod | ApiEventMethod;
-  fields: T;
+  fields: T & { message_ids: string[] };
```
undsoft, line 59: This does not seem to be needed.

src/app/modules/websocket/api.service.ts
```diff
@@ -122,10 +134,22 @@ export class ApiService {
...
+    method: M,
+    params?: ApiJobParams<M>,
+    callUuid?: string
+  ): Observable<number>;
+  private callMethod<M extends ApiCallMethod | ApiJobMethod>(
+    method: M,
+    params?: unknown[],
+    callUuid?: string,
+  ): Observable<unknown> {
+    const uuid = callUuid || UUID.UUID();
```
undsoft, line 152: I don't see `callUuid` being passed in from anywhere.

undsoft (approved):
Appears to be working correctly, thanks 👍🏼

PR #12045, opened 2025-05-26
NAS-135972 / 25.10 / Removes setup password dialog if stig enabled

RehanY147:
**Changes:**

Disables dialog to setup new password when stig enabled on the system

**Testing:**

See ticket details


undsoft (changes requested):
The problem here is not stig, it's setting passwords on AD users (check for `local` attribute in a user).

PR #12055, opened 2025-05-27
NAS-136000 / 25.10 / Remove validators from IP addresses in GUI form

undsoft:
**Changes:**

Removes validators from those fields, because they are coming from choices anyway and current validators create a problem.

**Testing:**

See that you can select and save multiple items in `Web Interface IPv4 Address` in `GUI Settings`.
Middleware error when 0.0.0.0 is something else.


AlexKarpov98 (approved):
👍

PR #12059, opened 2025-05-28
NAS-135958 / 25.10 / Refactors slide-ins to use cdk-overlay instead of manual ix-controller component

RehanY147:
**Changes:**

Introduces new function of opening slide-ins using cdk-overlay with a new service. Can reuse old modal-header and has the same interface as the previous service so switching should be simple enough.

**Testing:**

Code review. Go to Data Protection -> Replication Task Card -> Add.

Clicking the Advanced Options button actually opens the form slide-in on top of the wizard slide in. But clicking the wizard button on the form swaps the wizard to the form. This is only done to test both swap and open functionality.
Test swapping components and layering them.
Test for regressions.

TODOs before merging
- [x] Remove old slide-in and replace usage with new slide-in
- [x] Fix swap function of replication task wizard to form and form to wizard
- [x] Rename stuff for more readibility


src/app/modules/overlay-slide-ins/overlay-slide-in.service.ts
```diff
@@ -0,0 +1,130 @@
...
+export class OverlaySlideInService {
+  private overlays = signal<OverlayInstance<unknown>[]>([]);
+  openOverlays = computed(() => this.overlays()?.length);
+
+  constructor(
+    private overlay: Overlay,
+    private injector: Injector,
+  ) {}
+
+  open<C, T = unknown>(
```
undsoft, line 41: Previous implementation ensured that dataand return value was strongly typed. See ComponentInSlideIn

src/app/modules/overlay-slide-ins/components/overlay-container/overlay-container.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  selector: 'ix-overlay-container',
+  templateUrl: './overlay-container.component.html',
+  styleUrl: './overlay-container.component.scss',
+  standalone: true,
+  imports: [
+    CdkPortalOutlet,
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class OverlayContainerComponent {
```
undsoft, line 16: The name is too generic. If's a slide-in container, let's just call it that.

src/app/modules/overlay-slide-ins/components/overlay-container/overlay-container.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  @ViewChild(CdkPortalOutlet, { static: true }) portalOutlet!: CdkPortalOutlet;
+  constructor(private host: ElementRef<HTMLElement>) {}
+
+  startCloseAnimation(): Promise<void> {
```
undsoft, line 20: Can we not use Angualr's :enter and :leave animation instead of doing things manually?

src/app/modules/overlay-slide-ins/overlay-slide-in.service.ts
```diff
@@ -0,0 +1,130 @@
...
+      (overlayItem) => overlayItem.overlayId === overlayId,
+    );
+    let close$ = previousOverlay?.close$ as Subject<T | undefined>;
+    if (!close$) {
+      close$ = new Subject<T | undefined>();
+    }
+    const slideInRef = new SlideInOverlayRef<T>(overlayRef, containerRef, close$);
+
+    const injector = this.createInjector(config.data, slideInRef);
+    const contentPortal = new ComponentPortal(component, null, injector);
```
undsoft, line 66: Maybe move some of this logic inside of the OverlayContainerComponent.

src/app/pages/data-protection/replication/replication-form/replication-form.component.ts
```diff
@@ -107,18 +107,19 @@ export class ReplicationFormComponent implements OnInit {
     private replicationService: ReplicationService,
     private keychainCredentials: KeychainCredentialService,
     private authService: AuthService,
-    public slideInRef: SlideInRef<ReplicationTask | undefined, ReplicationTask | false>,
+    public slideInRef: SlideInOverlayRef,
+    @Inject(SLIDE_IN_DATA) protected existingReplication: ReplicationTask,
```
undsoft, line 111: I liked old approach better because it allowed us to have type safety for incoming and returning values.

undsoft (changes requested):
1. Margins look different in new slide ins and it seems like its width also changes based on the content.
2. Let's avoid introducing a new service. Let's call it SlideIn and make its signature as compatible with old implementation as possible.
3. We need to make sure that "Are you sure you want to close this form?" functionality is still there.
4. Let's make backdrop darker, like in old implementation.
5. Old implementation also closes on Escape.

src/app/modules/slide-ins/slide-in-ref.ts
```diff
@@ -11,7 +12,7 @@ export class SlideInRef<D, R> {
    * have the same purpose and return the same response type e.g, form to wizard and
    * wizard to form.
    */
-  swap?: (component: Type<unknown>, options?: { wide?: boolean; data?: unknown }) => void;
+  swap?: (component: ComponentInSlideIn<D, R>, options?: { wide?: boolean; data?: unknown }) => void;
```
undsoft: This is probably incorrect, because D and R already refer to first component show in slide-in. These values may be different for second component.
You can try:
```
  swap?: <D2, R2>(component: ComponentInSlideIn<D2, R2>, options?: { wide?: boolean; data?: D2 }) => void;
```

undsoft (approved):
Looks good, nice work.

Let's make it more similar to how old slide-ins looked and felt. Backdrop needs to be darker and both vertical and horizontal margins need to be made similar. Please compare old and new implementation side by side.

PR #12061, opened 2025-05-28
NAS-135871 / 25.10 / Allow sorting apps in the UI by popularity rank

AlexKarpov98:
**Testing:**
See ticket.

https://github.com/user-attachments/assets/9599d2ff-9565-4545-8f10-80cf0ab2dc9d

|Affects         |Reasoning
|Documentation   | Added popularity ranking on the available app card


undsoft (changes requested):
I get `'<' not supported between instances of 'NoneType' and 'int'` when I select sorting by popularity rank. (try after selecting more trains).

Also would be good to add "Higher value is more popular" to the tooltip or rename everything to just "Popularity". Right now the word `rank` makes me think that 1 means 1st place.

undsoft (approved):
Cool, looks good.

PR #12066, opened 2025-05-28
NAS-135826: Change bytes to bits before consumption in apps network traffic details

RehanY147:
**Changes:**

Change bytes to bits before consumption in apps network traffic details

**Testing:**

Check that bytes are converted to bits before being shown on widgets and apps table.


undsoft (approved):
Thanks, looks good.

PR #12069, opened 2025-05-29
NAS-135972 / 25.10 / Don't show option to set password for non-local users

RehanY147:
**Changes:**

Don't show option to set password for non-local users right after login.

**Testing:**

See ticket


undsoft:
@RehanY147 This needs to be backported to .2

RehanY147:
backport

PR #12081, opened 2025-06-03
NAS-136005 / 25.10 / Create instance: switching from "Upload ISO" to "Use a Linux" image won't allow to create an instance

AlexKarpov98:
Testing: see ticket.


undsoft (changes requested):
Doesn't send correct params when I reproduce steps in the ticket. Creates a VM using uploaded ISO, not linuxcontainers.org image I've selected.

AlexKarpov98:
@undsoft I refactored it a bit 😃

undsoft (approved):
Nice refactoring. Looks much cleaner.

PR #12082, opened 2025-06-03
NAS-135956: NVMe-oF namespaces

undsoft:
**Changes:**

- Adds ability to create and manage namespaces for an NVMe subsystem.

**Known issues:**
- If there are validation errors when creating a subsystem, the wizard may end up in a weird state.


undsoft:
Cannot reproduce. I'll try to get to it in next ticket.

PR #12084, opened 2025-06-03
NAS-127704 / 25.10 / Add warning to allowed IPs

undsoft:
**Testing:**

Saving the form will lock you out, so you may want to use mocks.


denysbutenko (approved):
Looks good.

PR #12089, opened 2025-06-04
NAS-136118 / 25.10 / Add STIG related logic for new users

AlexKarpov98:
Testing: 
See ticket.


src/app/pages/credentials/new-users/all-users/user-details/user-password-card/user-password-card.component.ts
```diff
@@ -20,8 +40,58 @@ import { YesNoPipe } from 'app/modules/pipes/yes-no/yes-no.pipe';
...
+    private errorHandler: ErrorHandlerService,
+  ) {}
+
+  ngOnInit(): void {
+    this.api.call('system.security.config').pipe(untilDestroyed(this)).subscribe((config) => {
+      this.isStigMode.set(config.enable_gpos_stig);
+    });
+  }
+
+  generateOneTimePassword(user: User): void {
```
undsoft: Passing in the argument seems unnecessary given how there is a user input.

undsoft (approved):
Actually, this ticket https://ixsystems.atlassian.net/browse/NAS-135760 will make it possible for users to user one-time password even if they already have a password, so we can remove the stig check.

1. There should be margin between icon and Generate One-Time Password.
2. Let's try a different icon. For example `mdi-lock-plus-outline`.
3. If the user has a password set, let's add a tooltip to the button saying: `Existing user password will not be removed`.
4. Let's update remove `User will be encouraged to choose their own password after they login for the first time.` in the dialog if user already has a password. If hard to implement a check there, just add `for the first time if they don't have a password.`.

PR #12095, opened 2025-06-05
NAS-136144: Offer to delete underlying file when namespace is deleted

undsoft:


AlexKarpov98 (approved):
Works good 👍 

Noticed an issue, trying to delete not a file, but UI proposes to delete a directory though:

https://github.com/user-attachments/assets/bb0195f5-001e-4949-af83-ea199677fccf

And as well found a bug:

https://ixsystems.atlassian.net/browse/NAS-136194

undsoft:
Okay, being able to select directories is probably something that should not be allowed on MW.

PR #12102, opened 2025-06-05
NAS-135972 / 25.04.2 / Don't show option to set password for non-local users

RehanY147:
**Changes:**

Don't show option to set password for non-local users

**Testing:**

See ticket

backport


undsoft (changes requested):
Not sure how this was tested before, but if you reproduce the steps described in the ticket, you will not be able to continue when setting up the password for the AD user.

1. Create a fangtooth VM connected to AD.
2. Enable STIG by:
- Creating another admin user (by adding it to `builtin_administrators` group). Log in with it and remove password from root.
- Enable 2FA in advanced settings and set 2FA for your user.
- Enable STIG in security settings.
3. Generate OTP for ACME/Administrator user and attempt to login.

ER: For this user I should only see the form to set up 2FA. Once I do that I should be able to login.
AR: I correctly see the 2FA form, but there is no button to continue.

This may need to be fixed in master too.

undsoft (approved):
👍🏼

PR #12103, opened 2025-06-06
Connect shell and ssh checkboxes to user form

denysbutenko:
**Changes:**

- If user has Roles, SMB, SSH or Shell Access checkboxes active.

**Testing:**

Try to create and edit a user.

### DIFF
--- a/src/app/pages/credentials/new-users/all-users/all-users.component.ts
+++ b/src/app/pages/credentials/new-users/all-users/all-users.component.ts
@@ -13,6 +13,7 @@ import {
 } from 'rxjs';
 import { UiSearchDirective } from 'app/directives/ui-search.directive';
 import { CollectionChangeType } from 'app/enums/api.enum';
+import { QueryParams } from 'app/interfaces/query-api.interface';
 import { User } from 'app/interfaces/user.interface';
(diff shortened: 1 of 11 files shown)


RehanY147 (changes requested):
I check the SSH and Shell Access when adding the user. Leave the Publish SSH Keys field empty and save the form. Then I open the edit form on the user and the fields aren't filled in correctly. The checkboxes are not what I saved. SSH is disabled and so is shell.

RehanY147 (changes requested):
I still don't see what I saved when I choose to edit the user.

https://github.com/user-attachments/assets/56fbd8e5-8cf6-472d-8e8b-fcaa53852f89

denysbutenko:
> I still don't see what I saved when I choose to edit the user.
> 
>  Screen.Recording.2025-06-20.at.5.41.50.AM.mov

It should raise error saying that you'll need to configure `home path` and `shell` to enable `Shell/SSH` access. I'll double-check and let you know.

denysbutenko:
@RehanY147 this problem requires more effort to resolve and it will be handled in other ticket(s). I added a temporary solution for this PR to set first shell option once `Shell Access` is enabled.

PR #12104, opened 2025-06-06
NAS-136202: Replace secondary fields with editables in zvol form

undsoft:
**Testing:**

There are some additional issues that I've discovered in zvol form for which I'll be creating additional tickets. See if you find them.

### Downstream

|Affects         |Reasoning
|Documentation   | Label changes for some fields in zvol form. Look and feel changes for the zvol form.


undsoft:
Yup, the loading needs to be refactored and we may need to introduce some sort of loading indication for editables, not sure. Won't solve in this PR.

PR #12106, opened 2025-06-06
NAS-136195 / 25.10 / Add ability to create zvols in ix-explorer

undsoft:
**Changes:**

- Adds ability to create zvols in ix-explorer.
- Decouples logic for creation buttons from ix-explorer.

**Testing:**

Check that zvols can be created from ix-explorer in nvme-of namespaces.
Check that dataset creation hasn't regressed.


src/app/modules/forms/ix-forms/components/ix-explorer/explorer-create-dataset/explorer-create-dataset.component.ts
```diff
@@ -0,0 +1,78 @@
...
+  imports: [
+    MatButton,
+    IxIconComponent,
+    TranslateModule,
+    RequiresRolesDirective,
+    TestDirective,
+  ],
+})
+export class ExplorerCreateDatasetComponent {
+  readonly datasetProps = input<Omit<DatasetCreate, 'name'>>({});
```
RehanY147: This name could be better. Also, since we override the name property in dialog submission when creating the dataset anyway, omitting name here seems unnecessary detail.

RehanY147 (changes requested):
I can't create a zvol on the nvme-of form. The button doesn't enable under `/dev/zvol/` which I imagine is because that's not a valid path. But you need to show valid dataset paths to create a zvol under.

src/app/modules/forms/ix-forms/components/ix-explorer/explorer-create-dataset/explorer-create-dataset.component.ts
undsoft: Cannot change the type, because `name` is required in `DatasetCreate` and I would have to cast here.

undsoft:
Added an ugly workaround.

RehanY147 (approved):

PR #12109, opened 2025-06-08
Check auth mechanisms for token approval

RehanY147:
**Changes:**

Calls to `auth.mechanism_choices` to determine if token login is allowed

**Testing:**

Test login page for regressions. Also see ticket for specific testing instructions

### DIFF
--- a/src/app/modules/auth/auth.service.ts
+++ b/src/app/modules/auth/auth.service.ts
@@ -9,6 +9,7 @@ import {
   Observable,
   of,
   ReplaySubject,
+  Subject,
   Subscription,
   switchMap,
   take,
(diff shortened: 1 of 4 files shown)


src/app/modules/auth/auth.service.ts
```diff
@@ -248,7 +250,9 @@ export class AuthService {
   }
 
   private setupPeriodicTokenGeneration(): void {
-    this.isTokenAllowed$.pipe(
+    this.checkIsTokenAllowed$.pipe(
+      switchMap(() => this.api.call('auth.mechanism_choices')),
```
undsoft: Let's add at least some error handling here. At least `console.error` it.

undsoft (changes requested):
The code is good, but something is wrong. Could be MW.
I've generated a one-time password via CLI and used to login. Tokens are not supported with OTP, so I get an error about it.

OTP can be generated with:

```
midclt call auth.generate_onetime_password '{"username": "root"}'
```

RehanY147:
To test this PR, create a VM that rebuilds middleware from the branch `NAS-136328`

undsoft (approved):
The UI code looks good, but middleware probably shouldn't send PASSWORD_CHANGE_REQUIRED if user already has a password.

PR #12116, opened 2025-06-09
Add a subscription internally so caller doesn't have to subscribe to open the slide-in

RehanY147:
…ubscribe

**Changes:**

Add a subscription internally so caller doesn't have to subscribe to open the slide-in

**Testing:**

See ticket. Also test when the form successfully closes, the response is emitted correctly by the observable.


undsoft (approved):
Thanks for the quick turnaround.

PR #12121, opened 2025-06-10
NAS-136168 / 25.10 / Update to Angular 20

denysbutenko:
**Changes:**

* Update to Angular 20
* Update date-fns

**Testing:**
Check pages for regressions introduced by adding new CSS variables in the material?

### Downstream

|Affects         |Reasoning
|Documentation   | mention in release note?

TODO:
- [x] Fonts in buttons are different.
- [x] Fonts in status pills (like in Shares) look different
- [x] Colors in radio groups are incorrect.
- [x] Toggle control looks different.
- [x] Apps warning Using 3rd party applications with TrueNAS... looks incorrect. Text too squished.
- [x] Header in job dialog and margins between other elements looks different.
- [x] Checkboxes have different color, but it's okay to keep them gray. They do however need to have light checkmark on light themes.
- [x] Boot Environments → Date Created now shows Invalid Date. Same for dates in reporting.
- [x] Button group checkmark used to be white, not blue.
- [x] There is a tiny scrollbar in confirmation dialog that says You have unsaved changes.


src/app/core/testing/mock-enclosure/enclosure-templates/utils/make-array-device-slots.utils.ts
```diff
@@ -18,15 +18,15 @@ export function makeArrayDeviceSlots(
...
-      type: null,
-      rotationrate: null,
-      pool_info: null,
+      drive_bay_light_status: null as null,
+      size: null as null,
+      model: null as null,
+      serial: null as null,
+      type: null as null,
+      rotationrate: null as null,
+      pool_info: null as null,
```
undsoft: `null as null` means that only `null` is allowed as a value, which is not true. It's better to case `as DashboardEnclosureSlot` on the object.

src/app/modules/alerts/store/alert.reducer.ts
```diff
@@ -38,7 +38,7 @@ export const alertReducer = createReducer(
   on(alertIndicatorPressed, (state) => ({ ...state, isPanelOpen: !state.isPanelOpen })),
   on(alertPanelClosed, (state) => ({ ...state, isPanelOpen: false })),
 
-  on(adminUiInitialized, (state) => ({ ...state, isLoading: true, error: null })),
+  on(adminUiInitialized, (state) => ({ ...state, isLoading: true, error: null as null })),
```
undsoft: Same here.

src/app/modules/jobs/store/job.reducer.ts
```diff
@@ -31,7 +31,7 @@ export const jobReducer = createReducer(
   on(jobIndicatorPressed, (state) => ({ ...state, isPanelOpen: !state.isPanelOpen })),
   on(jobPanelClosed, (state) => ({ ...state, isPanelOpen: false })),
 
-  on(adminUiInitialized, (state) => ({ ...state, isLoading: true, error: null })),
+  on(adminUiInitialized, (state) => ({ ...state, isLoading: true, error: null as null })),
```
undsoft: and here and in other places You should be able to use `as string | null`

undsoft (changes requested):
- Fonts in buttons is different.
- Fonts in status pills (like in Shares) looks different
- Colors in radio groups is incorrect
- Toggle looks different
- Apps warning `Using 3rd party applications with TrueNAS...` looks incorrect.
- Header in job dialog and margins between other elements looks different.
- Checkboxes have different color, but it's okay to keep them gray. They do however need to have light checkmark on light themes.
- Boot Environments → Date Created now shows `Invalid Date`. Same for dates in reporting.
- Button group checkmark used to be white, not blue.
- There is a tiny scrollbar in confirmation dialog that says `You have unsaved changes.`


Also I had to upgrade my node to make project build. We need to make sure our build process won't fail because of it. Please use Jenkins -> master -> custom with `TRY_BRANCH_OVERRIDE` to build your branch to make sure it builds.

package.json
```diff
@@ -44,24 +44,24 @@
     "url": "https://jira.ixsystems.com"
   },
   "engines": {
-    "node": ">=18.19.1"
+    "node": ">=22.0.0"
```
undsoft: We need to confirm that this is actually required by Angular and is supported by our build pipeline.

undsoft (changes requested):
Looks like something is wrong with translation extractor – I can see some messages removed from translation files, even though they are present in code.

package.json
denysbutenko: [angular.dev](https://angular.dev/reference/versions) says at least 20.19 or higher.

denysbutenko:
@undsoft I updated `fetch_node.sh` and `truenas_webui` package builds successfully. Instance available at http://10.220.37.22/

Dockerfile
```diff
@@ -1,39 +1,43 @@
...
 	nginx \
 	&& apt-get clean
 
-#Remove any extra packages we don't need from the container
-# Also cleanup any random things we don't want to distribute
-RUN export SUDO_FORCE_REMOVE=yes \
-	&& apt autoremove -y || true \
-	&& apt autoclean \
+# Install Node.js from NodeSource (exact version 20.19)
+RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash - && \
```
undsoft: Shouldn't these changes be reverted in favor of changing the first line in the old implementation?

jest.config.cjs
```diff
@@ -7,32 +7,39 @@ process.env.TZ = 'Europe/Kiev';
...
-  'ip-regex',
+  'ip-regex', 
   'cidr-regex',
   'lodash-es',
   'internmap',
   'd3',
   'delaunator',
   'cheerio',
   'robust-predicates',
+  '@angular',
```
undsoft: Seems to work just fine locally without some of those. Are you sure they are necessary?

src/assets/i18n/af.json
```diff
@@ -901,14 +894,11 @@
   "Choose an alert service to display options for that service.": "",
   "Choose an authentication method.": "",
   "Choose between a <i>Hex</i> (base 16 numeral) or <i>Passphrase</i> (alphanumeric) style encryption key.": "",
-  "Choose File": "",
```
undsoft: Those strings are also used.

undsoft (changes requested):
Most things look good. Some minor visual glitches:
1. Checkboxes used to have white tickmark.
Снимок экрана 2025-07-02 в 12 06 35

2. No button in unsaved changes dialog is yellow.
Снимок экрана 2025-07-02 в 12 05 48

3. When you press primary button it becomes yellow.

https://github.com/user-attachments/assets/3680c169-35c0-44b1-9a69-2e8072d5ea4a

undsoft (approved):
Looks good, some minor issues:

- `No` button in the unsaved changes dialog gets slightly Blue when focused.
- Disabled buttons like Save when form is not valid has a hover state when it didn't use to. 
- Hover colors for button is of different color now.
- Hovering over disabled checkboxes in ix-permissions (for example in old users form) shouldn't make them blue.
- Text on checkbox labels is now dimmer than it used to be.

PR #12125, opened 2025-06-12
NAS-136290: Do not override global ANA setting when creating a subsystem

undsoft:
**Testing:**

Settings is not actually exposed in the form. Code review is enough.


denysbutenko (approved):
Looks good.

PR #12128, opened 2025-06-12
NAS-136294: Move download key and other fixes

undsoft:
**Changes:**

- Changed `View API Keys` and `View Logs` into links so that user can open them in new tab.
- Added snackbar confirmation when locking/unlocking user.
- Add api key now opens the form right away when it is possible
- Download authorized keys is now in Access card
- Code clean up


undsoft:
time 2:30

RehanY147 (approved):

PR #12135, opened 2025-06-15
NAS-136297 / 25.10 / "Create New" Device Iscsi Extent does not show pool/datas…

undsoft:
**Testing:**

See ticket.


undsoft:
time 1:00

PR #12137, opened 2025-06-15
NAS-134619: App page uses both "update" and "upgrade" as terminology …

undsoft:
**Changes:**

`Upgrade` is still used by middleware and so I didn't rename interfaces that refer to middleware objects.
But I think `Update` sounds better for the UI, so UI-related things were updated.

### Downstream

|Affects         |Reasoning
|Documentation   | Minor label and test tag changes.


undsoft:
time 0:30

AlexKarpov98 (approved):
👍

PR #12138, opened 2025-06-15
NAS-136331 / 25.10 / Use `protected` in more places

undsoft:
**Changes:**

We want people to test components using API that was deliberately made public or by interacting with it via UI.
However, there are some cases where people just directly call methods that should be internal.
As an effort to combat this, I'd like to start using `protected` in more places to make it clearer where component boundaries are.


undsoft:
time 0:30

RehanY147 (approved):

PR #12142, opened 2025-06-16
NAS-136318 / 25.10 / Unsaved changes dialog appears after successfully submitting the full-width forms

AlexKarpov98:
Testing: see ticket.


undsoft (approved):
Thanks, warnings appears to be working as expected now.

PR #12146, opened 2025-06-17
NAS-136253 / 25.10 / Adjust webui for SMB share schema changes related to new API

AlexKarpov98:
Testing: see ticket.

Changes:
- Non-related to the preset fields are now hidden and not only disabled.

### DIFF
--- a/src/app/modules/forms/ix-forms/components/ix-errors/ix-errors.component.ts
+++ b/src/app/modules/forms/ix-forms/components/ix-errors/ix-errors.component.ts
@@ -1,7 +1,6 @@
 import { LiveAnnouncer } from '@angular/cdk/a11y';
 import {
   ChangeDetectionStrategy, ChangeDetectorRef, Component, input, OnChanges,
-  OnInit,
 } from '@angular/core';
 import { AbstractControl } from '@angular/forms';
 import { MatError } from '@angular/material/form-field';
@@ -32,7 +31,7 @@ export const ixManualValidateError = 'ixManualValidateError';
     TranslateModule,
   ],
 })
-export class IxErrorsComponent implements OnChanges, OnInit {
+export class IxErrorsComponent implements OnChanges {
   readonly control = input.required<AbstractControl>();
   readonly label = input<string>();
 
@@ -97,12 +96,6 @@ export class IxErrorsComponent implements OnChanges, OnInit {
     ).subscribe(() => {
       this.handleErrors();
     });
-  }
-
-  ngOnInit(): void {
-    if (this.control().errors) {
-      this.handleErrors();
-    }
   }
 
   private handleErrors(): void {


src/app/interfaces/smb-share.interface.ts
```diff
@@ -34,17 +34,41 @@ export interface SmbShare {
     watch_list?: string[];
     ignore_list?: string[];
   };
+  options: {
```
undsoft: I'd like to see a union for what `options` can really be.
E.g.

```
export type SmbShareOptions = 
 | LegacySmbShareOptions
 ...
 ```

**Review by undsoft — CHANGES_REQUESTED** [review]
Interesting changes, hopefully we can make the form even simpler.

- Can we check with Andrew whether we can force users to select a new share type on edit? E.g. user opens the form to edit the share and the first thing they have to do is to select a new share purpose?  I'd like to avoid having to support all those legacy fields.
- In either case, we should not show Legacy Share as an option when creating a share.
- There is something wrong with `Remote Path` for `External Share`. I cannot get it to accept any values.
- I can open many tooltips at the same time in Purpose dropdown. Can we use matTooltip instead?
- `VUID` tooltip says `Usually auto-assigned by server`, but MW requires some value.
- I also don't know what values are acceptable – everything I try is not accepted.

File: src/app/pages/sharing/smb/smb-form/smb-form.component.ts [hunk 2]
```diff
@@ -444,21 +452,44 @@ export class SmbFormComponent implements OnInit, AfterViewInit {
...
+    delete flatShare.options;
+
+    if (!flatShare.purpose) {
+      flatShare.purpose = SmbPresetType.LegacyShare;
+    }
+
+    if (share.purpose === SmbPresetType.LegacyShare) {
+      this.dialogService.warn(
+        this.translate.instant('Legacy Share'),
+        this.translate.instant('To save changes, choose a new purpose. Legacy shares cannot be edited without migrating.'),
```
undsoft:
```
In order for share to be saved, it needs to be migrated.
Please select a new share purpose to migrate the share.
````

File: src/app/pages/sharing/smb/smb-form/smb-form.component.ts [hunk 3]
```diff
@@ -194,7 +199,12 @@ export class SmbFormComponent implements OnInit, AfterViewInit {
   form = this.formBuilder.group({
     path: ['', [Validators.required]],
     name: ['', Validators.required],
-    purpose: [null as SmbPresetType | null],
+    purpose: [SmbPresetType.DefaultShare as SmbPresetType | null, [
```
undsoft: After you close the dialog, the legacy type is still there on the list. If you don't touch the list, validation is not triggered and you can save share with legacy type.

src/app/interfaces/smb-share.interface.ts
```diff
@@ -34,17 +34,108 @@ export interface SmbShare {
...
+  [SmbPresetType.TimeMachineShare, T('Time Machine Share')],
+  [SmbPresetType.MultiProtocolShare, T('Multi-Protocol Share')],
+  [SmbPresetType.TimeLockedShare, T('Time Locked Share')],
+  [SmbPresetType.PrivateDatasetsShare, T('Private Datasets Share')],
+  [SmbPresetType.ExternalShare, T('External Share')],
+]);
+
+export const smbPresetTooltips = new Map<SmbPresetType, string>([
+  [SmbPresetType.DefaultShare, T('Set the SMB share for best compatibility with common SMB clients.')],
+  [SmbPresetType.LegacyShare, T('Set the SMB share for compatibility with older TrueNAS versions. Automated backend migrations use this to help the administrator move to better-supported share settings. It should not be used for new SMB shares.')],
```
undsoft: Either never show this option on the list or change tooltip to:
```
Legacy share type for compatibility with older TrueNAS versions. Please select a new share purpose.
```

undsoft (changes requested):
Changes look fine, although we need to wait for confirmation from MW that we can actually do that.
If middleware agrees we could also remove fields that are only present in Legacy shares such as `Allow Guest Access`.

undsoft (approved):
Some issue that I trust you to resolve:

- Time Machine Share
  -  Dataset Naming Schema should only appear when Auto Dataset Creation is true. Should be sent as null otherwise.
  - I've tried to use values from the tooltip, but got validation error

- Auto Quota needs a better tooltip

- I probably shouldn't get "Configuring SMB exports of root-level datasets may lead to storage reconfiguration issues. Consider creating a dataset instead." warning when creating an external share.

- API docs now list VEEAM_REPOSITORY_SHARE as share type. I don't know if it needs to be added.

AlexKarpov98:
@undsoft changes addressed now. you may review if you have time. I'll merge a bit later.

PR #12147, opened 2025-06-17
NAS-136312: Improvements for table in new users

AlexKarpov98:
**AI GENERATED:**  + some additional small fixes.

## Summary
- add utility to get user type and use in profile card
- show API keys in user access column
- remove UID column and replace Built in with Type in new users list
- disable sorting on Access column
- set default sorting and filter for built-in users + root user
- update unit tests

------
https://chatgpt.com/codex/tasks/task_e_68513b995fd8832bbb55c6faf97a4c5e


undsoft (approved):
Nice 🙂 
I wonder if we can get Codex to use proper branch names and follow github PR templates 🤔

AlexKarpov98:
> Nice 🙂 I wonder if we can get Codex to use proper branch names and follow github PR templates 🤔

There is an option to copy changes and apply manually - that's pretty easy as well. Here I was just curious about creating it right away from codex.

src/app/pages/credentials/new-users/all-users/user-list/user-access-cell/user-access-cell.component.html
```diff
@@ -10,3 +10,7 @@
 @if (user().ssh_password_enabled) {
   <ix-icon name="mdi-ssh" [matTooltip]="'Has SSH Access' | translate"></ix-icon>
 }
+
+@if (this.user().api_keys?.length) {
```
denysbutenko: `this` is unnecessary here

PR #12152, opened 2025-06-18
NAS-136307 / 25.10 / Add R60 enclosure to the UI

undsoft:
**Testing:**

See machine IP in the ticket.


AlexKarpov98 (approved):
Works as expected.

PR #12153, opened 2025-06-18
NAS-130303 / 25.10 / Validation errors are not announced

undsoft:
**Testing:**

Enable screen reader functionality in your system (VoiceOver on mac) and trigger validation error in the form.


RehanY147 (changes requested):
With Voiceover on MacOS, the error is announced when I am on a required input. I type something in and then remove so it has the required validation error. But if I focus out of the input and then focus back into it, then the error is not announced.

undsoft:
I didn't really intend it to announce error every time you focus on the field, just once when the error appears.

RehanY147 (approved):

PR #12156, opened 2025-06-19
NAS-135587 / 25.10 / Windows (SMB) share Running should be green like other "running" states?

AlexKarpov98:
Testing: see ticket.

Shared colors introduced:

```
  --status-active: var(--green);
  --status-inactive: var(--grey);
  --status-error: var(--red);
  --status-in-progress: var(--yellow);
  ```


undsoft (approved):
Thanks, much cleaner.

PR #12159, opened 2025-06-19
NAS-136372: Various smaller fixes

undsoft:
**Changes:**

- Reworded validation message in SMB a bit.
- Fixes race condition in Cloud Credentials table https://ixsystems.atlassian.net/browse/NAS-136315
- Fix memory leaks with how we bind to events in various places.
- Other code cleanup


denysbutenko (approved):
Looks good

PR #12162, opened 2025-06-19
NAS-136373: Automate PR creation

undsoft:
**Changes:**

Probably not super useful, but it's somewhat entertaining.
You can use `claude` and  `/make-pr NAS-136373: Automate PR creation` to create a PR.


undsoft:
time 0:30

PR #12165, opened 2025-06-22
NAS-135972: Fixes issue with the ticket once login is successful

RehanY147:
**Changes:**

Fixes issue with the ticket once login is successful

**Testing:**

See ticket


undsoft (changes requested):
Still see 2FA setup when 2FA is not enabled globally.

1. Create a new admin user on a system that doesn't have 2FA enabled globally.
2. Generate OTP for it.
3. Login.

Снимок экрана 2025-06-23 в 13 42 34

undsoft (approved):
Thanks, everything appears to be working correctly now.

PR #12167, opened 2025-06-23
NAS-136348 / 25.04.2 / Old VM section

undsoft:
**Testing:**

See ticket for info.

There is a lot of dead code that would need to be cleaned up, but this will happen later in master when things stabilize.


denysbutenko (approved):
The page is loaded but has a few issues.

image

denysbutenko (approved):
Looks good on instance with old VM support.

image

PR #12168, opened 2025-06-23
NAS-136262 / 25.10 / Some slide-ins are not closed on logout

AlexKarpov98:
Testing: see ticket.

Preview:

https://github.com/user-attachments/assets/21fa1c38-b3e9-4783-a295-27bfe6c26bcd


src/app/app.component.ts
```diff
@@ -77,6 +77,7 @@ export class AppComponent implements OnInit {
   }
 
   private logOutExpiredUser(): void {
```
undsoft: This probably should already be handled in app.component.ts when listening to navigation changes.

undsoft (changes requested):
I've reproduced it in a different way by opening two slide-ins like in the ticket and then pressing back to navigate to previous page in the admin UI.
The slide-ins closed, but when I tried to open them again, the button didn't do anything.

PR #12172, opened 2025-06-24
NAS-136293 / 25.10 / Fix group fields in new user form

AlexKarpov98:
**Testing:**
See ticket.

Additionally Added: 
- Fix for form Submission disabled state.
- Confirmation dialog if user accidentally closes the form.


src/app/pages/credentials/new-users/user-form/additional-details-section/additional-details-section.component.ts
```diff
@@ -201,6 +203,19 @@ export class AdditionalDetailsSectionComponent implements OnInit {
     this.listenValueChanges();
   }
 
+  getGroupNameById(id: number): string {
+    if (!this.groupNameCache.has(id)) {
+      this.groupNameCache.set(id, '');
+
+      this.api.call('group.query', [[['id', '=', id]]]).pipe(
```
undsoft: It would be better to collect all ids and make one call with `id IN values`

RehanY147 (changes requested):
What Evgeny said

src/app/pages/credentials/new-users/user-form/additional-details-section/additional-details-section.component.ts
AlexKarpov98: @undsoft - you mean it's better to get all groups available eventually?
in current code case we first select a group - then request is made to get info regarding the selected group, to show it as NAME but not as ID for user.

I thought that for performance it's better to only get a group name for a selected group, but not for all.

This method is only used to get the selected group name, it can be as only one group selected.

undsoft (changes requested):
I somewhat misread the code.
Here's how it should work.
Groups editable should provide a simple overview of all groups.
- By default it should be `New <i>username</i> group` without the dot icon. Not sure where the dot came from.
- If I manually add aux group, it could change to: `New primary <i>username</i> group. <br> Auxgroup1, auxgroup2`. Right now it doesn't react to changes.
- It should correctly show all groups, including aux groups on edit.

AlexKarpov98:
@undsoft @RehanY147 - update

https://github.com/user-attachments/assets/9201dd65-276f-4b78-ae4e-1cf52edd71a8

undsoft (changes requested):
So what I meant by improving `getGroupNameById` is that if you have 1 primary group and 2 aux groups, the method will make 3 requests, when it can be one.

AlexKarpov98:
@RehanY147  you removed your review, but it left `requested changes` status which is blocking me.

undsoft (approved):
Looks okay, but if you add a group via TrueNAS Access checkbox -> select role, it appears as a number.

AlexKarpov98:
Final fix:

https://github.com/user-attachments/assets/0e16a0ab-6200-4ab3-89df-945f18de1399

RehanY147 (approved):

PR #12176, opened 2025-06-25
NAS-135333 / 25.10 / Adds user selection via navigation

RehanY147:
**Changes:**

Creates a new users specific data provider. Adds function to select user from the navigation url.

**Testing:**

Test for regressions. Also test the row selection function on the table. Reload the page and ensure that the selected user is selected agian after page reload. Add a new user and see that this user is selected on the table after creation. Code review.


RehanY147:
Will create new PR

PR #12187, opened 2025-06-28
NAS-135333 / 25.10 / Enable linking to a user via query params

RehanY147:
**Changes:**

Adds a new users-data-provider and uses query-params to select a user and remember them in the url

**Testing:**

Test selecting users and reloading page to see they are still selected

### DIFF
--- a/src/app/pages/credentials/new-users/all-users/all-users.component.ts
+++ b/src/app/pages/credentials/new-users/all-users/all-users.component.ts
@@ -1,10 +1,10 @@
-import { AsyncPipe } from '@angular/common';
+import { AsyncPipe, Location } from '@angular/common';
 import {
   ChangeDetectionStrategy, Component, OnInit,
   viewChild, OnDestroy,
   ChangeDetectorRef,
 } from '@angular/core';
-import { ActivatedRoute, Router } from '@angular/router';
+import { ActivatedRoute } from '@angular/router';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { TranslateModule } from '@ngx-translate/core';
 import {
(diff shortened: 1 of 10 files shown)


src/app/pages/credentials/new-users/all-users/all-users.component.ts
```diff
@@ -48,55 +48,73 @@ import { UserListComponent } from 'app/pages/credentials/new-users/all-users/use
...
-      dataProvider.expandedRow = this.masterDetailView().isMobileView() ? null : users[0];
-    });
-    dataProvider.load();
-    return dataProvider;
-  });
+  private readonly defaultParams = [
+    [['OR', [['builtin', '=', false], ['username', '=', 'root']]]],
+  ] as QueryParams<User>;
+
+  protected readonly dataProvider = new UsersDataProvider(this.api, 'user.query', this.defaultParams);
```
undsoft: `user.query` here seems unnecessary given how it is a `UsersProvider`.

src/app/pages/credentials/new-users/all-users/all-users.component.ts
```diff
@@ -48,55 +48,73 @@ import { UserListComponent } from 'app/pages/credentials/new-users/all-users/use
...
   ) { }
 
   ngOnInit(): void {
+    this.setupDataProvider();
     this.subscribeToUserChanges();
   }
 
+  setupDataProvider(): void {
+    this.dataProvider.paginationStrategy = new PaginationServerSide();
+    this.dataProvider.sortingStrategy = new SortingServerSide();
```
undsoft, line 74: Doesn't this already happen within dataProvider?

undsoft (approved):
Looks good, nice clean up. The only issue is that if you select a user and then search for something so that the list is empty, the url doesn't reset.

PR #12190, opened 2025-06-29
NAS-115533 / 25.10 / Suggested DNS / Gateway Filled Out

kmoore134:
**Changes:**

This has been a long-standing bug with initial TrueNAS setup that has irritated for me years now. We fixed part of it by allowing users to set a gateway when they switch from DHCP -> Static IPs a while back, but in the process we completely skipped over the DNS side of the equation. This required the end-user to manually go into the general network config and add back DNS, otherwise they get surprised later when TrueNAS can't resolve.

Additionally we fixed some general UI clumsiness here:

* Registration button wasn't being enabled properly until you unfocused the gateway input box
* The default gateway wasn't helpfully filled out in advance, which is a shame when 99% of all users will keep the same one. 

**Old Behaivor:**

[OldGatewayDialog.webm](https://github.com/user-attachments/assets/9c21ae34-0ec6-4c9d-8dbe-1ab829b810fa)

**New Behaivor:**

[NewGatewayDialog.webm](https://github.com/user-attachments/assets/ffed746e-ba65-46fc-a03d-5f16e6fbc2a8)

**Testing:**

Tested a few times in the situation where I have a fresh VM which of course was DHCP out of the box, but needed to change it to Static. 

**NOTE:** There is a bug on the MW side still, where if you start this process, then abort partway through, the DNS from DHCP is no longer working and doesn't show up in the network.general.summary endpoint. This is not UI related though and should be fixed separately. 

Middleware Bug:
https://ixsystems.atlassian.net/browse/NAS-136360

### Downstream

|Affects         |Reasoning
|Documentation   | Will need updated screenshots of the new dialog and process.


src/app/pages/system/network/components/default-gateway-dialog/default-gateway-dialog.component.ts
```diff
@@ -76,9 +106,53 @@ export class DefaultGatewayDialog {
...
+
+        const currentDns2 = state.value.nameservers[1];
+        if (currentDns2) {
+          this.form.patchValue({ dns2: currentDns2 });
+        }
+      }
+    });
+  }
+
+  private optionalIpValidator() {
```
undsoft: Original `ipv4Validator` already doesn't require value to be set.

src/app/pages/system/network/components/default-gateway-dialog/default-gateway-dialog.component.ts
```diff
@@ -76,9 +106,53 @@ export class DefaultGatewayDialog {
     private validatorsService: IxValidatorsService,
   ) {}
 
+  ngOnInit(): void {
+    this.currentGateway$.pipe(untilDestroyed(this)).subscribe((state) => {
```
undsoft: Since you are loading values for the form here, it's better to get rid of loading state approach in the template.
1. Extract this code to new method.
2. Just make normal call to `this.api.call` with error handling and loader like in other forms.
3. Collapse statements that patch form values.
4. Change `currentGateway`, `currentDns1` and `currentDns2` that are currently unused to signals and use them in the template instead of `ixWithLoadingState`

src/app/pages/system/network/components/default-gateway-dialog/default-gateway-dialog.component.spec.ts
```diff
@@ -5,54 +5,166 @@ import { MatButtonHarness } from '@angular/material/button/testing';
...
+import { throwError } from 'rxjs';
 import { mockCall, mockApi } from 'app/core/testing/utils/mock-api.utils';
 import { mockAuth } from 'app/core/testing/utils/mock-auth.utils';
 import { NetworkSummary } from 'app/interfaces/network-summary.interface';
 import { IxInputHarness } from 'app/modules/forms/ix-forms/components/ix-input/ix-input.harness';
 import { ApiService } from 'app/modules/websocket/api.service';
 import { DefaultGatewayDialog } from 'app/pages/system/network/components/default-gateway-dialog/default-gateway-dialog.component';
+import { ErrorHandlerService } from 'app/services/errors/error-handler.service';
+
+const createComponent = createComponentFactory({
```
undsoft: Needs to move back within `describe` block.

src/app/pages/system/network/components/default-gateway-dialog/default-gateway-dialog.component.spec.ts
```diff
@@ -5,54 +5,166 @@ import { MatButtonHarness } from '@angular/material/button/testing';
...
-      ]),
-      mockProvider(MatDialogRef),
-    ],
-  });
-
   beforeEach(() => {
     spectator = createComponent();
     loader = TestbedHarnessEnvironment.loader(spectator.fixture);
     api = spectator.inject(ApiService);
+    spectator.detectChanges();
```
undsoft: Not needed.

src/app/pages/system/network/components/default-gateway-dialog/default-gateway-dialog.component.spec.ts
```diff
@@ -5,54 +5,166 @@ import { MatButtonHarness } from '@angular/material/button/testing';
...
-    expect(spectator.query('h1')).toHaveText('Register Default Gateway');
-    expect(spectator.query('p')).toHaveText('Editing interface will result in default gateway being removed, which may result in TrueNAS being inaccessible. You can provide new default gateway now:');
+    expect(spectator.query('h1')).toHaveText('Set Gateway and DNS');
+    expect(spectator.query('p')).toHaveText('Editing interface will result in default gateway being removed, which may result in TrueNAS being inaccessible. You can provide new default gateway and DNS information now:');
     expect(byText('Current Default Gateway: 1.1.1.1')).toBeTruthy();
+    expect(byText('Current DNS #1: 8.8.8.8')).toBeTruthy();
+    expect(byText('Current DNS #2: 8.8.4.4')).toBeTruthy();
+  });
+
+  it('should pre-fill the input with current gateway and DNS', async () => {
```
undsoft: Here and below it's better to use IxFormHarness instead of querying each control individually.

src/app/pages/system/network/components/default-gateway-dialog/default-gateway-dialog.component.spec.ts
```diff
@@ -5,54 +5,166 @@ import { MatButtonHarness } from '@angular/material/button/testing';
...
-    await defaultGatewayInput.setValueAndTriggerBlur('192.168.1.1');
+    await defaultGatewayInput.setValue('192.168.1.1');
 
     const registerGatewayButton = await loader.getHarness(MatButtonHarness.with({ text: 'Register' }));
     await registerGatewayButton.click();
 
     expect(api.call).toHaveBeenCalledWith('interface.save_default_route', ['192.168.1.1']);
   });
+
+  it('should save DNS entries to session storage when provided', async () => {
```
undsoft: Only keep three cases:
1. One that checks that header shows correct current values.
2. One that makes sure that form gets pre-filled with correct values.
3. One that validates how values get saved when correct form is submitted (validate side-effects of how the form is saved here).

Remove separate test cases for validators and error handling.

src/app/pages/system/network/network.component.ts
```diff
@@ -345,16 +346,51 @@ export class NetworkComponent implements OnInit {
         untilDestroyed(this),
       )
       .subscribe(() => {
-        this.store$.dispatch(networkInterfacesChanged({ commit: true, checkIn: true }));
+        // Check for pending DNS entries from session storage
```
undsoft: This needs to be reworked.
Current approach doesn't jive well with how other network changes are saved and introduces unusual communication between components.

Instead:
1. Save new DNS settings when `default-gateway-dialog` is closed. Make call to `network.configuration.update`, show snackbar there, update tests.
2. Do not use session storage and remove everything related to session storage that was added in this branch.
3. I'm pretty sure most changes to this file can be reverted.

undsoft (changes requested):
There are some typical AI mistakes, but Claude also took a creative approach to how DNS settings are saved.
It waits for Test - Save changes flow to be completed before saving new DNS , which forced it to use session storage.

It would be much simpler if DNS settings were just saved when the dialog is closed. That would clean up the code and that's how Network Configuration form works anyway.
It is unexpected to the user that there are two different flows: one saves right away and another stages changes for later.

yocalebo:
This conflicts with an already opened ticket here: https://ixsystems.atlassian.net/browse/NAS-136360

src/app/pages/system/network/network.component.ts
kmoore134: Thanks Evgeny, good feedback. Removed that functionality and went to the approach you suggested. Will update further when the backend ticket lands with proper DNS support.

kmoore134:
Will update PR again once backend piece lands. Believe the various feedback has been addressed, diff is much cleaner now. Let me know if anything else looks off to your eyes.

src/app/pages/system/network/components/default-gateway-dialog/default-gateway-dialog.component.html
```diff
@@ -18,6 +39,18 @@ <h1 mat-dialog-title>{{ 'Register Default Gateway' | translate }}</h1>
         [required]="true"
         [tooltip]="helptext.ipv4gatewayTooltip | translate"
       ></ix-input>
+      
+      <ix-input
+        formControlName="dns1"
+        [label]="'DNS Server #1' | translate"
```
RehanY147: The label can be improved by just stating "Primary DNS Server" or maybe mentioning "Primary" in the label somewhere e.g., "DNS Server #1 (Primary)" to avoid a click to see the tooltip to know that's what it is. Same for the secondary DNS input below.

RehanY147 (changes requested):
The inputs could be ix-combobox inputs which allow new custom values. There's an example of this in `src/app/pages/datasets/modules/permissions/components/edit-nfs-ace/edit-nfs-ace.component.html` for the `formControlName="user"` `ix-combobox` field. This removes the need to show the values separately for the user to copy and paste into the input fields. They can just get the values again from the input itself. There's also a variation of `ix-input` which enables such functionality.

kmoore134:
Updated to address that feedback, thanks @RehanY147

kmoore134:
PR updated and adapted to the new middleware endpoints here:

https://github.com/truenas/middleware/pull/16713

Tested with Logan this morning, appears to be functional. 

@AlexKarpov98 - Appreciate a quick review on this when you have a moment, since the existing UI method is broken as of today with the backend change that went in.

PR #12196, opened 2025-06-30
NAS-133847: SED settings for ATA Security User 

AlexKarpov98:
Testing: see ticket.


RehanY147 (changes requested):
The password field is filled in when I open the slide-in and then when I enter the same letter into the other input, it complains that they don't match but they do.

https://github.com/user-attachments/assets/f948196a-626f-4bba-86ad-54e1e1b382c6

AlexKarpov98:
@RehanY147 - yeah, having `–` as a start value is probably a bug.

But what you wrote about that you typed down the same character is wrong. 
The starting character was `–` but you probably entered `-` 
These are different but visually similar. `– -`

RehanY147 (approved):

PR #12198, opened 2025-06-30
WebUI greek translation

chrislongros:
Added greek translation of the WebUI


undsoft (approved):

PR #12200, opened 2025-07-01
NAS-136346 / 25.10 / Restart SMB Service Dialog Displays even when not required

AlexKarpov98:
Testing: see ticket.

Additional fixes:
- interface cleanup 
- path & interface usage fixes


RehanY147 (changes requested):
Getting this error. This hapened after I created an SMB share, started the service. Created a new share and saved and got this error
image

RehanY147 (approved):

PR #12201, opened 2025-07-01
NAS-136036 / 25.10 / Scroll jumping issue in datasets

AlexKarpov98:
Testing: see ticket.

You can compare how it works on `m40` and then use that machine locally to see the fix.


undsoft (changes requested):
Issue is fixed, but only in datasets. Same issue exists in other master-details view like Installed apps list.

undsoft (approved):
New users page may also need this treatment.

AlexKarpov98:
> New users page may also need this treatment.

it works fine there.

PR #12205, opened 2025-07-01
(i18n) Updated Irish translations

aindriu80:
**Changes:**

This PR updates approximately 444 strings in Irish translation.  

**Testing:**

As with previous translation I checked with Poedit to ensure everything is ok and also used messageformat-parser, checks include:-

✔ Correct use of {n, plural, one {...} other {...}}
✔ No missing or extra {} braces
✔ No syntax errors like =1 where one should be used
✔ No missing plural forms (e.g., only one but missing other)


aindriu80:
thanks :)

PR #12214, opened 2025-07-02
NAS-136520 / 25.10 / Transport Type just spins, cannot Add Port on NVMeOF

undsoft:
**Changes:**

Previous issue appeared when trying to add a subsystem on Shares dashboard.


denysbutenko (approved):
It works well!

PR #12216, opened 2025-07-03
NAS-136526 / 25.10 / Some NVMe-oF Global Settings should be hidden

undsoft:
**Changes:**

- Disables `Enable Asymmetric Namespace Access (ANA)` for non-enterprise systems.
- Removes `Generate Cross-port Referrals for Ports On This System` checkbox.


undsoft:
time 0:30

PR #12219, opened 2025-07-03
NAS-127775 / 25.10 / UI offers to restart web services when no changes to allo…

undsoft:
…wed addresses has been made

**Changes:**

Does not prompt to restart web service when editing allowed addresses and no changes were made.


undsoft:
time 0:15

PR #12220, opened 2025-07-03
NAS-136495 / 25.04.2 / Replace "Instance(s)" with "Container(s)" on Containers s…

undsoft:
**Testing:**

String changes. Code review is enough.


undsoft:
> Shouldn't components also be renamed?

Will do later in master.

PR #12225, opened 2025-07-04
NAS-136371 / 25.10 / Finish support for update tagging

AlexKarpov98:
Testing: see ticket.

Preview:

https://github.com/user-attachments/assets/e557cb9b-3fc3-4eda-94af-6e36bb29eba5

- You need to use a custom VM with `update` middleware branch [https://github.com/truenas/middleware/pull/16495]

- Update profile section is only visible if `update.profile_choices` includes at least one profile with `available: true`

- these below methods were removed 
```
  'update.check_available': { params: void; response: SystemUpdate };
  'update.get_auto_download': { params: void; response: boolean };
  'update.get_pending': { params: void; response: SystemUpdateChange[] };
  'update.set_auto_download': { params: [boolean]; response: void };
  ```
  
  now we use:
  ```
  'update.status': { params: void; response: UpdateStatus };
   'update.config': { params: void; response: UpdateConfig };
  'update.update': { params: [Partial<UpdateConfig>]; response: UpdateConfig };
  ```
  
  V. Vinogradenko told me about such hack to make sure that there are some available profiles 
  > ( You won’t be able to set any other profile if you have a developer profile)
  
`midclt call test.set_mock system.is_enterprise null '{"return_value":false}'`

> If you do this, GENERAL and CONSERVATIVE will be available, but MISSION_CRITICAL wont

Remove mock with:
`midclt call test.remove_mock system.is_enterprise null`


william-gr:
I wonder if makes sense to show the all the trains, but disabling the ones not allowed, giving the user a reason so they dont have to know/guess. Thoughts?

undsoft (approved):
Added protections against some edge cases and will now show unavailable profiles.
Note that the list that is returned by middleware (including unavailable profiles) is different depending on the license.

Image

PR #12230, opened 2025-07-07
NAS-136574: Show network interface description in the table

undsoft:
**Changes:**

Simply shows description next to name if it's available.


undsoft:
time 0:15

PR #12235, opened 2025-07-08
NAS-136503: Be smarter about timemachine shares and SMB2/3 extensions

undsoft:
**Changes:**

Improves user experience around Timemachine shares.

Снимок экрана 2025-07-08 в 11 40 38


RehanY147 (approved):
Works as intended. But it doesn't capture the user's attention as it should. A modal stops their flow of inputs and demands an action. This can go unnoticed. Maybe the button should be a warning color or text or maybe a modal.

undsoft:
If user ignores the warning and submits, they'll be greeted with a middleware validation error.

PR #12241, opened 2025-07-09
NAS-136521: Return asterisk to required selects

undsoft:
**Changes:**

I had previously removed asterisk from selects with `[required]="true"` because it seemed redundant.
While this still may be true, having a required select without an asterisk is confusing for `[multiple]="true"`, hence I'm bringing it back.

Also adds functionality where empty required singular select with auto-select first loaded option.

**Testing**
Check if auto-selection of first option breaks something.


undsoft:
time 0:45

RehanY147 (approved):
Looks good

undsoft:
Auto-selection creates problems, so I'm reverting it.

PR #12247, opened 2025-07-11
NAS-136597: generates TNC claim token when `Get Connected` is clicked

gmboknat:
**Changes:**
'tn_connect.generate_claim_token' will only be called when the token does not exist and `Get Connected` is clicked.

**Testing:**
1. enable the TNC service
2. click TNC service icon in the toolbar
3. Click `Get Connected` and it will open a new tab that navigates to the TNC system registration page

### Downstream

|Affects         |Reasoning
|Documentation   |


aervin (approved):
Thanks @gmboknat

PR #12260, opened 2025-07-16
NAS-136586: New Users – form bugs (home directory)

AlexKarpov98:
Testing: see ticket.


src/app/modules/forms/editable/services/editable.service.ts
```diff
@@ -74,10 +74,14 @@ export class EditableService implements OnDestroy {
...
-    }
+    if (event.key !== 'Escape' || !this.hasOpenEditables) return;
+
+    // Defer the decision to end of event loop
+    setTimeout(() => {
+      if (!event.defaultPrevented) {
+        this.tryToCloseAll();
+      }
+    });
   }
```
AlexKarpov98: Fixed an issue I found: 
When tapping  `Escape` - editable closed first while the modal was opened and on the second `Escape` tap - Modal was closed - that was wrong.

Now it works as expected - tapping `Escape` only closes the modal first and editable is still opened. On the second `Escape` tap - the editable will be closed as expected.

denysbutenko (approved):

PR #12264, opened 2025-07-16
NAS-136133 / 25.10 / New directory services implementation catching up with new API

william-gr:


william-gr:
@RehanY147 please review this as soon as you can. Not perfect, I am sure there are edge cases but It does work. Tested fully joining directory services

PS: I will generate i18n files after review is done

william-gr:
Changing types is not a common workflow and will hardly ever happen in the field.
Thanks for the review

PR #12265, opened 2025-07-17
NAS-136471 / 25.10 / Adds check for HA to edit network interfaces

RehanY147:
**Changes:**

Adds check for HA to edit network interfaces

**Testing:**

On an HA system, if HA is enabled you should not be able to add, edit, delete or reset network interfaces and vise versa.


william-gr:
@RehanY147 can you remove translations, please?

william-gr (changes requested):
Also after changing translations

image
Can you remove that refresh icon, and use "Changing" instead of "Resetting"

william-gr (approved):

PR #12269, opened 2025-07-17
NAS-136615 / 25.10 / Fixes condition for Containers navigation menu item

RehanY147:
**Changes:**

Removes condition for Containers navigation menu item

**Testing:**

Confirm that you can see the Containers navigation menu item in both community and enterprise editions.


william-gr (approved):

william-gr:
Merging myself since we need it.

PR #12277, opened 2025-07-18
NAS-134499 / 25.10 / Update messaging for admin on STIG

denysbutenko:
**Changes:**

Update messaging for admin on STIG

**Testing:**

See ticket.


william-gr (changes requested):
Per the ticket comments, we are not supposed to make changes to the login page to warn about STIG.

Also please remove all the i18n in this PR.

PR #12278, opened 2025-07-18
First pass at implementing failover checks after login

william-gr:
**Changes:**

Implement several failover checks after a login to make sure the node is in a good enough state to allow a login.

**Testing:**

- Login when one node is master
- Login when a failover event is in progress
- Login on the standby node

### Downstream

|Affects         |Reasoning
|Documentation   |


RehanY147 (approved):
Looks good. Login is working in different conditions.

PR #12286, opened 2025-07-22
NAS-136649 / 25.10 / On Goldeye creating a zvol makes the datasets UI act weird.

AlexKarpov98:
Testing: see ticket.

### DIFF
--- a/src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
+++ b/src/app/pages/datasets/components/dataset-details-card/dataset-details-card.component.html
@@ -63,16 +63,18 @@
       <div class="label">{{ 'ZFS Deduplication' | translate }}:</div>
       <div class="value">{{ dataset().deduplication.value | orNotAvailable }}</div>
     </div>
-    <div class="details-item">
-      <div class="label">{{ 'Case Sensitivity' | translate }}:</div>
-      <div class="value">
-        {{ (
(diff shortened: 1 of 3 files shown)


william-gr:
Why is this changed needed? API is not returning case sensitivity?

AlexKarpov98:
> Why is this changed needed? API is not returning case sensitivity?

I assume it's not included in the api response for ZVOLs

AlexKarpov98:
> In that case whole case sensitivity item should be inside `if (!isZvol()) {` ?

We can do that 👌

william-gr:
@AlexKarpov98 did you test this? it seems to have a wrong div tag

AlexKarpov98:
> @AlexKarpov98 did you test this? it seems to have a wrong div tag

ofc, it's all right there, don't forget it's wrapped into 
```@if (!isZvol()) {```

william-gr:
@AlexKarpov98 look again, your eyes are tricking you : )  div opens 3 times within the if, and closes 4 times

AlexKarpov98:
@william-gr are you sure? I don't see any issue 😄 
Screenshot 2025-07-23 at 15 39 35

I've even involved an AI here 😄 
Screenshot 2025-07-23 at 15 44 41

william-gr:
@AlexKarpov98 I guess I am blind then :) Apologies!

PR #12298, opened 2025-07-24
NAS-136845 / 25.10 / Check for failover license as well for ha enabled check

RehanY147:
**Changes:**

Check for failover license as well for ha enabled check

**Testing:**

Non-HA system shouldn't warn the user about making changes to network interfaces. They should appear on HA systems when HA is enabled

### DIFF
--- a/src/app/services/network.service.ts
+++ b/src/app/services/network.service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@angular/core';
 import { combineLatest, Observable } from 'rxjs';
 import { filter, map } from 'rxjs/operators';
+import { FailoverStatus } from 'app/enums/failover-status.enum';
 import { Choices } from 'app/interfaces/choices.interface';
 import { Option } from 'app/interfaces/option.interface';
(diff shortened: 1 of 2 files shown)


src/app/services/network.service.ts
```diff
@@ -72,8 +72,8 @@ export class NetworkService {
   }
 
   getIsHaEnabled(): Observable<boolean> {
-    return this.api.call('failover.config').pipe(
-      map((config) => !config.disabled),
+    return combineLatest([this.api.call('failover.config'), this.api.call('failover.licensed')]).pipe(
```
william-gr: Here is the middleware logic:

```
    async def check_failover_disabled(self, schema, verrors):
        if not await self.middleware.call('failover.licensed'):
            return
        elif await self.middleware.call('failover.status') == 'SINGLE':
            return
        elif not (await self.middleware.call('failover.config'))['disabled']:
            verrors.add(schema, 'Failover must be disabled.')
```

Please mimic it here.

william-gr (changes requested):

PR #12300, opened 2025-07-24
NAS-136588: Add sudo commands fields

AlexKarpov98:
Testing: see ticket.

Preview:

https://github.com/user-attachments/assets/5e42b874-f41b-4096-b495-73e4e0b4d85f

### DIFF
--- a/src/app/modules/forms/editable/editable.component.scss
+++ b/src/app/modules/forms/editable/editable.component.scss
@@ -8,6 +8,7 @@
   font-size: 13px;
   padding: 0 5px;
   vertical-align: top;
+  word-break: break-word;
 
   &.readonly {
     background-color: transparent;
--- a/src/app/pages/credentials/new-users/all-users/user-details/user-access-card/user-access-card.component.html
+++ b/src/app/pages/credentials/new-users/all-users/user-details/user-access-card/user-access-card.component.html
@@ -80,7 +80,6 @@
           </span>
         </div>
 
-
         @if (user.api_keys?.length) {
          <a
             routerLink="/credentials/users/api-keys"
@@ -99,17 +98,6 @@
             {{ 'Add API Key' | translate }}
           </a>
         }
-      </div>
-
-      <div class="content-wrapper" [class.not-available]="noShellAccess()">
-        <ix-icon name="mdi-powershell"></ix-icon>
-        <span class="label">
-          @if (noShellAccess()) {
-            {{ 'No Shell Access' | translate }}
-          } @else {
-            {{ 'Shell Access' | translate }}: {{ user.shell }}
-          }
-        </span>
       </div>
 
       <div class="content-wrapper" [class.not-available]="!sshAccessStatus()">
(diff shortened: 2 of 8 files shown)


RehanY147 (changes requested):
I am getting this validation message on the old user form but not on the new one when I put in the sudo command "sudo this and that". And I think I'm supposed to. 

image

RehanY147 (changes requested):
Also, the chips input is not working. Adding a comma doesn't do anything.

AlexKarpov98:
> See the demonstration of both issues https://github.com/user-attachments/assets/34c7359f-0603-46ec-ac98-1730e3a7a990

Hey, thanks for the feedback.
I think You missed that it's a CHIPs input, you need to add chips. 
You may check the preview to see if data was actually applied.
 
Screenshot 2025-07-25 at 09 50 53

I think it's getting saved without an issue because it somehow does not save sudo commands. 
The form does not fit save them. 
But I can't reproduce it on my side, it all works as needed. 
If you compare both forms - on the old form - it's as well CHIPs input and comma does not do anything because you need to save with ENTER on Blur event.

But I can't reproduce what you showed, it works fine for me, see:

1. 

https://github.com/user-attachments/assets/0140899f-7cce-4bb0-9c91-44042c8f7dcb

2. It's chips:

https://github.com/user-attachments/assets/a2332628-a5c0-4183-9a9f-907d02e05796

Maybe pull latest changes and try again 🤔

william-gr:
I wonder if you need some validation to make sure each sudo comamnd starts with '/' ? or is that already there? I cant find it

AlexKarpov98:
@william-gr 

on UI side there are no validation but it's there once you submit the form.
We can make sure we do that essential validation:
> some validation to make sure each sudo comamnd starts with '/'

**on UI as well**

RehanY147 (changes requested):
It works. You are right. But it caused confusion for me. It will 100% cause confusion for layman user. Clicking outside the editable box causes the editable box to close. Giving the impression that the entered value was saved if you don't tap enter after entering a value in the ix-chips

AlexKarpov98:
> It works. You are right. But it caused confusion for me. It will 100% cause confusion for layman user. Clicking outside the editable box causes the editable box to close. Giving the impression that the entered value was saved if you don't tap enter after entering a value in the ix-chips

Okay, then we need to make sure that we save chips on blur as well. Thanks.

william-gr:
> @william-gr
> 
> on UI side there are no validation but it's there once you submit the form. We can make sure we do that essential validation:
> 
> > some validation to make sure each sudo comamnd starts with '/'
> 
> **on UI as well**

For a better user experience its best to avoid errors and warn users on the spot. If UI accepts an invalid value that can be easily identified prior to submission then we failed the task.

AlexKarpov98:
@william-gr @RehanY147 changes arrived: 🙌 

https://github.com/user-attachments/assets/794f710f-2708-40ca-9235-ae30f3ab9f35

william-gr:
Great, thanks!

william-gr (approved):

RehanY147 (approved):

PR #12306, opened 2025-07-24
NAS-136760 / 25.10 / Adds the `primary_kdc` field to Kerberos Realms form

RehanY147:
**Changes:**

Adds the `primary_kdc` field to Kerberos Realms form

**Testing:**

Check that the field is saved on create and on edit


src/app/pages/directory-service/components/kerberos-realms-form/kerberos-realms-form.component.ts
```diff
@@ -55,6 +55,7 @@ export class KerberosRealmsFormComponent implements OnInit {
   form = this.fb.group({
     realm: ['', Validators.required],
     kdc: [[] as string[]],
+    primary_kdc: [null as string, Validators.required],
```
william-gr: This does not seem to be required on the backend/API, is it?

PR #12313, opened 2025-07-25
NAS-136844 / 25.10 / Add product_id to iSCSI extents

RehanY147:
**Changes:**

Add product_id to iSCSI extents

**Testing:**

See ticket


src/app/pages/sharing/iscsi/iscsi-wizard/iscsi-wizard.component.ts
```diff
@@ -119,6 +119,7 @@ export class IscsiWizardComponent implements OnInit {
       dataset: ['', [Validators.required]],
       volsize: new FormControl(null as number | null, [Validators.required]),
       usefor: [IscsiExtentUsefor.Vmware, [Validators.required]],
+      product_id: [''],
```
william-gr, line 122: Is this field displayed in the wizard? If so, we can remove it.

william-gr:
image
I think you need to send null to the API, or not send anything at all when product id is not entered.

PR #12314, opened 2025-07-25
NAS-136863 / 25.10 / handle sign in case when failover status is SINGLE

william-gr:
**Changes:**

Handle sign in after license is applied for HA.

**Testing:**

failover.licensed is True and failover.status is SINGLE
Happens on fresh install and after license is applied.

### Downstream

|Affects         |Reasoning
|Documentation   |


denysbutenko (approved):
👍🏻

PR #12320, opened 2025-07-27
NAS-136844 / 25.10 / Add product_id

william-gr:
**Changes:**

Add product_id to extent and wizard form for iSCSI.

**Testing:**

Add a new extent with Product ID and verify it shows in the initiator side.

### Downstream

|Affects         |Reasoning
|Documentation   | New field in Extent form
|Testing      | New Product ID Field, its optional, if used it will show a different identifier in the initiator


src/app/pages/sharing/iscsi/iscsi-wizard/iscsi-wizard.component.ts
```diff
@@ -185,6 +186,7 @@ export class IscsiWizardComponent implements OnInit {
       insecure_tpc: true,
       xen: value.usefor === IscsiExtentUsefor.Xen,
       rpm: IscsiExtentRpm.Ssd,
+      product_id: value.product_id === '' ? null : value.product_id,
```
gmboknat, line 189: can be simplified to `value.product_id || null`

william-gr: The first option is better: value.product_id === '' ? null : value.product_id
Here's why:
The first approach is explicit and predictable - it only converts empty strings to null, leaving all other values unchanged (including 0, false, undefined, etc.).
The second approach (value.product_id || null) is problematic because it converts ALL falsy values to null, not just empty strings. This means:

0 becomes null
false becomes null
undefined becomes null
NaN becomes null
Empty string '' becomes null

If product_id could legitimately be 0 (like an ID of 0), the second approach would incorrectly convert it to null.

gmboknat: It makes sense. Also, even if you enter `0` in the input field, it wont be converted to `null` since the form control is initialized as string and not a number.

PR #12321, opened 2025-07-27
NAS-134499 / 25.10 / Update messaging for admin on STIG

william-gr:
**Changes:**

Add a warning if there are users without 2FA configured when enabling STIG.

**Testing:**

Have admin users without 2FA and enable STIG.

### Downstream

|Affects         |Reasoning
|Documentation   |Warn if there are users with roles and without 2FA
|Testing         |Enabling STIG has new potential dialog

### DIFF
--- a/src/app/pages/system/advanced/system-security/system-security-form/system-security-form.component.ts
+++ b/src/app/pages/system/advanced/system-security/system-security-form/system-security-form.component.ts
@@ -17,7 +17,7 @@ import { MatHint } from '@angular/material/form-field';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { TranslateService, TranslateModule } from '@ngx-translate/core';
 import {
-  filter, forkJoin, map, of,
+  filter, map, of,
 } from 'rxjs';
 import { stigPasswordRequirements } from 'app/constants/stig-password-requirements.constants';
 import { RequiresRolesDirective } from 'app/directives/requires-roles/requires-roles.directive';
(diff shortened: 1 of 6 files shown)


src/app/pages/system/advanced/system-security/system-security-form/system-security-form.component.ts
```diff
@@ -187,7 +191,65 @@ export class SystemSecurityFormComponent implements OnInit {
...
+  }
+
+  private checkUsersWithoutTwoFactorAuth(callback: () => void): void {
+    forkJoin({
+      users: this.api.call('user.query', [[
+        ['OR', [['builtin', '=', false], ['username', '=', 'root']]],
+        ['twofactor_auth_configured', '=', false],
+        ['locked', '=', false],
+        ['password_disabled', '=', false],
+        ['roles', '!=', []],
```
william-gr: @anodos325 can you let me know if this query is sane? I wonder if first filter should be removed? Probably so

src/app/pages/system/advanced/system-security/system-security-form/system-security-form.component.ts
```diff
@@ -112,6 +121,19 @@ export class SystemSecurityFormComponent implements OnInit {
   ngOnInit(): void {
     if (this.systemSecurityConfig()) {
       this.initSystemSecurityForm();
+      // Check global 2FA status if STIG is already enabled
+      if (this.systemSecurityConfig().enable_gpos_stig) {
+        this.api.call('auth.twofactor.config').pipe(
+          untilDestroyed(this),
+        ).subscribe((twoFactorConfig) => {
+          const enabled = twoFactorConfig?.enabled || false;
```
aervin, line 129: Can middleware actually return undefined/null here? Seems weird

aervin (approved):

src/app/pages/system/advanced/system-security/system-security-form/system-security-form.component.ts
william-gr: In theory no, but I guess it doesnt hurt to have a fallback

PR #12325, opened 2025-07-28
NAS-136811 / 25.10 / add cpu and proxy port validation to container creation

william-gr:
**Changes:**

add cpu and proxy port validation to container creation

**Testing:**

Try cpu "-1" and port 70000
Should give an error

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


gmboknat (approved):

gmboknat:
I can input `0` for the cpu and submit the form, but I receive an API response error stating that cpu value is invalid. It would be good to handle this type of error in the UI.

william-gr:
@gmboknat You're right, don't believe it should, good catch.

PR #12326, opened 2025-07-29
NAS-136879: Snackbar is overlapped by the backdrop

AlexKarpov98:
Testing: see ticket.

Preview:

https://github.com/user-attachments/assets/0fcec9b0-0922-45aa-8d63-3a74ca12da99

### DIFF
--- a/src/app/modules/snackbar/services/snackbar.service.ts
+++ b/src/app/modules/snackbar/services/snackbar.service.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { MatSnackBar, MatSnackBarRef } from '@angular/material/snack-bar';
+import { MatSnackBar, MatSnackBarConfig, MatSnackBarRef } from '@angular/material/snack-bar';
 import { iconMarker } from 'app/modules/ix-icon/icon-marker.util';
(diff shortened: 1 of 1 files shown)


william-gr:
Would it not make more sense to have the snackbar show below the backdrop? Or why did we prefer to manipulate dom directly to change z index?

william-gr:
Oh nevermind, I misunderstood the original issue. Still, it seems manipulating DOM is not a great approach, is there not a better alternative you think?

AlexKarpov98:
> Oh nevermind, I misunderstood the original issue. Still, it seems manipulating DOM is not a great approach, is there not a better alternative you think?

I haven't found any alternative. That's the only working solution so far. That's not a big issue actually.

william-gr:
@AlexKarpov98 Claude found a solution with CSS. Pushed to your branch. Let me know if you think thats better or not. You dont have to use it.

AlexKarpov98:
> @AlexKarpov98 Claude found a solution with CSS. Pushed to your branch. Let me know if you think thats better or not. You dont have to use it.

@william-gr 
That actually works, I tried it before, it was not working with CSS. We can use that solution indeed.

william-gr (approved):

PR #12327, opened 2025-07-29
NAS-136857 / 25.04.3 / Traceback Setting up External Sshare on 25.04.2 to a Share on 25.10

AlexKarpov98:
Testing: see ticket.
It's a fix for a Fangtooth version.

https://github.com/user-attachments/assets/5609c9cc-a3a0-4f36-bc21-4e83e3f5049a

### DIFF
--- a/src/app/pages/sharing/smb/smb-form/smb-form.component.spec.ts
+++ b/src/app/pages/sharing/smb/smb-form/smb-form.component.spec.ts
@@ -465,16 +465,6 @@ describe('SmbFormComponent', () => {
       )).getValue();
 
       expect(api.call).toHaveBeenCalledWith('filesystem.stat', [sharePath]);
-
-      const homeShare = await (await loader.getHarness(
-        IxCheckboxHarness.with({ label: formLabels.home }),
-      )).getValue();
-
-      expect(spectator.inject(Router).navigate)
-        .toHaveBeenCalledWith(
-          ['/', 'datasets', 'acl', 'edit'],
-          { queryParams: { homeShare, path: sharePath } },
-        );
     });
 
     it('should submit the form with the correct value and check service action is dispatched', async () => {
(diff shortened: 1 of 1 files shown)


william-gr:
Does this not happen on Goldeye?

AlexKarpov98:
> Does this not happen on Goldeye? @william-gr 

yes, but it's a little different there, on Goldeye there are no UI calls to `filesystem.listdir` or `filesystem.stat` as I see, but still same error modal appears on submit. (when submitting as Default Purpose). 
I will discuss that with @anodos325

src/app/interfaces/smb-share.interface.ts
```diff
@@ -45,6 +45,8 @@ export enum SmbPresetType {
   SmbWorm = 'WORM_DROPBOX',
 }
 
+export const externalSmbSharePath = 'EXTERNAL';
```
william-gr, line 48: It should be `EXTERNAL:`, i believe

src/app/pages/sharing/smb/smb-form/smb-form.component.ts
```diff
@@ -585,13 +586,13 @@ export class SmbFormComponent implements OnInit, AfterViewInit {
...
-    const datasetId = sharePath.replace('/mnt/', '');
+
+    if (!sharePath || sharePath.startsWith(externalSmbSharePath)) {
+      return of(false);
+    }
+
     return this.api.call('filesystem.stat', [sharePath]).pipe(
-      switchMap((stat) => {
-        return of(
-          stat.acl !== this.form.controls.acl.value && datasetId.includes('/'),
```
william-gr, line 592: This seems to be a change of behavior. Can you elaborate a little? What is this.form.controls.acl.value ? Is creating/editing a share to root dataset no longer possible?

AlexKarpov98: @william-gr 
I talked with Andrew:
```I think we should redirect to ACL edit if filesystem.stat shows acl False and the purpose is not EXTERNAL_SHARE.```

But I can revert it if you think it's incorrect.

src/app/interfaces/smb-share.interface.ts
AlexKarpov98: > It should be `EXTERNAL:`, i believe

I only use it as `startsWith()` - so it can stay `EXTERNAL`

william-gr (approved):

PR #12334, opened 2025-07-30
NAS-136889: iSCSI target screen does not show iSCSI Group information

AlexKarpov98:
**Changes:**
Added iSCSI groups card.
Messages are now standardized to have a `.` in the end, to be all similar.

**Testing:**
see ticket.

https://github.com/user-attachments/assets/210e5528-cee4-4049-814d-313550d2639c

### Downstream

|Affects         |Reasoning
|Documentation   | Added iSCSI groups card 
|Testing         |


william-gr:
The visual aspect of each group is a bit hard to read.
Can you try some iterations to make it more visually appealing and easier to read?
One idea is having the group as a title/header and each attribute properly spaced.

@dszidi any suggestions?

AlexKarpov98:
@william-gr @dszidi - how about?

Screenshot 2025-07-31 at 14 58 13

william-gr:
@AlexKarpov98 much better! I just dont know if the ID number for portal and initiator are meaningful to the user, maybe the names only. We need to verify if they are useful or not.

AlexKarpov98:
> @AlexKarpov98 much better! I just dont know if the ID number for portal and initiator are meaningful to the user, maybe the names only. We need to verify if they are useful or not.

How could we verify it? I see that in Edit mode we show ID's as well BTW 🙌

william-gr:
@AlexKarpov98 yeah, a problem for another day, would require API changes which we dont have time.

william-gr (approved):
Lets give @dszidi a chance to review then we can merge.

dszidi:
image

These blue headers look out of place and almost look like the tabbed navigation. Normal subheader font with normal color would probably be more appropriate. 


image

dszidi (approved):
The rest LGTM

AlexKarpov98:
Screenshot 2025-08-01 at 12 30 05



@dszidi - I think it should be even more compact & readable, what do you think? 
We have the same for other cards, so I believe it should work 

Screenshot 2025-08-01 at 11 54 03

william-gr:
Agreed, we can go with that for BETA, feel free to merge

Thanks!

PR #12337, opened 2025-07-30
Enclosure Mocking in Debug Panel

william-gr:
**Changes:**

Move enclosure mocking from the environment to the debug panel.

**Testing:**

Open debug panel, enable enclosure mocking and reload UI so the Enclosure menu will show up under System.

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


william-gr:
@aervin should be ready for review now, sorry, not a fan how github automatically asks for review before its ready =P

PR #12343, opened 2025-07-31
NAS-136761 / 25.10 / Disable function not exposed in the UI in the NVMe Namespace widget.

AlexKarpov98:
Testing: see ticket.

Preview:

https://github.com/user-attachments/assets/a5b93400-125a-442c-8c42-4e8c6147fdeb


william-gr:
@AlexKarpov98 that seems like a terrible UX. I wonder why one would care to disable it to edit it again to make it behave like a new namespace? Seems easier to not allow edit and only do delete / add?

EDIT: I am also consulting with Brian

AlexKarpov98:
> @AlexKarpov98 that seems like a terrible UX. I wonder why one would care to disable it to edit it again to make it behave like a new namespace? Seems easier to not allow edit and only do delete / add?
> 
> EDIT: I am also consulting with Brian

That's the possible logic allowed by API. I agree that it's not the best way of handling it. But I tried to follow the ticket logic. On the other hand - just disabling editing is as well not the best UX for the user. It requires extra steps for - deleting, creating new in that case.

william-gr:
@AlexKarpov98 lets remove Edit ability for namespaces for now. Only add/delete.

AlexKarpov98:
Updated code/pr: https://github.com/truenas/webui/pull/12349 (to keep this version as well)

PR #12348, opened 2025-07-31
NAS-136928 / 25.10 / Use `ix-code-editor` instead of `ix-textarea` for mockconfig form

AlexKarpov98:
Testing: see ticket.

Preview:

image


william-gr (approved):
commit / ticket title is wrong

thanks!

PR #12349, opened 2025-07-31
NAS-136761: Disable function not exposed in the UI in the NVMe Namespace widget.

AlexKarpov98:
Final decision: 
```lets remove Edit ability for namespaces for now. Only add/delete.```


william-gr (commented):
Should we not remove onEditNamespace and any related component/etc?

AlexKarpov98:
> Should we not remove onEditNamespace and any related component/etc?

yup, I forgot that part 😛

PR #12356, opened 2025-08-01
NAS-136953 / 25.10 / Fix reporting data immutability

william-gr:
**Changes:**

**Testing:**

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


william-gr:
Thanks, some improvements can be done here but we can do it after BETA ;)

PR #12362, opened 2025-08-04
NAS-136988 / 26.04 / Fix and improve External Share for Goldeye

AlexKarpov98:
Testing: see ticket.

We show purpose first.
But still if user types some external path - we switch to `External` Purpose.

Preview:

https://github.com/user-attachments/assets/a1f0800b-89b4-43ce-8b74-eb7be586cd1a

### DIFF
--- a/src/app/pages/sharing/components/shares-dashboard/smb-card/smb-card.component.ts
+++ b/src/app/pages/sharing/components/shares-dashboard/smb-card/smb-card.component.ts
@@ -1,5 +1,7 @@
 import { AsyncPipe } from '@angular/common';
-import { ChangeDetectionStrategy, Component, OnInit, inject } from '@angular/core';
+import {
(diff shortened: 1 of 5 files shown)


william-gr:
image

when adding External, its weird to allow multiple, is it supported?

image
on the table it only shows EXTERNAL, that is not very helpful.

william-gr:
Well, looking at the API it seems multiple are allowed, so nevermind on that one.

EDIT: just make sure to reflect that on the label. Remote Paths, plural

william-gr:
When changing from External share to some other there is a warning
image

That does not seem particularly relevant as its not applied for External shares.

AlexKarpov98:
@william-gr - could you check it once again? 🙌 
BTW - should we show that warning if user changes `DEFAULT` purpose to `EXTERNAL` upon edit? (currently we show)

william-gr (approved):
Looks good now, thanks!

PR #12363, opened 2025-08-04
NAS-136993 / 26.04 / Add method name filter to WebSocket debug panel

william-gr:
- Add filter input field in message list header using ix-input component
- Implement real-time filtering of messages by method name (case-insensitive)
- Position filter on the left, auto-scroll and clear button on the right
- Remove label and minimize padding on filter input for compact design
- Update component to use filtered messages array for display

**Changes:**

**Testing:**

Type in a filter and make sure only method names that includes the input are shown

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |

### DIFF
--- a/src/app/modules/websocket-debug-panel/components/message-list/message-list.component.spec.ts
+++ b/src/app/modules/websocket-debug-panel/components/message-list/message-list.component.spec.ts
@@ -280,5 +280,81 @@ describe('MessageListComponent', () => {
     });
   });
 
+  describe('filter functionality', () => {
+    beforeEach(() => {
+      // Initialize component with messages
+      spectator.component.ngAfterViewInit();
+      spectator.detectChanges();
+    });
+
+    it('should initialize with empty filter text', () => {
+      expect(spectator.component['filterText']).toBe('');
+    });
+
+    it('should show all messages when filter is empty', () => {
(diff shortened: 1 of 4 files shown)


src/app/modules/websocket-debug-panel/components/message-list/message-list.component.spec.ts
```diff
@@ -280,5 +280,81 @@ describe('MessageListComponent', () => {
     });
   });
 
+  describe('filter functionality', () => {
```
aervin, line 283: Much of the codebase takes a different approach to unit tests. User DOM interactions are mocked and behavior is verified via DOM queries.

aervin (approved):
I didn't notice any performance issues while filtering long message lists. Functionally it works well.

src/app/modules/websocket-debug-panel/components/message-list/message-list.component.spec.ts
william-gr: Which do you think is best?

aervin: I think it makes sense for tests to verify what the user "sees" with DOM queries and harnesses.

PR #12365, opened 2025-08-04
NAS-136997 / 26.04 / Only check failover reasons when HA is licensed

william-gr:
**Changes:**

The failover.disabled.reasons API should only be called when the system has an HA license. This change ensures the WebSocket subscription and status checks are conditional on HA being licensed, preventing unnecessary API calls on non-HA systems.

**Testing:**

Check failover.disabled.reasons is not called on login.

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


AlexKarpov98 (approved):
Looks good . Would be great to add some tests as well.

william-gr:
> Looks good . Would be great to add some tests as well.

Tests added

AlexKarpov98 (approved):
Thanks, _Claude_ 😄

PR #12368, opened 2025-08-05
NAS-136518 / 26.04 / Allow admins to clear 2FA settings for other users

AlexKarpov98:
**Changes:**
Added new button to clear 2FA setup for the user (for new users and old users pages)
API does not allow to clear 2FA if it's globally enabled, first disable it globally - clear 2FA for the user.

**Testing:**
See ticket.

https://github.com/user-attachments/assets/4e6e749d-c361-4773-86f9-e95261483b02

### Downstream

|Affects         |Reasoning
|Documentation   | Added new button to clear 2FA for user
|Testing         |


aervin (approved):
Code looks good. @AlexKarpov98 don't worry about fixes and features for the old Users page moving forward.

william-gr:
Why Global 2FA needs to be disabled to clear 2FA for a user? That does not seem to make much sense.

william-gr (changes requested):
Just for reference: https://github.com/truenas/middleware/pull/16893

AlexKarpov98:
@william-gr updates arrived, no more global auth enabled check 👌

src/assets/icons/sprite-config.json
```diff
@@ -1,3 +1,3 @@
 {
-  "iconUrl": "assets/icons/sprite.svg?v=4422b7cd42"
+  "iconUrl": "assets/icons/sprite.svg?v=728e610251"
```
william-gr, line 2: Curious why this changed

william-gr (approved):

src/assets/icons/sprite-config.json
AlexKarpov98: @william-gr - `mdi-cellphone-off` was used and added to the sprite.svg - so it's updated. `v={ID}` is the cached new version.
So - we only update `v={ID}`  once anything changed to the sprite.svg.

PR #12373, opened 2025-08-05
NAS-137031 / 26.04 / Do not have debug panel overlay the app when refreshing with it opened

william-gr:
**Changes:**

Do not allow debug panel to overlay app.

**Testing:**

Have debug panel opened then hit F5. It should not overlay rest of page.

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


william-gr:
@AlexKarpov98 I was addressing claude review, its ready now :P

AlexKarpov98 (approved):
Works as described 👍

PR #12376, opened 2025-08-06
Update ja.json

ReiKirishima:
**Changes:**

Translation more.


ReiKirishima:
Reviewed the words that keeping in their original notation

PR #12378, opened 2025-08-06
Fix missing translation marker

MDVAurelien:
Fix missing translation marker in all-instances.component.ts. I think that can fix it.

**Changes:**

**Testing:**

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


AlexKarpov98 (approved):
Merci.

PR #12379, opened 2025-08-06
NAS-117171: webui makes it too easy to try to set an invalid POSIX1E ACL

AlexKarpov98:
**Changes:**
Added `Copy Access to Default` & `Add Missing MASK Entries` buttons for `AclType.Posix1e` ACL Types.

Improvements:
* Users should not be able to remove USER_OBJ, GROUP_OBJ, OTHER entries. This also is true of case where default entries are specified
* Users should have easy way to replace default entries with what’s specified in access list (convenience feature)
* If user specifies USER or GROUP entries, a MASK entry should automatically be added to the relevant list so that user has ability to review it and set appropriately. 

**Testing:**
See ticket.

### Downstream

|Affects         |Reasoning
|Documentation   | Improved experience for POSIX ACLs

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,8 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "Bash(rm:*)"
-    ],
-    "deny": []
-  }
-}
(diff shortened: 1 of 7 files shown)


AlexKarpov98:
Creating a new draft for ACL editor - https://ixsystems.atlassian.net/browse/NAS-137108

PR #12381, opened 2025-08-07
NAS-136973 / 26.04 / Virtual Machines page doesnt refresh properly

AlexKarpov98:
**Changes:**
- Fixed slide in buttons state, previously it showed incorrect status if I haven't submitted an action
- Added `Suspended` state support. `Power Off & Resume` buttons are the only ones available for `Suspended` state as described.
- Refactored refresh state logic to depend on the `vm.query` subscription - to get all updates.

 

**Testing:**
See ticket.

### Downstream

|Affects         |Reasoning
|Documentation   | Suspended state support added


dszidi:
Tooltip is broken. I hover over the slide toggle button in the Running column and the tooltip is way up on the top left corner of the screen

 
> image

AlexKarpov98:
@dszidi - not sure how to reproduce it. What I see is:

https://github.com/user-attachments/assets/7e96aae2-5cf8-4175-ad62-ee7651e89360

dszidi (approved):
Ok. The rest LGTM. If I encounter that behavior again I will create a new ticket

PR #12388, opened 2025-08-08
NAS-136969: Can't save a new replication task (infinite loading & 403 sentry)

AlexKarpov98:
**Changes:**
- Fixed error handling. The bug was that the error handling was not working and didn't show the error modal.
- But the error which shows up says that Naming Schema is not available for Local Source with PUSH strategy, details below:
```
I’ve noticed 1 more issue - for Replication Wizard -> Local Source -> [x] Replicate Custom Snapshots -->  Naming Schema is shown and it creates API errors when I submit a form with it.
I tested and seems like only Snapshot Name Regular Expression is valid for above scenario.
For Remote Source - both work.
```
So now we are hiding non acceptable field (by API) on UI when appropriate 

image

- Added custom validators for `Naming Schema` and `Snapshot Name Regular Expression`
- Improved accessibility
- Added & Fixed tests

**Testing:**
See ticket, last video, now you should be able to create Replication Task without an error and infinite loader.

### Downstream

|Affects         |Reasoning
|Documentation   | Removed `Naming Schema` for Local source & PUSH strategy (in Replication Wizard)


william-gr:
image

I must be doing something wrong. Wasnt this supposed to be forbidden?

AlexKarpov98:
@william-gr for LOCAL & PUSH - regular expression is the only one valid option, it's correct 👌 
Could you provide a bit more context? 🙌 

see the difference below:

Local:

https://github.com/user-attachments/assets/97b4177d-c30e-4890-a22f-f2e5d089a46b

Remote:

https://github.com/user-attachments/assets/c9cfcc3d-aca4-4fa5-9195-b06e59f097b4

AlexKarpov98:
backport.

PR #12421, opened 2025-08-15
NAS-136335: Refactor ix-input to ix-field-input and disable material autocomplete in sign-in form

pgilad:
## Summary
  - Refactored ix-input component to a new ix-field-input component with proper autocomplete handling
  - Updated signin form to use the new ix-field-input component
  - Added comprehensive unit tests and testing harness for the new component

  ## Why this fixes 1Password integration
  The Angular Material input was adding `role="combobox"` and `aria-autocomplete="list"` attributes to the input fields. These attributes cause 1Password and
  other password managers to not recognize the fields for autocomplete functionality, as they interpret these attributes as a different input type.

  By creating a new ix-field-input component that properly handles autocomplete attributes without triggering Material's built-in autocomplete behavior, we
  ensure password managers can correctly identify and fill login fields.

  A previous attempt in PR #12395 tried to address password manager recognition by modifying the `getType()` method to maintain the `type="password"` attribute
   while showing/hiding passwords. However, that approach didn't address the root cause of the issue - the Material autocomplete attributes interfering with
  password manager detection.

  ## Test plan
  - Verify that the signin form works correctly with the new component
  - Check that autocomplete behavior works as expected with 1Password and other password managers
  - Confirm that password fields can be auto-filled by password managers
  - Run unit tests for the new ix-field-input component


AlexKarpov98 (commented):
Thanks for the update. Would be good if we could update existing component instead of creating a brand new one just for this use case.

pgilad:
Totally agree however that would significantly complicate the existing component logic (html + js) so I thought it would be simpler to create a new component type that is without the dropdown auto-complete

dszidi (commented):
I agree with Alex. If there is a problem with the ix-input component then it should be addressed there

william-gr:
@pgilad Thanks for the contribution. Are you willing to try and fix this inside the ix-input component instead?

pgilad:
Yeah I'll try to get it handled over the next couple of days

william-gr:
Closing this PR now for house keeping purposes. We can reopen once we have the changes. Thank you!

pgilad:
Hi sorry, didn't have time to get to it, feel free to implement the required changes, in the meantime the angular material input is blocking password managers (like 1Password) from auto-filling login details 😢

AlexKarpov98:
@pgilad @william-gr - I tried on the master branch with `1Password` Chrome Ext., this is how it works for me, seems fine:


https://github.com/user-attachments/assets/5d07b876-2d75-4416-9079-c2b398ec764d

@pgilad - could you provide some steps or tools which I should check?

pgilad:
Can you try w/o the extension ? I'm using the native OSX 1Password and it doesn't autocomplete. On my branch I created a local dummy file to preview the login w/o the angular material dropdown and then it works.

AlexKarpov98:
@pgilad I'll try, thanks.

AlexKarpov98:
@pgilad - could you show me how it works for you?
How do you see/use the password manager, I don't see it on any login form (any web app)

pgilad:
Not sure I understand - are you talking about how I use the password manager (1Password Native OSX) or how I setup a local reproduction html to test out the change ?

AlexKarpov98:
@pgilad 

could you show me how it suggest autocomplete values to you? For me it's not working at all, do you use any shortcut combination or just focus an input?

I use this app:
Screenshot 2025-09-10 at 13 01 21

pgilad:
I set Universal Autofill on OSX app
image

Then when you are at a site that has a saved item in 1Password, I use their keyboard shortcut (CMD+\) which pops up autofill, it's immediate if there's a single matching item (based on url), and let's you choose between options if there are several.
In any case - on the login for TrueNAS screen, it doesn't autofill at all (doesn't even suggest it).

On my local test page, after removing angular material input (that sets `aria-autocomplete="list"` which blocks 1Password from working, at least natively), it autofills

PR #12423, opened 2025-08-18
NAS-137150 / 26.04 / Private Key and Passphrase Validation Errors when Importing CSR

AlexKarpov98:
**Testing:** 
see ticket and use CSR and Private key from the Comment here -> https://ixsystems.atlassian.net/browse/NAS-137150?focusedCommentId=317319

The form should be submitted successfully now.

Added more tests + util method.


william-gr:
Anything preventing this getting merged?

PR #12437, opened 2025-08-21
NAS-137247 / 26.04 / remove calls to virt plugin apis

aervin:
Removing our calls to the `virt` plugin as middleware is removing that API namespace. This PR also removes the Containers feature from the webui. We will be redesigning that experience once the new containers plugin is ready.

|Affects         |Reasoning
|Documentation   |
|Testing         |


AlexKarpov98 (commented):
I talked with Caleb about it, let's wait for @william-gr to return first. 🙌 This page should be refactored (soon) as I understand, not fully removed.

PR #12439, opened 2025-08-22
NAS-137270: Navigation is trapped and does not allow to focus on the next elements

AlexKarpov98:
Testing: see ticket.
Accessibility/focus improvements.

Preview:

https://github.com/user-attachments/assets/6763e28e-8fda-4099-9af3-a2ed0ee9fb48

### DIFF
--- a/src/app/modules/layout/admin-layout/admin-layout.component.html
+++ b/src/app/modules/layout/admin-layout/admin-layout.component.html
@@ -70,8 +70,6 @@
     [style.flex-basis]="'calc(100% - ' + sidenavWidth + ')'"
     [style.margin-left]="sidenavWidth"
   >
-    <ix-topbar></ix-topbar>
-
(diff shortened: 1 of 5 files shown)


william-gr:
I dont follow why we have web socket debug panel changes here? Do they interfere?
Also Claude seems to think we could use cdkTrapFocus instead, is he wrong?

AlexKarpov98:
> I dont follow why we have web socket debug panel changes here? Do they interfere?
> 
> Also Claude seems to think we could use cdkTrapFocus instead, is he wrong?

@william-gr 
- cdkTrapFocus - It didn't work for me.
- debug panel has a trap as well and it's OK - but the problem is that the panel is rendered even if it's not actually opened - we only hide it with CSS - but it's actually rendered - and can take some invisible focus and trap there.

william-gr:
Works here, thanks

PR #12440, opened 2025-08-22
NAS-137239: reset user search state when toggling between Basic and Advanced

aervin:
**Changes:**
Clears query filters and reverts to default filtering when switching between Basic and Advanced search.

**Testing:**
See ticket


dszidi (approved):
Code looks good. No regressions I have noticed and the search field properly resets to local users only as requested in the ticket. LGTM

PR #12441, opened 2025-08-22
NAS-137241 / 26.04 / warn users about adding special/dedup VDEVs with no redundancy

aervin:
**Changes:**
During add VDEV workflow, adds warnings about special and dedup VDEVs lacking redundancy.

**Testing:**
* Create a pool
* Navigate to that pool's "Add VDEV" workflow.
* Verify that adding special or dedup VDEVs without redundancy results in a red warning message.


william-gr:
Chris P comments on the ticket about also warning a user about permanent vdevs. Could be another ticket/PR if we prefer but it seems to make sense.

aervin:
I'll make a follow-up ticket for Chris P's suggestion.

AlexKarpov98 (approved):
Works as described 👍 

Image

PR #12444, opened 2025-08-25
NAS-135162 / 26.04 / Simplify Storage Dashboard

AlexKarpov98:
**Changes:**

**Testing:**
See ticket. 

Screenshot 2025-08-25 at 14 02 55

### Downstream

|Affects         |Reasoning
|Documentation   | Button text changes

### DIFF
--- a/src/app/helptext/storage/volumes/volume-list.ts
+++ b/src/app/helptext/storage/volumes/volume-list.ts
@@ -9,7 +9,7 @@ export const helptextVolumes = {
     unableToTerminate: T('Unable to terminate processes which are using this pool: '),
   },
 
-  exportError: T('Error disconnecting/deleting pool.'),
+  exportError: T('Error disconnecting pool.'),
 
   exportDialog: {
     warningSysDataset: T('This pool contains the system dataset that stores critical data like debugging \
--- a/src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.html
+++ b/src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.html
@@ -57,21 +57,23 @@
           }
         </div>
 
-        @if (shouldShowHaWarning) {
+        @if (isLastPoolInHaSystem()) {
           <mat-expansion-panel class="expansion-panel ha-warning-panel" [expanded]="true">
             <mat-expansion-panel-header>
               <mat-panel-title>
                 <div class="warning-icon">
                   <ix-icon name="warning"></ix-icon>
                 </div>
-                {{ 'Critical: High Availability will be disabled' | translate }}
+                {{ 'Critical: High Availability should be disabled' | translate }}
               </mat-panel-title>
             </mat-expansion-panel-header>
 
             <div class="expansion-content">
               <div class="ha-warning">
                 <p><strong>{{ 'This is the last pool in your HA system.' | translate }}</strong></p>
-                <p>{{ 'Exporting this pool will disable High Availability functionality because no pools will remain for failover operations. Your system will continue to operate, but redundancy will be lost until another pool is added.' | translate }}</p>
+                <p>{{ 'Removing this pool will leave no pools available for failover operations, effectively disabling High Availability functionality. Your system will continue to operate, but redundancy will be lost until another pool is added.' | translate }}</p>
+                <p><strong>{{ 'Action Required:' | translate }}</strong> {{ 'You must first disable High Availability before you can proceed with this operation.' | translate }}</p>
+                <p><em>{{ 'The disconnect button will remain disabled until HA is turned off.' | translate }}</em></p>
                 <p>{{ 'Please ensure you understand the implications before proceeding.' | translate }}</p>
               </div>
             </div>
@@ -108,7 +110,7 @@
           <mat-expansion-panel class="expansion-panel">
             <mat-expansion-panel-header>
               <mat-panel-title>
-                {{ 'These services and processes depend on this pool' | translate }}
+                {{ '{count} services and processes depend on this pool' | translate: { count: dependencyCount } }}
               </mat-panel-title>
             </mat-expansion-panel-header>
 
@@ -150,7 +152,7 @@
                   <ul>
                     @for (process of process.unknownProcesses; track process) {
                       <li>
-                        {{ process.pid || ('Unknown PID' | translate) }} -
+                        {{ process.pid }} -
                         {{ process.cmdline?.substring(0, 40) }}
                       </li>
                     }
--- a/src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
+++ b/src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
@@ -20,6 +20,7 @@ import { PoolStatus } from 'app/enums/pool-status.enum';
 import { Role } from 'app/enums/role.enum';
 import { isFailedJobError } from 'app/helpers/api.helper';
 import { helptextVolumes } from 'app/helptext/storage/volumes/volume-list';
+import { FailoverConfig } from 'app/interfaces/failover.interface';
 import { Job } from 'app/interfaces/job.interface';
 import { PoolAttachment } from 'app/interfaces/pool-attachment.interface';
 import { isServicesToBeRestartedInfo, ServicesToBeRestartedInfo } from 'app/interfaces/pool-export.interface';
@@ -117,6 +118,7 @@ export class ExportDisconnectModalComponent implements OnInit {
   systemConfig: SystemDatasetConfig;
   totalPoolCount = 0;
   isHaEnabled = false;
+  failoverConfig: FailoverConfig;
 
   isFormLoading = false;
   form = this.fb.nonNullable.group({
@@ -161,6 +163,13 @@ export class ExportDisconnectModalComponent implements OnInit {
 
   cancel(): void {
     this.dialogRef.close(false);
+  }
+
+  isLastPoolInHaSystem(): boolean {
+    return this.isHaEnabled
+      && this.failoverConfig
+      && !this.failoverConfig.disabled
+      && this.totalPoolCount === 1;
   }
 
   startExportDisconnectJob(): void {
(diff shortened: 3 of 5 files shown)


william-gr:
How about changing the button label in the dialog based on the action?
There are two different actions there: Disconnect (destroy unchecked) and Delete (destroy checked).

@dszidi thoughts?

dszidi:
I would be good with that if we also changed the radio label of the confirmation checkbox to match what the action button says.

While we're changing labels, I would also consider changing "Delete saved configurations from TrueNAS?" to "Forget this pool"

william-gr:
> I would be good with that if we also changed the radio label of the confirmation checkbox to match what the action button says.
> 
> While we're changing labels, I would also consider changing "Delete saved configurations from TrueNAS?" to "Forget this pool"

Delete saved configurations means to remove all shares or other configs that depend on this pool, it does not depend on the label. Perhaps there is a better way to say that

dszidi:
@AlexKarpov98 I added to the ticket a comment and video outlining suggested changes to the trigger button and the dialog.

AlexKarpov98:
@dszidi @william-gr please check the updated version now 🙌 

Preview:

https://github.com/user-attachments/assets/db705145-fed3-4a16-8621-8e2581e21fdb

I used mock for `pool.processes` and `pool.attachments`, see:

[mock-configs-2025-09-03.json](https://github.com/user-attachments/files/22116329/mock-configs-2025-09-03.json)

src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
```diff
@@ -240,51 +267,116 @@ export class ExportDisconnectModalComponent implements OnInit {
...
+        this.poolEncryptionStatus[poolName] = encrypted;
+      });
+      this.cdr.markForCheck();
     });
+  }
 
-    this.form.controls.destroy.valueChanges
-      .pipe(untilDestroyed(this))
-      .subscribe(() => this.resetNameInputValidState());
+  private determineDestinationPool(): void {
```
william-gr: How do you know this is the destination pool middleware is going to choose?

src/app/pages/storage/components/dashboard-pool/vdevs-card/vdevs-card.component.html
```diff
@@ -121,8 +121,13 @@ <h3 mat-card-title>
...
       @if (noOtherVdevTypes()) {
-        <div class="vdev-line">
-          <b>{{ 'No other VDEV types.' | translate }}</b>
+        <div class="vdev-line empty-vdev-state">
+          <div class="empty-message">
+            {{ 'Pool configuration complete.' | translate }}
+          </div>
+          <div class="empty-description">
+            {{ 'Optional VDEV types like cache, log, or spare can improve performance.' | translate }}
+          </div>
```
william-gr, line 130: Why is this change in this PR?

src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
AlexKarpov98: It's the logic for multi pool provided in the mockup.

src/app/pages/storage/components/dashboard-pool/vdevs-card/vdevs-card.component.html
AlexKarpov98: This is a part of improvement. Vdevs card is available on Storage dashboard.
Mentioned in the original ticket description.

src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
william-gr: Please check with backend team, dont think we can guess what the destination is going to be. We need their input.

william-gr:
@AlexKarpov98 it seems we missed handling the case of HA systems.
More importantly when we are disconnecting the last pool we need a big warning saying the pool is fundamental for HA operation and that HA will be disabled.

AlexKarpov98:
> @AlexKarpov98 it seems we missed handling the case of HA systems. More importantly when we are disconnecting the last pool we need a big warning saying the pool is fundamental for HA operation and that HA will be disabled.

is that a new functionality, it wasn't there before?

src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
william-gr: Discussed with Caleb. Lets simplify the logic and just say the system dataset will have to be moved to another pool.
That means, not try to determine where its going to go.

william-gr:
> > @AlexKarpov98 it seems we missed handling the case of HA systems. More importantly when we are disconnecting the last pool we need a big warning saying the pool is fundamental for HA operation and that HA will be disabled.
> 
> is that a new functionality, it wasn't there before?

Yes, something we have always overlooked

william-gr:
As mentioned before we cannot determine the destination pool for system dataset, all we can say is that it will move away from that pool.

src/app/helptext/storage/volumes/volume-list.ts
```diff
@@ -9,34 +9,22 @@ export const helptextVolumes = {
     unableToTerminate: T('Unable to terminate processes which are using this pool: '),
   },
 
-  exportError: T('Error exporting/disconnecting pool.'),
+  exportError: T('Error disconnecting/deleting pool.'),
```
william-gr: I think this can be only disconnecting

william-gr:
@Mrt134 could you or someone in docs team please help us review all the text changes here? Any suggestions are more than welcomed.

src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
```diff
@@ -240,51 +269,80 @@ export class ExportDisconnectModalComponent implements OnInit {
...
+  get shouldShowSystemDatasetWarning(): boolean {
+    return this.showSysDatasetWarning && this.isExportSelected;
+  }
+
+  get isLastPool(): boolean {
+    return this.totalPoolCount <= 1;
+  }
+
+  get shouldShowHaWarning(): boolean {
+    return this.isHaEnabled && this.isLastPool && this.isExportSelected;
```
william-gr: This should show not only for export for also delete

src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
```diff
@@ -240,51 +269,80 @@ export class ExportDisconnectModalComponent implements OnInit {
...
+  get isExportSelected(): boolean {
+    return this.selectedOption() === DisconnectOption.Export;
+  }
+
+  get isDeleteSelected(): boolean {
+    return this.selectedOption() === DisconnectOption.Delete;
+  }
+
+  get shouldShowSystemDatasetWarning(): boolean {
+    return this.showSysDatasetWarning && this.isExportSelected;
```
william-gr: this should probably show for delete too?

src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
```diff
@@ -240,51 +269,76 @@ export class ExportDisconnectModalComponent implements OnInit {
       this.api.call('pool.attachments', [this.pool.id]),
       this.api.call('pool.processes', [this.pool.id]),
       this.api.call('systemdataset.config'),
+      this.api.call('pool.query'),
```
william-gr: That call is too expensive, please check with middleware team if there is some other call or arguments when you only want to know how many pools there are in the system.

src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
```diff
@@ -240,51 +269,76 @@ export class ExportDisconnectModalComponent implements OnInit {
...
+      this.store.select(selectIsHaEnabled).pipe(take(1)),
```
william-gr, line 282: Can you confirm what selectIsHaEnabled does? Was it only failover.licensed? If so we might to revisit this.

On a second thought I will need to discuss with Caleb on that and the behavior on HA, I will do that tomorrow.

src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.html
```diff
@@ -12,107 +12,199 @@ <h1 mat-dialog-title class="export-disconnect-modal-title">
...
-                  {{ process.pid || ('Unknown PID' | translate) }} -
-                  {{ process.cmdline?.substring(0, 40) }}
-                </li>
+              @if (process.unknownProcesses.length) {
+                <div class="unknown-processes">
+                  {{ 'These unknown processes are using the pool:' | translate }}
+                  <ul>
+                    @for (process of process.unknownProcesses; track process) {
+                      <li>
+                        {{ process.pid || ('Unknown PID' | translate) }} -
```
william-gr: Not your bug but this is odd. Every process has a pid otherwise its not a process.

tonyriv3:
@william-gr would you rather I make suggestions here in a comment here, or within the files?

william-gr:
> @william-gr would you rather I make suggestions here in a comment here, or within the files?

@tonyriv3 no preference, up to you

src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
william-gr: `'pool.query', [], {'count': True}`

src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
william-gr: So what really needs to happen here is:
check isHaEnabled and also check failover.config['disabled'] is false, if so show error about being the last pool in the system and that HA needs to be administratively disabled before proceeding.

Disconnect cannot proceed.

dszidi (changes requested):
image

The expansion panel header currently says "These services and processes depend on this pool". This is not very useful until the panel is expanded. It would be better if it could offer a tally and say "<n> services and processes depend on this pool". That way the user gets a total and they can expand for details.

Beyond that I would say it looks good. I'll leave any further text content suggestions to docs team review.

src/app/pages/storage/components/dashboard-pool/export-disconnect-modal/export-disconnect-modal.component.ts
AlexKarpov98: Adding such logic:

https://github.com/user-attachments/assets/a2257373-76e9-48df-840d-6949a1f55931

AlexKarpov98:
@william-gr @dszidi updates arrived 🫴

william-gr:
Lets merge it. If docs have suggestions we can implement later, this has been sitting for a week already.

PR #12455, opened 2025-08-26
NAS-137291: VM list does not remember column state

AlexKarpov98:
Testing: see ticket.

Now VM list remembers column state.

https://github.com/user-attachments/assets/70662903-996d-4830-a4e3-a5868dccac5a


dszidi (approved):
Code looks good and works as intended. LGTM

PR #12459, opened 2025-08-26
NAS-137320 / 26.04 / Yarn 4

william-gr:
**Changes:**

Change from yarn 1.x to 4

**Testing:**

Build dev and production

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


william-gr:
I dont think so. The commands are all the same in package.json, nothing changes.

PR #12486, opened 2025-09-02
NAS-137332 / 26.04 / During vm iso upload clicking away from the progress window results in an error

AlexKarpov98:
Testing: 
See ticket. 
More tests added.

No error should appear now. 
As well I improved the logic - now the request is cancelled if user clicks outside or `ESC` pressed (with a notification).
I thought that blocking user with non-clickable loader - until request will succeed or fail - is not the desired UX.

Previously the request was still going and I could started a new one on top of the previous one. 

Preview:
`(locally request will fail)`

https://github.com/user-attachments/assets/3ce40299-a9f4-45a4-ae44-3cef6157bf82

### DIFF
--- a/src/app/modules/loader/loader.service.ts
+++ b/src/app/modules/loader/loader.service.ts
@@ -133,16 +133,12 @@ export class LoaderService {
   private handleConfirmationClose(): void {
     // Check if handler is still active (not removed)
     if (this.onBeforeClose && this.handlersSetup) {
-      console.info('DEBUG: Showing confirmation dialog');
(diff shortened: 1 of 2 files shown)


william-gr:
is it possible to ask for a confirmation before cancelling the upload?

AlexKarpov98:
@william-gr should be like that now:

https://github.com/user-attachments/assets/783d7daf-8dfc-4a8c-8157-cfbe068be69b

aervin (approved):
Works well here. I ran into this issue recently. This fix is a nice improvement.

PR #12488, opened 2025-09-02
NAS-137395 / 26.04 / Do not require Port in NVMe-oF and show 4420 as placeholder

william-gr:
**Changes:**

image

**Testing:**

Try to add nvme Port with a TCP port.

### Downstream

|Affects         |Reasoning
|Documentation   |Port no longer required
|Testing         |Test without and with port


nasbdh9:
8009 should also be added, according to the standard, 4420 used for data transmission, 8009 used for discovery service.

dszidi (approved):
Code review only. LGTM

PR #12497, opened 2025-09-03
NAS-137378: do not pass user/group unless box is checked

william-gr:
**Changes:**

**Testing:**

Try to set ACL with boxes checked and unchecked. Verify dataset permissions changes via shell.

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


AlexKarpov98 (approved):
Nice, I was thinking about the same solution 👌

PR #12501, opened 2025-09-03
NAS-137407 / 26.04 / added back the the refresh directory services cache button

gmboknat:
**Changes:**
- added `Rebuild Directory Service Cache` to directory-services-form
image

**Testing:**
1. Navigate to Directory Service page
2. Configure and enable directory service
3. In the card, click the the menu icon 
4. Click `Rebuild Directory Service Cache`

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |

### DIFF
--- a/src/app/helptext/directory-service/dashboard.ts
+++ b/src/app/helptext/directory-service/dashboard.ts
@@ -32,4 +32,9 @@ export const helptextDashboard = {
     title: T('Warning'),
     message: T('Changing Advanced settings can be dangerous when done incorrectly. Please use caution before saving.'),
   },
+  rebuildCache: {
+    success: T('Directory Service cache has been rebuilt.'),
+    error: T('Failed to rebuild directory service cache.'),
+    errorTitle: T('Error'),
+  },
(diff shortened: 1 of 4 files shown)


src/app/pages/directory-service/components/directory-services-form/directory-services-form.component.html
```diff
@@ -96,6 +96,16 @@
         >
           {{ 'Save' | translate }}
         </button>
+        <button
```
aervin: I don't think it makes sense to expose this button if directory services are not configured. In that case, the middleware job completes instantly with a description of "Directory services are not enabled."

william-gr: Agreed. Maybe it could live in the directory services card and not in the form, possibly in a three dot menu.

gmboknat: Okay, I'll move it to the card.

PR #12502, opened 2025-09-03
NAS-137417 / 26.04 / Fix parent slide-in not reappearing after closing child slide-in

william-gr:
When closing a child slide-in (e.g., namespace form in NVMe-oF), the parent slide-in was not becoming visible again, leaving only the overlay backdrop blocking interaction.

The fix ensures the parent container's host element has correct visibility classes when animating back in after a child slide-in closes.

**Changes:**

**Testing:**

Add a namespace when adding a new NVMe oF entry.

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


aervin (approved):
No issues with stacked panels here

PR #12536, opened 2025-09-12
NAS-137520 / 26.04 / Fix IPMI VLAN validation preventing form save

AlexKarpov98:
IPMI network configuration form was incorrectly requiring VLAN ID even when "Enable VLAN" was unchecked, preventing users from saving configuration.

- Remove unconditional required validator from vlan_id field
- Add dynamic validation that only requires VLAN ID when VLAN is enabled
- Clear VLAN ID value when VLAN is disabled
- Add tests to verify correct validation behavior

Fixes issue where Save button remained disabled due to invalid VLAN ID field.

**Changes:**

**Testing:**

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


anodos325:
I think your bug number is wrong. NAS-137505 is a PAM error on backend. This PR reopened a backend ticket.

AlexKarpov98:
> I think your bug number is wrong. NAS-137505 is a PAM error on backend. This PR reopened a backend ticket.

oh, yes, sorry.. gonna fix it now
I trusted Claude... 😿

PR #12546, opened 2025-09-16
Fix translate pipe error

MDVAurelien:
This will fix the translation error due to the character : stuck to the translate keyword

### DIFF
--- a/src/app/pages/system/update/update.component.html
+++ b/src/app/pages/system/update/update.component.html
@@ -26,7 +26,7 @@
             </div>
             @if (doesNotMatchProfile() && currentVersionProfile()) {
               <div class="profile-mismatch">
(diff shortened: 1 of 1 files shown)


MDVAurelien:
I'm a little surprised, it doesn't work with the FR version on ISO 25.10.0-MASTER-20250912-234154 - Goldeye . fr.json is OK !

Edit: Your're right, it's work with other translate file ! I don't understand

MDVAurelien:
How is the nightly version built? The json is always integrated with new updates, are they injected from time to time? Thks

william-gr:
> How is the nightly version built? The json is always integrated with new updates, are they injected from time to time? Thks

We push new translations weekly on Monday.

I will close this PR for now since it seems it was handled elsewhere.

PR #12548, opened 2025-09-16
NAS-137543 / 26.04 / New Users Page Edit UI showing Validation errors instead of the inline red text errors

AlexKarpov98:
Testing: see ticket.
Added inline error handling for new users form.
Introduced logic to notify ix-editable if there are validation errors on submit - to show inline error handling when possible.

Preview:

https://github.com/user-attachments/assets/7e8b6917-2698-4012-b10e-7ee3636035c5

### DIFF
--- a/src/app/pages/credentials/users/user-form/additional-details-section/additional-details-section.component.html
+++ b/src/app/pages/credentials/users/user-form/additional-details-section/additional-details-section.component.html
@@ -191,6 +191,11 @@
             }
 
             <div edit class="extra-controls-container">
+              <ix-checkbox
+                formControlName="sudo_commands_all"
+                [label]="'Allow all sudo commands' | translate"
+              ></ix-checkbox>
+
               @if (!form.controls.sudo_commands_all.value) {
                 <ix-chips
                   formControlName="sudo_commands"
@@ -199,8 +204,8 @@
               }
 
               <ix-checkbox
-                formControlName="sudo_commands_all"
-                [label]="'Allow all sudo commands' | translate"
+                formControlName="sudo_commands_nopasswd_all"
+                [label]="'Allow all sudo commands with no password' | translate"
               ></ix-checkbox>
 
               @if (!form.controls.sudo_commands_nopasswd_all.value) {
@@ -209,11 +214,6 @@
                   [label]="'Allowed sudo commands with no password' | translate"
                 ></ix-chips>
               }
-
-              <ix-checkbox
-                formControlName="sudo_commands_nopasswd_all"
-                [label]="'Allow all sudo commands with no password' | translate"
-              ></ix-checkbox>
             </div>
           </ix-editable>
         </ix-details-item>


src/app/modules/forms/ix-forms/services/form-error-handler.service.ts
```diff
@@ -146,9 +148,18 @@ export class FormErrorHandlerService {
...
+    let element = this.formService.getElementByControlName(field);
+    if (!element && this.document?.querySelector) {
+      // Fallback: try to find element by formControlName attribute
+      element = this.document.querySelector(`[formControlName="${field}"]`) as HTMLElement;
+    }
+
     if (!element) {
-      console.warn(`getElementByControlName Could not find element for ${field}.`);
+      console.warn(`Could not find DOM element for field ${field}.`);
       this.needToShowError = true;
```
william-gr: Perhaps if we cant find we can display at t he top of the form instead of a error dialog?

AlexKarpov98: I tried it, seems like modal is more clear and visible to the user. We can consider about it in a separate ticket if needed.

william-gr: Please create the ticket then, showing dialog is horrible experience.

AlexKarpov98: @william-gr - for all other forms we are displaying the dialogs on some errors (without field mentioning) - should we replace it across all forms? (to show the error inline at the top of the form or near the submit button)

william-gr: I am talking about validation errors exclusively
In theory it should never happen and go along with the field, but this is the case where it doesnt and its a bug.

AlexKarpov98: @william-gr  ah, in this case I'll try to fix it here, if API mentions the field. Seems like there are some issues with conditions..

AlexKarpov98:
@william-gr please check it out once again 🙌

src/app/modules/forms/ix-forms/services/form-error-handler.service.ts
```diff
@@ -63,33 +70,35 @@ export class FormErrorHandlerService {
...
     originalError: unknown,
   ): void {
     this.isFocusedOnError = false;
     const extra = (error as ApiErrorDetails).extra as string[][];
     for (const extraItem of extra) {
-      const field = extraItem[0].split('.').pop();
+      const fullFieldPath = extraItem[0];
+      const field = this.extractFieldName(fullFieldPath);
       if (!field) {
         return;
```
william-gr, line 92: Should this be continue and not return?

AlexKarpov98: correct, good catch 👍

PR #12584, opened 2025-09-23
NAS-137646 / 26.04 / User UI: Enabling SSH access does not save although the text says that the user was updated

AlexKarpov98:
Testing: see ticket.
No validation was present for SSH checkbox logic. Now it's added and form works as expected. Tests added.

Preview:

https://github.com/user-attachments/assets/6c3996f5-8797-40c6-a196-d8a4265965ce

### DIFF
--- a/src/app/pages/credentials/users/user-form/auth-section/auth-section.component.spec.ts
+++ b/src/app/pages/credentials/users/user-form/auth-section/auth-section.component.spec.ts
@@ -118,12 +118,19 @@ describe('AuthSectionComponent', () => {
       expect(await loader.getHarness(IxCheckboxHarness.with({ label: 'Allow SSH Login with Password (not recommended)' }))).toBeTruthy();
     });
 
-    it('disables "Allow SSH Login with Password" when password is disabled', async () => {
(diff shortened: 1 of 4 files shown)


src/app/pages/credentials/users/user-form/auth-section/auth-section.component.spec.ts
```diff
@@ -134,6 +141,16 @@ describe('AuthSectionComponent', () => {
       });
     });
 
+    it('automatically enables SSH password authentication for new users when SSH access is enabled', async () => {
```
william-gr: Password should not be encouraged. Can we just run validation so user knows either is needed, perhaps?

AlexKarpov98: Sure, it was made to avoid instant red error message 🙌

AlexKarpov98: Screenshot 2025-09-24 at 13 19 07

Updated.

william-gr (approved):

PR #12585, opened 2025-09-23
NAS-137661 / 26.04 / User UI: Back button is not working

AlexKarpov98:
Not sure why there is no lint error for such cases.
`(close)` instead of `(closed)` used


william-gr:
angular options wont do much, you may want to look for this in eslint rules.

william-gr:
Seems like this could use a test too

AlexKarpov98:
> angular options wont do much, you may want to look for this in eslint rules.

haven't found any good ESLint rule

AlexKarpov98:
> Seems like this could use a test too

what do you mean? `.spec.ts` test?

aervin (approved):

PR #12588, opened 2025-09-23
NAS-137568 / 26.04 / NAS-137611: Add service worker for cache busting on upgrades

william-gr:
Implement service worker to ensure users get the latest WebUI version after OS upgrades without manual cache clearing. The service worker:

- Automatically detects new versions and reloads the page
- Clears old caches when a new version is deployed
- Uses version string with timestamp and git commit hash
- Checks for updates when user returns to tab (no constant polling)
- Works with both development (/) and production (/ui/) paths
- No offline support (appropriate for NAS management UI)

Changes:
- Add service worker (sw.js) with cache versioning strategy
- Add service worker registration module
- Add build script to update version on each build
- Configure Angular to include service worker in assets
- Update build process to inject version automatically

**Testing:**

Unfortunately since we cant reliably reproduce the issue I dont know how to test other than just make sure it doesnt breaking anything.

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |

### DIFF
--- a/src/app/pages/signin/signin.component.ts
+++ b/src/app/pages/signin/signin.component.ts
@@ -117,6 +117,14 @@ export class SigninComponent implements OnInit {
     this.isConnected$
       .pipe(filter(Boolean), untilDestroyed(this))
       .subscribe(() => {
+        // Check if this is after a system update - reload page to get latest UI
+        const isUpdateRestart = this.window.sessionStorage.getItem('updateCompleted') === 'true';
+        if (isUpdateRestart) {
+          this.window.sessionStorage.removeItem('updateCompleted');
+          this.window.location.reload();
+          return;
+        }
+
         this.signinStore.init();
       });
   }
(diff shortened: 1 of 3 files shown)


gmboknat:
@william-gr how about using `@angular/pwa` or `@angular/service-worker`? I think it would simplify the implementation and handle version updates more effectively.

william-gr:
> @william-gr how about using `@angular/pwa` or `@angular/service-worker`? I think it would simplify the implementation and handle version updates more effectively.

What are the advantages? They seem to focus on the opposite that we want, which is offline application and we only want to avoid cache.

I am not opposed to the idea but I don't see much benefits right away, please elaborate if you can.

gmboknat:
@william-gr  Yes, you are right that PWA focuses on offline functionality. However, for tracking version updates or deployments, I think we can implement `@angular/service-worker`. It's similar to your current approach and offers a `versionUpdates` observable that the UI can use to monitor updates.

william-gr:
@gmbok19 still dont see the benefit. Version is not the problem as scripts are hashed.

would you please take over the ticket and implement your proposal for comparison as it seems you have familiarity with that approach?

thanks!

gmboknat:
@william-gr okay, will do it.

william-gr:
time 6:00

PR #12597, opened 2025-09-24
NAS-137673 / 26.04 / fix default value for role on user edit

william-gr:
**Testing:**

See ticket. Add existing group to full admin privilege and edit that user to see role.

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


william-gr:
time 0:20

PR #12612, opened 2025-09-29
NAS-137665 / 26.04 / feat: add dynamic validation to the cloudflare DNS auth form

mattwyatt-ix:
**Changes:**

when configuring a Cloudflare ACME DNS authenticator on the `credentials/certificates` page, the email and API key fields are mutually exclusive with the API token field. these requirements weren't actually formalized in the UI and weren't revealed until a user submitted the form. this PR adds dynamic validation to the page to properly show requirements as the user edits the form.

**Testing:**

1. navigate to the "Credentials" menu on the left and choose "Certificates".
2. click the "Add" button on the ACME DNS-Authenticators section. the "Cloudflare" DNS authenticator is already selected.
3. edit the fields on the form and observe that the required fields change as you input and clear each field.

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |

### DIFF
--- a/src/app/pages/credentials/certificates-dash/acmedns-form/acmedns-form.component.spec.ts
+++ b/src/app/pages/credentials/certificates-dash/acmedns-form/acmedns-form.component.spec.ts
@@ -13,6 +13,7 @@ import { DnsAuthenticator } from 'app/interfaces/dns-authenticator.interface';
 import { Option } from 'app/interfaces/option.interface';
 import { Schema } from 'app/interfaces/schema.interface';
 import { DialogService } from 'app/modules/dialog/dialog.service';
+import { IxInputHarness } from 'app/modules/forms/ix-forms/components/ix-input/ix-input.harness';
 import { IxFormHarness } from 'app/modules/forms/ix-forms/testing/ix-form.harness';
 import { SlideInRef } from 'app/modules/slide-ins/slide-in-ref';
 import { ApiService } from 'app/modules/websocket/api.service';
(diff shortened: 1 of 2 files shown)


william-gr (commented):
It seems the validators are not cleaned if we switch to another authentication (not cloud flare).

For the tests in this project we do prefer use of harness instead of query, its usually more robust.

We also need to support i18n, see how we use translate.instant elsewhere.

Otherwise it looks good!

mattwyatt-ix:
thanks for the feedback, everyone! @william-gr, i'll fix the validators and tests and support that one error string i added for i18n. @AlexKarpov98 those errors from the start of the form are incongruent with the rest of the DNS auth types, but wasn't sure whether we liked the obvious-ness or not. i think i dislike them too, so i'll get them suppressed until the user types anything.

AlexKarpov98 (changes requested):
@mattwyatt-ix - after some manual testing, I noticed a couple of bugs, seems like validation is not consistent:

1. When email & API key set (in edit mode) and i update email or API key.

https://github.com/user-attachments/assets/f3bee3d9-461e-4b8c-ab85-df549576b920

2. No email validator for email field

https://github.com/user-attachments/assets/c627fa57-5713-489d-9aa1-308689035ff1

3. UI allows submitting with no email + API key setup OR API token

https://github.com/user-attachments/assets/25680550-a85d-4034-be03-ae36a0c5385e

Additionally:

Long method complexity & hard to understand (src/app/pages/credentials/certificates-dash/acmedns-form/acmedns-form.component.ts:140-218)

```
private setupConditionalValidation(): void {
  // 78 lines of complex logic
}
```
Consider breaking this into smaller, focused methods:

setupCloudflareControls()
createValidationHandlers()
handleFormValueChanges()

AlexKarpov98:
@william-gr @mattwyatt-ix - I was so curious about this ticket, so that I created an alternative PR 🙈 
here it is: https://github.com/truenas/webui/pull/12627

william-gr:
Closing this as it seems this was sorted out in Slack and fixed in another PR.

PR #12613, opened 2025-09-29
NAS-137674 / 26.04 / fix: reverse sorting arrow; up means ascending, down means descending

mattwyatt-ix:
**Changes:**

the sorting direction indicator (up or down arrow) was backwards in our sorting menu. when sorting in ascending order, the arrow should point upwards and the items should be ordered low-to-high. when sorting in descending order, the arrow should point downwards and the items should be ordered high-to-low.

**Testing:**

1. navigate to any page with a sortable table view. (the ticket mentions the snapshots view, but the jobs view works just as well)
2. observe that the arrow points upwards when the smallest items are first, and downwards when the largest items are first. 

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


AlexKarpov98 (approved):
👍

william-gr (approved):

PR #12617, opened 2025-09-30
NAS-137568 / 26.04 / added angular service-worker

gmboknat:
**Changes:**
- added @angular/service-worker

**Testing:**

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component, OnInit, inject } from '@angular/core';
+import { ChangeDetectionStrategy, Component, OnInit, inject, OnDestroy } from '@angular/core';
 import { MatSnackBar } from '@angular/material/snack-bar';
 import { Title } from '@angular/platform-browser';
 import { Router, NavigationEnd, RouterOutlet } from '@angular/router';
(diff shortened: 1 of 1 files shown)


src/app/app.component.ts
```diff
@@ -80,6 +83,16 @@ export class AppComponent implements OnInit {
 
   ngOnInit(): void {
     this.setupScrollToTopOnNavigation();
+
+    if (this.swUpdate.isEnabled) {
+      this.swUpdate.versionUpdates
+        .pipe(takeUntil(this.destroy$))
+        .subscribe((event) => {
+          if (event.type === 'VERSION_READY') {
```
william-gr, line 91: This seems to work when a new version is available from the upstream side. This is not supposed to happen with TrueNAS as the version is always the same. It will only change on a software upgrade which then we take care of reloading ourselves by nature.

gmboknat:
@william-gr Regarding these changes, the code checks for new deployments, such as during a NAS upgrade. I had planned to implement a check for the "Loading chunk" error, which occurs with new deployments, but it's already implemented.

gmboknat:
The way I tested it is that I have a terminal run the `yarn build` and the other one is `npx http-server -p 8080 -c-1 dist`.
Make changes like in `app.component.html` then run again `yarn build`. After the build, it still requires page reload before it shows the changes. We can also add here some periodic check.

william-gr:
> @william-gr Regarding these changes, the code checks for new deployments, such as during a NAS upgrade. I had planned to implement a check for the "Loading chunk" error, which occurs with new deployments, but it's already implemented.

Please make me understand why I am wrong. Again, we dont care if there is a new version on the same IP because thats not an online aplication. We already reload the application on upgrade.
The problem is browser cache. How can you guarantee with your service worker that index.html (/ui/) is being fetched without any cache involved when you load the UI for the first time?

gmboknat:
@william-gr my bad, I thought it's similar implementation on what you did on other PR. Will just close this.

PR #12619, opened 2025-09-30
NAS-137506 / 26.04 / added search protocols field in smb settings

gmbok19:
**Changes:**

**Testing:**

https://github.com/user-attachments/assets/5a630863-89fe-48d4-a1f4-ce5c5ea499c8

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |

### DIFF
--- a/src/app/interfaces/smb-config.interface.ts
+++ b/src/app/interfaces/smb-config.interface.ts
@@ -1,4 +1,5 @@
 import { SmbEncryption } from 'app/enums/smb-encryption.enum';
+import { SmbProtocol } from 'app/enums/smb-protocol.enum';
 
 export interface SmbConfig {
   aapl_extensions: boolean;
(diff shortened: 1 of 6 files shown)


william-gr:
@gmbok19 there is no reason to have a field with only one option. Please switch it to "Enable Search (Spotlight)" and a checkbox, or something similar.
There are still some reviews to address from Claude too, apparently.

Thanks

william-gr (changes requested):
As comment before

gmboknat:
@william-gr okay, I implemented a selection field for future `WSP` value inclusion but for now I'll change it to a checkbox.

src/app/interfaces/smb-config.interface.ts
```diff
@@ -1,5 +1,7 @@
 import { SmbEncryption } from 'app/enums/smb-encryption.enum';
 
+export const smbProtocolSpotlight = 'SPOTLIGHT' as const;
```
william-gr: smbProtocolSpotlight is confusing, please use smbSearchProtocolSpotlight or just  smbSearchSpotlight

william-gr (approved):

PR #12622, opened 2025-10-01
Fix mocking job responses

william-gr:
The first collection update event is always using "added" type with the job id.

**Testing:**

Create a mocked job response by adding events to the call and see the UI behaves correctly.
One idea is to use update.run
Screenshot 2025-10-01 at 11 15 00 AM

https://github.com/user-attachments/assets/947246c1-6680-4a8b-be8d-45fd4c342c8f

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


william-gr:
time 0:30

AlexKarpov98 (approved):
Cool 👍

PR #12631, opened 2025-10-02
NAS-137694 / 26.04 / fix: update on `pool.snapshottask.query`

mattwyatt-ix:
**Changes:**

* previously, the snapshot tasks page only updated once when first opened or when a UI element was added.
* this fix subscribes to `pool.snapshottask.query` and calls `getSnapshotTasks` when receiving an event.

**Testing:**

here are the testing instructions directly from the ticket, modified to note the expected and observed behavior:
1.  Configure a new Periodic Snapshot Task from Data Protection page  (Note: if edit a previously configured Snapshot that was successfully Finished, it doesn’t always get updated to Pending state so may not see the issue without creating a new task)
    * For ease, set the Schedule to “Custom” and enter any time within the next few minutes to execute.  Select any dataset and leave all other parameters at default.  Save.
2. Monitor the Periodic Snapshot Tasks table to confirm the task is created in the table
3. Confirm that the next run accurately shows the countdown and the State of the task shows “PENDING”.  Wait for the task to execute.
4. **With this PR:** Observe that the task should update to "FINISHED" when it completes

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |

### DIFF
--- a/src/app/interfaces/api/api-event-directory.interface.ts
+++ b/src/app/interfaces/api/api-event-directory.interface.ts
@@ -9,6 +9,7 @@ import { DockerStatusData } from 'app/interfaces/docker-config.interface';
 import { FailoverDisabledReasonEvent } from 'app/interfaces/failover-disabled-reasons.interface';
 import { Group } from 'app/interfaces/group.interface';
 import { Job } from 'app/interfaces/job.interface';
+import { PeriodicSnapshotTask } from 'app/interfaces/periodic-snapshot-task.interface';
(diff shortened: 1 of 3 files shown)


mattwyatt-ix:
time 06:00

AlexKarpov98 (approved):
Works as described. 👌 

As an enhancement:
When we're subscribing - we get updates within the incoming event, so we can update state without a new request.
`(we have more places where we could bring such improvement)` @william-gr

PR #12633, opened 2025-10-02
NAS-137847 / 26.04 / Handle several deprecated usages

william-gr:
For some where its legacy code do not use @deprecated to not interfere with actual API changes which is the intent of deprecation.


AlexKarpov98 (approved):
👌

PR #12638, opened 2025-10-03
NAS-137765: fixed fade away issue of network interface graph

gmboknat:
**Changes:**

**Testing:**

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |

### DIFF
--- a/src/app/modules/charts/view-chart-area/view-chart-area.component.html
+++ b/src/app/modules/charts/view-chart-area/view-chart-area.component.html
@@ -1,1 +1,3 @@
-<canvas #canvas></canvas>
+<div class="canvas-container">
+    <canvas #canvas></canvas>
+</div>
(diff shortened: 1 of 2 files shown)


AlexKarpov98 (commented):
Unable to reproduce the original issue, I believe we fixed it some time ago previously. 
See This ticket (https://ixsystems.atlassian.net/browse/NAS-134337)

gmboknat:
@AlexKarpov98 Based on the ticket you provided, the changes were backported to 25.04.0, but the issue persists in 25.04.2.4, as reported.

AlexKarpov98:
@gmbok19 are you able to reproduce it in this case?

gmboknat:
@AlexKarpov98 I wasn't able to reproduce it in a Windows VM with a non-4k resolution.

AlexKarpov98:
@william-gr - what do you think about it? (check the above messages & ticket)
maybe it's some windows issue only IDK, but seems like it's impossible to reproduce it 🤷‍♂️

william-gr:
Lets not push a fix if we can't reproduce it. Or at least have great confidence that it will fix.

PR #12640, opened 2025-10-03
NAS-137884 / 26.04 / Use newer version of tn-common-typescript with sonarjs/deprecation turned on

william-gr:


gmboknat (approved):

william-gr:
time 0:30

PR #12646, opened 2025-10-06
NAS-137806 / 26.04 / Add stale data detection and user feedback for dashboard widgets

william-gr:
Implement a comprehensive solution to detect and notify users when dashboard widget data becomes stale (no updates received within 5 seconds). This improves user experience by providing clear visual feedback when the system is still generating data.

Key changes:
- Add detectStaleData RxJS operator with distinctUntilChanged to efficiently track data freshness
- Create WidgetStaleDataNoticeComponent with accessible ARIA live region for screen readers
- Integrate stale detection across all dashboard widgets (CPU, memory, network, storage)
- Extract PoolUsage interface for better type safety in widget-resources.service
- Add takeUntilDestroyed to all widget subscriptions to prevent memory leaks
- Update all widget templates to display stale notice instead of indefinite loading
- Add comprehensive test coverage including error handling scenarios

This ensures widgets gracefully handle scenarios where backend data generation is delayed while maintaining optimal performance through change detection optimizations.

https://github.com/user-attachments/assets/080905c4-fb4a-490d-9729-8bbf43f31e28

**Testing:**

systemctl stop netdata\

### DIFF
--- a/src/app/pages/dashboard/widgets/storage/widget-storage/widget-storage.component.scss
+++ b/src/app/pages/dashboard/widgets/storage/widget-storage/widget-storage.component.scss
@@ -18,7 +18,14 @@
   }
 
   .container {
+    flex: 1;
     padding: 4px 16px 16px;
+
+    // Only use flex display when showing stale notice
+    ix-widget-stale-data-notice {
+      display: flex;
+      height: 100%;
+    }
   }
 }
 


xMarlonNx:
Just to be clear, the old WebUI didn't have the stale dashboard widgets for me.  Just snappy instant loading, even after a restart.

Why is this an acceptable solution rather than fixing why they're stale for so long in 25.10 after every reboot or randomly? Why do we see this degradation in function in the first place? Wasn't the WebUI overhauled and supposed to be faster?

william-gr:
> Just to be clear, the old WebUI didn't have the stale dashboard widgets for me. Just snappy instant loading, even after a restart.
> 
> Why is this an acceptable solution rather than fixing why they're stale for so long in 25.10 after every reboot or randomly? Why do we see this degradation in function in the first place? Wasn't the WebUI overhauled and supposed to be faster?

There are challenges on the backend which I cannot speak for. This PR is for handling the situation better in the frontend.
If you feel this is not optimal you can raise another ticket.

xMarlonNx:
I personally find your solution not a solution but a band-aid. The problem lays in the backend. Why isn't it fixed there? 

Your answer indicates that the devs are operating isolated and there's a lack of intercommunication. 

Also, why is this band-aid 'solution' pushed to 26.04. It's a UI breaking bug. Shouldn't it be pushed forward into the 25.10.0 release at the least.  Are we going to have a broken dashboard for the remainder of 25.10? 

I find the approaches not optimal. Where should I file another ticket for fixing the actual issue in the backend.

gmboknat:
Some dashboard widgets are missing their labels.
image
image

william-gr:
> Some dashboard widgets are missing their labels. image image

Which widgets are these?

gmboknat:
@william-gr interface and memory widgets

william-gr:
> @william-gr interface and memory widgets

Good catch, updated

gmboknat:
@william-gr the text of the storage widget should be centered vertically
image

william-gr:
> @william-gr the text of the storage widget should be centered vertically image

Fixed

gmbok19 (approved):
please resolve the

william-gr:
time 0:30

AlexKarpov98:
Noticed an issue with this PR.
`Both "refresh" and "mdi-refresh" are used in the application. Consider only using the 'mdi' version.`
As well icon is not even visible.

I am preparing a quick fix 🙌 
[https://github.com/truenas/webui/pull/12692]

PR #12648, opened 2025-10-06
NAS-137713: Convert Containers pages to use new API (next updates)

AlexKarpov98:
Next updates.


william-gr (changes requested):
I am not a big fan on how pools is presented in the create form

Image

I have a preferred pool set that means I think it should not show up there. It should just default to preferred pool. It probably should be hidden under Advanced settings if I have one set. What do you think?

william-gr:
Also unrelated to these changes but I think it was discussed before:
image

I dont need all CPU options to show by default, I only need the first field by default.

william-gr:
Lastly, it seems webshell is supposed to be supported already, we are still using virt_instance_id for the shell, there is container_id.

william-gr:
Did the shell work for you? its not for me.

AlexKarpov98:
> Did the shell work for you? its not for me.

nope, I just updated it to use new fields. ATM - I am even unable to create a new container, got some errors that image does not exist or so.

PR #12650, opened 2025-10-07
NAS-137770: SSH Password login error message does not clear automatically

AlexKarpov98:
Testing: see ticket.
Local validation added to prevent API validation errors for checkboxes. (because for checkboxes we need to tick/untick in order to make the form valid again)

Preview:

https://github.com/user-attachments/assets/dcdb42c2-cbc0-4479-963d-88cdfdcdcb27


AlexKarpov98:
@aervin - yup, added the tests.

PR #12654, opened 2025-10-07
NAS-137913: Fix filesystem.file_tail_follow WebSocket subscription parameter format

william-gr:
Update WebSocket subscriptions to use JSON-encoded parameters instead of colon-separated strings for filesystem.file_tail_follow calls in job progress dialog and console messages store.

**Testing:**

Enable console messages in General Settings -> UI


william-gr:
time 0:20

AlexKarpov98 (approved):
👌

PR #12655, opened 2025-10-07
NAS-137861 / 26.04 / Fix editing a zvol

william-gr:
`inherit_encryption` should not be sent to the API on update.

**Testing:**

Add a zvol then try to edit it (e.g. change size)


william-gr:
time 0:20

aervin (approved):
Works fine here

PR #12663, opened 2025-10-08
NAS-137683: Main nav menu navigation issue

AlexKarpov98:
Testing: see ticket.
I implemented the desired behaviour, the idea:

The behaviour of clicking the left side Datasets button should always be the same, regardless of what is currently loaded in the main window. It's something of a POLA (path of least astonishment) violation to have that button press act in a different way just because the datasets page is currently loaded and showing details of some specific dataset/zvol...

Preview:

https://github.com/user-attachments/assets/d57ee6ba-c156-40c8-912d-6b29b360f9f6

As well - i've cleaned up, removed unnecessary code, improved mobile view accessibility.

### DIFF
--- a/src/app/modules/master-detail-view/master-detail-view.component.html
+++ b/src/app/modules/master-detail-view/master-detail-view.component.html
@@ -7,7 +7,7 @@
     <div
       class="details-container"
       ixDetailsHeight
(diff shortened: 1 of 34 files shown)


AlexKarpov98:
backport

PR #12900, opened 2025-12-03
NAS-138641 / 26.04 / Fix validation errors not displaying on form init

william-gr:
- Handle errors immediately in ix-errors when control already has errors
- Update directory services timeout validation to 5-60 seconds

**Changes:**

Screenshot 2025-12-03 at 10 18 48 AM

**Testing:**

Use a form (in this case directory services) and mock the API response to return a value outside the validation, in this case we saw the issue on directory services.

### DIFF
--- a/src/app/modules/forms/ix-forms/components/ix-errors/ix-errors.component.html
+++ b/src/app/modules/forms/ix-forms/components/ix-errors/ix-errors.component.html
@@ -1,4 +1,4 @@
-@if (control().touched || control().dirty) {
+@if (control().touched || control().dirty || showErrorsForUntouched) {
   <div class="form-error" role="alert" aria-live="polite">
     @for (message of messages; track trackMessage(message)) {
       <mat-error>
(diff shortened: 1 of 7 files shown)


AlexKarpov98 (commented):
when I set invalid data, SAVE button is enabled and not clickable. (bad UX)

https://github.com/user-attachments/assets/102ca458-0ee3-41df-a92f-77c5d2481133

william-gr:
> when I set invalid data, SAVE button is enabled and not clickable. (bad UX)
> 
>  Screen.Recording.2025-12-05.at.14.44.39.mov

good catch! claude must have broken it addressing reviews

AlexKarpov98 (commented):
All forms across the APP now show the required field error on init:

https://github.com/user-attachments/assets/fb834956-dc18-462b-a53e-110dfddffeea

william-gr:
> All forms across the APP now show the required field error on init:
> 
>  Screen.Recording.2025-12-08.at.11.04.37.mov

Great catch! That would have been a mess!
I was skeptical to change ix-errors because of regressions like that but it does seem like a generic problem of not showing errors when editing things for data coming from backend that we want to flag on early.
Tried to make a change please give it a try now.

AlexKarpov98 (approved):

william-gr:
time 2:00

william-gr:
backport

PR #13476, opened 2026-04-14
NAS-140311 / 27.0.0-BETA.1 / Improve pool vdev construction for special (metadata) vdevs

AlexKarpov98:
Testing: see ticket.

  ## Summary
  - Centralize vdev layout resolution into reusable resolveTopologyLayout() and existingVdevLayout() utilities, eliminating    duplicated topology-type-to-layout conversion logic across data/log/metadata/dedup wizard steps                          
  - Filter dRAID layouts from special (metadata) and dedup vdev steps — these vdev types don't support dRAID, so now they only show non-dRAID equivalents with matching parity                                                                             
  - Harden topologyToPayload() to skip empty categories and throw on missing layout, preventing invalid API payloads             
  - Add tests for existing-pool layout locking in metadata and dedup wizard steps                                   
  - Remove now-unused getLayoutsForVdevType() from pool manager store 
  
  ### Downstream

|Affects         |Reasoning
|Documentation   | New helper text added (  - Filter dRAID layouts from special (metadata) and dedup vdev steps — these vdev types don't support dRAID, so now they only show non-dRAID equivalents with matching parity                                                                             )
|Testing         |


aervin (approved):
Very nice

PR #13482, opened 2026-04-15
NAS-139220 / 27.0.0-BETA.1 / Improve chunk load error handling

AlexKarpov98:
Summary
                                                                                                                                 
  - Broadened chunk error detection to cover both Webpack (Loading chunk N failed) and esbuild/Vite (failed to fetch dynamically 
  imported module) error formats.
  - Added a reload guard using sessionStorage timestamps to prevent infinite reload loops — if a reload was already attempted    
  within 30 seconds, shows a user-facing fallback page instead of looping.                                                       
  - Wrapped logic in try/catch so restricted sessionStorage environments (e.g., some private browsing modes) fall back to a
  single reload.                                                                                                                 
  - Extracted handler into handle-chunk-load-error.ts for testability; added cleanup initializer to clear the guard on successful
   app boot.                                                                                                                     
                                                                                                                               
  Test plan                                                                                                                      
                                                                                                                               
  - Unit tests added (6 tests covering: first failure reload, dynamic import failure, recent-attempt fallback, expired-attempt   
  retry, sessionStorage exception, non-chunk error passthrough)
  - Verify on real hardware: update the server while a UI tab is open, navigate to a lazy route, confirm it reloads and recovers 
  - Verify fallback page appears if the reload doesn't fix it (e.g., disconnect network, navigate to a lazy route twice)

### DIFF
--- a/src/app/helpers/handle-chunk-load-error.spec.ts
+++ b/src/app/helpers/handle-chunk-load-error.spec.ts
@@ -1,4 +1,4 @@
-import { NavigationError, Event as RouterEvent } from '@angular/router';
+import { NavigationError } from '@angular/router';
 import { chunkReloadKey, handleChunkLoadError, resetReloadedFlag } from 'app/helpers/handle-chunk-load-error';
 
 describe('handleChunkLoadError', () => {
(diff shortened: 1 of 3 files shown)


aervin (approved):
Nice fix. Works as described

PR #13485, opened 2026-04-16
NAS-140660 / 27.0.0-BETA.1 / Snapshot created by a Replication Task defaults to a 2 week Lifetime

AlexKarpov98:
Testing: see ticket.

Preview:

https://github.com/user-attachments/assets/73a253f5-29e0-441f-b236-0cd2e22451a6

### Downstream

|Affects         |Reasoning
|Documentation   | Yes - new "Source Snapshot Lifetime" field added to the Replication Wizard's "When" step; user-facing  help text and tooltip need to be documented 
|Testing         |  Yes - unit tests added for the new source snapshot lifetime controls in both ReplicationWhenComponent and  ReplicationWizardComponent; QA should verify the field appears only when source is local and schedule is cron, and that the lifetime value is passed to pool.snapshottask.create


aervin (approved):

aervin:
@AlexKarpov98 I suppose docs and QE may like a heads up about these changes.

PR #13506, opened 2026-04-21
i18n: add Greek translations for common UI actions, pagination, and status states

chrislongros:
Adds Greek translations for commonly-reused UI items.


aervin (approved):
many thanks

PR #13511, opened 2026-04-22
NAS-140753: Introduce ix-form wrapper and migrate initial forms

AlexKarpov98:
## Summary                                                                                                                     
                                                                  
  Addresses the per-form scalability concern raised on the original NAS-139313 revert: rather than touching every form           
  individually to change behavior, this PR introduces a generic `<ix-form>` wrapper that centralizes common form chrome and
  lifecycle, and migrates a first batch of forms to prove out the API.                                                           
                                                                  
  ### New component: `IxFormComponent`                                                                                           
  
  Located at `src/app/modules/forms/ix-forms/components/ix-form/`. Handles:                                                      
                                                                  
  - **Common chrome** — modal header, card wrapper, save button, form actions slot, optional `ixExtraActions` slot for inline    
  buttons (e.g. "Advanced Options").                              
  - **Auto title** — `addTitle` / `editTitle` inputs resolve based on mode; explicit `title` still wins.                         
  - **Change tracking** — captures an initial snapshot (from `editData` auto-patch or an external `initialFormSnapshot` for async
   setups) and provides `changedValues` (diff vs. snapshot) alongside `allValues` to the submit handler. This is what unblocks   
  the original "only send changed properties" goal without per-form plumbing.                                                    
  - **Loading state** — combines internal submit loading with an `externalLoading` input; drives header progress bar and disables
   save.                                                                                                                         
  - **Dirty confirmation** — auto-registers with `SlideInRef` when present.
  - **Submit lifecycle** — `submitHandler` returns `{ request$, successMessage, onSuccess?, onError? }`; component handles       
  loading toggle, success snackbar, close, and delegates validation errors to `FormErrorHandlerService`.                         
                                                                                                                                 
  ### Forms migrated in this PR                                                                                                  
                                                                  
  - Group form
  - Cloud Sync form
  - Zvol form
  - Kernel form
  - NTP Servers form                                                                                                             
                     
  Remaining ~190 forms will be migrated in follow-ups.                                                                           
                                                                                                                                 
  ## Test plan
                                                                                                                                 
  - [ ] Group add/edit — create, edit, verify only changed fields are sent
  - [ ] Cloud Sync add/edit — including advanced-options toggle via `ixExtraActions`                                             
  - [ ] Zvol add/edit — covers the async-setup path using `initialFormSnapshot`     
  - [ ] Kernel form submit                                                                                                       
  - [ ] NTP Servers add/edit                                                                                                     
  - [ ] Static Route add/edit                                                                                                    
  - [ ] Dirty-state confirmation prompt when closing slide-in with unsaved changes                                               
  - [ ] Error handling surfaces validation errors on the right form controls      
  - [ ] Loading state disables save button and shows header progress bar


AlexKarpov98:
@aervin I added few changes to the code. found few more things we needed to have.

  New <ix-form> capabilities:
  - dirtyPredicate — override the dirty-confirmation check for multi-FormGroup forms.
  - requireDirty — gate Save while pristine.
  - extraDisabled — extra Save-disabled signal for cross-form gates.
  - suppressSuccessSnackbar — skip the toast for config-builder forms.
  - transformEditData, preSubmit, onCancel lifecycle hooks.

  Forms migrated to <ix-form>:
  - storage-settings-form, alert-service (with extraDisabled + dynamic child form), base-namespace-form (used by
  add-subsystem-namespace and namespace-form).
  
  I created another ticket related to your q-n regarding partial API updates support: https://ixsystems.atlassian.net/browse/NAS-140973

william-gr:
This is definitely an improvement on what we have, I think.
However I think this is not exactly what I was envisioning.
We are handling all the form high level handling this way, however the forms content is still very much a free for all, being up to each page to define each field type/component.

I was thinking of something way more rigid where the form definition is all defined up front by the form interface.
Otherwise we are still going to have every form potentially looking very different from one another.

I dont know if this is the way to do it or not.

AlexKarpov98:
@william-gr @aervin 

I talked to Claude about it and here are thoughts I agree with:

  **1.** TrueNAS already tried this and moved away from it. The old EntityFormComponent + FieldConfig[] system was exactly "the form is defined up front by a config object." The whole ix-forms effort was a deliberate migration away from that, because the config  approach didn't scale: weak typing of the form value, giant config blobs, hard-to-debug indirection, and an ever-growing pile of  escape hatches for anything non-trivial. Proposing full rigidity is, in effect, proposing to walk that back — so the burden of proof   is "what's different this time?"
  **2.** Rigidity and TrueNAS forms are in tension. A lot of our forms are not flat field lists — they have conditional fields, cross-field    validation, async options from the API, custom widgets (explorer, lists, dynamic sections). A truly rigid schema either can't   express those, or it grows custom field types + conditional config until the "schema" is just a less-typed, harder-to-debug   programming language. At that point you've reabsorbed all the flexibility but lost the type-safety and clarity that templates give   you — and ironically the escape hatches reintroduce the inconsistency you were trying to kill.
  **3.** Consistency is mostly a building-blocks + conventions problem, not necessarily an engine problem. A huge amount of "forms looking   different" is solvable by making the field and layout components opinionated and standard (which this PR's high-level handling is   already doing), plus conventions/lint, rather than forcing every form through one renderer.

So my honest opinion: don't commit the whole codebase to a rigid schema-driven engine on faith. The high-level handling in this PR is  a net win and is orthogonal — land it.

william-gr:
I am skeptical about that review from Claude, it seems to be missing some context:
 - Why previous attempt didn't succeed? Wasn't it a bad designed abstraction layer combined with poor coding skills?
 - Dozens of forms repeat the same pattern, with CRUD, which seems to make sense having a more rigid structure for these forms. It does not mean every single form has to use this abstraction layer, but the ones that repeat the same pattern can/should. It significantly reduces the amount of code and allows for more contained fixes that apply to a huge chunk of the UI. e.g. If we were to change how we present some set of fields it would be one targeted change, not a change of dozens of forms separately.
 - Abstraction layer could still define a pattern for the most common workflows -- if we have that many different workflows that seems to be leaking a design problem and bad UX experience -- why there are so many different ways to handle forms in the first place?
 - We could argue that consistency is solved by enforcing this in code, no amount of convention or linting will solve the problem of repeating the same patterns over and over again. And if we want to eventually change a pattern have to touch every single file that follows the same pattern. That is to some extent code duplication and architecture leakage.
 - In the past API was way more rigid, backend team is more willing to accept changes to support our UX.
 - Lastly if the UX is very similar for most of the forms, why would we need a custom implementation for every one of them? Majority has: slide in visual, a list of fields (input texts, select boxes, path selector), each with validation (some local, some through API) and possibly a help text.


I think its all worth more consideration before we make a decision.
Perhaps you feed that into Claude and see what it says?

AlexKarpov98:
@aervin so together with @william-gr we discussed it further.

I came up with an example of the FormDefinition: (see below message)
  
  **SO the ultimate goal is to :** 
- avoid repeating ourselves as much as we are today, many forms look very much the same, only field names/types changing and the API calls we use

  And I did some (few rounds of) research - how many forms can be simplified.
  
Screenshot 2026-05-29 at 16 38 03
Screenshot 2026-05-29 at 16 39 41

It's +- 43% of the forms - so basically a good number! 

@aervin what do you think about such `FormDefinition` approach for simpler forms - to get rid of lot's of similar repeatable code?

`review the audit below: 👇 `

AlexKarpov98:
# ix-forms: declarative `FormDefinition` — feasibility audit

> Context: follow-up to the `ix-form` wrapper PR (NAS-140753). The wrapper standardizes
> form **chrome/lifecycle** but each form still hand-writes its **fields** as a template.
> The open question (raised by @william-gr) is whether the field content itself should be
> declared up front as a typed `FormDefinition` rendered by a single `<ix-form-renderer>`,
> e.g.:
>
> ```ts
> protected formDefinition: FormDefinition<NtpServer> = {
>   addTitle: T('Add NTP Server'),
>   editTitle: T('Edit NTP Server'),
>   requiredRoles: [Role.NetworkGeneralWrite],
>   submit: (values) => this.api.call('system.ntpserver.create', [values]),
>   fields: [
>     { name: 'address', type: 'input',    label: T('Address'),  required: true },
>     { name: 'burst',   type: 'checkbox', label: T('Burst') },
>     { name: 'minpoll', type: 'input',    label: T('Min Poll'), inputType: 'number' },
>     { name: 'maxpoll', type: 'input',    label: T('Max Poll'), inputType: 'number' },
>   ],
> };
> ```
> ```html
> <ix-form-renderer [definition]="formDefinition" />
> ```
>
> This memo quantifies **how many of our forms this could actually replace**, what it saves,
> and where the boundary is.

---

## TL;DR

Audited **all 70 slide-in forms** against the `FormDefinition` idea (deterministic script,
then manual verification of every borderline case):

- **~21% (15 forms)** work with the *simple* definition **today** — zero new renderer features.
- **+~14% (10 forms)** become feasible with **one** added capability: declarative conditional
  visibility/validation (`visibleWhen` / `requiredWhen`). → **~36% reachable at low risk.**
- **Hard ceiling ≈ 43%.** The other **~57%** should stay on the current template approach
  (custom widgets, FormArrays, wizards, **schema-driven dynamic fields**, payload reshaping).
- **5 forms that *looked* simple are actually schema-driven/dynamic** (exporters, acmedns,
  dataset, widget-group, system-security) — i.e. the exact shape that sank the old
  `EntityForm`/`FieldConfig[]`. These are the line we must **not** cross.
- **LOC:** ~4–5k lines removed across the easy+reachable set, **almost entirely from
  eliminating templates** — not from deleting business logic.

**Suggested path:** ship the simple definition and migrate the ~15 easy forms (Phase 1, no
risk). Treat conditional-visibility support (Phase 2, unlocks ~36%) as a **separate decision**.
Never force the hard ~57% through the engine.

---

## Method (reproducible)

Population = the 70 slide-in forms (components matching `*form*.component.ts` under `pages/`
that inject `SlideInRef` **and** build a `FormGroup`). 28 other `*form*` components are
excluded (non-slide-in / inline page forms / no FormGroup).

A ~90-line script scans each form's `.ts` + `.html` and flags signals:

| Signal | Meaning |
|---|---|
| custom widget (`ix-explorer`, `ix-scheduler`, `ix-list`, `ix-file-input`, `code-editor`, …) | essential complexity |
| `FormArray` / `formArrayName` | dynamic collection |
| `MatStepper` / wizard | multi-step |
| `.valueChanges` subscriptions | cross-field / dependent logic |
| `@if` / `*ngIf` in template | conditional rendering |
| `setValidators` / `addValidators` | runtime validator juggling |
| `transformEditData` / `setupForm` | payload/edit reshape |

Forms with **only** `@if` and/or `.valueChanges` (no hard signals) were marked *Reachable*
and then **read by hand** to confirm — that manual pass is where the picture sharpened
(see below). Numbers are reproducible; the classifier is deterministic.

---

## Results — the spectrum

| Bucket | Count | % of 70 | Meaning |
|---|---:|---:|---|
| **Easy / Easy+** | 15 | 21% | Flat field list, primitives only. Drop-in with the simple definition **today**. |
| **Clean reachable** | 10 | 14% | Only blocker is conditional visibility / one simple dependency. The Phase-2 sweet spot. |
| **Heavy reachable** | 5 | 7% | Reachable in principle, but at the painful end (mode switches, side-effects, reshape). |
| **Complex — keep bespoke** | 40 | 57% | Custom widgets, FormArrays, wizards, schema-driven dynamic fields, payload reshaping. |

- **Drop-in today:** ~21%
- **Low-risk target (Easy + Clean reachable):** ~36%
- **Hard ceiling (incl. Heavy reachable):** ~43%

### The key manual-verification finding

The mechanical scan put **20** forms in *Reachable*. Reading them, **5 are not reachable —
they are hard cases hiding behind a cheap-looking `@if`/`valueChanges`**, and they are exactly
the EntityForm trap shape:

| Form | Why it's actually hard |
|---|---|
| `reporting-exporters-form` | **Schema-driven** — `type` change rebuilds the field set from a fetched schema |
| `acmedns-form` | **Schema-driven** attributes + custom cross-field validation (cloudflare rules) |
| `dataset-form` | Composed of **child section sub-form components**; preset-driven |
| `widget-group-form` | Dynamically **builds slot structure** from the chosen layout |
| `system-security-form` | Rich conditional **non-field UI** (STIG requirement lists, action links) |

Takeaway: a purely mechanical count **over-states** reachability, and every form it
over-stated is a dynamic/schema form — the cases that must stay bespoke. Only manual review
finds the real line.

---

## Lines of code

The 35 candidate forms (Easy + the mechanically-flagged Reachable) total **9,399 lines**
(`.ts` 7,045 / `.html` 2,354).

- **Certain — templates vanish:** 2,354 lines of `.html` → ~35 lines (one `<ix-form-renderer>`
  each). **~2,300 lines gone, not an estimate.**
- **`.ts` — partial:** the FormGroup block + submit/lifecycle boilerplate collapses into the
  definition, but imports, options-loading, and API wiring survive.
  - Worked example, `ntp` (Easy): 66 `.ts` + 57 `.html` = 123 → ~45 lines (**~63% off**).
  - Reachable forms average ~250 `.ts` lines of surviving logic, so they shrink far less.
- **Realistic total: ~4,000–5,000 lines (~45–55%), dominated by template elimination.**

So @william-gr's "thousands of lines" is **validated** (~4–5k). It does **not** come from
deleting business logic — boundary forms (see `email-form` below) come out roughly
**LOC-neutral**; their payoff is *uniformity*, not line count. The big LOC wins are the
genuinely-easy forms.

---

## The one design decision that matters

Phase 1 (the `ntp` shape) needs only `fields: [{ name, type, label }]`. Reaching the
conditional forms needs `visibleWhen` / `requiredWhen` — and the entire "how complex does this
get" question lives in **one type**:

```ts
type Condition<T> =
  | { field: keyof T & string; equals: unknown }   // declarative — SAFE
  | { field: keyof T & string; in: unknown[] }     // declarative — SAFE
  | { field: keyof T & string; truthy: true }      // declarative — SAFE
  | ((value: T) => boolean);                        // predicate  — ESCAPE HATCH

interface BaseField<T> {
  name: keyof T & string;
  label: TranslatedString;
  tooltip?: TranslatedString;
  validators?: ValidatorFn[];
  visibleWhen?: Condition<T>;
  requiredWhen?: Condition<T>;
  disabledWhen?: Condition<T>;
}
```

The first three `Condition` forms are declarative and analyzable. The `(value) => boolean`
predicate is the escape hatch. **The discipline rule writes itself: predicate conditions are a
review smell — lint can flag them.** When most forms reach for the predicate, the "definition"
has become the template again, in a worse language. That regression is precisely what killed
`EntityForm`.

Beyond conditions, a **small, fixed** set of top-level escape hatches covers the
form-specific bits without reinventing a config language:

- `custom` field type (render an arbitrary component)
- `extraActions` (custom buttons)
- `toPayload` (payload reshape)
- `disableSubmitWhen` (custom submit-enabled rule)

> **The EntityForm tripwire:** hold that set *fixed*. The day someone proposes a *fifth*
> hatch type to fit one more form is the day that form stays bespoke instead.

---

## Concrete examples

### 1. Clean reachable — `host-form` (all upside, zero escape hatches)

```ts
fields: [
  { name: 'hostnqn', type: 'input', label: T('Host NQN'), required: true },
  { name: 'requireHostAuthentication', type: 'checkbox', label: T('Require Host Authentication') },
  { name: 'dhchapKey', type: 'input', label: T('Host Key'),
    visibleWhen: { field: 'requireHostAuthentication', truthy: true } },        // ✅
  { name: 'addDhKeyExchange', type: 'checkbox', label: T('Add DH Key Exchange'),
    visibleWhen: { field: 'requireHostAuthentication', truthy: true } },        // ✅
  { name: 'dhchapCtrlKey', type: 'input', label: T('Controller Key'),
    visibleWhen: { field: 'addDhKeyExchange', truthy: true } },                 // ✅
]
```

### 2. Boundary — `email-form` (declarative skeleton + 4 escape hatches)

Picked deliberately because it sits *on the line*: it shows both the win and where the hatch
kicks in. (Field names/types are illustrative.)

```ts
protected definition: FormDefinition<EmailFormValue> = {
  title: T('Email Options'),
  requiredRoles: [Role.AlertWrite],
  loadData: () => this.api.call('mail.config'),   // singleton: renderer fetches + patches

  fields: [
    { name: 'sendMethod', type: 'radio', label: T('Send Mail Method'), options: sendMethodOptions },

    { name: 'fromemail', type: 'input', label: T('From Email'), validators: [emailValidator()],
      requiredWhen: { field: 'sendMethod', in: [MailSendMethod.Smtp, MailSendMethod.Outlook] } },  // ✅ declarative
    { name: 'fromname', type: 'input', label: T('From Name') },

    // SMTP section: one group, shown only for SMTP
    { type: 'group', visibleWhen: { field: 'sendMethod', equals: MailSendMethod.Smtp }, fields: [   // ✅ declarative
      { name: 'outgoingserver', type: 'input', label: T('Outgoing Mail Server') },
      { name: 'port', type: 'number', label: T('Mail Server Port'), validators: [portRangeValidator()],
        requiredWhen: (v) => v.sendMethod === MailSendMethod.Smtp && v.smtp },   // ⚠️ predicate (two-field)
      { name: 'security', type: 'select', label: T('Security'), options: securityOptions },
      { name: 'smtp', type: 'checkbox', label: T('SMTP Authentication') },
      { name: 'user', type: 'input', label: T('Username'),
        visibleWhen: { field: 'smtp', truthy: true }, requiredWhen: { field: 'smtp', truthy: true } },  // ✅
      { name: 'pass', type: 'password', label: T('Password'),
        visibleWhen: { field: 'smtp', truthy: true } },  // ✅
    ]},

    // OAuth UI: NOT a field → custom component
    { type: 'custom', component: EmailOauthSectionComponent,                    // 🚪 escape hatch
      visibleWhen: { field: 'sendMethod', in: [MailSendMethod.Gmail, MailSendMethod.Outlook] } },
  ],

  extraActions: [{ label: T('Send Test Mail'), roles: [Role.MailWrite], handler: () => this.sendTestEmail() }], // 🚪
  disableSubmitWhen: (v) => v.sendMethod !== MailSendMethod.Smtp && !this.hasOauthCreds(),                       // 🚪
  toPayload: (v) => this.prepareConfigUpdate(v),                                                                // 🚪

  submit: (payload) => this.api.call('mail.update', [payload]),
  successMessage: T('Email settings updated.'),
};
```

**Deleted (the win):** the entire 118-line template; both `.valueChanges` blocks (they only
existed to call `updateValueAndValidity()` — the renderer owns conditional-required now); the
manual `onSubmit` lifecycle, `loadEmailConfig`, and FormGroup boilerplate. The standalone
`sendMethodControl` (currently outside the FormGroup) folds back in as a normal field — cleaner
than today.

**Stayed (the catch):** four escape hatches — the OAuth section becomes a new component, plus
`sendTestEmail`, `prepareConfigUpdate`, and the custom validity survive as component methods.
Net result for a form like this: roughly **LOC-neutral**; the payoff is *consistency + deleted
plumbing*, not line count.

---

## Recommendation

1. **Phase 1 (no risk):** land the simple `FormDefinition` + renderer; migrate the ~15 Easy
   forms. Proves the pattern, kills boilerplate, commits to nothing.
2. **Phase 2 (separate decision):** add declarative `visibleWhen`/`requiredWhen` + the fixed
   escape-hatch set. Unlocks up to ~36% at low risk (~43% if we also take the heavy tier).
   Enforce the **predicate-condition lint** so the engine doesn't drift into a DSL.
3. **Never:** push the ~57% hard tail (custom widgets, FormArrays, wizards, schema-driven
   dynamic forms, payload reshaping) through the engine. Those keep the current
   `ix-form` + template approach, which is testable, readable, and debuggable.

Both positions in the thread are right, quantified: there's a real ~21% now and a plausible
~36–43% ceiling — and a hard ~57% floor that must remain bespoke.

---

### Appendix — bucket membership

**Easy / Easy+ (15):** ntp-servers, kernel, console, audit, tunable, self-encrypting-drive,
isolated-gpus, static-route, bootenv, jbof, dataset-quota-edit, pull-image, ssh-keypair,
gui (general-settings), kerberos-realms.

**Clean reachable (10):** host (nvme-of), access, api-key, dataset-quota-add,
storage-settings, disk, vmware-snapshot, docker-registry, port (nvme-of), tier-config.

**Heavy reachable (5):** email, ssh-connection, failover, privilege, snapshot-add.

**Moved Reachable → Hard on manual review (5):** reporting-exporters, acmedns, dataset,
widget-group, system-security.

**Complex / bespoke (remainder, 40 total)** — incl. smb, nfs, cloudsync, cloud-backup,
rsync-task, snapshot-task, user, group, interface, ipmi, vm-edit, device, container,
custom-app, cloud-credentials, iscsi (portal/target/extent/authorized-access), syslog,
zvol, init-shutdown, allowed-addresses, directory-services, kerberos-keytabs, cron, scrub, …
(custom widgets, FormArrays, wizards, schema-driven, or payload reshaping).

aervin:
After discussing with @AlexKarpov98 , we're agreed that the declarative object approach is a net win for simple CRUD forms. We will incrementally migrate the simple forms to this approach for 27. As for the more complex forms, we will reassess our current approach after the simpler forms are migrated.

src/app/modules/forms/ix-forms/components/ix-form-renderer/ix-form-renderer.component.html
```diff
@@ -0,0 +1,76 @@
...
+  [initialFormSnapshot]="loadedSnapshot()"
+  [title]="title"
+  [addTitle]="addTitle"
+  [editTitle]="editTitle"
+  [requiredRoles]="requiredRoles"
+  [submitHandler]="definition().submit"
+>
+  @for (section of sections; track $index) {
+    @if (isSectionVisible($index, section)) {
+      <ix-fieldset [title]="section.title" [tooltip]="section.tooltip">
```
aervin: there's a new form-section component in the library. should we use it here?

aervin (changes requested):

src/app/modules/forms/ix-forms/components/ix-form-renderer/ix-form-renderer.component.html
AlexKarpov98: Yeah, gonna use it!

AlexKarpov98:
@aervin please have a look at it now 🙌

src/app/modules/forms/ix-forms/components/ix-form-renderer/form-definition.interface.ts
```diff
@@ -0,0 +1,224 @@
...
+ * attributes should stay on the hand-written `<ix-form>` + template approach.
+ *
+ * The host component owns the definition as an instance property, so it can
+ * close over injected services (api, translate, slide-in data) in `submit`
+ * and in any `validators`.
+ *
+ * @example
+ * protected definition: FormDefinition<CreateNtpServer> = {
+ *   addTitle: T('Add NTP Server'),
+ *   editTitle: T('Edit NTP Server'),
```
william-gr, line 26: Do we need one for each? Perhaps we can default `Add {title}` and `Edit {title}` ?

AlexKarpov98: I believe we should keep it, I don't think we have all the remaining parts after Add or Edit translated in our app. 
I'll claude it and get back :)

AlexKarpov98: @william-gr 
  ▎ Looked into it — we should keep both. Two reasons:
  ▎
  ▎ 1. We already have 84 whole-phrase T('Add X') / T('Edit X') markers extracted and translated across all locales; Add {item} 
  ▎ interpolation exists in exactly one spot (an aria-label in ix-list), and there's no Edit {item} key at all.
  ▎ 2. More importantly, Add {item} / Edit {item} only works for English word order. Other locales put the verb last, add articles, or 
  ▎ decline the noun (e.g. German "NTP-Server hinzufügen", French "Ajouter un serveur NTP"), so translators need the full phrase to get
  ▎ it right — interpolation would regress titles that are already correctly localized.
  ▎
  ▎ For the rare case where one string is preferred, the interface already has an explicit title? that overrides both. Keeping the 
  ▎ example as-is.

william-gr (approved):
I think this is great, thanks for working on this.

Perhaps drop the system/advanced changes for now?

AlexKarpov98:
@william-gr I resolved all the conflicts, I think we can go with all the changes provided 🙌

PR #13518, opened 2026-04-23
i18n: add Greek translations for core UI actions, states, and field labels

chrislongros:
Adds Greek translations for a batch of core UI strings: common actions (lock/unlock, hold/keep, push/pull, sign in/out, review, renew, wipe, prune) etc

`yarn validate-translations` passes.


aervin (approved):
thank you

PR #13553, opened 2026-05-05
NAS-140864 / 27.0.0-BETA.1 / Revert / relax special-vdev validation introduced in NAS-140311 to unblock ER-72 testing

AlexKarpov98:
## Summary                                                                                                                     
  - Replaces the strict redundancy lock on special/dedup vdev pickers with a permissive one: any non-dRAID layout is now
  selectable regardless of the data vdev's parity (Stripe still gated on data being Stripe).                                     
  - Surfaces non-blocking warnings in `PoolManagerValidationService` when (a) the special/dedup vdev's parity is lower than the
  data vdev's, or (b) a special/dedup vdev being added doesn't match the layout already present in that category on the existing 
  pool.                                                           
  - Simplifies `topology.utils.ts` accordingly: drops `parityLockForMinParity` and the existing-category branch in               
  `resolveParityLock`, and removes the now-unused `vdevType` arg from `parityLock$`.                                             
  
  ## Context                                                                                                                     
  Per ER-72 / NAS-140839, the wizard should no longer block users from picking a lower-redundancy metadata/dedup layout — it
  should warn instead. Pool integrity guidance (mirror of metadata vdev failure destroying the pool) is preserved as a warning   
  rather than a hard constraint.
                                                                                                                                 
  ## Test plan                                                    
  - [ ] Create a new pool with RAIDZ2 data + Mirror special — expect the metadata step to allow it and a redundancy-mismatch
  warning to appear in the review step.                                                                                          
  - [ ] Create a new pool with Stripe data — confirm Stripe is selectable for special/dedup; pick non-Stripe data and confirm
  Stripe disappears from the special/dedup layouts.                                                                              
  - [ ] Add a special vdev to a pool whose existing special class is Mirror, but pick RAIDZ1 — expect a mixed-layout warning, no
  block.                                                                                                                         
  - [ ] Add a dedup vdev with parity lower than data parity — expect dedup redundancy-mismatch warning.
  - [ ] dRAID is never offered for special/dedup.                                                                                
  - [ ] `yarn test src/app/pages/storage/modules/pool-manager/store/pool-manager-validation.service.spec.ts` and                 
  `…/utils/topology.utils.spec.ts` pass.                                                                                         
  - [ ] `yarn lint` clean on changed files.


aervin (approved):
Works as described. thanks a lot

PR #13559, opened 2026-05-07
NAS-140929 / 27.0.0-BETA.1 / Add inline "Create Dataset" to forms that select datasets/paths

AlexKarpov98:
Testing: see ticket.


aervin (approved):
works well

PR #13592, opened 2026-05-15
NAS-141074 / 27.0.0-BETA.1 / migrate shares dashboard to truenas components

aervin:
**Changes:**

**Testing:**

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


aervin:
time 4:00

aervin:
time 8:00

AlexKarpov98 (commented):
Looks great visually with the new component.

Few things:

`tn-icon-button` seems this component lacks accessibility - i can't use keyboard for it.

https://github.com/user-attachments/assets/2fb8d31c-055a-4f6c-958d-9bff42effd90

_____

For show actions - in all cards - will we use `tn-icon-button` component as well?

Image

Same for tables, switches?

____

seems like a repetetive pattern, would be nice to extract it somehow and not copy ourselves many times:

Screenshot 2026-06-04 at 20 14 45

_____

AlexKarpov98 (approved):
Looks good, all prev issues addressed, it looks like a brand new page 😄  👍 

Image

**NOTE:** Table header font size looks a big bigger than before

PR #13596, opened 2026-05-18
NAS-140955 / 27.0.0-BETA.1 / FAULTED Drive but VDEV appears fine

AlexKarpov98:
Testing: see ticket.

Preview:

Screenshot 2026-05-27 at 15 57 47


william-gr:
@AlexKarpov98 I am not sure I understand what we are doing here? Doesnt the status is reflected in ZFS? We should not show different output than zpool status.

AlexKarpov98:
> @AlexKarpov98 I am not sure I understand what we are doing here? Doesnt the status is reflected in ZFS? We should not show different output than zpool status.

Per-row statuses still match zpool status exactly — I only added a rolled-up effectiveStatus on the parent VDEV so it 
reflects its worst child. That's the ticket symptom: FAULTED disk under a VDEV that rendered ONLINE in the collapsed tree. Zpool status's flat output makes the leaf visible inline; our tree hides it under the parent, so the header has to surface it.

maybe it worth a fix on API side 🤔 @william-gr

william-gr:
@AlexKarpov98 seems something is off here. This is not what is showing in the ticket screenshot, he has a DEGRADED raidz with a UNAVAIL disk within it. The VDEV is DEGRADED.

AlexKarpov98:
@william-gr I changed it now.
updates:
Screenshot 2026-05-22 at 15 01 03

william-gr:
@AlexKarpov98 what do you think auto expanding a vdev when there is an error?

AlexKarpov98:
@william-gr good idea, implemented 👌

william-gr:
@AlexKarpov98 great, but i just noticed we are not flagging UNAVAIL? thats probably one of the most important but we are focused on errors?

william-gr (changes requested):
ping

AlexKarpov98:
> ping

Yeah
Sorry - been busy with other tickets 🙈
Gonna switch to this one today

AlexKarpov98:
@william-gr it's updated now - please have a look at it. 👌

william-gr (approved):
Looks good, thanks!

PR #13622, opened 2026-05-22
NAS-141063 / 27.0.0-BETA.1 / Migrate audit/ to tn-*

AlexKarpov98:
Before: 
Screenshot 2026-05-22 at 14 22 03

After:
Screenshot 2026-05-22 at 17 31 58


aervin (approved):
looks good! we need to decide whether to use the bordered tn-card variant or not.

PR #13652, opened 2026-06-02
NAS-141209 / 27.0.0-BETA.1 / Storage Widget doesnt show last scrub when more than 4 VDEVs

AlexKarpov98:
Testing: see ticket.

Preview: 
see ticket.

Screenshot 2026-06-02 at 17 02 52

You can use mock:
[mock-storage-5-pools.json](https://github.com/user-attachments/files/28513637/mock-storage-5-pools.json)


william-gr:
I am not sure this is even necessary? I mean, space is limited, there is only so much we should cram things in.
@aervin thoughts?

aervin:
I agree it's too cramped. We provide scrub status on the single pool widget anyway.

AlexKarpov98:
Okay, I'll close the ticket then. @aervin @william-gr

PR #13664, opened 2026-06-08
NAS-141274 / 27.0.0-BETA.1 /  Metadata and Dedup VDEV layout dropdowns allow selection of incompatible type when a VDEV of that type already exists in the pool

AlexKarpov98:
Testing: see ticket.

Preview:

https://github.com/user-attachments/assets/262d7aef-072d-45e6-bf82-ae354108ca43


william-gr (commented):
I am not sure this is the expected behavior for special vdevs. Did you verify the backend is raising an error?

AlexKarpov98:
@william-gr yes!
I've seen this error: 
image

https://github.com/user-attachments/assets/2ea313ce-4edd-4acf-aa25-3f3e430947fc

william-gr:
@AlexKarpov98 that should not be the case, so we need to:
 - Try with "force" parameter, if it still fails, raise a backend ticket
 - Do not enforce this for special vdevs

william-gr (changes requested):

AlexKarpov98:
Waiting for updates from @anodos325

AlexKarpov98:
@william-gr an API ticket created: https://ixsystems.atlassian.net/browse/NAS-141411

PR #13665, opened 2026-06-08
NAS-141262 / 27.0.0-BETA.1 /  Est. Usable Raw Capacity displays "NaN B" when dRAID spare count is 0

AlexKarpov98:
Preview:

NAS-141262


aervin (approved):
While we're touching this section of the codebase, do you mind cleaning up this warning message:

pool-manager-validation.service.ts:358:

Before: "In order for dRAID to overweight its benefits over RaidZ the minimum recommended number of disks per dRAID vdev is 10."
After: "For dRAID's benefits to outweigh those of RaidZ, we recommend at least 10 disks per dRAID vdev."

PR #13671, opened 2026-06-08
NAS-141277 / 27.0.0-BETA.1 / Allow dedicated spares for dRAID pools

william-gr:
**Changes:**

Self explanatory


aervin (approved):
works as expected

PR #13698, opened 2026-06-16
NAS-141062 / 27.0.0-BETA.1 / migrate reports dashboard to `tn-*`

mattwyatt-ix:
**Changes:**
migrate the reporting dashboard to use the components library.

| image |
| :---: |
| example screenshot |
| image |
| loading state |

**Testing:**
just examine the reporting dashboard.

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


aervin (changes requested):
Looking good. Some thoughts:

Alignment is a bit off here

Image

Should we go ahead and migrate these bits to tn-toggle and tn-select?
Image

`/migration-review` flagged some test ID changes. Try a run of `/migration-review-testid`

AlexKarpov98 (commented):
@mattwyatt-ix ping me once you address Aaron's review comments.

mattwyatt-ix:
@AlexKarpov98 would you mind to check this branch out now? thanks!

AlexKarpov98 (commented):
Previous comments fixed, I found some new things to fix:

1. Pull latest changes from master.
2. Reporting exporting table & elements is not migrated. (tn-table & related)
3. Add Reporting Exporter form should be migrated as well.
4. Add button does not look good ( no proper spacings ) :

Image

AlexKarpov98 (commented):
Looks good now!

Can you change how we open the form?
To open it with tn slideIn - please use `this.formPanel.open` 
Then save button will be in the footer (right corner)

see how groups form works: 

Image

mattwyatt-ix:
> Can you change how we open the form? To open it with tn slideIn - please use `this.formPanel.open` Then save button will be in the footer (right corner)

done! thanks for the tip. just let me know if there's anything else you notice that needs to change.

AlexKarpov98 (commented):
Almost there, sorry for one more comment, I believe this one should be the last one!

Would be nice to align form elements greater.

Image

Gaps between elements are different and top form has some padding on the left side.

________

As well don't forget to run `/migration-review` for each migration 😄 
I did it and pushed updates to save you some tokens haha

Image

mattwyatt-ix:
> Would be nice to align form elements greater.

good catches, and thank you for the help with the test IDs! I'm not as familiar with the NAS UI as I am with the TNC UI, so I appreciate it haha :heart: 

a couple things i'll address real quick:
* finally fixed the reporting exporters form. see screenshot below.
* there are a few test IDs that are still problematic since I changed the `tn-button` to `tn-select` for selecting a reporter, but I believe we'll just have to live with those, since we definitely shouldn't try to alter how `tn-select` generates IDs.
* the claude review above states "Shared dynamic-form SCSS change has blast radius beyond this page". this is intentional, since the changes to `ix-dynamic-form-item` in f1e353252655077f6ef881b258dee6332703652c were fighting the form changes on this branch. it *does* affect the ACME DNS form, but that form is unmigrated anyway, so it's not worth worrying about it yet.


| image |
| :---: |
| fixed reporting exporters form |

aervin:
Looking good @mattwyatt-ix . in the exporter form, only the "Name" field has a tooltip. is that the preexisting behavior? of all the form fields, Name needs the least explanation lol

mattwyatt-ix:
> Looking good @mattwyatt-ix . in the exporter form, only the "Name" field has a tooltip. is that the preexisting behavior? of all the form fields, Name needs the least explanation lol

@aervin yup, that's the previous behavior in both 25.10 and 26 haha. if we'd like, i can add tooltips to the rest of the fields in this PR, or i can leave it as-is and can create a separate ticket to add them. what do you think?

**edit:** after a short conversation, we decided to just make a ticket for the tooltips: https://ixsystems.atlassian.net/browse/NAS-141622

aervin (approved):

AlexKarpov98 (approved):

PR #13699, opened 2026-06-16
NAS-141423 / 27.0.0-BETA.1 / Restore disabled read-only state for shares dashboard actions

aervin:
Screenshot 2026-06-16 at 12 04 10 PM

Screenshot 2026-06-16 at 12 04 30 PM

### DIFF
--- a/src/app/pages/sharing/components/shares-dashboard/nvme-of-card/nvme-of-card.component.html
+++ b/src/app/pages/sharing/components/shares-dashboard/nvme-of-card/nvme-of-card.component.html
@@ -40,7 +40,7 @@
       *ixRequiresRoles="requiredRoles"
       color="primary"
       variant="filled"
-      testId="button-nvme-of-share-add"
+      testId="nvme-of-share-add"
       [label]="'Add' | translate"
       (onClick)="openForm()"
     ></tn-button>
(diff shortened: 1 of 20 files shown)


aervin:
time 0:45

src/app/pages/sharing/components/shares-dashboard/nvme-of-card/nvme-of-card.component.html
```diff
@@ -25,6 +23,29 @@ <h3 class="card-title">
...
+      ></tn-slide-toggle>
+    }
+  </ng-template>
+
+  <ng-template tnCardFooterActions>
+    <tn-button
+      *ixRequiresRoles="requiredRoles"
+      color="primary"
+      variant="filled"
+      testId="button-nvme-of-share-add"
```
AlexKarpov98: hm, can we use `testId="nvme-of-share-add"` ? 
'button' part should be added automatically, same for other tn-buttons

aervin: good catch. thanks

AlexKarpov98 (approved):
👍

PR #13703, opened 2026-06-17
NAS-141024 / None / Migrate webshare to tn-*

aervin:
Screenshot 2026-06-17 at 10 53 28 AM

Screenshot 2026-06-17 at 11 11 03 AM


aervin:
time 0:45

AlexKarpov98 (approved):
Looks good 👍 

A few things to address:

  ---
  🛑 Unrelated dead CSS added to nfs-form.component.scss

  src/app/pages/sharing/nfs/nfs-form/nfs-form.component.scss:1-11

  This file is in pages/sharing/nfs/, not webshare, and is the only change to the NFS area on this branch. It adds styling for
  tn-form-field / tn-checkbox, but the NFS form template uses neither (grep tn-form-field nfs-form.component.html → 0 matches). This
  looks like it leaked in from another migration branch or was committed by accident — it's dead CSS in an unrelated file. Please drop
  it from this PR (or move it to the branch that actually migrates the NFS form).

  ---
  🧹 Dead SlideIn mock in the list spec

  webshare-list.component.spec.ts:68, 280, 361

  WebShareListComponent no longer injects SlideIn (the private slideIn = inject(SlideIn) line was removed), so the three
  mockProvider(SlideIn, { open: … }) blocks — plus the SlideIn/SlideInResult imports — are now inert scaffolding. Worth removing so the
  spec reflects the new side-panel flow.

  ---
  🔁 Two divergent uniqueRowTag definitions

  webshare-list.component.ts

  createTable(...) is still given uniqueRowTag: (row) => row.name and ariaLabels: (row) => […], but tn-table renders cells from the
  template now, so those options are dead — and the template instead calls the new component methods uniqueRowTag(row) (→
  convertStringToId(row.name)) and ariaLabel(row). So the raw-name version and the slugified version coexist, and the createTable
  options mislead a reader into thinking they still drive rendering. Consider dropping the now-unused uniqueRowTag/ariaLabels from the
  createTable options. Also note the resolved data-test values shift from row.name to convertStringToId(row.name) versus master —
  intended, but flagging in case any e2e selectors depended on the old value.

PR #13712, opened 2026-06-19
NAS-141051 / 27.0.0-BETA.1 / Migrate credentials/Users to tn-*

AlexKarpov98:
Screenshot 2026-06-19 at 12 29 52


william-gr (changes requested):
Item 1 and 3 from Claude deserve an attention. For 3 we decided we dont want to hide buttons due to missing roles permission, disabled should be preferred.

william-gr (approved):
Image

Icon does not seem to be working, i tried restarting dev server, worth checking, might be local issue.

AlexKarpov98:
@william-gr fixed:

Screenshot 2026-06-25 at 14 56 56

PR #13713, opened 2026-06-19
NAS-141045 / 27.0.0-BETA.1 / Migrate system/advanced pages and forms to tn-* components

william-gr:


AlexKarpov98 (commented):
ix-manage-configuration-menu - is not migrated 

Image

And few form visual issues I found:

Image

Image

AlexKarpov98 (commented):
1. I see an issue with card titles:

Image
(have bigger font size)

- Allowed IP Addresses
- [Init/Shutdown Scripts](http://localhost:4203/system/initshutdown)
- Cron Jobs 


2. Manage configuration misses the caret icon:
before:
Image
Now:

Image

3. Found an issue with the tn-select component - it creates really large unscrollable viewport. Would be better to follow material logic - providing max height with scrollbar.
As well I can't click outside to close selectable options - only hitting ESC button closes the panel.
see:

https://github.com/user-attachments/assets/b89b18d9-3734-46c2-bd70-146ddeb79aa1

4. The slide in has some sort of gradient or set of different colors on the right side, not sure if that's intended.

Image

5. The list element is too tight to top 

Image

usually we have it like this:

Image

william-gr:
> 1. I see an issue with card titles:
> 
> Image (have bigger font size)
> * Allowed IP Addresses
> * [Init/Shutdown Scripts](http://localhost:4203/system/initshutdown)
> * Cron Jobs
> 
> 2. Manage configuration misses the caret icon:
>    before:
> 
> Image Now: Image
> 3. Found an issue with the tn-select component - it creates really large unscrollable viewport. Would be better to follow material logic - providing max height with scrollbar.
>    As well I can't click outside to close selectable options - only hitting ESC button closes the panel.
>    see:
> 
>  Screen.Recording.2026-06-22.at.21.55.47.mov 
> 4. The slide in has some sort of gradient or set of different colors on the right side, not sure if that's intended.
> 
> Image
> 5. The list element is too tight to top
> 
> Image
> usually we have it like this:
> 
> Image

1, 2 and 5 should have been resolved.
3 should be handled by component library (PR up).
4. does not seem to have been introduced by this PR, we can revisit later as its a generic side panel question.

AlexKarpov98 (approved):
Perfect!
Thanks for the updates 🚀

PR #13726, opened 2026-06-24
NAS-141056 / 27.0.0-BETA.1 / Migrate all-containers to @truenas/ui-components

william-gr:
image


AlexKarpov98 (commented):
Same primary delete - probably not ok ? 😄 

Image
_gonna review later today or tomorrow_

william-gr:
> Same primary delete - probably not ok ? 😄
> 
> Image _gonna review later today or tomorrow_

good catch!

AlexKarpov98 (commented):
@william-gr 
1. Will you migrate all forms related to containers in the separate PR? (for example - I see Create/Edit Container form/slidein is not migrated)

2. Maybe we can make delete red outlined button as we did for Apps?

3. Form spacing have large gaps - we need to get rid of margins for form elements because they are a part of a flex box which has a gap property (for every form).

Image

Image

4. Card elements spacing is huge - too big padding. (all similar cards)

Image

william-gr:
> @william-gr
> 
> 1. Will you migrate all forms related to containers in the separate PR? (for example - I see Create/Edit Container form/slidein is not migrated)
> 2. Maybe we can make delete red outlined button as we did for Apps?
> 3. Form spacing have large gaps - we need to get rid of margins for form elements because they are a part of a flex box which has a gap property (for every form).
> 
> Image Image
> 4. Card elements spacing is huge - too big padding. (all similar cards)
> 
> Image

1. Yes, there is a draft PR up
2. I dont think we want to use red for delete since we have a confirmation dialog
3. Thank you, i completely forgot to test the cards with items.

AlexKarpov98 (approved):
Good work 👍 

Fix these issues before merge:

Image

PR #13727, opened 2026-06-24
NAS-141548 / 27.0.0-BETA.1 / fix yaml boolean casing in event-data-details-card spec

william-gr:
A test broke after a dependency upgrade.


mattwyatt-ix (approved):
YAML seems like it's always causing problems 🙄. lgtm!

PR #13728, opened 2026-06-25
NAS-141057 / 27.0.0-BETA.1 / Migrate containers common/form/wizard to tn-* components

william-gr:


AlexKarpov98 (commented):
Works well but worth fixing UI visual issues I found:

1. Advanced options buttons is not aligned well to the left side of the form.

Image 

2. Env variables card lacks padding at the bottom.

Image

3. Select image filters - gap is too big between elements.

Image

4. Default network title no proper padding on top.

Image

AlexKarpov98 (approved):
Perfect 👍 

1 improvement may be made though while we are here:
Let's stick the filters panel 

https://github.com/user-attachments/assets/38c26e27-fb35-4b23-80c9-fd388feb06f7

william-gr:
Good catch

PR #13731, opened 2026-06-26
NAS-141042 / 27.0.0-BETA.1 / Migrate system/bootenv, update and alerts to tn components

william-gr:


AlexKarpov98 (commented):
1. Selecting category - I don't see full text for each option, as well mat-menu does not rememberize latest selected element, like select:

Image

2. I see Status Finished - but should not see it because it is collapsed.

Image

3. Boot env listing table is not migrated to tn-table.

Image

**Migration review results:**

Image

william-gr:
> 1. Selecting category - I don't see full text for each option, as well mat-menu does not rememberize latest selected element, like select:
> 
> Image
> 2. I see Status Finished - but should not see it because it is collapsed.
> 
> Image
> 3. Boot env listing table is not migrated to tn-table.
> 
> Image
> **Migration review results:**
> 
> Image

1. I had noticed this, but I imagine this is a component library fix
2. component library fix pushed
3. ugh, must have reverted my change, thanks for noticing

AlexKarpov98:
@william-gr 
for: `1. I had noticed this, but I imagine this is a component library fix` - let's create a ticket for that to not forget about it.

**Would be good to merge after verifying it here with component library fixes 🙌**

AlexKarpov98 (approved):

PR #13734, opened 2026-06-26
polish translation update

warchlak:
just more translated strings


william-gr:
Please open PR again once its ready. Thank you.

PR #13736, opened 2026-06-29
NAS-141393 / 27.0.0-BETA.1 / Migrate credentials/Backup Credentials to tn-*

AlexKarpov98:
https://github.com/user-attachments/assets/43d5ac41-6dfd-4a99-93cc-79edc388c8ee


aervin (approved):
Looks great. 

There's a library update which allows content projection into tn-banner. It'd be good to use it here imo

Image

AlexKarpov98:
@aervin - good point!

Screenshot 2026-07-01 at 00 00 51

PR #13744, opened 2026-06-29
NAS-141024 / None / migrate Shares -> NVMe-oF to tn-* components

aervin:


aervin:
time 5:00

william-gr (commented):
Image
Is this supposed to be changed by your PR?

i think the actions should be in the footer.

william-gr (approved):
It was a local devserver problem, looks good.

PR #13747, opened 2026-06-30
NAS-141422 / 27.0.0-BETA.1 / Show restricted-access warning instead of infinite shell spinner

AlexKarpov98:
Screenshot 2026-06-30 at 11 25 05

### DIFF
--- a/src/app/core/testing/classes/mock-auth.service.ts
+++ b/src/app/core/testing/classes/mock-auth.service.ts
@@ -1,5 +1,6 @@
 import { Injectable } from '@angular/core';
 import { BehaviorSubject, of } from 'rxjs';
+import { filter, map } from 'rxjs/operators';
 import { Role } from 'app/enums/role.enum';
 import { LoggedInUser } from 'app/interfaces/ds-cache.interface';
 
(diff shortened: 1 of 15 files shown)


william-gr (approved):

william-gr:
Should we not allow to user click in the shell to begin with?

AlexKarpov98:
> Should we not allow to user click in the shell to begin with?

Not sure what you mean, hmm

william-gr:
> > Should we not allow to user click in the shell to begin with?
> 
> Not sure what you mean, hmm

I mean, how did the user get to that screen in the first place if he has no access?

AlexKarpov98:
@william-gr , ah I see 
I thought about that scenario as well - I may try restricting user (with that button) - do you think it's better? 
But I'll keep the current solution as well as a direct interaction

william-gr:
> @william-gr , ah I see I thought about that scenario as well - I may try restricting user (with that button) - do you think it's better? But I'll keep the current solution as well as a direct interaction

Yeah, I would expect the button to be disabled with a lock icon or somesuch.

AlexKarpov98:
@william-gr updates arrived:

https://github.com/user-attachments/assets/0a138db7-4b48-4679-a6e3-b3fcb4ab618e

william-gr (approved):

PR #13762, opened 2026-07-01
NAS-141726 / 27.0.0-BETA.1 / Update pl.json (fill in missing Polish translations)

pht31337:
Filled in 828 missing translations in the Polish locale file (src/assets/i18n/pl.json) — every key that previously had an empty value now has a translation.

- Translations follow terminology already established in the file (e.g. "Zbiór danych" for Dataset, "Pula" for Pool, "Migawka" for Snapshot).
- All placeholders ({name}, {count}, etc.), HTML tags, and ICU plural syntax were preserved unchanged.
- The <poolname> placeholder in two strings was left untranslated, matching the convention used in other locales (fr/es/it).
- No previously existing translations were modified — only the untranslated entries were filled in.

**Changes:**

**Testing:**

### Downstream

|Affects         |Reasoning
|Documentation   |
|Testing         |


aervin (approved):
Dziękuję!

PR #13769, opened 2026-07-03
NAS-141392 / 27.0.0-BETA.1 / Migrate credentials/Directory services to tn-*

AlexKarpov98:
https://github.com/user-attachments/assets/5f8aa976-887e-471f-9a69-4999af4e8614


william-gr (commented):
What do you think of moving Settings to the actions buttons in the bottom? Seems to be something people will want to use more often.

AlexKarpov98:
@william-gr - yeah, I think it should be nice 👌

AlexKarpov98:
@william-gr will that work or you think it's better to remove 3 dots at all?

Screenshot 2026-07-03 at 17 11 25

william-gr (approved):
I like it, thanks! I am on the fence about button being primary or just a regular button unfilled.
Up to you.

PR #13770, opened 2026-07-03
NAS-141656 / 27.0.0-BETA.1 / Migrate credentials/Privileges to tn-*

AlexKarpov98:
https://github.com/user-attachments/assets/aca4b154-d291-40cc-8ab4-9994929d1958


aervin:
I think the table pager looks better with top border and bottom border radius. What do you think? In any case, it doesn't block this PR.

Screenshot 2026-07-07 at 10 52 00 AM

aervin (approved):

PR #13772, opened 2026-07-03
NAS-141391 / 27.0.0-BETA.1 / Use formPanel.open instead of duplicating <tn-side-panel />

AlexKarpov98:


william-gr (approved):
Nice improvement!
Can you review the first comment from claude about translate of saveLabel? Seems to make sense

PR #13777, opened 2026-07-06
NAS-141617 / 27.0.0-BETA.1 / Migrate mat-vertical-stepper to tn wizard/stepper

AlexKarpov98:
https://github.com/user-attachments/assets/b455a638-ad16-408d-a888-3540fb09c6d4


william-gr (approved):
Impossible to review the code with that amount of changes but I tested it and looks good. Good job

PR #13778, opened 2026-07-06
NAS-141040 / 27.0.0-BETA.1 / Migrate system / network to tn-*

AlexKarpov98:
https://github.com/user-attachments/assets/60f48107-f238-4494-b123-382ffe085530


william-gr (changes requested):
Looks good! I found one issue so far:

Image

Save is enabled even tough required fields have not been filled.
The NONE for protocol also feels strange, like its a selected value.

AlexKarpov98:
@william-gr good catch! thanks!
Please take a look at it now :)

william-gr (approved):
Thanks!

PR #13781, opened 2026-07-06
NAS-141626 / 27.0.0-BETA.1 / Remove High-Availability, Alert Setting options from non-HA Enterprise systems

AlexKarpov98:


william-gr (commented):
This looks odd. We seem to be hardcoding stuff.
Does alert.list_categories return these classes on non-HA?

AlexKarpov98:
@william-gr - ah, I agree!
yeah, would be good to fix  that on the API side indeed.

PR #13785, opened 2026-07-06
NAS-141032 / 27.0.0-BETA.1 / Migrate alerts, jobs and page-header modules to tn components

william-gr:


william-gr:
@AlexKarpov98 i am unsure about the button tags, do you think we should keep it that way?

AlexKarpov98 (approved):

PR #13796, opened 2026-07-08
NAS-141642 / 27.0.0-BETA.1 / Migrate datasets/ quotas to tn-*

AlexKarpov98:
https://github.com/user-attachments/assets/ab2eb234-7858-4626-8e0a-84f663b621a2


aervin (approved):
Looks good. 

I agree with Claude here:
```2. Empty SCSS still referenced — dataset-quotas-list.component.scss is now 0 bytes but styleUrls still points at it. Safe to delete the file and drop the styleUrls entry.```

PR #13797, opened 2026-07-08
NAS-141643 / 27.0.0-BETA.1 / Migrate datasets/ permissions/acl to tn-*

AlexKarpov98:
https://github.com/user-attachments/assets/48271bc3-00cf-4028-9abd-2d6141844a83


aervin (approved):
Nice work. Just a minor note:

`Restore accessible names on the four tn-selects (edit-nfs-ace ×2, edit-posix-ace, select-preset-modal) — one-line [ariaLabel]="… | translate" each; every select currently announces as generic "Select an option" to screen readers.`

PR #13812, opened 2026-07-10
NAS-141720 / 27.0.0-BETA.1 / TrueCloud Backup Task – Source Path and Cache Paths Fields Do Not Auto-Update with Newly Created Datasets

AlexKarpov98:


src/app/services/filesystem.service.ts
```diff
@@ -62,14 +62,17 @@ export class FilesystemService {
 
   private getTreeNodeProvider(options: ProviderOptions): TreeNodeProvider {
     return (node: TreeNode<ExplorerNodeData>) => {
+      // Return fresh copies of the shared root node constants. angular-tree-component caches loaded
+      // children on `node.data`, so handing out the singletons by reference would let a stale dataset
+      // list survive across form instances (only cleared by a full page reload). See NAS TrueCloud bug.
```
aervin: Let's reference the ticket number here instead of "TrueCloud bug"

aervin (approved):

PR #13820, opened 2026-07-13
NAS-141024 / None / migrate sharing/ iscsi to tn-* components

aervin:


william-gr (approved):
Image

Looks good. But shouldnt the associate button here be at the bottom of the card?

PR #13822, opened 2026-07-14
NAS-141024 / None / sharing/ tn-* migration (final pass)

aervin:
NAS-141024-final — review-driven cleanup of the sharing/ tn-* migration
Four commits on top of the merged iscsi work, driven by a full six-lane migration review of the branch. Net effect: every review blocker and warning is resolved, and the sharing area is fully off Angular Material (one documented hold), off | async provider bindings, and off the legacy [ixTest] directive.

Correctness fixes (review blockers)

Column picker i18n — the shared ix-table-column-picker now translates option labels; the raw column title stays as the option value since it's the saved-preferences key.
nvme-of test-id parity — dynamic test-id segments (NQNs, device paths, subsystem names) are pre-split with lodash kebabCase so digit-bearing values resolve byte-identically to the legacy ids (8 sites), and the text-*-row-text cell ids dropped during the tn-table migration were restored in subsystems-list and both manage dialogs.
Spec hygiene — the one forbidden data-test locator replaced with TnButtonHarness.
Signals over async pipe — every remaining list binds rows() / isLoading() / emptyType() via the shared dataProviderRows/dataProviderLoading helpers, which now also accept a Signal<provider> for input.required providers (target-list, subsystems-list) and rebuilt-on-input providers (smb-open-files). Provider construction moved to field initializers throughout.

Accessibility — the iscsi tn-tabs migration regained its role="tabpanel" container; webshare-list's name cell dropped the nonstandard role="text" and its home-share icon is now exposed to AT (role="img" + label); and — enabled by the library bump to 0.3.23, whose tn-form-field now auto-associates its label via aria-labelledby — the redundant [ariaLabel] overrides and stale TODO comments were removed from nfs-form, service-smb, smb-form, and smb-acl.

Consistency/debt — smb-list and the five smb-status stylesheets deduped onto the shared tn-table-in-card mixin; last [ixTest] consumers swapped to [tnTestId]; member scoping tightened (spec-driven seams kept public with comments); new spec coverage for the side-panel footerActions contract.

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "@smarttools/eslint-plugin-rxjs": "~1.0.22",
     "@stylistic/eslint-plugin": "~2.9.0",
     "@truenas/common-typescript": "github:truenas/tn-common-typescript#master",
-    "@truenas/ui-components": "~0.3.21",
+    "@truenas/ui-components": "~0.3.23",
     "@types/cheerio": "~0.22.35",
     "@types/d3": "~7.4.3",
     "@types/dygraphs": "^2.1.10",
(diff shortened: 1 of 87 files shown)


william-gr (approved):
Looks good, pending the actions for the wizard as we discussed

PR #13829, opened 2026-07-20
NAS-141481 / 27.0.0-BETA.1 / Migrate Data Protection slide-ins & modal-header to tn-*

AlexKarpov98:


aervin (changes requested):
Looks and works well. We need to be consistent with the styling of tables in the Shares and Data Protection dashboard cards. Which do you prefer?

Image

Image

AlexKarpov98:
@aervin i think I like top one more, where table is full (card) width and table title text aligned with the title of the card (left side padding)

AlexKarpov98:
@aervin please take a look at it now :)

aervin (approved):
looks great thanks

PR #13838, opened 2026-07-21
NAS-141831 / 27.0.0-BETA.1 / Change ACL traverse check from an error to a warning

AlexKarpov98:
**Problem**

When editing a dataset's ACL, the save is rejected outright if any user/group in the ACL lacks traverse (execute) permission on a parent path leading to the dataset. This is too strict for legitimate nested-dataset use cases — e.g. a container volume mounted from a user's home directory — forcing users to build more complex ACLs starting above the mount point.

Per backend guidance (Andrew Walker), the middleware already supports skipping this sanity check (validate_effective_acl: false). Validation should stay on by default, but instead of refusing to save, the UI should warn the user that some ACL entries won't grant the expected access and let them proceed.

**Solution**

The dataset ACL editor now saves with validation on as before. If the backend rejects the save specifically because a user/group can't traverse to the dataset, the UI shows a "Traverse Access Warning" dialog (including the backend's specific detail) with a Save Anyway action. Confirming re-submits with validate_effective_acl: false. Any other failure still surfaces the normal error dialog — no bypass offered.

**How the traverse failure is detected**

The backend's filesystem.setacl → filesystem.check_acl_execute raises a CallError (errno EPERM) whose message contains the stable text execute permissions on the path (NFSv4 propagates the CallError; POSIX wraps it as a filesystem_acl.path validation error with the same text). The failed job surfaces as a FailedJobError, so detection matches on that marker in job.error. This correctly excludes unrelated failures such as the ENOENT "path component does not exist" case.

**Changes**

- dataset-acl-editor.store.ts — split makeSaveRequest into runSetAclJob / handleSaveError / isTraverseValidationError / onSaveSucceeded; added the warn-and-retry flow (retry with validate_effective_acl: false on confirm).
- dataset-acl.ts (helptext) — added effectiveAclDialog strings (title, interpolated message with backend detail, "Save Anyway" button).
- dataset-acl-editor.store.spec.ts — 3 new tests: retry on confirm, no retry on decline, normal error modal for non-traverse failures.
- i18n — extracted the 3 new strings into all locale files.

**Testing**

- Unit: dataset-acl-editor.store.spec.ts — 10 tests pass (incl. the 3 new branches).
- Lint clean; validate-translations passes.
- Manual E2E against a live backend: created a nested dataset whose parent denies traverse to a user, added that user to the child's ACL, and saved:
  - Save → "Traverse Access Warning" dialog appears (not a hard error) with the backend detail and a Confirm-gated Save Anyway button.
  - Confirm → Save Anyway → ACL saves (bypass applied) and the entry persists.
  - Removing the un-traversable user and re-saving goes through cleanly with no dialog (normal saves unaffected).

**Notes**

No debug file needed — this is not a configuration issue. Related prior ticket: NAS-135444 (closed as a cleanup duplicate).

### DIFF
--- a/src/app/helptext/storage/volumes/datasets/dataset-acl.ts
+++ b/src/app/helptext/storage/volumes/datasets/dataset-acl.ts
@@ -82,4 +82,14 @@ export const helptextAcl = {
     message: T('This process continues in the background after closing this dialog.'),
   },
 
+  effectiveAclDialog: {
+    title: T('Traverse Access Warning'),
+    message: T('One or more users or groups do not have permission to traverse to this dataset,\
+ so some Access Control Entries will not grant them the access shown:<br><br>{details}<br><br>\
+This can be expected when a dataset is reached through a path the user cannot traverse, such as a\
+ container volume mounted from another user\'s home directory. Saving will apply the ACL without\
+ validating this access.<br><br>Do you want to save anyway?'),
+    buttonText: T('Save Anyway'),
+  },
+
 };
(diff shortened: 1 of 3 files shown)


aervin (approved):
works well

PR #13841, opened 2026-07-21
NAS-141474 / 27.0.0-BETA.1 / Migrate System/Network slide-ins & modal-header to tn-*

AlexKarpov98:


aervin (approved):
We should probably make the cards with tables match Shares and Data Protection. Looks good though

PR #13845, opened 2026-07-22
NAS-141773 / 27.0.0-BETA.1 / Show files in TrueCloud restore snapshot path selectors

AlexKarpov98:
### Problem
In the TrueCloud "Restore from Snapshot" form, the **Included Paths** and **Excluded Paths** tree selectors only displayed folders, even though the tooltip states files and directories can be selected. Users could not select individual files to restore.

### Cause
`getSnapshotNodeProvider()` only mapped directory entries into tree nodes; file entries returned by `cloud_backup.list_snapshot_directory` were silently discarded.

### Fix
Map both files and directories into explorer nodes — files as selectable leaf nodes (`type: File`, `hasChildren: false`), directories unchanged. The `ix-explorer` already supports file selection, and the restore params pass selected paths through as-is, so no backend change is needed.

### Testing
- Added a spec asserting the node provider returns both file and directory nodes.
- All specs pass; lint clean.


william-gr (approved):
Code review only, looks fine.

PR #13847, opened 2026-07-22
NAS-141791 / 27.0.0-BETA.1 / data-test missing in the search of the Applications Discover search input

AlexKarpov98:
Code review is enough. 
Found more missing test ids & added them.


william-gr (approved):
Looks fine. Dont we auto generate testId for tn components?

AlexKarpov98:
@william-gr 

Not blanket — but form controls do auto-derive it. A `tn-input/tn-select` with a **formControlName** and no explicit testId falls back to the control name `(formControlName="sshPort" → data-testid="input-ssh-port")`. Non-control components (buttons, icons, menu items) still need an explicit testId. So for the form fields here we can drop the hand-written ones; the action buttons still need them.

PR #13848, opened 2026-07-22
NAS-141779 / 27.0.0-BETA.1 /  Snapshots UI: Search bar does not filter by Used, Date Created, or Referenced fields

AlexKarpov98:
https://github.com/user-attachments/assets/c9c35c1c-6a00-44b8-ba68-61767fa1264c

### DIFF
--- a/src/app/pages/datasets/modules/snapshots/snapshot-list/snapshot-list.component.spec.ts
+++ b/src/app/pages/datasets/modules/snapshots/snapshot-list/snapshot-list.component.spec.ts
@@ -252,6 +252,49 @@ describe('SnapshotListComponent', () => {
     });
   });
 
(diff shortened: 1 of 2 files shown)


aervin (approved):
I can understand filtering by date, but used/referenced seems a bit strange. Is this the normal pattern for filtering tables in the app?

AlexKarpov98:
@aervin 
Yes — normal. preprocessMap is the standard way to filter on a column's displayed value; already used by initiator-list, tunable-list, VM device-list, docker-images, and services. Filtering by size also has precedent (disk-list filters by size, quotas by quota). Only difference here: I match the exact rendered string (same pipes as the cells) instead of the built-in fuzzy-bytes path, which only fires for a column literally named size.

PR #13849, opened 2026-07-22
NAS-141877 / 27.0.0-BETA.1 / Migrate ix-explorer to tn-file-picker

william-gr:
Replaces the `@bugsplat/angular-tree-component` tree in `ix-explorer` with `tn-file-picker` from `@truenas/ui-components` (bumped to ~0.4.3).

- `ix-explorer` becomes a thin adapter: node providers are bridged through `FilePickerCallbacks`, provider errors surface via the picker's error output, and manual input is committed verbatim (comma-separated lists, dataset names, SMB external paths) through a capture-phase listener.
- Relative-path providers (replication, `pool.filesystem_choices`) are supported via the picker's `valueRoot` mapping; static `/mnt` and `/dev/zvol` roots confine popup browsing, other root sets are synthesized at `/`.
- The Create Dataset dialog is replaced by the picker's inline creation row; create actions are now projected components implementing the new `ExplorerCreateAction` contract (`ix-explorer-create-dataset`, `ix-explorer-create-zvol`).
- `lastSelectedNode` now exposes `ExplorerNodeData` directly; `FileValidatorService` and VM device form updated accordingly.

image
image
image


william-gr:
@aervin @AlexKarpov98 this one has been a bit difficult. I still need to test across the UI and validate zvol works, but would appreciate another pair of eyes this far.

We still need to replace all ix-explorer usages with tn-form-field+ix-explorer and remove internal ix-label, etc but I think I will handle that in a follow up PR.

AlexKarpov98 (commented):
Looks really cool! Good job!
I noticed few visual issues I'd like to address:

Selection panel is not always fully visible and I am unable to select any item in such case:

https://github.com/user-attachments/assets/12f6db45-acad-4aeb-9a26-8948c7200c24

As well buttons does not look good, not aligned, different sizes:

Image

see checkboxes, they are cut on the left side and not fully rendered:

Image

william-gr:
@AlexKarpov98 thanks for reviewing

For the multiple checkbox, thats a component library issue I can fix separately.
For the buttons at the end, that happens for constrained space so it wraps text, what is your suggestion to fix these? I am out of ideas.
For the file picker hiding, I will fix that separately in the components library.

EDIT:
image
no wrap makes it look worse IMO

william-gr:
Components PR https://github.com/iXsystems/truenas-ui-components/pull/162

william-gr:
@AlexKarpov98 components library updated

william-gr:
@AlexKarpov98 ping

AlexKarpov98 (commented):
@william-gr should we allow custom value to be typed in, seems it's creating an issue ?

https://github.com/user-attachments/assets/34553a05-c68f-4f12-b29c-ed5f9965198d

william-gr:
> @william-gr should we allow custom value to be typed in, seems it's creating an issue ?
> 
>  Screen.Recording.2026-07-31.at.09.47.35.mov

I dont see a problem with that. What is the concern? Missing validation before submitting?

william-gr:
@AlexKarpov98 ping

AlexKarpov98 (approved):
Looks good.
I just thought that would be nice to prevent submitting the form when we know that the error will be shown in the dialog. (Maybe show it inline onhand)

william-gr:
> Looks good. I just thought that would be nice to prevent submitting the form when we know that the error will be shown in the dialog. (Maybe show it inline onhand)

We can probably improve that in the future but seems to be the current behavior. I will merge as-is and we can follow up with an improvement.

PR #13857, opened 2026-07-23
NAS-141778 / 27.0.0-BETA.1 / Expose force_topology option for vdevs on Community Edition

AlexKarpov98:
Screenshot 2026-07-23 at 13 21 28

Adds a CE-only **Force** checkbox to the pool wizard's Review step that sends `force_topology` to `pool.create` / `pool.update`, bypassing the topology policy checks added in [middleware#19282](https://github.com/truenas/middleware/pull/19282) (equal data-vdev width, RAIDZ/mirror width caps, special/dedup redundancy). Hidden on Enterprise, where middleware rejects the flag.


aervin (approved):
Claude's first point seems like the safest thing to do

PR #13860, opened 2026-07-24
NAS-141489 / 27.0.0-BETA.1 / Migrate Services config forms to tn-* and ix-form wrapper

AlexKarpov98:


william-gr:
@aervin i think you are better suited to review this than me

src/app/pages/services/services.component.ts
aervin, line 52: while we're working on this component, let's give the `tn-card` a `display: block` styling so it shrinks to its content.

Image

aervin (changes requested):
I think Claude's # 3 point is worth addressing

aervin (approved):

PR #13869, opened 2026-07-27
NAS-141034 / 27.0.0-BETA.1 / Migrate shared control modules to tn-*

AlexKarpov98:


william-gr:
@AlexKarpov98 mixins/tn-list.scss idea and the oauth-button flag seems to make sense to me, what do you think?

AlexKarpov98:
@william-gr issues addressed

william-gr (approved):

PR #13879, opened 2026-07-29
NAS-141035 / 27.0.0-BETA.1 / Migrate scheduler and terminal to tn-*

aervin:


aervin:
time 2:00

AlexKarpov98 (approved):
Works well! 👍 

Worth checking Claude's comments: 

[Fix now — cheap, and the reviewer is right]

- Omit<TnCalendarIntl, 'monthGridLabel'> — substantively the best comment in the review. Partial<> means the next library version that adds a string ships untranslated, which defeats the provider's whole purpose. Five lines, and it deletes a test.

- Dead [disabled] in terminal.component.html:29 — deleting one attribute. Worth doing precisely because the element is already being rewritten.

- The scss comment — it's a comment edit. Misleading comments cost more than missing ones.

PR #13881, opened 2026-07-31
NAS-141483 / 27.0.0-BETA.1 / Migrate Datasets slide-ins and modal-header to tn-*

AlexKarpov98:


aervin (commented):
Should we display the advanced options by default when editing a dataset? Seems weird to have a form with a single disabled field.

Image

AlexKarpov98:
@aervin good catch!
It's fixed now!

aervin (approved):

PR #13902, opened 2026-08-06
NAS-141486 / 27.0.0-BETA.1 / Migrate VM slide-ins & modal-header to tn-*

AlexKarpov98:


aervin (approved):
While we're here, can you please update this banner? The "Note" heading should be more informative, something like "Requires separate client software". And I think we should remove the password requirements from this banner, that requirement is unrelated to the banner's purpose.

Image

Otherwise it looks good and works well.

PR #13903, opened 2026-08-06
NAS-141065 / 27.0.0-BETA.1 / Migrate remaining small pages to tn-*

AlexKarpov98:


src/app/pages/two-factor-auth/two-factor.component.spec.ts
```diff
@@ -93,34 +96,34 @@ describe('TwoFactorComponent', () => {
     expect(copyButton).toHaveProperty('text', 'KYC123');
   });
 
-  it('shows warning when global setting is disabled', () => {
+  it('shows warning when global setting is disabled', async () => {
```
aervin, line 99: This series of tests is green only by accident — the first banner test's jest.spyOn doesn't affect its own assertion but leaks into the next two, which fail in isolation. Drive globalTwoFactorEnabled / userTwoFactorAuthConfigured directly.

aervin (approved):

PR #13914, opened 2026-08-10
NAS-142058 / 27.0.0-BETA.1 / Adopt tn-table 0.4.9 inputs and drop table workarounds

AlexKarpov98:


william-gr (commented):
Can you rebase this after the new claude review? I have a feeling some of the review flags are more than LOW and could deserve fixes.

william-gr (approved):

PR #13916, opened 2026-08-10
NAS-142083 / 27.0.0-BETA.1 / Add search, sort and keyboard navigation to Update Members

AlexKarpov98:
Preview:

https://github.com/user-attachments/assets/f4dc4afa-d7d9-42d9-8560-cfa139433227


aervin (approved):
Works well. 

1) This is a pretty large changeset for a backport. I'd prefer we just improve this for 27 alone.

2) Perhaps in a follow-up we could add the "Unsaved changes" guard to the cancel action?

AlexKarpov98:
@aervin, sure - 1. I won't' backport it. 2. I implemented an improvement in this PR since we already here.

PR #13920, opened 2026-08-12
NAS-142127 / 27.0.0-BETA.1 / Add Rsync Task Connect using option data-test shows numbers

AlexKarpov98:


aervin (changes requested):
This feels like we're working around some unwanted behavior on the truenas-ui-components side. Do you think we should make tn-select's default test ID behavior more useful instead?

AlexKarpov98:
Will be resolved on the https://github.com/iXsystems/truenas-ui-components/pull/182 side.

PR #13944, opened 2026-08-19
NAS-142225 / 27.0.0-BETA.1 / Fix Force checkbox not enabling Save on the NTP Server form

AlexKarpov98:
**Changes:**

Adding an unreachable NTP server via the documented **Force** escape hatch was a dead
end: the backend's "Server could not be reached. Check 'Force' to continue regardless."
was pinned onto the `address` control with `setErrors()`, which — unlike a validator
result — never re-evaluates. Angular drops such a pin when *that* control changes, but
an error the user is meant to answer from a *different* field has no way out, so the
form stayed invalid and Save stayed disabled. The only escape was retyping the Address.

Two fixes:

1. **`form-error-handler.service.ts` — the pin now retires itself.** Each backend
   verdict describes one submitted payload, and the next edit anywhere in the form moves
   the payload on. So when the service pins an error it also subscribes one-shot to the
   control's `root.valueChanges` and, on the next edit, re-runs that control's
   validators. Re-validating rather than blanking is the point: the pinned set is
   replaced by the control's *real* validation state, so a field that is genuinely
   empty-but-required goes back to saying `required` instead of falling silently valid,
   and live client-side validators keep their say. If the verdict still stands, the next
   save pins it again.

   Putting this in the service rather than in `<ix-form>` means it covers all 67 callers
   of `handleValidationErrors`, including forms `<ix-form>` never wraps, and needs no
   per-field declaration — nothing to remember, nothing to forget. Unsubscribing before
   the clearing lets `updateValueAndValidity()` emit normally, which is what refreshes
   `<ix-form>`'s status signal and so the host's Save button.

   *Behaviour change worth flagging:* a pinned backend message now also disappears if
   the user edits an unrelated field. That is deliberate — the verdict describes a
   payload that no longer exists — and matches what the directory-services form already
   does by hand in `clearFormControlErrors()`. Validator-driven errors are untouched:
   `image-validator` and `file-validator` return `ixManualValidateError` from real
   validator functions, so they re-evaluate normally and never carry the
   `manualValidateError` flag this keys on.

2. **`form-error-handler.service.ts` — a failed *element* lookup no longer escalates to
   an error modal.** The service pins the message on the control (rendered inline by
   `tn-form-field`) and then looks up the DOM element purely to scroll/focus it. Both
   lookup strategies missed for tn-* forms — `IxFormService` only knows `ix-*` controls,
   which register via `RegisteredControlDirective`, and `[formControlName]` is a
   property binding that leaves no attribute behind — so every `<ix-form-renderer>` form
   showed the message twice, inline *and* in a modal. The lookup now also matches the
   `data-control-name` the renderer already stamps on each control, so scroll and focus
   work there too, and focus lands on the native control inside the tn-* host rather
   than the host itself. The genuine fallback (no *control* found, nothing renderable
   inline) still opens the modal.

The three `manualValidateError` keys also get one home in
`manual-validate-error.constants.ts`, instead of being re-spelled at each reader.

**Testing:**

Automated — `ntp-servers.form-config.spec.ts` drives the reported path end to end
through `<ix-form-renderer>` with the real `FormErrorHandlerService`, so the error is
genuinely pinned by a rejected `system.ntpserver.create`:

- backend rejects → message pinned on Address, Save blocked
- tick Force → Save freed, Address value untouched *(fails without the fix)*
- Max Poll 99 + Force → Save still blocked, live validators keep their say
- tick **Burst** instead → also frees Save, documenting that no field is special-cased

Plus `form-error-handler.service.spec.ts` covers the retiring directly: re-validation vs
blanket-clearing, a sibling's live validator surviving, a nested control retired by an
edit in a sibling group, and re-pinning when the next save is rejected again — as well
as the message staying inline-only when the element is missing.

Manual — System → Advanced → NTP Servers → **Add**, Address `192.0.2.1` (RFC 5737,
never routable), **Save**. Expect the inline error and **no** error modal. Tick
**Force** → error clears and Save enables; saving adds the server. Untick Force and
save again → the error returns, confirming the verdict is retired, not suppressed.

### Downstream

|Affects         |Reasoning
|Documentation   | No — the Force option already works as documented; this makes the UI match.
|Testing         | Yes — NTP add/edit, plus any form surfacing backend field errors: those no longer show a duplicate error modal alongside the inline message, and a pinned message now clears on the next edit anywhere in the form rather than only on the flagged field.


src/app/modules/forms/ix-forms/components/ix-form-renderer/form-definition.interface.ts
```diff
@@ -169,6 +169,20 @@ interface BaseFieldDefinition<T extends object> {
...
+   * from a DIFFERENT field would keep Save disabled forever. Name that escape hatch's
+   * target here: on every change of this field the listed controls are re-validated,
+   * which replaces the stale server verdict with their real validation state (a field
+   * that is genuinely empty-but-required goes back to saying `required`). Client-side
+   * validators are untouched.
+   *
+   * Example: the NTP form's `force` ("add this server even though it could not be
+   * reached") declares `clearsServerErrorsFor: ['address']`.
+   */
+  clearsServerErrorsFor?: (keyof T & string)[];
```
william-gr: I am a bit skeptical about a new explicit flag for this. Seems very prone to issues.
I wonder if we could simply clear and re-evaluate all fields once the form changes, or some other clever way.

AlexKarpov98: Yes, that's the initial option I was thinking about but decided to keep the existing logic for all forms and simply fix the issue which is basically not a bug but an improvement we should have @william-gr

AlexKarpov98: @william-gr so you think this
> if we could simply clear and re-evaluate all fields once the form changes

Should work?

william-gr: I am not sure, just a thought. Perhaps discuss some options with claude and have it defend this implementation versus something else, doing adversarial review.

AlexKarpov98: @william-gr 

Good call — you were right, and this is a bit of a confession: the form-wide clearing was actually my first commit on this branch. I narrowed it to a per-field flag for tidiness, not because anything broke.

Two things pushed me back. clearsServerErrorsFor only exists inside ix-form-renderer (20 configs) while handleValidationErrors has 67 call sites, so most forms keep the bug — and being opt-in, the default stays broken and forgetting the flag fails silently. There's also precedent: directory-service-validation.service.ts already hand-rolls exactly this.

One change from my first attempt: the clearing now lives in FormErrorHandlerService, not ix-form. It already has the FormGroup, so when it pins an error it subscribes one-shot to the control's root valueChanges and retires the pin on the next edit. Covers all 67 call sites including legacy SidePanelForm hosts, and since it unsubscribes before clearing, updateValueAndValidity() emits normally — which drops the emitEvent: false + manual formStatus.set() hack I disliked the first time.

Not a casualty: image-validator/file-validator return ixManualValidateError from real validator functions, so they never carry the manualValidateError flag we key on. The one real behaviour change is the message disappearing if you edit an unrelated field — correct, I think, since the verdict describes a payload that no longer exists.

william-gr (approved):

PR #13950, opened 2026-08-20
NAS-142162 / 27.0.0-BETA.1 / Consolidate alerts and shorten alert messages

AlexKarpov98:
https://github.com/user-attachments/assets/36978dd7-4680-4e63-b4b4-444d4082ceac

**Changes:**

Alerts of the same kind are now consolidated into one entry, and only a concise summary is shown up front — the full text moves behind Show More / View More.

**Consolidation.** Grouping previously keyed on `alert.key`, which is unique per object, so four pools needing a ZFS upgrade produced four near-identical banners and four separate dismisses. Grouping now keys on **class + level + target page** (`alert-consolidation.utils.ts`). The group carries every alert's id, so one Dismiss clears all of them. Level and target page are part of the key on purpose: classes with conditional enhancements (boot pool vs data pool capacity) point at different pages and must not merge into an entry that links to the wrong one.

**Shorter messages.** `getAlertSummary()` (`alert-summary.utils.ts`) reduces a message to its first sentence, or a word-boundary truncation at 120 chars when that sentence is still long. It ignores periods inside version numbers and initialisms, and strips markup so a cut can't sever a tag. What was hidden:

- *Page banners* — summary line; Show More reveals each consolidated alert's full message plus the contextual help and docs link, which previously sat in the banner unconditionally.
- *Alerts dropdown* — same, behind View More. This also replaced the `scrollHeight > offsetHeight` measurement (`ViewChild` + `setTimeout` + `afterNextRender`) with a computed signal.

**Group headlines.** `SmartAlertEnhancement` gains an optional `groupSummary` with a `{count}` placeholder, filled in for the 28 classes that actually repeat per pool / disk / task / certificate — e.g. `A new ZFS version or feature flags are available for {count} pools`. Classes without one fall back to the newest alert's own summary plus the count badge.

Counts stay per alert instance — the severity filters, the category header counts and the nav badges still report individual alerts, so they remain consistent with the topbar bell.

**Testing:**

Mock config for the WebSocket Debug Panel (Mock Configurations → Import, then reload — `alert.list` only fires on `adminUiInitialized`):

<details>
<summary>nas-142162-alert-mocks.json</summary>

[nas-142162-alert-mocks.json](https://github.com/user-attachments/files/31258733/nas-142162-alert-mocks.json)

Mocks `alert.list` with the seven alerts from the ticket screenshot — `LicenseIsExpiring`, `RsyncSuccess`, `PoolUpgraded` ×4 (newpool / ggdraid / basicpool / Sanity, each with a **distinct** `key`), `ReplicationSuccess` — plus `alert.dismiss` and `alert.restore` so the dismiss flow works without hitting the appliance.

</details>

Expected on `/storage`: one banner reading *"A new ZFS version or feature flags are available for 4 pools"* with a `4` badge instead of four separate banners; Show More lists all four pool messages. In the dropdown, 7 alerts render as 4 rows, and a single Dismiss on the Storage row clears all four.

Specs: `alert-summary.utils.spec.ts` and `alert-consolidation.utils.spec.ts` are new (20 tests). `alert.component.spec.ts` and `page-alerts.component.spec.ts` cover the summary, the group headline, the expanded detail list and group dismiss.

⚠️ `alerts-panel.component.spec.ts` fails wholesale on **master**, before this branch — every test dies at component construction with `TypeError: Cannot read properties of undefined (reading 'length')` inside ng-mocks' `getNgType`/`funcExtractDeps`. I ruled out the spec's `OverlayContainer` factory, the `MockComponent(AlertComponent)` declaration, each `mockProvider`, `mockAuth()`, and reverting the general-settings tn-* migration. I updated that spec's expectations for the new behaviour so they're correct once it runs, but I could not execute them — the panel changes are covered at the util level and in `alert.component.spec.ts` instead. Worth its own ticket.

### Downstream

|Affects         |Reasoning
|Documentation   | No — no new settings or workflows; alert text is unchanged, only how much of it is shown before expanding.
|Testing         | Yes — a repeated alert class now renders as **one** row/banner rather than one per object, and the page-banner dismiss acts on the whole group. Tests that count alert rows or dismiss them individually need updating. Severity-filter and nav-badge counts are unchanged (still per alert).


aervin (approved):
working well

PR #13954, opened 2026-08-21
NAS-142223 / 27.0.0-BETA.1 / Warn that cancelling a data migration does not undo the tier change

AlexKarpov98:
**Changes:**

Users reasonably assume that cancelling a dual action cancels both halves of it. A data migration *can* be cancelled (`zfs.tier.rewrite_job_cancel`), but cancelling only stops the rewrite — `dataset_set_tier` has already been applied and is not reverted, so the dataset stays on the new tier. Nothing in the UI said so.

This takes option (a) from the ticket: warn up front, rather than adding a "cancel Tiering and Migration / Migration only" choice. Option (b) needs middleware support to actually revert a tier, so it isn't something the UI can deliver alone.

- **Change Storage Tier dialog** — warning banner directly above Cancel/Apply: *"Changing the tier takes effect immediately"*. When **Move existing data** is checked it adds *"Data migration can be cancelled while it runs, but cancelling does not undo the tier change: the dataset stays on the new tier."*
- **Cancel-migration confirmation** — second sentence: *"The storage tier change has already been applied and is not reverted by cancelling."*

The heading was originally taken almost verbatim from the ticket ("Tiering change cannot be cancelled once started") — out of that context it read as "there is no abort button", which is wrong. Reworded in ec5f8cac2c to say what the ticket is actually after.

Two defects found while testing the above against a real migration, both pre-existing since NAS-139979:

- **Migration progress was measured in bytes.** `count_bytes` is the byte count the job has *claimed*, not written — it reaches `total_bytes` as soon as the dataset is enumerated, and can overshoot it (the ticket's own screenshot shows `2.08 GiB / 1.66 GiB`). A 4 GiB single-file rewrite therefore rendered a full progress bar with `0 / 1` items done. Progress now comes from `(success + failures) / total_items` — items *processed*, so a job that ends having failed an item still reaches 100% next to its terminal badge — clamped, and floored rather than rounded so 999 of 1000 reads 99, not a premature 100.
- **The ETA rendered as the start time.** It extrapolated from that same byte fraction: fraction 1.0 × 0s elapsed → ETA = start. It's now derived from item progress and suppressed when it can't say anything — at >=100%, or when the first status tick lands in the same second the job started. A single-item job has no expressible middle (`success` is 0, then 1), so it runs the bar indeterminate instead of pinning it at 0%.

Since the bar deliberately ignores `count_bytes`, the two stat cards were relabelled so they don't contradict it on screen: **Data Transferred → Data Scanned** (claim-shaped, matching what the counter actually reports) and **Items Completed → Items Succeeded** (the card counts `success` only, while the bar counts processed).

Also: left-aligned the tier chips (they were centered over their "N GiB available" captions, so they sat indented from every other row in the dialog), and dropped the box on the migration stat cards — padding, border and radius — so they read as plain columns aligned with the rest of the dialog body.

### Downstream

|Affects         |Reasoning
|Documentation   | Tier-change docs should state that a tier change is not reversible and that cancelling stops only the data migration.
|Testing         | .


william-gr (commented):
Hmm I am confused. We can call abort, can't we?

AlexKarpov98:
@william-gr I updated it now, thanks for the comment!

 You can, yes — the Cancel button in the Data Migration dialog calls zfs.tier.rewrite_job_cancel. The bad wording is mine to fix: I took the heading almost verbatim from the ticket's suggested text ("Tiering change cannot be cancelled once started. Migration of data can."), and out of that context it reads as "there is no abort".

What the ticket is actually after is that abort doesn't revert dataset_set_tier — cancelling stops the rewrite, but the dataset stays on the new tier, and users cancelling a dual action assume both halves are undone.

Reworded so the banner says that without denying abort exists (ec5f8cac2c):
- heading: "Changing the tier takes effect immediately"
- with Move existing data on: "Data migration can be cancelled while it runs, but cancelling does not undo the tier change: the dataset stays on the new tier."

william-gr (approved):

william-gr:
@AlexKarpov98 dont forget to update the PR title and message when merging

PR #13971, opened 2026-08-27
NAS-142373 / 27.0.0-BETA.1 / Migrate Sharing/iSCSI forms to tn-* form controls

AlexKarpov98:
## Changes

Migrates the Sharing → iSCSI forms to `@truenas/ui-components` form controls and bumps the library from `~0.5.3` to `~0.7.1`, adopting what that release makes possible.

### iSCSI forms
- Extent, Target, Portal, Associated Target and the iSCSI wizard steps move to `tn-form-field` + `tn-input` / `tn-select` / `tn-checkbox` / `tn-radio-group`.
- Initiator form actions move into the card footer (`[primaryAction]` / `[secondaryAction]`), and the page now guards unsaved changes via `UnsavedFormGuard`.
- FC port rows take an `[index]` so every test id — and the DOM ids `tn-select` derives from it — is scoped per row.

### What 0.7.1 unlocked
- **Extent `filesize` back on `tn-input`.** 0.7.x stops re-parsing a Size field's own blur-formatted display, so the model keeps the exact byte count typed. `2049 KiB` round-trips as 2 098 176 again, and the spec expectation is the original arithmetic rather than a rounded one.
- **`tn-radio-group [options]` instead of projected `<tn-radio>` children.** Since 0.7.0 `tn-radio` declares the `radio-button` prefix and the group scopes each option's id by label, so `[options]` reproduces the legacy `radio-button-mode-*` ids on its own. The projection workaround and its hand-written `['radio-button', …]` bases are gone.
- **`TN_SELECT_LABELS` at the app root** (`provideTnSelectLabels()`), replacing the `[placeholder]`/`[noOptionsLabel]` pair repeated on every `<tn-select>`. Removed from the files this PR touches; the `tnSelectLabels` constant survives as the single source of the translation keys and now feeds the provider.
- **`TN_FALLBACK_LABELS` at the app root** (`provideTnFallbackLabels()`) — translated accessible names for an unnamed `tn-spinner` / `tn-progress-bar` / `tn-dialog-shell` / `tn-side-panel` / `tn-drawer`. Required by the bump: 0.7.x warns in dev mode on every unnamed instance, and `jest-fail-on-console` turned that into 31 failures across `ix-select`, `ix-combobox`, `ix-user-picker` and `data-migration-status-dialog`. A per-instance `[ariaLabel]` still wins where the name can say *what* is loading.

Both providers are mirrored in `setup-jest.ts` so specs read the copy the app actually renders, and both are covered by their own specs.

### Playbook
`tn-migration/SKILL.md` drops the two caveats 0.7.1 resolved (Size re-parse, untranslated select defaults), rewrites the radio-group test-id recipe, and documents the app-root label providers in one table.


william-gr (approved):
Great job!

PR #13978, opened 2026-08-28
NAS-142746 / 27.0.0-BETA.1 / Align master-detail layouts on Apps, Containers, Audit and Users

AlexKarpov98:
Layout polish across the pages that use the shared master-detail view, so they read the same way the Datasets page does.

https://github.com/user-attachments/assets/94cbff90-4747-40ba-8485-080f54409d95

### Shared master-detail view
- The master card sizes to its content instead of stretching to the details column's height. The details column gets an explicit viewport height from `ixDetailsHeight`, and the row's default `align-items: stretch` matched the card to it, leaving a large empty card below the last row whenever the list was short.
- The detail-header row collapses when a page projects no title into it (Installed Apps shows one only in mobile view), so both columns start on the same line.

### Apps — Installed
- Dropped the "Applications" heading; the card starts with the search field. The bulk-actions header now renders only while rows are checked, padded to line up with the search field.
- Table fits the master pane at every width — no horizontal scroll — and row height no longer changes as the stats tick. Each half of a Block I/O / Network pair is an unbreakable `.stat-part` inside a `.stat-pair` flex column, so the cell is exactly two lines for any value; previously the separator was its own break opportunity and the cell flipped between one, two and three lines.
- Names wrap at the hyphen rather than being cut, with the full name on the cell's tooltip; RAM, Block I/O and Network tooltips now carry the value.
- Selection column widened to 38px — the 18px checkbox was being clipped by the library's `overflow: hidden`.
- Detail panel's Delete button is outlined/default instead of red, matching Containers and the rest of the app.
- Workloads "Containers" rows wrap onto a second line instead of chopping a state into `Runnin` / `g` (the details-cards mixin sets `word-break: break-word` panel-wide).

### Containers
- Same header treatment: the bulk-actions row only exists when rows are checked, which removes the extra padding above the search field.

### Audit
- Details cards lay out in two columns. The shared mixin uses CSS multi-column, which kept both cards in the first column and grew the container instead of balancing them, leaving half the panel empty; the panel now uses a grid that collapses back to one column when narrow.

### Users
- Removed the `:host ::ng-deep .table-container` width override. It had the same specificity as the shared layout mixin's rule, so which one won depended on stylesheet insertion order: the page rendered at 1458px after a client-side navigation and 704px after a reload. The mixin's width now holds in both cases.

### DIFF
--- a/src/app/modules/master-detail-view/master-detail-view.component.scss
+++ b/src/app/modules/master-detail-view/master-detail-view.component.scss
@@ -84,3 +84,21 @@
     width: 100%;
   }
 }
+
+// The master list sizes to its content instead of stretching to the details column's
+// height. The details column is given an explicit viewport-based height (ixDetailsHeight),
+// and the row's default `align-items: stretch` made the master card match it — leaving a
+// large empty card below the last row whenever the list is short. Hugging the content
+// keeps the card ending at its last row; a list taller than the viewport still scrolls
+// with the page (same approach as the datasets tree, see dataset-management.component.scss).
+.table-container {
+  align-self: flex-start;
+}
+
+// Collapse the detail-header row when the page projects no title into it (Installed Apps
+// shows one only in mobile view). The empty h3 still reserved ~30px, which pushed the
+// details cards below the top of the master card; hiding it lines both columns up.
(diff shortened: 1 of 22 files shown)


aervin (approved):
looks great

PR #13982, opened 2026-08-31
NAS-142381 / 27.0.0-BETA.1 / Migrate credentials forms to the shared ix-form wrapper

AlexKarpov98:
**Changes:**

T11 of the credentials migration: put every side-panel-hosted credentials form on the shared
`<ix-form>` wrapper, and replace the remaining `ix-*` form controls with their
`@truenas/ui-components@0.7.1` equivalents.

**Moved onto `<ix-form>` / `IxFormHostForm`.** All three are opened through
`FormSidePanelService.open`, so they qualify under the "only SlideIn / panel hosts adopt
`ix-form`" rule; the wrapper now owns the submit lifecycle, error handling, snackbar,
dirty tracking and close.

| Form | Notes |
|---|---|
| `certificate-acme-add` | `onSubmit`/`isLoading`/`canSubmit` replaced by `handleSubmit`; the domain-name load moved to the base's `loadFormConfig`, clearing the `FormArray` first so the retry replay stays idempotent |
| `privilege-form` | `handleSubmit` returns the same `combineLatest` → `switchMap` request; gained a success snackbar it never had |
| `api-key-form` | `handleSubmit` + `onSuccess`; `LoaderService` dropped (the panel shows its own progress). `successMessage` is a function returning `null` when a key was generated, since `KeyCreatedDialog` is the stronger confirmation, and the plain "API Key Updated" snackbar only fires when it wasn't |

**Controls migrated to tn-\*:**

- `ix-file-input` → `tn-file-input` (google-cloud-provider-form). The value shape changes
  `File[]` → `File`, so the control type and the `switchMap` that reads it changed with it.
- `ix-list` / `ix-list-item` → `tn-form-list` / `tn-form-list-item` (ACME domains). The SCSS
  that compensated for `ix-list`'s standalone top margin and its zero bottom padding is gone —
  the library component sizes projected content itself.
- `ix-chips` → `tn-chip-input` (csr-subject, Subject Alternative Name).
- `ix-group-chips` → `tn-chip-input` + a provider stream (privilege-form, Directory Services
  Groups), matching how the Local Groups field on the same form already worked.
  `UserGroupExistenceValidationService.validateGroupsExist()` is re-registered on the control
  directly — after the edit patch and without `updateValueAndValidity()`, so opening the form
  for edit doesn't immediately flag the loaded groups.
- `ix-user-picker` → `tn-form-field` + `tn-autocomplete` (api-key-form). `(searchChange)` /
  `(loadMore)` / `(opened)` drive `UserPickerProvider.fetch` / `nextPage`, `[requireSelection]`
  replaces the old reset-on-blur, `[filterFn]` disables the client-side pass (the provider
  already filtered server-side, and the default label filter would hide matches and drop the
  "Add New" row as soon as the user typed), and selecting "Add New" still opens the user form.

**Deliberately left alone:**

- `additional-details-section` and `allowed-access-section` are sections of a parent form, not
  forms of their own, so there is no `<ix-form>` to add. Both are already Material-free — the
  audit found nothing to change.
- `ix-errors` in `allowed-access-section` stays until T17, which owns it.
- `ix-combobox` (Primary Group) is shared infrastructure with 5 consumers including two
  wrappers, so it is outside this ticket's file set. It is the last migratable `ix-*` control
  left in these templates; `ix-explorer`, `ix-permissions`, `ix-details-table` and
  `ix-editable` have no tn-* replacement.

**Testing:**

`yarn test src/app/pages/credentials src/app/modules/forms/ix-forms` — 130 suites / 1140 tests
passing. `tsc -p src/tsconfig.app.json` and `yarn lint` clean on every changed file.

Specs were updated to the new surface rather than around it:

- `csr-subject` and `csr-add` drove the SAN field through `IxFormHarness.fillForm`, which only
  resolves `ix-*` controls — both now use `TnChipInputHarness`.
- `api-key-form` gained a test covering the new username autocomplete: that it offers the
  privileged users (plus "Add New") and commits the picked one to the form.
- The three `<ix-form>` hosts now register `ixFormTestingProviders()`, which also zeroes
  `ixFormMinSubmitFeedbackMs` so a spec can assert `closed` synchronously after `submit()`.

Manual check worth doing on the appliance: the ACME domains list (each domain's authenticator
select now sits in a `tn-form-list-item`) and the API-key Username picker, including scrolling
the dropdown to page in more users and the "Add New" → create-user round trip.

### Downstream

|Affects         |Reasoning
|Documentation   | No — no user-facing behaviour or copy changes beyond two added success snackbars.
|Testing         | Yes — two `data-test` values necessarily rename, since the library owns the element prefix and neither component allows a host pin (both fall back to the bound control name, so pinning would emit two ids): `input-san` → `chip-input-san` (CSR Subject Alternative Name) and `input-username` → `autocomplete-username` (API key form). Neither value is referenced anywhere in `e2e/`. All other test ids resolve unchanged, including `input-upload-credentials`, which is pinned on the `tn-file-input` host.


src/app/modules/forms/ix-forms/components/ix-user-picker/ix-user-picker.component.html
```diff
@@ -1,86 +0,0 @@
-<div class="full-width container has-tooltip">
```
william-gr, line 1: Not sure I agree with this change.

Previously we had a single component to reuse for user selection, now we are splitting that in every component?

AlexKarpov98: @william-gr yeah, or maybe it's better to migrate these components to tn-* components and use a tn-* instances instead of ix-* ?

william-gr: Yeah, thats my thought

AlexKarpov98: @william-gr agree, let's do it 👌

william-gr (approved):
Seems to work great!

PR #13991, opened 2026-09-01
NAS-142517 / 27.0.0-BETA.1 / Fix Storage Dashboard failing when zpool.query returns null properties

AlexKarpov98:
I used this file to reproduce the bug (on master branch):

[nas-142517-zpool-null-properties.mock.json](https://github.com/user-attachments/files/31690929/nas-142517-zpool-null-properties.mock.json)

One thing you must edit before importing: replace REPLACE_WITH_BROKEN_POOL_NAME and REPLACE_WITH_HEALTHY_POOL_NAME with actual pool names from your system. The store keys the zpool list by name and only overlays a pool when zpoolsByName[pool.name] hits (pools-dashboard-store.service.ts:118) — a non-matching name is silently ignored and nothing reproduces. If you only have one pool, point the null entry at it and leave the second entry as-is.

**Changes:**

The Storage Dashboard threw `can't access property "class_special_usable", T.properties is null` and rendered no pools at all.

`zpool.query` returns `properties: null` for a pool ZFS can't read properties from (unavailable/faulted). `PoolsDashboardStore.loadPoolsAndRootDatasets()` dereferenced it unguarded — the existing `?.` only covered a *missing* property (a pool with no special vdev), not a null `properties` object. The throw happens inside the `tap`, so the `catchError` wrapping the whole `forkJoin` swallows the entire load: one bad pool hides every pool.

- `zpool.interface.ts` — `Zpool.properties` is now `Record<string, ZpoolProperty> | null`, so the compiler forces the guard at every call site.
- `pools-dashboard-store.service.ts` — hoists `properties` and returns the plain `pool.query` pool when it's null. The pool still renders with the used/available numbers `pool.query` supplied; only the tier overlay is skipped.
- `change-tier-dialog.component.ts` — same unguarded deref, guarded the same way. (`dataset-capacity-management-card` already used `?.properties?.` and needed no change.)
- Spec: added *"keeps showing pools when zpool.query returns null properties"*.

### Downstream

|Affects         |Reasoning
|Documentation   | No — no user-facing behavior or wording change.
|Testing         | Yes — worth an E2E/manual case for a dashboard with a faulted or unavailable pool; previously untested path.

### DIFF
--- a/src/app/interfaces/zpool.interface.ts
+++ b/src/app/interfaces/zpool.interface.ts
@@ -18,7 +18,9 @@ export interface Zpool {
   warning: boolean;
   status_code: string;
   status_detail: string | null;
-  properties: Record<string, ZpoolProperty>;
+  // Null for pools ZFS cannot read properties from, e.g. an unavailable or
+  // faulted pool, so every access has to be guarded.
+  properties: Record<string, ZpoolProperty> | null;
   topology: PoolTopology | null;
   scan: PoolScanUpdate | null;
   // `expand` and `features` are present on the wire but not consumed in the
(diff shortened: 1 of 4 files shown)


aervin (approved):

AlexKarpov98:
backport

PR #14004, opened 2026-09-02
Restore Spanish and Galician translations for 26.0.0 (`gl.json` is currently 0% on this branch)

rtajes-max:
### What this fixes

`src/assets/i18n/gl.json` on `release/26.0.0-RC.1` has **0 of 6075 strings translated** — every value in the file is an empty string. As it stands, 26.0.0 will ship with Galician selectable in the UI and completely untranslated. `es.json` on this branch is in good shape (99.5%) but is missing 28 strings.

Both locales are complete on `master` (98.3%). The gap exists because `.github/workflows/update-translations.yml` hardcodes `base: master`, so the weekly translation sync only ever lands on `master`; nothing merged after this branch was cut has reached it.

### What this PR does

For each locale, every **empty** value in this branch's file is filled with the translation for that same key from the current `master` file.

| file | before | after | strings added |
|---|---|---|---|
| `es.json` | 6033 / 6062 (99.5%) | 6061 / 6062 (100.0%) | +28 |
| `gl.json` | 0 / 6075 (0.0%) | 5655 / 6075 (93.1%) | +5655 |

Deliberately conservative:

- **No keys added, removed or reordered.** The key set of this branch is preserved exactly, so the file still matches the strings this release actually uses.
- **No existing translation is overwritten.** Only empty values are filled.
- **No guessing.** Keys that do not exist on `master` are left empty rather than machine-translated (420 in `gl`, 1 in `es`).

One extra change: `es.json` contained a duplicate key — `"At least {min} disk(s) are required for {vdevType} vdevs"` appeared twice, once empty and once translated. It is now a single translated entry.

### Verification

- Both files parse as JSON and round-trip byte-identically under the repository's formatting (2-space indent, LF, unescaped UTF-8), so the diff contains only changed values.
- Key list and key order are identical to the branch's current files (verified programmatically).

I maintain both locales and am happy to redo this against whichever branch 26.0.0 will actually be cut from, if this is not the right target.


william-gr:
Please target stable/26.

PR #14005, opened 2026-09-02
Restore Spanish and Galician translations for 25.10.x (unchanged since 2025-08-03)

rtajes-max:
### What this fixes

`src/assets/i18n/es.json` and `gl.json` have not been touched on any 25.10.x branch since **3 August 2025** (`aa2835d`, "Update translations"). Every point release from 25.10.1 through 25.10.7 has therefore shipped the same year-old translation files:

- Spanish: 441 / 5717 strings translated (**7.7%**)
- Galician: 27 / 5736 strings translated (**0.5%**)

Meanwhile both locales are complete on `master` (98.3%). The weekly sync in `.github/workflows/update-translations.yml` targets `base: master` only, so the stable line never picks these up.

### What this PR does

For each locale, every **empty** value in this branch's file is filled with the translation for that same key from the current `master` file.

| file | before | after | strings added |
|---|---|---|---|
| `es.json` | 441 / 5717 (7.7%) | 4859 / 5717 (85.0%) | +4418 |
| `gl.json` | 27 / 5736 (0.5%) | 4851 / 5736 (84.6%) | +4824 |

Same conservative rules as the companion PR for 26.0.0:

- **No keys added, removed or reordered** — the key set of this release is preserved exactly.
- **No existing translation is overwritten** — only empty values are filled.
- **No guessing** — strings that no longer exist on `master` (858 in `es`, 885 in `gl`) are left untranslated rather than machine-translated.

### Verification

- Both files parse as JSON and round-trip byte-identically under the repository's formatting, so the diff contains only changed values.
- Key list and key order are identical to the branch's current files (verified programmatically).

I picked `release/25.10.7` as the newest branch of the stable line. If point releases are cut from somewhere else, tell me the right base and I will retarget — the change is mechanical and I can regenerate it for any branch in a couple of minutes.


william-gr:
Sorry, its too late for this branch. You can target stable/26 if you want it to make it in .8.

PR #14011, opened 2026-09-03
NAS-142255 / 27.0.0-BETA.1 / Stop long migration stats from overlapping each other

AlexKarpov98:
### Problem

In the **Data Migration** dialog the three stats (`Data Scanned`, `Items Succeeded`, `Failures`) sat in a flex row whose values were `white-space: nowrap`. A value wider than its column could neither wrap nor shrink the column, so it overflowed its box and painted on top of the next column's text — the overlap reported in NAS-142255.

The fixed `width: min(640px, 90vw)` on the dialog hid it at desktop width for typical values (a ~187px column vs ~172px of text), but it still collides with a longer pair such as `1023.99 MiB / 1023.99 GiB` (~192px), and at any window under ~675px where the `90vw` cap shrinks the columns.

### Fix

- `.stats-cards` becomes a grid: `repeat(auto-fit, minmax(150px, 1fr))`. When the columns can no longer hold their contents they reflow onto another row instead of being squeezed under content width. `.stat-card` gets `min-width: 0` so a long value can't floor its track.
- `.stat-value` drops `nowrap` and gains `overflow-wrap: anywhere`. Each half of an `x / y` pair is a `.stat-part` span that stays `nowrap`, so the only break opportunity is at the slash — a value wraps as `222.61 KiB /` / `171.33 GiB`, never between a number and its unit.

### DIFF
--- a/src/app/pages/sharing/components/data-migration-status-dialog/data-migration-status-dialog.component.html
+++ b/src/app/pages/sharing/components/data-migration-status-dialog/data-migration-status-dialog.component.html
@@ -46,7 +46,7 @@
       <div class="stat-card">
         <div class="stat-label">{{ 'Data Scanned' | translate }}</div>
         <div class="stat-value">
-          <span class="stat-part">{{ stats.count_bytes | ixFileSize }} /</span>
+          <span class="stat-part">{{ stats.count_bytes | ixFileSize }} /</span>&ngsp;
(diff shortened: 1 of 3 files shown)


william-gr (approved):

AlexKarpov98:
backport

PR #14047, opened 2026-09-06
NAS-143224 / 27.0.0-BETA.1 / Add S3 share type with bucket, access key and service configuration UI

william-gr:
Master port of truenas/webui#14043 (merged to stable/26 as 72b7be1ebb), the UI counterpart of truenas/middleware#19610.

## What's included

- **Shares dashboard**: S3 card (name, dataset, owner, enabled toggle, edit/delete) with the service status, header toggle and menu from `ServiceActionsMenuService`. The menu carries an "Access Keys" shortcut.
- **Shares > S3**: bucket list with search, column picker and pager.
- **Bucket form**: basic mode is Name, Parent Dataset, Owner, Enabled. Advanced adds permissions model, grants, versioning and snapshot versions, object lock with default retention, multipart ETag, and auditing (license-gated). Validates that the composed dataset does not already exist and that `/mnt` is not chosen as parent.
- **Credentials > S3 Access Keys**: list (status, expiry, rotate, delete) and form. Created or rotated keys open a dialog showing the key pair with a once-only warning.
- **S3 service config** from the Services page and the card menu: listeners, certificate (defaults to UI certificate), servers, region, log level, global grants, default audit (license-gated).
- Plumbing: `truenas_s3` service name, `SHARING_S3_READ/WRITE` roles, API directory entries, interfaces and enums, helptext, `S3BucketDatasetMissing` alert enhancement, search elements.
- `ix-user-picker`: fixed a blank option being added when the control value is an empty string.

## Migration to @truenas/ui-components

Unlike the stable/26 version, every screen follows the master patterns from the tn-migration playbook:

- Card and lists: `tn-card` + `tn-table` with `TableToggleCellComponent` / `TableActionsCellComponent`, `tn-empty` inline, `tn-table-pager` and the column picker.
- Bucket and service config forms: `IxFormHostForm` wrapping `<ix-form>` with `tn-form-section` / `tn-form-field` / `tn-input` / `tn-select` / `tn-checkbox`; the Advanced toggle is a side-panel footer action. `ix-explorer`, `ix-user-picker`, `ix-chips`, `ix-list` and `ix-combobox` stay per the playbook.
- Access-key form: `SidePanelForm` with `tn-date-input`. Credentials dialog: `TnDialogShell` with a warning `tn-banner`.
- All forms open through `FormSidePanelService`; S3 is registered in the Services page config map and in `ServiceActionsMenuService`.

## Tests

[...]

PR #14055, opened 2026-09-07
NAS-143345 / 27.0.0-BETA.1 / Make object lock a basic S3 bucket option, expire access keys by default, show versioning in the list

william-gr:
Follow-up to #14047 after review of the first S3 build.

## Changes

- **Object lock is a basic option.** It has its own section next to the bucket fields instead of living under Advanced Options, since backup targets are the primary use case. Checking it switches Versioning to Enabled and holds it there (the select shows a hint while locked); unchecking releases it. The Multiprotocol permissions model is disabled in the picker while object lock is on, and object lock is disabled with a hint when Multiprotocol is already selected.
- **Retention mode defaults to Compliance** when object lock is turned on. A stored "no default rule" on an existing bucket is left alone; only a user enabling object lock gets the default.
- **New access keys expire by default.** Non-expiring is now opt-in, so the form asks for a date up front.
- **Bucket list shows Versioning and Object Lock** columns by default, both sortable.

## Tests

Bucket form: object lock in basic mode, versioning forced and locked, Compliance default, Multiprotocol exclusion, retention days required/released. Access key form: date required by default until Non-expiring is checked. List: new columns in the row text. 25 tests, lint and AOT clean.

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
     "@smarttools/eslint-plugin-rxjs": "~1.0.22",
     "@stylistic/eslint-plugin": "~2.9.0",
     "@truenas/common-typescript": "github:truenas/tn-common-typescript#master",
-    "@truenas/ui-components": "~0.7.2",
+    "@truenas/ui-components": "~0.7.4",
     "@types/cheerio": "~0.22.35",
     "@types/d3": "~7.4.3",
     "@types/dygraphs": "^2.1.10",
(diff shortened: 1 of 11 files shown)


william-gr:
The expires on issue with calendar seems to be a components library issue.

AlexKarpov98 (approved):
👍

PR #14056, opened 2026-09-07
NAS-143108 / 26.0.0 / Don't reset the page when a checkbox is ticked or the list reloads

AlexKarpov98:
## Problem

Ticking a snapshot checkbox on page 2 of **Datasets → Snapshots** sends the user back to page 1. The row they aimed at vanishes and a page-1 row sits there looking selected instead — which reads as "the wrong snapshot got selected", right before a batch Delete.

There turned out to be **two independent ways** to lose your place on this page. One commit each.

## 1. Ticking a checkbox

The ix-table checkbox columns keep their state **on the row** (`row.selected`) and mutate it in place, so nothing in the table's OnPush tree learns about it. To force the re-render they nuked the rows and re-ran the search filter:

```ts
this.dataProvider.setRows([]);
this.onListFiltered(this.searchQuery());
```

`setFilter` resets to the first page. `setRows([])` also drops `totalRows` to zero on the way through, which makes the pager independently correct itself off a now-out-of-range page — a second route to the same place.

Selecting the **whole page** was worse than reported: it snapped to page 1 *and* left the user looking at 50 unticked boxes over a 50-row selection.

**Fix:** `BaseDataProvider.refreshCurrentPage()` does the re-render on its own and leaves the filter, the sort and the pagination alone. The empty emission it starts with is load-bearing — it is what rebuilds the row cell components, and without it select-all stops rendering as checked at all.

Snapshots is the reported page. **Disks**, **Boot Environments** and **Docker Images** carry the same copy-pasted handlers and the same bug, so they move over too — happy to trim those back to snapshots-only if you'd rather keep this narrow.

## 2. A background reload

`getSnapshots()` subscribes to the store, and the effects subscribe to `pool.snapshot.query` — so a snapshot created, changed or destroyed **anywhere on the system** re-emits here, and the subscription re-ran the search filter, resetting the pagination just the same.

This one matters for retesting: the repro steps call for a dataset with 100+ snapshots, which is exactly the kind of system that has periodic snapshot tasks running. Without this commit, QA could retest the checkbox fix, have a task fire mid-test, and see "snapped back to page 1" again.

**Fix:** `setFilter` takes a `keepPage` option for the case where the same query is re-run against refreshed rows. A query the user *typed* still starts at page 1, which is the right behaviour there.

## Known limitation

**Selections still don't survive a reload.** The reload rebuilds every row with `selected: false`, so a background snapshot event drops whatever the user had ticked. Retaining it is a further change and is deliberately not in this PR — say the word if you want it here rather than on master.

The equivalent work for master is truenas/webui#14057 (blocked on a `@truenas/ui-components` release), which fixes cross-page selection along with a pager bug that doesn't exist on this branch. Neither is a prerequisite for this PR.

## Testing

Three regression tests on the snapshot list, each verified to fail with its fix reverted:

- ticking a row on page 2 leaves the pager and the provider on page 2
- selecting the whole page stays on page 2 **and** every row renders checked
- a store re-emission leaves the user on page 2

561 tests pass across `ix-table`, datasets, disks, bootenv, docker-images and jobs. Lint and `yarn buildng` clean.

### DIFF
--- a/src/app/modules/ix-table/classes/async-data-provider/async-data-provider.ts
+++ b/src/app/modules/ix-table/classes/async-data-provider/async-data-provider.ts
@@ -30,7 +30,7 @@ export class AsyncDataProvider<T> extends BaseDataProvider<T> {
     );
   }
 
-  override setFilter(filter: TableFilter<T>): void {
-    super.setFilter({ ...filter, list: filter.list || this.loadedRows });
+  override setFilter(filter: TableFilter<T>, options?: { keepPage?: boolean }): void {
+    super.setFilter({ ...filter, list: filter.list || this.loadedRows }, options);
   }
 }
--- a/src/app/modules/ix-table/classes/base-data-provider.ts
+++ b/src/app/modules/ix-table/classes/base-data-provider.ts
@@ -73,8 +73,19 @@ export class BaseDataProvider<T> implements DataProvider<T> {
     this.sortingOrPaginationUpdate.emit();
   }
 
-  setFilter(filter: TableFilter<T>): void {
-    this.resetPaginationToFirstPage();
+  /**
+   * Applies a filter and, by default, sends the user back to the first page — the right
+   * move when the QUERY changed, since the rows they were reading are no longer the ones
+   * the table holds.
+   *
+   * Pass `keepPage` when the same query is being re-run against refreshed rows, e.g. a
+   * background reload the user did not ask for. Resetting there reads as the table losing
+   * their place mid-task, which is what NAS-143108 reported.
+   */
+  setFilter(filter: TableFilter<T>, { keepPage = false }: { keepPage?: boolean } = {}): void {
+    if (!keepPage) {
+      this.resetPaginationToFirstPage();
+    }
     const filteredRows = filterTableRows(filter);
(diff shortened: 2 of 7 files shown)


AlexKarpov98:
Added a second commit: the checkbox was one way to lose your place on this page, but not the only one.

`getSnapshots()` subscribes to the store, and the effects subscribe to `pool.snapshot.query` — so a periodic snapshot task firing anywhere on the system re-emits here, and the subscription re-ran the search filter, which resets the pagination. The repro steps call for a dataset with 100+ snapshots, which is exactly the kind of system that has snapshot tasks running, so this was reachable while testing the original fix.

`setFilter` now takes `keepPage` for the case where the same query is re-run against refreshed rows. A query the user typed still starts at page 1.

**It only holds the page.** The reload still rebuilds every row with `selected: false`, so a selection does not survive it — retaining that is a further change and I've left it out of this PR.

New regression test, verified to fail without the fix. 561 tests across `ix-table`, datasets, disks, bootenv, docker-images and jobs; lint and `yarn buildng` clean, all run in a worktree with this branch's own dependencies.

aervin (approved):

PR #14087, opened 2026-09-10
NAS-142357 / 27.0.0-BETA.1 /  Delete dead ix-forms components and migrate unused-disk-select to tn-*

AlexKarpov98:
## What

Continues the ix-forms teardown by removing components that no longer have any
call sites, and converting the last `ix-*`-based custom select.

### Deleted (zero remaining usages)
- `ix-date-picker` (component, styles, spec, harness) — also dropped from
  `supportedFormControlSelectors` and `setup-jest.ts`
- `ix-form-glossary`
- `ix-form-section` — with it, `form-sections.interface.ts`, the section
  registration in `RegisteredControlDirective`, and the
  `registerSectionControl` / `unregisterSectionControl` half of `IxFormService`

### `ix-unused-disk-select` → tn-*
- `ix-combobox` → `tn-form-field` + `tn-autocomplete`, `ix-checkbox` →
  `tn-form-field` + `tn-checkbox`
- The component is now an explicit `ControlValueAccessor`: an inner
  `diskControl` drives the autocomplete and mediates to the host's
  `formControlName` (replaces `SimpleAsyncComboboxProvider` and the
  `viewChild`/`AfterViewInit` wiring)
- Required state is mirrored onto the inner control so `tn-form-field` renders
  the inline error — the field reads validity from the control it wraps, not
  from the host's
- New `testId` input: the inner autocomplete is bound via a bare `[formControl]`
  and has no control name to fall back on, so each of the five call sites
  (replace-disk, extend, raidz-extend, boot-pool attach/replace) passes its own
  base to keep `data-test` values distinct

### Harness
`TnFormControlHarness` now supports a bare `tn-autocomplete` (read + set).
It is probed *after* the `ix-user-*` / `ix-group-*` pickers, which render an
autocomplete of their own and must not be driven through it. `setValue` types
the label before clicking the option, since the panel filters by the input's
current contents.


AlexKarpov98:
Tests failing on master, this PR will resolve issue: https://github.com/truenas/webui/pull/14088

william-gr (approved):

PR #14095, opened 2026-09-10
NAS-e2e: Add the with-pool baseline as a configured template

william-gr:
Stacked on #14094 (templates), which is stacked on #14093 (ISO resolver). Needs iXsystems/api-ci-testbed#337 (`create --template --leave-running` and `freeze`) on the lab box before its first run.

## Summary

Every journey that is not about building a pool has to build one before it can start. This bakes the pool into the appliance instead.

- **One template per baseline**, nicknamed `<prefix>-<baseline>` (`e2e-fresh-install`, `e2e-with-pool`). `TN_GUEST_TEMPLATE` becomes `TN_GUEST_TEMPLATE_PREFIX` (variable `E2E_TN_GUEST_TEMPLATE_PREFIX`, default `e2e`).
- **`with-pool`** is `fresh-install` left running (`create --template --leave-running`), `e2e/ci/baselines/with-pool.py` run against the guest's API to create one striped pool named `tank` on the first unused disk, then `tn_guest.py freeze` for the shutdown and snapshot. A baseline script that fails, or a freeze that fails, deletes the half-built template so the next claim rebuilds rather than trips over it; an unfrozen template is never treated as present.
- **The pool is `tank`, not `e2e_shared_tank`**, so the suite's pool fixture adopts it as the appliance's and never exports it.
- The workflow claims `with-pool` by default and gains a `baseline` dispatch input; `fresh-install` remains for first-run journeys.
- Adding a baseline is a script under `e2e/ci/baselines/` and a line in `appliance.sh`.

Both templates are ISO-keyed the same way as before (#14094).

## Test plan

- [x] `bash -n`, and the branch's `templateCreated` now requires a `snapshot` property so an unfrozen leftover is rebuilt.
- [ ] After api-ci-testbed#337 is on the box: dispatch on this branch builds `e2e-with-pool` (install, pool, freeze) and clones it; the clone shows `tank` ONLINE.
- [ ] `baseline: fresh-install` dispatch builds `e2e-fresh-install` and the suite passes without a pool.
- [ ] Delete the old `e2e-template` on the box by hand once nothing needs it: `tn_guest.py delete e2e-template`.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -81,6 +81,13 @@ on:
         type: choice
         default: built
         options: [built, shipped]
+      baseline:
+        description: >-
+          Which baseline to claim. `with-pool` boots with a pool named `tank`
+          present; `fresh-install` has nothing, for first-run journeys.
+        type: choice
+        default: with-pool
+        options: [with-pool, fresh-install]
       refresh_iso:
         description: >-
           Install from the newest v27 nightly now, instead of the one chosen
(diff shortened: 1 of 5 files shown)


william-gr:
Dropped: the suite's pool fixture already provides a pool over the API, so a baked-in pool is not worth a second moving part in api-ci-testbed.

PR #14097, opened 2026-09-10
NAS-143545 / 27.0.0-BETA.1 / Remove the Docker image builds from CI

truenas-agent-tars:
Every pull request pushed an image tagged with its number, and every merge to master refreshed `ixsystems/truenas-webui:latest`. Neither is part of how anyone tests the UI any more.

Measured on DockerHub before writing this:

| | |
|---|---|
| Repository size | **1.87 TB** |
| Tags | **≥1000** — the listing API caps there, and 999 of that first page are bare PR numbers going back to 2025-10-18 |
| Last push | 2026-09-10 15:14 UTC (PR #14094) |
| Pulls | 194k all-time since 2021, 18 stars |

The pull count is a lifetime counter, so it neither confirms nor refutes current usage on its own.

## What goes

- `.github/workflows/docker_internal.yml` — per-PR build and push
- `.github/workflows/docker_latest.yml` — master → `:latest`
- `Dockerfile`, `docker/start.sh`, `.dockerignore` — nothing else builds the image
- `docs/testing_ui.md` and its README link — it told readers to `docker container run ixsystems/truenas-webui:<PR number>`, which stops being true the moment the builds stop

## What stays, and why

`docker/nginx.conf`. It is no longer part of an image build, but `e2e.yml` mounts it into a stock nginx container to serve the production build during an end-to-end run, filling the same `%%HOSTNAME%%` and `%%VERTAG%%` placeholders `start.sh` used to.

Two comments in `e2e.yml` named the `Dockerfile` and `start.sh`, so they now describe what the `sed` does instead. One of them also claimed this file is what serves the shipped UI — `debian/rules` installs only `dist/`, so it is a stand-in for the appliance's routing rather than the config an appliance runs. That comment now says so.

## Two things this PR does not do

1. **It reclaims none of the 1.87 TB.** Deleting the workflows stops the growth; pruning the ~1000 existing tags needs DockerHub org access, which I do not have.
2. **`DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` are left unreferenced** by any workflow — I checked the rest of `.github/`. A push-capable token that nothing uses is worth revoking rather than leaving parked.

## Before merging

The PR job reports under the check name **`docker`** (the job carries no `name:`, so GitHub uses the job id). **If that name is in master's required-status-checks list, merging this leaves every open PR waiting on a check that will never report again.** I cannot read branch protection, so someone with admin should confirm it is not required — or remove it there first. `main.yml` already warns that branch protection is coupled to reported check names, which is what prompted the check.

Draft pending a NAS ticket number for the title.


truenas-agent-tars:
Both LOW findings reviewed. One taken, one is not a code change and I have partial evidence on it now.

### LOW — `docker/nginx.conf` is an orphan with no in-file note · **taken**, `1f67576`

Checked the convention cited before accepting it, and it is real and exactly where the review said: `check-ticket.yml:4-6` and `claude.yml:22-26` both record cross-file coupling at the coupling. This file had nothing, and after this PR its directory holds only it.

The header did more harm than the missing note, though. It reads `TrueNAS nginx configuration file`, and that is the claim I corrected in `e2e.yml` — `debian/rules` installs only `dist/`, so an appliance serves the UI through middleware's own nginx. A reader landing in this file would have taken the title at face value and over-read what an E2E run proves. So the note covers both: what the file is not, and who reads it.

```
#    Not the config an appliance runs. The deb installs only `dist/` (see
#    debian/rules), so a real appliance serves the UI through middleware's own
#    nginx; this mirrors that routing for testing.
#
#    One consumer: the "Serve the UI" step in .github/workflows/e2e.yml, which
#    mounts this file into a stock nginx container to serve a production build.
#    That step fills %%HOSTNAME%% and %%VERTAG%%, points the log directives at
#    the container's stderr, and rewrites the plain-HTTP `listen` below to TLS.
#    It asserts that `listen` line is still here, so restructuring the server
#    block fails the E2E job rather than silently serving the wrong thing.
```

### LOW — the `docker` check name stops reporting · **no change; still needs an admin**

Correctly raised and correctly rated. I went looking for a way to settle it from outside branch protection and could not, but I did turn up something adjacent that is worth recording.

PR #14091 merged this morning with **five** check runs: `Checkout and Install`, `Validate translation strings`, `check-ticket` (once failing, once passing), and `docker` (`skipped`). No `Build`, no `Validate code style`, no `Run tests (…)`. If any of those were required, that PR could not have merged — so master's required set is either very small or routinely bypassed.

That is suggestive, **not proof about `docker` specifically**: `docker` did produce a check run on that PR, as `skipped`, and GitHub counts a skipped job as satisfying a requirement. The case that would settle it — a merged PR where the workflow never triggered, so no `docker` check exists at all — needs a PR touching only `tests/**`, which is this workflow's sole `paths-ignore` entry and a directory that no longer exists on master. So there is nothing recent to test against.

Someone with admin still has to confirm `docker` is not in master's required checks, or drop it there first. A required check that never reports stays pending rather than failing, so the failure mode is every open PR silently unmergeable with nothing red to explain it.

william-gr (approved):

PR #14113, opened 2026-09-11
NAS-143514 / 27.0.0-BETA.1 / Do not prefill the bucket parent dataset from the managed root

william-gr:
Follow-up to #14101: a new bucket should start with an empty Parent Dataset, not one prefilled from the S3 service's managed root dataset. Removes the prefill and its spec. Same change is going to stable/26 (#14111) and release/26.0.0-RC.1 (#14112).


william-gr:
backport

PR #14144, opened 2026-09-16
NAS-143801 / 26.0.0 / Tag unentitled S3 features Premium instead of hiding them

william-gr:
Backport of #14139 to the 26 line. Middleware backported the entitlement keys to both 26 targets (truenas/middleware#19727, truenas/middleware#19728), so the UI gates them here too.

Closes [NAS-143801](https://ixsystems.atlassian.net/browse/NAS-143801).

## Not a cherry-pick

The 26 line is a different UI generation, so this is a port. `git cherry-pick` of the master commit conflicts in 11 files; three differences are structural:

| | master | this branch |
|---|---|---|
| S3 form sections | `tn-form-section` | `ix-fieldset` |
| Section spacing in `_tn-styles.scss` | `tn-form-section + tn-form-section` rule | no such rule |
| Entitlements reducer on failed load | `failedLoadEntitlements` + `assumed` | fell back to `{}` |

Adapted accordingly, and kept on webui's own components rather than `@truenas/ui-components` (0.1.57 here), which has neither `scopeTestId` nor `TnTestIdValue`:

- The badge uses `matTooltip` and `ixTest`, mirroring `missing-access-wrapper`, this wrapper's direct sibling on this branch.
- The wrapper's dimming and pointer-blocking name the `ix-*` controls, exempting `ix-tooltip` so the help stays readable.
- `PremiumFeatureDirective.testId` is a plain `string`, since `ixTest` composes the `button-` prefix itself.

The e2e half of #14139 is omitted — the suite does not exist on this branch.

## Groundwork this needed

`selectIsEntitledStrictly` fails closed, so an absent key reads as denied. The reducer here left no way to tell a real answer from the permissive map a failed fetch leaves behind, so an `assumed` flag was added first: without it, a licensed appliance whose `truenas.entitlements.info` call failed once at boot would tag every S3 section Premium and drop its audit settings from a create.

## Test plan

- [x] `yarn test` across directives, entitlements store/service and both S3 pages — 93 pass, 19 suites
- [x] Strict reading reverted to `?? true` → 2 tests fail, confirming they guard it
- [x] `yarn lint`, `yarn build`
- [ ] Click-through on a 26 appliance

## Notes

- The `keepFocusable` exemption has no keyboard half here: a field's help is a `tn-icon` with no tabindex, which the focus sweep never selects. What it restores is `pointer-events`, which jsdom does not compute — noted in the spec rather than covered by a test that would pass regardless.


william-gr:
ported

PR #14153, opened 2026-09-16
NAS-143804 / 27.0.0-BETA.1 / Restore per-row test IDs on tn-table lists and enforce them

AlexKarpov98:
**Changes:**

Credentials → Users emitted no `data-test` anywhere in the table body: `tbody [data-test]` was 0
while the rest of the page had 45. `002b2ad8a7` replaced `<tbody ix-table-body>` — which tagged
every row from the column model's `uniqueRowTag` — with bare-interpolation `tnCellDef` bodies, and
`tn-table` writes nothing on the `<tr>` of its own. This is the sixth accepted ticket of that class
(NAS-141047, NAS-141484, NAS-142069, NAS-141791, NAS-141186), so this PR fixes the defect, sweeps
the app for the rest of the class, and puts a check in front of the seventh.

### 1. Per-row test IDs on tn-table lists

Every cell now carries the row's identity, at the value `[ixTest]` resolved before the migration:

| List | Row tag | Was |
|---|---|---|
| `credentials/users/all-users/user-list` | `user-<username>` | nothing |
| `credentials/groups/group-list` | `group-<group>` | toolbar only |
| `audit/components/audit-list` | `audit-<service>-<username>-<event>-<audit id>` | nothing |
| `datasets/modules/snapshots/snapshot-list` | `snapshot-<id>` | toolbar / bulk only |
| `system/advanced/cron/cron-list` | `cron-<command>-<description>` | detail row only |

Each of those lists also binds `[rowTestId]`, which puts the tag on the `<tr>` itself
(`row-user-truenas-admin`) — the input added for this in `@truenas/ui-components` 0.7.7
(iXsystems/truenas-ui-components#318, merged and released). The row id is the handle for *acting*
on a row; the cell ids are how a suite reads or asserts one column, since descending into a column
by `[data-column]` is a CSS selector the e2e rules do not allow. Both halves are needed.

`disk-list` joins them: twelve bare-interpolation columns, no actions column, and its only row tag
in the detail row — which a test can only open by clicking a row it cannot select. It now binds
`[rowTestId]` too, so its rows are `row-disk-sda`.

So `text-username-user-truenas-admin-row-text` selects the username cell of that row, and clicking
it activates the row exactly as clicking anywhere else in it does. Cron was not in the ticket's
audit — its only per-row ids sit on detail-row buttons, which you can reach only by first clicking
a row you cannot address, so it belongs to the same class. Users' Access column also tags each
access indicator (`icon-user-<name>-smb-access`, `…-ssh-access`, …), which is the only place that
state appears in the table.

Existing text/yes-no columns go through the shared `<ix-table-text-cell>`; date, size and the
composite audit cells carry `tnTestIdType` + `[tnTestId]` directly. Row tags are built with
`memoizedRowTag`, so the value is normalized the way `[ixTest]` normalized it and is computed once
per row rather than per cell per change-detection pass.

Two ids are renames rather than restorations. The snapshots "Date created" column resolved to
`date-date-created-…` under `[ixTest]`, and the library's `composeTestId` refuses to repeat a type
prefix the base already starts with, so it is now `date-created-snapshot-<id>-row-date`. There is
no way to pin it back through the directive. Noted in the spec and in `e2e/CLAUDE.md`. And the
multi-error dialog's "Download Logs" button was `button-download-logs` on every error it rendered,
so it is now scoped by the error's index, `button-download-logs-<index>` — a repeated id resolves
to whichever element the selector happens to reach first, which is not something a suite can pin.

### 2. The same gap everywhere else in the app

A sweep of all 794 templates for elements handling a `(click)`/`(keydown)` with no `data-test` on
themselves or on any descendant found 25. Four are genuinely not automation targets and are
allowlisted with their reason; the other 21 are now addressable:

| Surface | Added | Why it matters |
|---|---|---|
| ACL editor (`acl-editor-list`) | `row-ace-<i>`, `button-remove-ace-<i>`, `button-ace-error-<i>` | Dataset ACL is the *next* e2e story (`e2e/docs/status.md`), and its rows could not be selected at all |
| Pool export dialog | `option-export-pool`, `option-delete-pool` | "Deleting a pool" is one of the journeys the old known-gap note called blocked; the choice between the two was unaddressable |
| Page alerts | `button-alert-details-<key>`, `button-dismiss-alert-<key>`, `link-alert-docs-<key>` | Per alert, so a test can dismiss the one it caused |
| Jobs panel (`job-item`) | `row-job-<description>-<method>` | The row body; only its icon buttons were tagged |
| Apps | `option-<sort value>`, `link-section-<name>` | Repeated entities in the sort list and the install wizard's section nav |
| Error dialogs | `link-show-detail-<label>`, `link-hide-detail-<label>`, `button-more-info` | Per-detail trace toggles |
| Advanced search | `input-advanced-search` | The editor you type a query into |
| Global search | `button-clear-global-search` | |
| Session-expiring dialog | `button-open-preferences` | |
| System security form | `link-configure-<title>` | Per unmet STIG requirement |
| Network | `link-go-to-ha` | |
| Tier status, dashboard product image | `text-migration-status`, `image-<product>` | |

Allowlisted, in `scripts/check-test-ids.ts` with the reason: the two `admin-layout` scrims, the
alerts-panel container (listens for Escape; its contents are tagged), and the dataset tree's inner
click target, which sits inside `<tn-tree-node [testId]="['dataset', name]">` and resolves to that.

### 3. Enforcement, so the seventh ticket does not happen

`e2e/CLAUDE.md` already bound the *test author* to `[data-test]` selectors; nothing bound the PR
that removed the attribute.

- `scripts/check-test-ids.ts` (`yarn check-test-ids`, a lint-job step) enforces both rules above:
  something that renders INSIDE a row must carry the row's identity — a cell body, a row action, or
  `[rowTestId]` on the table, and explicitly not the detail row — and a clickable must resolve an id
  on itself or a descendant. Text scan, deliberately shallow: it catches the mistake that actually
  happens and accepts anything that plausibly resolves an id — a false negative is the acceptable
  failure here, a false positive is not.
- `e2e/CLAUDE.md` gains an "Addressing a table row" section — how a row is reached through its
  cells, the two rules that keep that true, and a fourth entry in the "things that will catch you
  out" list for the un-doubled type prefix. The "Known gap" note and its counterpart in
  `e2e/docs/status.md` are retired: list-driven journeys are no longer blocked.
- The PR template's Testing block now asks what happened to a touched table's test IDs.
- The tn-migration playbook's table recipe leads with this, since that is what a migrating agent
  reads.

### Not in scope

- **Per-column coverage, measured and deliberately deferred.** Across all 81 tn-table templates
  there are 450 columns with a cell body, and 117 of them (26%) still carry no id — so a row can be
  acted on but one column's value cannot always be asserted. Every row is now addressable, which is
  what was blocking; the remaining columns are worth tagging where a suite actually asserts them,
  and that is a follow-up rather than 117 speculative edits in this PR.
- `tn-table` no longer consumes `ariaLabels`, so a row's accessible name is the concatenation of its
  cells (icon names included) and is locale-dependent. `[rowTestId]` closed the automation half of
  that gap; the `aria-label` half would need the same shape of input library-side and is still open.
- 11 `<a href>` links have no id — all of them `target="_blank"` links to truenas.com docs, forums
  and support. A journey does not click those, so the clickable check covers handlers, not links.
  Happy to tag them too if Release Engineering wants to assert the hrefs.

**Testing:**

- `yarn jest` over every touched area: 77 suites, 667 tests, all pass. Includes a new
  "tags every cell with the row it belongs to" assertion per list, pinning the exact resolved ids.
- `yarn check-test-ids` — 81 tn-table templates tag their rows; no unaddressable clickable in 794
  templates. Both rules were tightened after review found them too loose: the clickable scan
  stopped at the first `>` in a tag, so a handler written after `[class.wide]="n > 3"` was skipped
  entirely (proven with a probe template — silently green before, reported after), and the row rule
  accepted a tag in the detail row.
- `yarn lint` and `tsc -p src/tsconfig.app.json` clean.
- Verified in a browser against a real appliance: Users `tbody [data-test]` 202 across 50 rows
  (was 0), Groups 6 per row, Audit 250 across 50 rows, ACL editor `row-ace-0/1/2` (note index 0
  survives — the legacy directive dropped a falsy segment, `[tnTestId]` does not), page alerts
  `button-alert-details-…` / `button-dismiss-alert-…`. Snapshots and Cron had no records on the
  box; both are covered by spec. Screenshots on the ticket.

No behaviour, layout or copy changes — cells render the same values through a wrapping element, and
the rest is attributes.

### Downstream

|Affects         |Reasoning
|Documentation   | No — internal working notes only (`e2e/CLAUDE.md`, `e2e/docs/status.md`, PR template, migration playbook).
|Dependencies    | Yes — `@truenas/ui-components` `~0.7.6` → `~0.7.7`, for the `[rowTestId]` input added in iXsystems/truenas-ui-components#318. Patch release, additive input, nothing else in it.
|Testing         | Yes, in the intended direction. Five lists and 21 clickables that emitted no `data-test` are now addressable, at the values `[ixTest]` produced where one existed before; nothing that resolved before changes, except two renames: `date-date-created-…` → `date-created-…` on the snapshots list, which the library will not let us pin back, and `button-download-logs` → `button-download-logs-<index>` in the multi-error dialog, which renders one template per error and so repeated a single id across every one of them. Adds a lint-job step that fails a PR dropping a row id or leaving a clickable unaddressable.


James-OReilly-TrueNAS:
Acking the snapshot test ID rename flagged in the bot review:

> `snapshot-list.component.html:113` renames a resolved `data-test` that downstream suites select on (`date-date-created-…` → `date-created-…`).

Snapshot E2E are in active development on our end, as such any change to their tags are easily addressable.

AlexKarpov98:
Both findings from the last round are fixed, and the library half landed, so this PR now carries the row ids too.

**MEDIUM — `check-test-ids.ts:58`, the attribute group ended at the first `>`.** Fixed: quoted values are consumed whole, so a `>` inside a binding no longer ends the tag. Verified rather than assumed — dropping the reviewer's exact example into `src/app` as a probe:

```
--- with the fix:  ❌ src/app/__gate-probe.component.html:1  <div class="tiles">
--- before the fix: ✅ 795 templates carry no unaddressable clickable
```

Nothing in the app was hiding behind the hole today — the gate is still green with it closed.

**LOW — rule 1 was satisfied by a tag anywhere in the file.** Agreed, and it was worse than "usually an actions button": it counted a tag in the **detail row**, which a test can only open by first clicking a row it cannot select. Rule 1 now looks only at what renders inside the `<tr>` — cell bodies, row actions, or `[rowTestId]`.

Exactly one table in the app failed the stricter rule: `disk-list` (twelve bare-interpolation columns, no actions column, its only `uniqueRowTag` in the detail row). It now binds `[rowTestId]` and its rows are `row-disk-sda`.

**The library input shipped**, so the five migrated lists bind `[rowTestId]` as well — `@truenas/ui-components` 0.7.7, bumped here. Rows now carry `row-user-<name>`, `row-group-<name>`, `row-audit-<…>`, `row-snapshot-<id>`, `row-cron-<…>`, `row-disk-<name>`.

**Still not covered, measured rather than guessed:** of 450 columns with a cell body across all 81 tn-table templates, 117 (26%) still carry no id. Every *row* is addressable now; those columns are about asserting a particular value, and tagging them is worth doing where a suite actually needs it rather than as 117 speculative edits here. Follow-up ticket.

PR #14171, opened 2026-09-17
NAS-143906 / 26.0.0 / Default the multipart ETag to Minted

william-gr:
Backport of #14170.

Closes [NAS-143906](https://ixsystems.atlassian.net/browse/NAS-143906).

A bucket created from the form now defaults its **Multipart ETag** to **Minted** rather than Composite.

## Changes

- **Default.** `multipart_etag` initialises to `MINTED`. Only new buckets are affected: an existing bucket is loaded from its stored value, so editing one leaves whatever it has.
- **Tooltip.** The old copy read "Choose Minted only where nothing writing the bucket reads its ETags", which argued against the new default. It now presents Minted as the default — an opaque token, skipping the MD5 pass over every part — and Composite as the choice for a client that reads multipart ETags and expects the S3 standard construction.

The 26 form is the older `ix-*` generation, so the change applies to the same two lines but its test reads the select through `IxFormHarness` rather than `TnSelectHarness`.

## Notes

- This is the UI's default only. A bucket created through `sharing.s3.create` elsewhere still takes middleware's default, so if Minted is meant to be the product default, middleware needs the same change.

## Test plan

- [x] `yarn test src/app/pages/sharing/s3` — 39 pass, incl. 1 new asserting the form shows Minted (on both branches)
- [x] The create-payload test now expects `MINTED`; the edit tests keep a stored `COMPOSITE` bucket, so they cover an existing bucket keeping its value
- [x] Mutation-checked: putting the default back to Composite fails exactly those two tests
- [x] `yarn lint`
- [ ] Click-through on a 26 appliance


william-gr (commented):
backported, trivial

PR #14177, opened 2026-09-18
NAS-143957 / 27.0.0-BETA.1 / Let an error dialog name itself, and the shell render its title

aervin:
**Changes:**

Every error dialog opens the same component with the same `testId="error"`, so nothing could tell one from another except its wording. The E2E harness had to match the development build's "Max Concurrent Calls" diagnostic by its prose in order to dismiss it — a selector that breaks the moment the message is reworded or translated.

Two changes close that:

- **`ErrorReport` gains an optional `testId`**, bound to `tn-dialog-shell`. Unset, a dialog keeps the shared `error` base it has always had; set, it names itself. `websocket-handler.service.ts` passes `concurrent-calls` for the concurrency diagnostic.
- **The dialog shell renders the error title.** `error-dialog.component.html` was passing `[title]="''"` and rendering its own `<h1>` — an icon beside the title — so the shell's heading, and the `dialog-title-*` id that comes with it, never existed. That heading is now the shell's, and the harness matches `[data-test="dialog-title-concurrent-calls"]` instead of prose.

Removing the custom heading makes three things dead, all removed: the `icon` field on `ErrorReport`, its styles, and the `[ariaLabel]` binding (with a real title the shell points `aria-labelledby` at its own heading).

Needs `@truenas/ui-components` >= 0.7.8, already on master, for `dialog-title-<base>` (iXsystems/truenas-ui-components#319).

**Testing:**

`data-test` values — **nothing removed or changed**; two new ones appear only when a caller opts in via `testId`:

| Value | When |
|---|---|
| `dialog-title-<testId>` | an error dialog given a `testId` — `dialog-title-concurrent-calls` today |
| `button-close-<testId>` | same, on the shell's own ✕ |

`button-close-error-dialog` (the footer Close) and `button-close-error` are untouched, so nothing downstream moves.

Unit tests: the error-dialog spec now asserts the title through `TnDialogHarness.getTitle()` rather than a CSS class — 17 passing. Three icon tests were deleted along with the behaviour. The error-parser spec drops three `icon` expectations — 22 passing.

Full E2E suite against a TrueNAS v27 nightly from a golden snapshot: **26 passed, 2 skipped**. The two skips are deliberate `test.skip` calls (the HTTPS-warning assertion cannot hold under the `branch` profile; S3 versioning needs the `S3_VERSIONING` entitlement).

**One behaviour change worth a reviewer's eye:** network errors lose their distinct icon. `ECONNRESET`, `ETIMEDOUT` and `ENETUNREACH` set `icon: cloud-off` to mark themselves as connectivity rather than appliance problems, and now show no icon like every other error. The "Network Error" title, the hint, and the "Network Settings" action are unchanged, so nothing is unsayable — but the at-a-glance cue is gone.

### Downstream

|Affects         |Reasoning
|Documentation   | None. No user-facing behaviour changes beyond the error dialog losing its title icon, covered above.
|Testing         | Adds `dialog-title-*` / `button-close-*` ids for opted-in error dialogs; removes none. The E2E harness now matches the concurrency dialog by id instead of its wording. Any suite asserting the error dialog's icon or its `.err-title` element needs updating — in tree, only the two specs changed here did.


AlexKarpov98 (approved):
👍

